PR #11, opened 2013-08-14
Removed unused exception, and in order to avoid import I0Itec zkclient

yuyijq:
Removed unused exception, and in order to avoid import I0Itec zkclient


teaey:
预防代码。感谢

PR #56, opened 2014-06-16
修复dubbo admin部署在子目录相对路径不正确问题

ddatsh:


chickenlj:
It's been fixed by referencing your code, thanks.

PR #107, opened 2015-02-28
修复当设置了"dubbo.protocol.host"参数后启动脚本无限等待的问题

sonicwu:
Fix 'start.sh' waits indefinitely when specified the 'dubbo.protocol.host' parameter.


beiwei30:
is 'dubbo.protocol.host' an official parameter in dubbo.properties?

chickenlj:
'dubbo.protocol.host' is an official parameter @beiwei30 

I'll accept this PR

PR #209, opened 2016-03-02
optimize some static codes

wanaright10:
some code will be cause NullPointerException or boolean check is alway true and so on...


takeseem:
代码自动格式化，导致git比较更改时，太难看了。。。

PR #449, opened 2017-05-31
增加kendo对zookeeper acl digest 认证的支持

zhangrp:


kimmking:
i think you will create a testcase for a new feature

PR #520, opened 2017-08-14
shutdown with count down latch

johnraychina:
Notify - wait mechanism is fragile,  Main.class is global, anyone could wait on or notify it, which makes shutdown fragile too.

Shutdown with count down latch can avoid evil notify on Main.class(which causes abnormal shutdown) or evil wait on Main.class(which causes non-stop).

### DIFF
--- a/dubbo-container/dubbo-container-api/src/main/java/com/alibaba/dubbo/container/Main.java
+++ b/dubbo-container/dubbo-container-api/src/main/java/com/alibaba/dubbo/container/Main.java
@@ -90,8 +90,7 @@ public class Main {
             try {
                 running.await();
                 break;
-            } catch (InterruptedException e) {
-                e.printStackTrace();
+            } catch (Throwable t) {
             }
         }
     }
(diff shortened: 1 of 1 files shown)


qinliujie:
@johnraychina I think it is better to put running.countDown() outside the for loop, how do you think?

johnraychina:
@qinliujie Sometimes container.stop() may throw exceptions, 
it denpends on how you want shutdown the application: 
- I prefer to solve the exceptions, and shutdown again;
- If running.countDown() is outside of the loop,  main thread is stopped while container is still running, then the application is in a very strange status.

bigcai:
why do Main.class start so many container? --" for( Container container: containers ) { .....}  
BUT the countDownLatch just count one

johnraychina:
@bigcai 
Because Multiple containers is added AFTER my fork, and this branch is is conflict with master.

johnraychina:
Since simple object lock will be ok to avoid evil wait on Main class, I 'll just create another pull request.

chickenlj:
@johnraychina 
Current logic do have problem. 
1. Main.class.wait() may never receive a notify.
2. But because of `running` flag, i think the main process will not shutdown unexpectedly.

Looking forward to your new PR

chickenlj:
I have fixed this, see [Fixed #520 change Main.class wait/notify to lock/condition
](https://github.com/alibaba/dubbo/commit/49af203aa20667a0568a19566c1e9892166435ae)

PR #597, opened 2017-09-05
fix LruCache accessOrder set false

helloworldwt:
#517 针对缓存命名与行为不一致的修复，目前实现为LRU方式


qinliujie:
this problem has been fixed, thanks you!

PR #804, opened 2017-11-01
Optmize ServiceConfig、ReferenceConfig

plx927:
1. extract common methods and fields  to superclass
2. correct some spelling error
3. delete some unnessary boxing and unboxing
4. add some comment for Class


chickenlj:
thx, We'll consider this refactor later

PR #822, opened 2017-11-06
使用redis作为注册中心时，1.添加密码和指定dbindex的支持2.修复monitorservice服务注册的异常

klboke:
@chickenlj 请查看下，使用reids作为注册中心时指定dbindex，注册服务key和业务key可以区分存储


beiwei30:
1. password support has already in.
2. I don't agree with by-pass logic for MonitorService, it doesn't look right

Because of the above reasons, I file one new issue https://github.com/apache/incubator-dubbo/issues/1830 to track db.index, and close the current PR.

PR #859, opened 2017-11-15
Update ReferenceConfig.java

wujisheng:
setInterface(Class<?> interfaceClass)在泛化调用时不应该再去调用setInterface(String interfaceName)方法


chickenlj:
Actually, this is an api designed for user, i think there's no need to check whether we are in a generic context, because we usually don't have the definition of the interface when we use generic reference.

PR #884, opened 2017-11-19
添加method class缓存

xgqiuq:
在高并发情况下（QPS大于1000），RPC调用发起线程validation操作阻塞在`Class.forName(methodClassName, false, Thread.currentThread().getContextClassLoader())`上，特别是在与spring boot一起使用时，TomcatEmbeddedWebappClassLoader加载methodClass通常是耗时操作，不使用分组的情况下methodClass通常不存在

### DIFF
--- a/dubbo-filter/dubbo-filter-validation/src/main/java/com/alibaba/dubbo/validation/support/jvalidation/JValidator.java
+++ b/dubbo-filter/dubbo-filter-validation/src/main/java/com/alibaba/dubbo/validation/support/jvalidation/JValidator.java
@@ -64,6 +64,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * JValidator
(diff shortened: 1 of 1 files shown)


dubbo-filter/dubbo-filter-validation/src/main/java/com/alibaba/dubbo/validation/support/jvalidation/JValidator.java
```diff
@@ -89,6 +92,7 @@ public JValidator(URL url) {
             factory = Validation.buildDefaultValidatorFactory();
         }
         this.validator = factory.getValidator();
+        this.methodClassMap = new ConcurrentHashMap<String, Class>();
     }
```
hepin1989: AtomicReference+ImmutableMap

chickenlj:
Can we cache the `groups` for each method?  I have noticed that `groups` for a method is never changed.

diecui1202:
@chickenlj The cost is `Class.forName` and throw ClassNotFoundException. So cache groups is not necessary.

@xgqiuq Before merging this PR, you need to sign the license. By the way, consider my reviews. Thanks.

PR #932, opened 2017-11-27
Hessian bugfix, revert the override sequence of hessian, field of subclass should override that of parent class, not the other way around.

lovepoem:

https://github.com/dubbo/dubbo-feature-test/issues/1

## Brief changelog

update JavaDeserializer.class , reverse the field list.  Because the old type have a bug: the filed of parent class will cover the filed of sub class

because traversing fileds by while loop, the filed of sub-class is in the top of the List，when being transmitted to Map，it will be covered by the  the filed of parent-class 
## Verifying this change

https://github.com/dubbo/dubbo-feature-test/pull/2


qinliujie:
I have encountered this problem before, but I forgot to fix it.  -_-|||.  I would like to make a suggestion that you can move UT to this project @lovepoem . @chickenlj please review

lovepoem:
The unit test is added . @qinliujie @chickenlj

PR #940, opened 2017-11-28
#938 添加丢失的getter setter

lovepoem:
https://github.com/alibaba/dubbo/issues/938 修改代码，添加statisticsDirectory和chartsDirectory的getter setter

## Brief changelog
 dubbo-simple/dubbo-monitor-simple/src/main/java/com/alibaba/dubbo/monitor/simple/SimpleMonitorService.java


qinliujie:
@chickenlj it is ok,please review

chickenlj:
I deleted the getter&setter when fix issue #847
I think we can config these two properties in dubbo.properties

lovepoem:
@chickenlj chi ok

PR #995, opened 2017-12-05
Improve the annotation configuration

meixinbin:
## 完善注解配置

新增@Method，@Argument，@Parameter注解


chickenlj:
We have refactored the annotation part, `AnnotationBean` will deprecate soon. You can reimplement based on the new program.

chickenlj:
I am sorry, you have to refactor your code based on the new annotation implementation.

PR #1112, opened 2017-12-25
Fix bug in router

zhuyucheng52:
修复选择路由bug


chickenlj:
We check `enabled` immediately after  receiving notification, we check `runtime` here only to  decide whether we should run the router rule every time.

zhuyucheng52:
com.alibaba.dubbo.registry.common.domain.Route#toUrl method return runtime value is always false and not change any where.

PR #1118, opened 2017-12-26
bugfix: throw a NPE in deserialization process when one property of java.sql.Time type happens to be null

windlaughing:

RPC调用反序列化过程中，如果对象的成员变量的类型为java.sql.Time，且其值为NULL，则抛出NPE。

## Brief changelog
添加了必要的非空检查


chickenlj:
That's a hessian serialization problem, and i think your solution can works fine.
But can you add a  unit test for this change?

PR #1139, opened 2017-12-28
fix compile failure cause by missing jetty-util and servlet-api

linuke:

add dependency jetty-util and servlet-api 

## Brief changelog

add dependency jetty-util and servlet-api


chickenlj:
How come there is a compilation problem.
You get this problem when you try to compile the dubbo source or when you are using dubbo in you own projects? 
What is your jetty version?

linuke:
try to compile the dubbo source .

chickenlj:
I have transfer this transferred this part to a separate project. Now i am sure we won't face this problem.

PR #1144, opened 2017-12-29
[Dubbo-#1123] Fix several problems in ConsistentHashLoadBalance

kirayunfei:

Fix several problems in ConsistentHashLoadBalance

## Brief changelog

1. It didn't support generic invoke
2. The identityHashCode always change when invoking, it cause the rebuilding of hash cycle
3. I want to replace the reference in virtualInvokers by the invoke address. It seems to be my personal style, I don't hope a tool like ConsistentHashLoadBalance to catch the reference of Invoker which is important in the whole flow
4. Add a test to run loadbalance

## Verifying this change

1. Start the LoadbalanceProvider and LoadbalanceProvider2 to be the two provider. Only one provider the loadbalance will not work
2. Run LoadbalanceConsumer. And the same param invoke will send to the same provider, no matter whether the invoke is send by generic or not


kirayunfei:
### What is the purpose of the change
Here a new problem is found.
When using GenericService，and add a MethodConfig with loadbalance config，the loadbalance on method will be no effect. 

### Brief changelog
1. Use the tool method getRealMethodName() and getArgument() in RpcUtils instead of the code writen by myself.
2. Fix the problem that the methodName used in the AbstractClusterInvoker is not correct. Using RpcUtils.getRealMethodName(invocation) instead of invocation.getMethodName()
3. Change the test case of loadbalance consumer to adapt the new case，and cover the new logic

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/ConsistentHashLoadBalance.java
```diff
@@ -32,114 +33,156 @@
...
-        }
-
-    }
+	private final ConcurrentMap<String, ConsistentHashSelector<?>> selectors = new ConcurrentHashMap<String, ConsistentHashSelector<?>>();
+
+
+	@Override
+	@SuppressWarnings("unchecked")
+	protected <T> Invoker<T> doSelect(List<Invoker<T>> invokers, URL url, Invocation invocation) {
+		String methodName = RpcUtils.getMethodName(invocation);
```
aspirewit: Please indent all code by 4 spaces

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/ConsistentHashLoadBalance.java
```diff
@@ -32,114 +33,156 @@
 
 /**
  * ConsistentHashLoadBalance
- *
+ * Fixed by:  yunfei.gyf
+ * After fixed, the ConsistentHashLoadBalance will surpport to used in GenericService,
+ * and it will reduce the times of rebuilding hash cycle when the provider restart 
+ * but nothing to change
+ * 
+ * @author yunfei.gyf
```
aspirewit: Please remove the author's information, git will records your contribution faithfully. More Info, see #1044

kirayunfei: It has been removed

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/ConsistentHashLoadBalance.java
kirayunfei: I'm so sorry that the code has been indented by 4 space in my IDE...emmm...

chickenlj:
Sorry for late reply, i have checked your PR, it contains some really valuable improvements:
* Support Generic Invoke
* Avoid keeping reference of invokers

But i still got some questions:
1. identityhashcode will only change when providers(invokers) change(increase/reduce/override, anything that will cause recreation of invokers), this will trigger the recreation of `ConsistentHashSelector`, but it won't reschedule the hash circle, the existing invoker will still be in the same position.
2. Instead of calculating hashcode of invokers, how about add customized `hashcode()` for Invoker?

chickenlj:
```java
String key = invokers.get(0).getUrl().getServiceKey() + "." + method;
        int identityHashCode = System.identityHashCode(invokers);
        ConsistentHashSelector<T> selector = (ConsistentHashSelector<T>) selectors.get(key);
        if (selector == null || selector.identityHashCode != identityHashCode) {
            // Every time we create a new ConsistentHashSelector, selectors replace the reference with the new one, won't the old one be Garbage Collected later?
            selectors.put(key, new ConsistentHashSelector<T>(invokers, method, identityHashCode));
            selector = (ConsistentHashSelector<T>) selectors.get(key);
        }
        return selector.select(invocation);
```

kirayunfei:
@chickenlj Exactly sorry for late to reply，I am ill in hospital recently。
I will answer the question you left before，and I want to answer the second question first：
1. Q：Instead of calculating hashcode of invokers, how about add customized hashcode() for Invoker?
A：In my code，I use the IP address list of invokers to calculate the hash code。Because I think if the providers provide the same interface and the providers list has not changed（maybe shutdown，restart or new startup），the selector is not need to be rebuild。Therefore，if the invokers IP address list or ports has changed（no matter add，reduce and change），the hash code must be changed。Then selector will be rebuild，and hash circle will be changed too

2. Q：identityhashcode will only change when providers(invokers) change(increase/reduce/override, anything that will cause recreation of invokers), this will trigger the recreation of ConsistentHashSelector, but it won't reschedule the hash circle, the existing invoker will still be in the same position.

A：I‘m sorry that I didn't really understand the fact that when ConsistentHashSelector trigger the recreation, the hash circle won't be rescheduled. When the hashcode of ConsistentHashSelector is different in one invocation, the ConsistentHashSelector will be rebuild. And the virtualInvokers are also rebuild according to the address list of invokers. The address list will be different when such happened. I'm not sure I really understand your question, keep in touch, thx！

diecui1202:
@chickenlj I have checked this code, and for your two questions:
1. When identityhashcode changes, ConsistentHashSelector will be rebuilt, and the hash circle will also  be rebuilt. So I don't understand this question too.    : (
2. At this occasion, we need List<Invoker> 's hashcode, not single Invoker, so If we add hashcode() for Invoker, we also need to calculate List<Invoker> 's hashcode. Feel unnecessary.

@kirayunfei Your PR has some code style problems, we suggest use 4 spaces instead of one tab. So please format your code. Thanks.

diecui1202:
Another question, for changelog 3:
> I want to replace the reference in virtualInvokers by the invoke address. It seems to be my personal style, I don't hope a tool like ConsistentHashLoadBalance to catch the reference of Invoker which is important in the whole flow

After discussing with @chickenlj, we think this way will take more O(n) cost for each invoke, so, this enhancement is unnecessary.

dubbo-cluster/src/test/java/com/alibaba/dubbo/rpc/cluster/loadbalance/LoadBalanceTest.java
```diff
@@ -200,5 +200,33 @@ public void testLoadBalanceWarmup() {
         Assert.assertEquals(100, AbstractLoadBalance
             .calculateWarmupWeight(20 * 60 * 1000, Constants.DEFAULT_WARMUP, Constants.DEFAULT_WEIGHT));
     }
+    
+    public Invocation createInvocation(int i) {
+	    	Invocation invocation = EasyMock.createMock(Invocation.class);
+	    	EasyMock.expect(invocation.getMethodName()).andReturn("method").anyTimes();
```
diecui1202: Could you post a new PR while our test framework has changed to Mockito from EasyMock ?

chickenlj:
I have fileda new PR #1872 to solve the generic invoke problem with ConsitenthashLoadbalance.

But for the other problems, i think we still can't reach an agreement. 
And for my own opinion, now i don't think the left two in this PR are real problems. Welcome start a new discussion if you think it's necessary.

Issue #1142, opened 2017-12-29
Annotation configuration don't take effect.


lilanlong:
dubbo注解开发，@Service这个标签设置重试次数为0，但是在断点调试时，还会重试3次

chickenlj:
This issue reported the same problem, #1085
Can you post your detailed configuration here? I tried but can't reproduce your problem.

chickenlj:
see #1085

Issue #1085, opened 2017-12-22
retries property in @Service don't take effect when i set `retries=0`


brainhunt:
2.5.8中，使用@Service注解，配置retries为0时，根据文档描述将不重试。实际测试中，设置为0后相当于retries=2的结果。设置-1时不重试。

chickenlj:
I have tried and it works fine. Maybe there's something else wrong in your project, you can post your detailed configuration here.

TwoWaterLee:
I've located this bug, I'll pull a request.

chickenlj:
@brainhunt If you don't want retry, i recommend use `failfast` policy instead.
Besides, i think `retries=-1` means don't retry is also acceptable, since failover means retry, we can not figure out `retries=0` is a default value or user explicitly set 0.

TwoWaterLee:
@chickenlj Yes, you are right. When use the `failover` policy, retries must be more than one.
Thanks a lot.

PR #1148, opened 2018-01-02
[Dubbo-#1085]Change the DEFAULT_RETRIES from 2 to 0.

TwoWaterLee:

Fix the #1085 and #1142 

## Brief changelog

Change the DEFAULT_RETRIES from 2 to 0.

## Reason
When  retreis property isn't setup in both Reference and Service, the URL will have no retreis property.
So below code been executed, the Constans.DEFAULT_RETRIES will be used.

``` 
// file: FailoverClusterInvoker.java  line: 56
int len = getUrl().getMethodParameter(invocation.getMethodName(), Constants.RETRIES_KEY, Constants.DEFAULT_RETRIES) + 1;
```
So if we setup the retreis property of Refercence or Service, even set it's equal 0, the URL will have retries property, the Constans.DEFAULT_RETRIES will not be used.

### DIFF
--- a/dubbo-common/src/main/java/com/alibaba/dubbo/common/Constants.java
+++ b/dubbo-common/src/main/java/com/alibaba/dubbo/common/Constants.java
@@ -133,7 +133,7 @@ public class Constants {
 
     public static final int DEFAULT_CONNECT_TIMEOUT = 3000;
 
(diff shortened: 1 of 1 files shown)


chickenlj:
I don't think we should change the `DEFAULT_RETRIES ` value.
I give my reason here  #1085

PR #1189, opened 2018-01-09
Improve/improve qos code

qinliujie:
- refactor qos module,before implement is so ugly
 - change qos configuration to system property and application config
 - use wrapper to start qos server


dubbo-plugin/dubbo-qos/src/main/java/com/alibaba/dubbo/qos/protocol/QosProtocolWrapper.java
```diff
@@ -0,0 +1,95 @@
...
+        protocol.destroy();
+    }
+
+    private void startQosServer(URL url) {
+        if (!hasStarted.compareAndSet(false, true)) {
+            return;
+        }
+
+        try {
+            boolean qosEnable = Boolean.parseBoolean(url.getParameter(QOS_ENABLE,"true"));
```
nobodyiam: Why is this option default to true? Now if we start a dubbo client and a dubbo server in one machine, there is always an error message occurred...

ERROR [com.alibaba.dubbo.qos.server.Server]  [DUBBO] qos-server can not bind localhost:22222, dubbo version: 2.0.1, current host:
java.net.BindException: Address already in use

EdwardLee03:
How `DubboLogo` generate? @qinliujie

PR #1204, opened 2018-01-12
[Dubbo-1201] Fix condition router

haodut:

Fix condition router to process default.xxx vaule in url.

## Brief changelog

Modify test case of conditon router, add default.xxx=yyy parameter in url, add route codition xxx=yyy.
Modify condition router. Try to get default.xxx value when xxx is null.

## Verifying this change

Start two providers, one with configuration <dubbo:provider xxx="yyy" />.
Add condition router  "=> xxx=yyy".
Start one consumer, all request should send to provide whit  configuration <dubbo:provider xxx="yyy" />.


chickenlj:
I think this change is quite reasonable. Anyone has other opinion?

Issue #1141, opened 2017-12-28
change the prefixes of properties when @EnableDubboConfig binds multiple Config Bean 


mercyblitz:
Since Dubbo `2.5.8` , `@EnableDubboConfig` supports binds multiple Config Bean  when its' `multiple()` equals `true` , however,  the prefixes of properties as below ：
* `dubbo.application`
* `dubbo.module`
* `dubbo.registry`
* `dubbo.protocol`
* `dubbo.monitor`
* `dubbo.provider`
* `dubbo.consumer`

According to English convention , those prefixes should be plural noun ， thus they are required to be changed :

* `dubbo.applications`
* `dubbo.modules`
* `dubbo.registries`
* `dubbo.protocols`
* `dubbo.monitors`
* `dubbo.providers`
* `dubbo.consumers`

Issue #1122, opened 2017-12-26
DubboConfigConfiguration's Single and Multiple to be public classes


mercyblitz:
`DubboConfigConfiguration.Single` and `DubboConfigConfiguration.Multiple` to be public classes

mercyblitz:
Their will exposed in  2.5.9

Issue #1121, opened 2017-12-26
ReferenceBean should be exposable for production-ready features


mercyblitz:
for [production-ready features](https://github.com/dubbo/dubbo-spring-boot-project/issues/330)

Issue #1120, opened 2017-12-26
@DubboComponentScan is not compatible Spring @Service completely


mercyblitz:
If Service bean annotated Dubbo `@Service` and Spring `@Service` and `@ComponentScan`  executes earlier than `@DubboComponentScan` , current been will not be a candidate of `@Service` Bean.

`@DubboComponentScan` should consider this scenario:

There are two classes : `A`  and `B` under packages `com.acme` : 
A with Dubbo `@Service
B with Spring `@Service` and Dubbo `@Service`

`@ComponentScan` scans this package first, and B will be a candidate  of `@Service` Bean , and then `@DubboComponentScan` also scans same package , just A will be exported as Dubbo Remote interface , B will be missing.

PR #1205, opened 2018-01-12
2.5.9 Annotation Enhancement and Bugfix

mercyblitz:

Annotation Enhancement and bugfix

### Bugfix

* `@DubboComponentScan` is not compatible Spring @Service completely #1120

### Enhancement

* `ReferenceBean` should be exposable for production-ready features #1121
* `DubboConfigConfiguration`'s Single and Multiple to be public classes #1122

### Break Change

* change the prefixes of properties when @EnableDubboConfig binds multiple Config Bean #1141

## Verifying this change

Run use-tests


chickenlj:
@mercyblitz  Would you please revert the dubbo version? There are too many irrelevant changes that bothers when reviewing.

mercyblitz:
@chickenlj Done

PR #1237, opened 2018-01-18
fix: service config listener check should check ExporterListener

plan454:

<dubbo:service listener="xxx" />  , AbstractServiceConfig should check ExporterListener ,not InvokerListener


beiwei30:
this issue has been addressed by 04c93e8b98250230c4d22423372acd1c29e2c770, pls. check.

PR #1258, opened 2018-01-23
compatible:compatible with dubbox

qinliujie:

compatible with dubbox,because dubbox had change dubbo protocol, we compatible with dubbox,so dubbox user can update their dubbo version

## Brief changelog

judge dubbo version to decide whether is dubbox or not

### DIFF
--- a/dubbo-common/src/main/java/com/alibaba/dubbo/common/utils/DubboxUtil.java
+++ b/dubbo-common/src/main/java/com/alibaba/dubbo/common/utils/DubboxUtil.java
@@ -1,0 +1,29 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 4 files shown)


beiwei30:
should we really need to consider the compatibility to dubbox? is it possible to provide this ability with dubbo's extension mechanism, instead of invading into the core?

chickenlj:
@beiwei30 The original dubbox has changed the core dubbo RPC protocol, so it's impossible now to achieve compatibility with dubbox with extension. 
For users using dubbox to transfer smoothly to dubbo, I think we can create a separate branch and a separate release, for example `2.5.x-dubbox-compatible`. This way, the compatible code would not merged into master branch, it will only exists as a transitional code.

kimmking:
you can use this one:
https://github.com/kimmking/dubbo

1. compatible with dubbox
2. updating newer dependencies
3. integrating with swagger ui

zonghaishang:
@kimmking 
https://github.com/kimmking/dubbo， website is inaccessible.

---------------------
@chickenlj
I think the dubbox code merged into the dubbo trunk was a bad decision. For partial changes to the dubbox protocol, we want to find out whether the repair bug is a private extension.

ralf0131:
1. should be merged into master branch.
2. Since dubbox has been merged into dubbo, is this still something we need to consider?

PR #1264, opened 2018-01-24
add config for travis ci fail when test case fail

lovepoem:

add config for travis ci fail when test case fail

## Brief changelog

update .travis.yml

## Verifying this change

When one unit test case fail, travis CI should get red (fail)


lovepoem:
After the test , the new config is ok

PR #1291, opened 2018-01-29
[Dubbo-1279]Allow @Reference to support custom annotations

charpty:
Use custom annotations we can rename the annotation as more obvious name and add many default configs.
So I think support custom annotations is necessary.

Signed-off-by: Bo Cai charpty@gmail.com
请您查看: https://github.com/alibaba/dubbo/issues/1279


charpty:
image
上述图片是对@Service注解的处理示例, 想在@Reference上做类似处理时发现了不能注解在自定义注解上

charpty:
@chickenlj hi~
Nice suggestion. I have modified that. Please check it.

PR #1315, opened 2018-02-03
Collection.isEmpty() should be used to test for emptiness

igor-suhorukov:
Polish: use Collection.isEmpty()

## Brief changelog

Using Collection.size() to test for emptiness works, but using Collection.isEmpty() makes the code more readable and can be more performant. The time complexity of any isEmpty() method implementation should be O(1) whereas some implementations of size() can be O(n).

## Verifying this change

XXXX

- [ ] Make sure there is a [GITHUB_issue](https://github.com/alibaba/dubbo/issues) filed for the change (usually before you start working on it). Trivial changes like typos do not require a GITHUB issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue. 
- [ ] Format the pull request title like `[Dubbo-XXX] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
- [ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/alibaba/dubbo/tree/master/dubbo-test).
- [x] Run `mvn clean install -DskipITs` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.
- [x] If this contribution is large, please file an [Apache Individual Contributor License Agreement](http://www.apache.org/licenses/#clas).

### DIFF
--- a/dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/directory/AbstractDirectory.java
+++ b/dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/directory/AbstractDirectory.java
@@ -72,7 +72,7 @@ public abstract class AbstractDirectory<T> implements Directory<T> {
         }
         List<Invoker<T>> invokers = doList(invocation);
         List<Router> localRouters = this.routers; // local reference
-        if (localRouters != null && localRouters.size() > 0) {
+        if (localRouters != null && !localRouters.isEmpty()) {
             for (Router router : localRouters) {
                 try {
                     if (router.getUrl() == null || router.getUrl().getParameter(Constants.RUNTIME_KEY, false)) {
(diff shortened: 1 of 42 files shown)


chickenlj:
That's a good point, i will have a review on your changes later.

ralf0131:
Hi, welcome to the Dubbo community and great works!

Few comments here:

> Using Collection.size() to test for emptiness works, but using Collection.isEmpty() makes the code more readable and can be more performant. 

+ 1 to the readability. 

> The time complexity of any isEmpty() method implementation should be O(1) whereas some implementations of size() can be O(n).

I am not convinced by the conclusion. In my assumption, testing isEmpty() should be no slower that testing size(), because we need the actual size of the collection before we know whether it is empty. I've checked at least ArrayList, LinkedList, CopyOnWriteList, HashMap, HashSet, TreeMap, ConcurrentHashMap, and there is is no violation. Could you give an example for that?

igor-suhorukov:
@ralf0131 as an example ConcurrentLinkedQueue size complexity O(n), ConcurrentSkipListSet under the hood use ConcurrentSkipListMap with size complexity O(n)

ralf0131:
@igor-suhorukov Thanks for your clarification! That make sense to me.

beiwei30:
thanks for your PR, the change looks good.

PR #1340, opened 2018-02-07
Do not merge attachments when set attachments in RpcResult (#1166)

### LINKED ISSUE #1166: Unify the behavior of setAttachments method in RpcInvocation and RpcResult
RpcInvocation 中的 setAttachments 是设置新的 Attachments，是替换，RpcInvocation 中的 setAttachments 是添加新的 Attachments 到原 Attachments中，是追加，这两处应该保持一致。

liketic:
Replace attachments in setAttachments of RpcResult, instead of merging the new attachments and old attachments. 

Closes #1166

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcResult.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcResult.java
@@ -95,10 +95,8 @@ public class RpcResult implements Result, Serializable {
         return attachments;
     }
 
-    public void setAttachments(Map<String, String> map) {
-        if (map != null && map.size() > 0) {
-            attachments.putAll(map);
-        }
+    public void setAttachments(Map<String, String> attachments) {
+        this.attachments = attachments == null ? new HashMap<String, String>() : attachments;
     }
 
     public String getAttachment(String key) {


dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcResult.java
```diff
@@ -95,10 +95,8 @@ public boolean hasException() {
         return attachments;
     }
 
-    public void setAttachments(Map<String, String> map) {
-        if (map != null && map.size() > 0) {
-            attachments.putAll(map);
-        }
+    public void setAttachments(Map<String, String> attachments) {
+        this.attachments = attachments == null ? new HashMap<String, String>() : attachments;
```
kimmking: I dont suggest this modified code,

kimmking: getAttachments() can get this map, and removeAll it to achieve this replacement and these modified codes:
1、change an empty map when param is null
2、modify the meanings of  setAttachments

chickenlj: setAttachments() is not referenced by other codes currently, so the cost of the change is relatively low. Should we change `setAttachments()` and add `addAttachements()` so that we can keep consistent with that in RpcInvocation?

beiwei30: I think the original impl has no problem, it is just a convenient API to add attachment in batch. I will close this PR and add the following javadoc upon this method:

```java
    /**
     * Append all items from the map into the attachment, if map is empty then nothing happens
     *
     * @param map contains all key-value pairs to append
     */
    public void setAttachments(Map<String, String> map) {
        if (map != null && map.size() > 0) {
            attachments.putAll(map);
        }
    }
```

beiwei30:
9c2e2fa6d3c4e5893deb930bffa469993c2161a6

PR #1351, opened 2018-02-09
remove unuse code of RpcContext

zhaojigang:

remove unuse code of RpcContext

## Brief changelog

 https://github.com/alibaba/dubbo/issues/1346

## Verifying this change

 remove unuse code of RpcContext

[GITHUB_issue](https://github.com/alibaba/dubbo/issues/1346)

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcContext.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcContext.java
@@ -610,51 +610,24 @@ public class RpcContext {
     @SuppressWarnings("unchecked")
     public <T> Future<T> asyncCall(Callable<T> callable) {
         try {
-            try {
-                setAttachment(Constants.ASYNC_KEY, Boolean.TRUE.toString());
-                final T o = callable.call();
-                //local invoke will return directly
-                if (o != null) {
-                    FutureTask<T> f = new FutureTask<T>(new Callable<T>() {
-                        public T call() throws Exception {
-                            return o;
-                        }
-                    });
-                    f.run();
-                    return f;
-                } else {
-
-                }
-            } catch (Exception e) {
-                throw new RpcException(e);
-            } finally {
-                removeAttachment(Constants.ASYNC_KEY);
+            setAttachment(Constants.ASYNC_KEY, Boolean.TRUE.toString());
+            final T o = callable.call();
+            //local invoke will return directly
+            if (o != null) {
+                FutureTask<T> f = new FutureTask<T>(new Callable<T>() {
+                    public T call() throws Exception {
+                        return o;
+                    }
+                });
+                f.run();
+                return f;
             }
-        } catch (final RpcException e) {
-            return new Future<T>() {
-                public boolean cancel(boolean mayInterruptIfRunning) {
-                    return false;
-                }
-
-                public boolean isCancelled() {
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcContext.java
```diff
@@ -610,51 +610,24 @@ public RpcContext setInvocation(Invocation invocation) {
...
-                if (o != null) {
-                    FutureTask<T> f = new FutureTask<T>(new Callable<T>() {
-                        public T call() throws Exception {
-                            return o;
-                        }
-                    });
-                    f.run();
-                    return f;
-                } else {
-
```
chickenlj: What if `o == null` is true? Shouldn't we return `RpcContext.getFuture()`?

dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcContext.java
```diff
@@ -610,51 +610,24 @@ public RpcContext setInvocation(Invocation invocation) {
...
+            if (o != null) {
+                FutureTask<T> f = new FutureTask<T>(new Callable<T>() {
+                    public T call() throws Exception {
+                        return o;
+                    }
+                });
+                f.run();
+                return f;
             }
-        } catch (final RpcException e) {
```
chickenlj: I think we should keep this catch clause, so we can pretend to be a real async call and throw exception with `future.get`

dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcContext.java
zhaojigang: image
从当前master的dai'ma'lai代码来看，如果o==null，只会走一个空的else块，而第二个catch块是应该是永远无法走到的吧

zhaojigang: 奥，现在明白了，第二个catch用来捕捉第一个catch抛出的异常，那是否可以直接将第二个catch体内的逻辑直接放在第一个catch内？

chickenlj: Don't you think we should return a future for the `else clause`?

kimmking: yes, you can wrap a Future for ex in the inner catch.

dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcContext.java
kimmking: can we make a testcase to trigger the second catch statement?

chickenlj: Let's wait for @zhaojigang's reply, i think it would be great for himself to add these changes.

chickenlj:
I think there's no need to change the code. 
I will add a UT and close this PR now

PR #1378, opened 2018-02-20
Simplify string concatenation in log

liketic:
We can simplify string concatenation in log without ```StringBuilder``` cause a single naive string concatenation cost is also cheap.

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/com/alibaba/dubbo/remoting/transport/DecodeHandler.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/com/alibaba/dubbo/remoting/transport/DecodeHandler.java
@@ -55,16 +55,11 @@ public class DecodeHandler extends AbstractChannelHandlerDelegate {
             try {
                 ((Decodeable) message).decode();
                 if (log.isDebugEnabled()) {
-                    log.debug(new StringBuilder(32).append("Decode decodeable message ")
-                            .append(message.getClass().getName()).toString());
+                    log.debug("Decode decodeable message " + message.getClass().getName());
                 }
             } catch (Throwable e) {
                 if (log.isWarnEnabled()) {
-                    log.warn(
-                            new StringBuilder(32)
-                                    .append("Call Decodeable.decode failed: ")
-                                    .append(e.getMessage()).toString(),
-                            e);
+                    log.warn("Call Decodeable.decode failed: " + e.getMessage(), e);
                 }
             } // ~ end of catch
         } // ~ end of if


dubbo-remoting/dubbo-remoting-api/src/main/java/com/alibaba/dubbo/remoting/transport/DecodeHandler.java
```diff
@@ -55,16 +55,11 @@ private void decode(Object message) {
...
+                    log.debug("Decode decodeable message " + message.getClass().getName());
                 }
             } catch (Throwable e) {
                 if (log.isWarnEnabled()) {
-                    log.warn(
-                            new StringBuilder(32)
-                                    .append("Call Decodeable.decode failed: ")
-                                    .append(e.getMessage()).toString(),
-                            e);
+                    log.warn("Call Decodeable.decode failed: " + e.getMessage(), e);
```
kimmking: the capacity of StringBuilder in modified statement is 16, not equals with 32 in the original one

liketic: Thanks @kimmking , what do you think about this change?

kimmking: remain original codes

chickenlj: I think this is a good optimization, let's wait for other suggestions before merge.

liketic: Thanks @chickenlj . For those cases, the naive summation will be much faster than ```StringBuilder```. I made a simple test:
```

        for (int t = 0; t < 10; t++) {
            long st = System.currentTimeMillis();

            for (int i = 0; i < 100000; i++) {
                String s = "Decode decodeable message " + "xxx";
            }

            long e = System.currentTimeMillis();
            System.out.print(e - st);
            System.out.print(" ");
            
            st = System.currentTimeMillis();

            for (int i = 0; i < 100000; i++) {
                new StringBuilder(32).append("Decode decodeable message ").append("xxx").toString();
            }
            e = System.currentTimeMillis();
            System.out.println(e - st);
        }
    }
``` and I got output:
```
1 13
0 11
0 4
0 3
0 4
0 4
0 4
0 1
0 1
0 1
```.

beiwei30: this is not an equivalent comparison. the compiler may optimize `"Decode decodeable message " + "xxx"` into `"Decode decodeable message xxx"` directly.

liketic: Thanks @beiwei30 ! That's my mistake. The test is uncorrect. The two strings cannot both be constants. 👍

PR #1396, opened 2018-02-25
[Dubbo-#1362] cache provider always lru cache

ma-xiao-guang-64:
fix [#1362](https://github.com/alibaba/dubbo/issues/1362)

### DIFF
--- a/dubbo-filter/dubbo-filter-cache/src/main/java/com/alibaba/dubbo/cache/support/AbstractCacheFactory.java
+++ b/dubbo-filter/dubbo-filter-cache/src/main/java/com/alibaba/dubbo/cache/support/AbstractCacheFactory.java
@@ -18,6 +18,7 @@ package com.alibaba.dubbo.cache.support;
 
 import com.alibaba.dubbo.cache.Cache;
 import com.alibaba.dubbo.cache.CacheFactory;
+import com.alibaba.dubbo.common.Constants;
 import com.alibaba.dubbo.common.URL;
 import com.alibaba.dubbo.rpc.Invocation;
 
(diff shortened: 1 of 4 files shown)


lovepoem:
Please async the origin/master branch to resolve the problem of CI check failure

ma-xiao-guang-64:
@lovepoem All checks have passed, please review it

dubbo-filter/dubbo-filter-cache/src/main/java/com/alibaba/dubbo/cache/support/AbstractCacheFactory.java
```diff
@@ -30,7 +32,8 @@
 
     private final ConcurrentMap<String, Cache> caches = new ConcurrentHashMap<String, Cache>();
 
-    public Cache getCache(URL url) {
+    public Cache getCache(URL url, Invocation invocation) {
+        url = url.addParameter(Constants.METHOD_KEY, invocation.getMethodName());
```
kimmking: can you attach a testcase for these modified codes, thx.

ma-xiao-guang-64: ok, i need some help because i don't know how to test.
l debug in idea run ok.

kimmking: you can mock an invocation and cacheFactory,
then assert the result hitting your cache.

ma-xiao-guang-64: how to confirm use annother cache not lru

ma-xiao-guang-64:
@kimmking   l add some test for cache, please review it thanks

dubbo-filter/dubbo-filter-cache/src/main/java/com/alibaba/dubbo/cache/support/AbstractCacheFactory.java
beiwei30: The change looks good, except for this:

```java
url = url.addParameter(Constants.METHOD_KEY, invocation.getMethodName());
```
I don't think we need addParameter for method name in getCache() method

Issue #1399, opened 2018-02-26
Number type config data is lost in yaml file


daocr:
spring-boot-version : 1.5.9.RELEASE
dubbo.version : 2.6.0

bug 出现场景：
1、使用了@EnableDubbo 注解
2、配置文件使用的是 “.yml” 格式的文件 
![image](https://user-images.githubusercontent.com/28719772/36659592-edbbc7d4-1b0f-11e8-8dbf-d938ed49a840.png)

bug出现原因：
![image](https://user-images.githubusercontent.com/28719772/36659746-6ecd6ef4-1b10-11e8-87b5-942cb4ee1af1.png)

在yml 文件中，值是有数据类型的区分的，如果直接这样判断，会导致有些值被忽略

ralf0131:
Hi, welcome to Dubbo community!

Please consider using English, Dubbo is a global community, and all the issues may be viewed and discussed by people all over the world.

Meanwhile, you can send email to dev@dubbo.apache.org or subscribe it, which is preferred way for discussing questions like this.

lovepoem:
Fixed

PR #1401, opened 2018-02-26
fix number type is lost in yaml config file

lovepoem:

#1399  fix number type is lost in yaml config file

## Brief changelog

Remove String type filter when read yaml file data

## Verifying this change
Run referenced unit test


luffyke:
@lovepoem Release 2.5.10 has not this fix, may i know which release would include this fix? as there are many issues about this fix in dubbo-spring-boot-starter project.

lovepoem:
https://github.com/alibaba/dubbo/milestone/13  -->v2.6.1
https://github.com/alibaba/dubbo/milestone/16  -->v2.5.11
Now the two milestones contains this patch, I think v2.6.1 will be published soon.

lovepoem:
Now https://github.com/alibaba/dubbo/milestone/13  is released.

PR #1407, opened 2018-02-28
[Dubbo-1510] dubbo monitor missing parameters

### LINKED ISSUE #1510: dubbo monitor missing parameters
When collect dubbo invoke data, these parameters( `version`,`group`) are missing.
After check monitor module, it will allow user custom monitor interval,  but no place to override the default value (60s).

YoungHu:
fix #1510 

1.  dubbo monitor collection missing the `group` `version` `dubbo` parameters
2. dubbo monitor interval default set to 60s and can't override by outside setting

## Brief changelog

1. dubbo monitor parameters add `group`, `version`,`dubbo` parameters
2. monitor interval can be override by outside setting

## Verifying this change

XXXX

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/MonitorConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/MonitorConfig.java
@@ -41,6 +41,8 @@ public class MonitorConfig extends AbstractConfig {
 
     private String version;
 
+    private String interval;
+
     // customized parameters
     private Map<String, String> parameters;
 
@@ -121,6 +123,14 @@ public class MonitorConfig extends AbstractConfig {
 
     public void setDefault(Boolean isDefault) {
         this.isDefault = isDefault;
+    }
+
+    public void setInterval(String interval){
+        this.interval = interval;
+    }
+
+    public String getInterval(){
+        return interval;
     }
 
 }
--- a/dubbo-monitor/dubbo-monitor-default/src/main/java/com/alibaba/dubbo/monitor/dubbo/DubboMonitor.java
+++ b/dubbo-monitor/dubbo-monitor-default/src/main/java/com/alibaba/dubbo/monitor/dubbo/DubboMonitor.java
@@ -16,6 +16,7 @@
  */
 package com.alibaba.dubbo.monitor.dubbo;
 
+import com.alibaba.dubbo.common.Constants;
 import com.alibaba.dubbo.common.URL;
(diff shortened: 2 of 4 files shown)


dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/MonitorConfig.java
```diff
@@ -123,4 +125,12 @@ public void setDefault(Boolean isDefault) {
         this.isDefault = isDefault;
     }
 
+    public void setInterval(String interval){
+        this.interval = interval;
+    }
+
+    public String getInterval(){
```
diecui1202: Who calls `getInterval()`?

dubbo-monitor/dubbo-monitor-default/src/main/java/com/alibaba/dubbo/monitor/dubbo/DubboMonitor.java
```diff
@@ -92,6 +93,7 @@ public void send() {
             long maxOutput = numbers[7];
             long maxElapsed = numbers[8];
             long maxConcurrent = numbers[9];
+            String version = getUrl().getParameter(Constants.DEFAULT_PROTOCOL);
```
diecui1202: IMO, this variable is named `protocol` may be better.

YoungHu: `DEFAULT_PROTOCOL` value is `dubbo`, this code is to get dubbo version but not protocol name. dubbo now upgrade fast, for a team, different project may use different dubbo version(eg 2.5.3, 2.6.0, 2.6.1), IMO, monitor need to know each project's dubbo version.

dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/MonitorConfig.java
YoungHu: no one call `getInterval()` direct, this is used for fastjson serialize, so it can be passed to `URL`

dubbo-monitor/dubbo-monitor-default/src/main/java/com/alibaba/dubbo/monitor/dubbo/DubboMonitor.java
beiwei30: sounds a reasonable data to collect, let me think about it more.

Issue #1411, opened 2018-02-28
Locale deserialize 'zh-hant_CN'


takeseem:
`zh-hant_CN` will convert to `zh_HANT_cn`

Locale include (language, country, variant), Locale.toString() use '\_' to split then `{language}_{country}_({variant}_# | #){script}-{extensions}`

but Jackson deserialize Locale use '-' or '\_'  : (
code: [LocaleHandle.readResolve()](https://github.com/alibaba/dubbo/blob/master/hessian-lite/src/main/java/com/alibaba/com/caucho/hessian/io/LocaleHandle.java#L63)

[iana language-tags](https://www.iana.org/assignments/language-tags/language-tags.xml#Mark_Davis)
```
zh-Hant	Chinese, in traditional script	[Mark_Davis][Mark_Davis_2]	
zh-Hant-CN	PRC Mainland Chinese in traditional script	[Mark_Davis][Mark_Davis_2]	
zh-Hant-HK	Hong Kong Chinese in traditional script	[Mark_Davis][Mark_Davis_2]	
zh-Hant-MO	Macao Chinese in traditional script	[Mark_Davis][Mark_Davis_2]	
zh-Hant-SG	Singapore Chinese in traditional script	[Mark_Davis][Mark_Davis_2]	
zh-Hant-TW	Taiwan Chinese in traditional script	[Mark_Davis][Mark_Davis_2]
```

[rfc4647 language lookup](https://tools.ietf.org/html/rfc4647#section-3.4)
```
  Example of a Lookup Fallback Pattern

   Range to match: zh-Hant-CN-x-private1-private2
   1. zh-Hant-CN-x-private1-private2
   2. zh-Hant-CN-x-private1
   3. zh-Hant-CN
   4. zh-Hant
   5. zh
   6. (default)
```

```java
	public static void main(String[] args) {
		System.out.println(new Locale("zh-hant", "CN"));
	}
```
in jdk1.5 -> 1.8
```
jdk1.5	zh-hant_CN
jdk1.6	zh-hant_CN
jdk1.7	zh-hant_CN
jdk1.8	zh-hant_CN
```

beiwei30:
one more comments: if user starts to complain the inter-op compatibility between old dubbo version and current version, we may need to consider to rollback change 0423219, or consider to introduce a system property to control the behavior.

takeseem:
[...]

PR #1412, opened 2018-02-28
Locale deserialize for java {language}_{country}_({variant}_#| #){script}-{extensions}

takeseem:
see #1411


beiwei30:
I use an alternative approach to fix this issue, pls. check 0423219

PR #1430, opened 2018-03-07
Simplify check consumer side in RpcContext

whanice:
```
url.getPort() != address.getPort() ||
                !NetUtils.filterLocalHost(url.getIp()).equals(NetUtils.filterLocalHost(host));
```
There is no need to compare address to confirm whether they are consumer or not in RpcContext.
the remote address come from url and compare with url.

## Brief changelog
RpcContext


lovepoem:
Please synchronize the origin/master branch to resolve the problem of CI check failure

PR #1438, opened 2018-03-09
[Dubbo- support unit router feature] May be a nice feature.

zonghaishang:

Support unit router feature for dubbo.

What problems can this feature solve?

case 1:
  If we have two instance of services or more, we call it serviceA(192.168.0.1), serviceA(192.168.0.2)
 
 image

We want to invoke serviceA (192.168.0.2) first, if serviceA (192.168.0.2) is not available then invoke  serviceA (192.168.0.1) automatically.

case 2:

If we have many relative services, We call it serviceA(192.168.0.1), serviceA(192.168.0.2), serviceB(192.168.1.1), serviceB(192.168.1.2), serviceC(192.168.3.1), serviceC(192.168.3.2)

dependencies of service: 
serviceA -> serviceB -> serviceC

image

We can flexibly build call paths and automatically downgrade calls.

case 3:

We can take advantage of this feature to realize the priority routing of the same machine room service.
 
Notes:

if use unit feature, unit router will be last invoked, 
because if the unit router has no match, it will return the previous available invokers, this is very important !

Easy to use:

 Set service and reference are same unit value. or use global configuration.
 case 1：
 consumer.xml:  `<dubbo:reference id="XXX" check="false" interface="XXX" unit="anything"/>`
 provider.xml:     `<dubbo:service interface="XXX" ref="bean" unit="same as reference"/>`

 case 2：(global)
 consumer.xml & provider.xml : `<dubbo:provider unit="anything"/>`， "anything"  suggest same both side.

## Brief changelog

1. add unit property for dubbo.xsd
2. add new UnitRouter and so on

## Verifying this change

Here is more detail unit tests:
1. com.alibaba.dubbo.rpc.cluster.router.unit.UnitRouterTest
2. com.alibaba.dubbo.registry.dubbo.RegistryDirectoryTest

I have verified that the unit test has been correct and passed.

- [ ] Make sure there is a [GITHUB_issue](https://github.com/alibaba/dubbo/issues) filed for the change (usually before you start working on it). Trivial changes like typos do not require a GITHUB issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue. 
- [x] Format the pull request title like `[Dubbo-XXX] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
- [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/alibaba/dubbo/tree/master/dubbo-test).
- [x] Run `mvn clean install -DskipITs` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.
- [ ] If this contribution is large, please file an [Apache Individual Contributor License Agreement](http://www.apache.org/licenses/#clas).


chickenlj:
If this PR only support specify a specific unit at startup, I think that makes no difference with Dubbo's group.

I think we should consider the following aspects as a whole:
* The initial call decides the whole invoke chain automatically.
* We may need some support from third-party infrastructures, such as systems that can group instances automatically based on instance IP.
* Support change group/unit of an instance dynamically at runtime.
* Support adds routing rules dynamically at runtime.

chickenlj:
I will close this PR temporarily, and will reopen it or create a new ticket to track this feature later.

PR #1444, opened 2018-03-09
[Dubbo-RpcContext] Simplify check consumer side in RpcContext

whanice:
```
url.getPort() != address.getPort() ||
                !NetUtils.filterLocalHost(url.getIp()).equals(NetUtils.filterLocalHost(host));
```
There is no need to compare address to confirm whether they are consumer or not in RpcContext.
the remote address comes from url and compare with url.

## Brief changelog
RpcContext

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcContext.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcContext.java
@@ -154,22 +154,7 @@ public class RpcContext {
      * @return provider side.
      */
     public boolean isProviderSide() {
-        URL url = getUrl();
-        if (url == null) {
-            return false;
-        }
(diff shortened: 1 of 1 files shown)


cvictory:
where is the usage of isProviderSide and isConsumerSide?    
I cannot judge whether it is right.

cvictory:
Usually the commit is right. 

Is there some special situation happened？

whanice:
the isProviderSide and isConsumerSide seem not use in dubbo framework. just a api for user.
and the origin implementation is almost right.
but when we set DUBBO_IP_TO_REGISTRY use hostname not ip , the check  is wrong. 
and maybe  lookup dns to resolve hostname in implementation.

whanice:
It's depends on the side key in new implementation.  we can also see other place in dubbo code to use  side key to check is consumer side or not.

chickenlj:
I think SIDE_KEY is enough to determine provider or consumer.

PR #1506, opened 2018-03-25
Update ServiceConfig.java

bearDream:
replace string by Constants

don't direct use String.

## Brief changelog

replace string by Constants

## Verifying this change

passed

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/ServiceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/ServiceConfig.java
@@ -491,7 +491,7 @@ public class ServiceConfig<T> extends AbstractServiceConfig {
                 }
                 if (registryURLs != null && !registryURLs.isEmpty()) {
                     for (URL registryURL : registryURLs) {
-                        url = url.addParameterIfAbsent("dynamic", registryURL.getParameter("dynamic"));
+                        url = url.addParameterIfAbsent(Constants.DYNAMIC_KEY, registryURL.getParameter("dynamic"));
                         URL monitorUrl = loadMonitor(registryURL);
                         if (monitorUrl != null) {
                             url = url.addParameterAndEncoded(Constants.MONITOR_KEY, monitorUrl.toFullString());
--- a/dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/ServiceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/ServiceConfig.java
@@ -491,7 +491,7 @@ public class ServiceConfig<T> extends AbstractServiceConfig {
                 }
                 if (registryURLs != null && !registryURLs.isEmpty()) {
                     for (URL registryURL : registryURLs) {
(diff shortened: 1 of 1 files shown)


justinmclean:
Hi,
>  <https://cla-assistant.io/apache/incubator-dubbo?pullRequest=1506> 
> Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our Contributor License Agreement <https://cla-assistant.io/apache/incubator-dubbo?pullRequest=1506> before we can accept your contribution.
> 

This needs to be edited so that it points to the Apache CLA not the Alibaba one. Also just because someone signed an ICLA doesn’t mean they can contribute, their employment contract may for instance state that all IP belongs to their company we may need to spell this out in a little more detail.

Thanks,
Justin

ralf0131:
For the CLA issue, I've created a ticket here: https://issues.apache.org/jira/browse/DUBBO-29

bearDream:
@ralf0131 Thank you. I've finished it.

ralf0131:
The CLA-assistant has been removed, further PRs won't need to sign the Alibaba CLA any longer.

bearDream:
@ralf0131 It's mean that I don't neet signed CLA?

ralf0131:
@bearDream Yes, if the contribution is small, there is no need to sign CLA any longer since Dubbo has entered Apache incubator.

If the contribution is large, you may need to sign ICLA[1], or SGA if it is a one time donation.

In any of the case, the Alibaba CLA is no longer needed.

[1] https://www.apache.org/dev/new-committers-guide.html#cla

dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/ServiceConfig.java
```diff
@@ -491,7 +491,7 @@ private void doExportUrlsFor1Protocol(ProtocolConfig protocolConfig, List<URL> r
                 }
                 if (registryURLs != null && !registryURLs.isEmpty()) {
                     for (URL registryURL : registryURLs) {
-                        url = url.addParameterIfAbsent("dynamic", registryURL.getParameter("dynamic"));
+                        url = url.addParameterIfAbsent(Constants.DYNAMIC_KEY, registryURL.getParameter("dynamic"));
```
ralf0131: Why is the string in `registryURL.getParameter("dynamic")` not getting replaced with `Constants.DYNAMIC_KEY`?

bearDream: I think key should be managed by Constants, Instead of writing directly to the code. Like this ,Post-manitenance comfortably.

ralf0131: Yes, I agree with you. My question is, why it is not some sth. like this:

```java
url = url.addParameterIfAbsent(Constants.DYNAMIC_KEY, registryURL.getParameter(Constants.DYNAMIC_KEY));
```

bearDream: emmm... you are right, I will update it.

ralf0131:
Looks good to me! Thanks for the pull request!

PR #1507, opened 2018-03-25
Mirror improvement for test cases in dubbo-rpc-api module

laddcn:
In dubbo-rpc-api module:
1.  Since junit.framework.Assert are deprecated since Junit 4 and other test case classes in this module have already use org.junit.Assert for instead, it is necessary to update related codes in StatItemTest  and RpcUtilsTest.
2.  Make codes more easy to read

## Brief changelog
In dubbo-rpc-api module:
1. Use org.junit.Assert instead of junit.framework.Assert(@Deprecated since Junit 4)
2. Use assertTrue and other assert methods through static import to make codes easier to read

## Verifying this change

Test cases run with success


ralf0131:
LGTM, thanks for the pull request!

PR #1511, opened 2018-03-26
Add feature：require Reference service dynamically

zhouzhipeng:
## require Reference service dynamically

Supports dynamic retrieval of dubbo reference services, facilitating multiple dubbo implementation classes on the same interface. (You can use group to distinguish, eg: dynamically obtain the group name from the configuration to dynamically call the specified group service, see below example code)

## Service Provider Code Example
```java

    @Service(group = "abctest")
    public class TagSearchServiceImpl implements TagSearchService {
    ...
    }


    @Service(group = "abctest2")
    public class TagSearchServiceImpl2 implements TagSearchService {
    ...
    }
```

## Consumer code example
```java
    //inject the util service
    @Autowired
    private DynamicReferenceService dynamicReferenceService;
```

```java

    // Business code, dynamically obtain the specified interface and group implementation class
    RefConf rc=new RefConf();

    rc.group="abctest"; //Or abctest2, suggesting dynamic acquisition

    TagSearchService tagSearchService = dynamicReferenceService.getDubboService(TagSearchService.class,rc);

```

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/RefConf.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/RefConf.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 1999-2012 Alibaba Group.
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *  
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *  
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.alibaba.dubbo.config;
+
+/**
+ * Reference  注解对应的实体配置类
+ * 
+ * @author zhouzhipeng
+ */
+public class RefConf {
+
+   public Class<?> interfaceClass = void.class;
+
+   public String interfaceName = "";
+
+   public String version = "";
+
+   public String group = "";
+
+   public String url = "";
+
+   public String client = "";
+
+   public boolean generic = false;
+
+   public boolean injvm = false;
+
+   public boolean check = true;
+
+   public boolean init = false;
+
+   public boolean lazy = false;
+
+   public boolean stubevent = false;
+
+   public String reconnect = "";
+
+   public boolean sticky = false;
+
+   public String proxy = "";
+
+   public String stub = "";
+
+   public String cluster = "";
+
+   public int connections = 0;
+
+   public int callbacks = 0;
+
+   public String onconnect = "";
+
+   public String ondisconnect = "";
+
+   public String owner = "";
+
+   public String layer = "";
+
+   public int retries = 0;
+
+   public String loadbalance = "";
+
+   public boolean async = false;
+
+   public int actives = 0;
+
+   public boolean sent = false;
+
+   public String mock = "";
+
+   public String validation = "";
+
+   public int timeout = 0;
+
+   public String cache = "";
+
+   public String[] filter = {};
+
+   public String[] listener = {};
+
+   public String[] parameters = {};
+
+   public String application = "";
+
+   public String module = "";
+
+   public String consumer = "";
+
+   public String monitor = "";
+
+   public String protocol = "";
+
+   public String[] registry = {};
+
+}
--- a/dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/AnnotationBean.java
+++ b/dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/AnnotationBean.java
@@ -96,6 +96,14 @@ public class AnnotationBean extends AbstractConfig implements DisposableBean, Be
                 // spring 2.0
             }
         }
+
+
+        //
+        if(!beanFactory.containsBean(DynamicReferenceService.class.getName())) {
+            //注册服务bean
+            beanFactory.registerSingleton(DynamicReferenceService.class.getName(), new DynamicReferenceService(this));
+        }
+
     }
 
     public void destroy() throws Exception {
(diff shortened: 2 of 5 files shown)


dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/AnnotationBean.java
```diff
@@ -288,6 +296,73 @@ private Object refer(Reference reference, Class<?> referenceClass) { //method.ge
...
+        if (! "".equals( reference.interfaceName)) {
+            interfaceName =  reference.interfaceName;
+        } else if (! void.class.equals( reference.interfaceClass)) {
+            interfaceName =  reference.interfaceClass.getName();
+        } else if (referenceClass.isInterface()) {
+            interfaceName = referenceClass.getName();
+        } else {
+            throw new IllegalStateException("The @Reference undefined interfaceClass or interfaceName, and the property type " + referenceClass.getName() + " is not a interface.");
+        }
+        String key =  reference.group + "/" + interfaceName + ":" +  reference.version;
```
zonghaishang: Please see com.alibaba.dubbo.common.URL#getServiceKey , you should check property value if exists.

dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/AnnotationBean.java
```diff
@@ -96,6 +96,14 @@ public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
                 // spring 2.0
             }
         }
+
+
+        //
+        if(!beanFactory.containsBean(DynamicReferenceService.class.getName())) {
+            //注册服务bean
+            beanFactory.registerSingleton(DynamicReferenceService.class.getName(), new DynamicReferenceService(this));
```
zonghaishang: Why do you need to hard code the injected object ？ Do not allow multiple DynamicReferenceServices and different ids ? Or my understanding is wrong?

dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/RefConf.java
```diff
@@ -0,0 +1,107 @@
...
+ * limitations under the License.
+ */
+package com.alibaba.dubbo.config;
+
+/**
+ * Reference  注解对应的实体配置类
+ * 
+ * @author zhouzhipeng
+ */
+public class RefConf {
```
zonghaishang: "RefConf.java" does not meet the dubbo code specification, new configuration object should extend existing objects or reuse, see ReferenceConfig and so on. Current writing will break compatibility, eg: adding an attribute, your object cannot support, must modify again.

dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/AnnotationBean.java
```diff
@@ -288,6 +296,73 @@ private Object refer(Reference reference, Class<?> referenceClass) { //method.ge
...
+    /**
+     * refer 方法的公开版，将注解参数Reference换为 ReferenceConf 类
+     * @param reference
+     * @param referenceClass
+     * @return
+     */
+    public Object referNew(RefConf reference, Class<?> referenceClass) { //method.getParameterTypes()[0]
+        String interfaceName;
+
+        if (! "".equals( reference.interfaceName)) {
```
zonghaishang: Too much duplication of code with "com.alibaba.dubbo.config.spring.AnnotationBean#refer".

dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/AnnotationBean.java
zhouzhipeng: DynamicReferenceService is like a util class. It is the entrance to acquire a Reference service dynamic . like blow: 
    

SomeDubboService service = dynamicReferenceService.getDubboService(SomeDubboService.class,rc);

dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/RefConf.java
zhouzhipeng: "RefConf.java"   is not a full new configuration class. It is the normal  javabean version for "@Reference"   annotation . Because the "@Reference"   annotation  can not be initialized manual .

dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/AnnotationBean.java
zhouzhipeng: yes , for convenience, i just copied the refer method to referNew. I  will fix it by extract a common method.

dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/AnnotationBean.java
zonghaishang: What I am worried about is that the DynamicReferenceService injection is overrided， eg: I have two reference bean, 
``` java
     Consumer A (ReferenceBean #A): 
    // inject service
    @Autowired
    private DynamicReferenceService dynamicReferenceService;

     Consumer B (ReferenceBean #B): 
    // inject service
    @Autowired
    private DynamicReferenceService dynamicReferenceService;
```
Spring container contains "new DynamicReferenceService(ReferenceBean #B)" only, previous new DynamicReferenceService(ReferenceBean #A) instance maybe lost, when you invoke any service method in Consumer A (ReferenceBean #A), you hold DynamicReferenceService(ReferenceBean #B) instance actually, maybe trigger some bugs?

ralf0131:
Hi, welcome!

Would you please update the title and description using English? Dubbo is a global community and we would like the PR to be reviewed by people all over the world.

kimmking:
Unit Tests for a new feature is necessary

dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/RefConf.java
```diff
@@ -0,0 +1,107 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.alibaba.dubbo.config;
+
+/**
+ * Reference  注解对应的实体配置类
```
ralf0131: Hi, would you please update the comments in English?

zhouzhipeng: ok

dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/AnnotationBean.java
zonghaishang: @zhouzhipeng Existing refer methods is private, you can rename refer to refer0,  I think use refer better than referNew.  Of course, it's all depends on you, thank you very much for your contribution.

beiwei30:
I am afraid I cannot accept this change because the following reasons:

1. From dubbo's user doc, the group is designed to distinguish the different impls as you planned to give the support here, see http://dubbo.apache.org/books/dubbo-user-book/demos/service-group.html. That is, you could always look up the desired ReferenceBean with the given id, that means `DynamicReferenceService` is unnecessary

```java
<dubbo:reference id="feedbackIndexService" group="feedback" interface="com.xxx.IndexService" />
<dubbo:reference id="memberIndexService" group="member" interface="com.xxx.IndexService" />
``` 

2. I am not convinced that RefConf should be introduced, since it looks very familiar with com.alibaba.dubbo.config.ReferenceConfig#ReferenceConfig(com.alibaba.dubbo.config.annotation.Reference)
3. com.alibaba.dubbo.config.spring.AnnotationBean is deprecated, we should not add more functionality on this class
4. some minor issues including:
    * missing license header
    * comments in Chinese
    * author info should be removed

Feel free to reopen and let's discuss further if appropriated.

PR #1520, opened 2018-03-28
comments in URL class are incorrect

gavincook:

correct the comments in URL class

## Brief changelog

correct the protocol description for file url

## Verifying this change

correct the protocol description for file url


ralf0131:
Looks good to me! 
Thanks for the PR!

PR #1561, opened 2018-04-07
exclude dubbo-demo when calc coverage

htynkn:

Exclude `dubbo-demo` project when calc coverage

## Brief changelog

Add ignore config for codecov

## Verifying this change

CI Passs


ralf0131:
Looks good to me!

PR #1574, opened 2018-04-09
circuitBreaker

wanhuhou:
Add CircuitBreaker

### DIFF
--- a/all/pom.xml
+++ b/all/pom.xml
@@ -370,6 +370,16 @@
             <scope>test</scope>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>com.alibaba</groupId>
+            <artifactId>dubbo-flowcontrol-api</artifactId>
+            <version>${project.parent.version}</version>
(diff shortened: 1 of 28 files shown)


ralf0131:
1. The CI has failed. 
2. This is a significant change,  you need to sign the SGA before merging.

jvz:
Hello,

Your software grant has been filed. Thanks!

chickenlj:
Hello wanhuhou,

Thanks for your contribution, instead of a circuitbreaker i think your implementation is more like a new policy of LoadBalance. 

And after having a discussion with the dubbo team, i think below is the enhancements we should do:
* A foundamental Metrics module.
* Add a new LoadBalance Policy which can detect node healthy.
* Add a new Cluster Policy which can short-circuit and switch to fallback automically. More like a embedded circuitbreaker.
* Provide a demo of integrating Hystrix with Dubbo.

chickenlj:
#1924 #1923 #1922

PR #1575, opened 2018-04-09
[dubbo-1521]add java 9 in travis yml

### LINKED ISSUE #1521: Add java9 in CI build
Currently dubbo support java6, java7 and java8.

All unit test pass in java9, we should consider add official support for java9 and even java10.
Currently travis support java9. java10 is in progress (https://github.com/travis-ci/travis-ci/issues/9368)

As discussed in mail list, the first step is to add java9 in travis ci.

htynkn:
add java9 in ci

## Brief changelog
use jacoco
fix reflection issue
## Verifying this change
ci pass
coverage in codecov.io

close #1521

### DIFF
--- a/dubbo-common/src/main/java/com/alibaba/dubbo/common/utils/ReflectUtils.java
+++ b/dubbo-common/src/main/java/com/alibaba/dubbo/common/utils/ReflectUtils.java
@@ -910,6 +910,9 @@ public final class ReflectUtils {
                 while (cls != null && cls != Object.class) {
                     Field[] fields = cls.getDeclaredFields();
                     for (Field field : fields) {
+                        if (field.isSynthetic()) {
+                            continue;
+                        }
                         Object property = getEmptyObject(field.getType(), emptyInstances, level + 1);
                         if (property != null) {
                             try {
--- a/pom.xml
+++ b/pom.xml
@@ -111,10 +111,10 @@
         <maven_deploy_version>2.8.2</maven_deploy_version>
         <maven_compiler_version>3.6.0</maven_compiler_version>
         <maven_source_version>3.0.1</maven_source_version>
-        <maven_cobertura_version>2.7</maven_cobertura_version>
         <maven_javadoc_version>3.0.0</maven_javadoc_version>
         <maven_jetty_version>6.1.26</maven_jetty_version>
         <maven_checkstyle_version>3.0.0</maven_checkstyle_version>
+        <maven_jacoco_version>0.8.1</maven_jacoco_version>
     </properties>
 
     <modules>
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -422,16 +421,27 @@
...
-                <configuration>
-                    <formats>
-                        <format>html</format>
-                        <format>xml</format>
-                    </formats>
-                    <check/>
-                </configuration>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <version>${maven_jacoco_version}</version>
```
kimmking: 从 cobertura换成了jacoco了？

lovepoem:
When you have tested the program error or the test case doesn't pass, will the CI process terminate? @htynkn

htynkn:
@lovepoem please refer this build: https://travis-ci.org/diamondblack/incubator-dubbo
If test fail, CI will fail too. logs here: https://api.travis-ci.org/v3/job/364388877/log.txt

lovepoem:
Look good to me

dubbo-common/src/main/java/com/alibaba/dubbo/common/utils/ReflectUtils.java
```diff
@@ -910,6 +910,9 @@ private static Object getEmptyObject(Class<?> returnType, Map<Class<?>, Object>
                 while (cls != null && cls != Object.class) {
                     Field[] fields = cls.getDeclaredFields();
                     for (Field field : fields) {
+                        if (field.isSynthetic()) {
```
beiwei30: is it necessary to include this change?

lovepoem: What's the purpose to add this filter?

htynkn: @lovepoem @beiwei30 
please refer to jacoco faq http://www.jacoco.org/jacoco/trunk/doc/faq.html
screen shot 2018-04-10 at 12 15 57 pm
We need this change to skip $jacocoData, otherwise test will fail.

Also skip synthetic field in reflection is also a good practice. 
Refer to those two links:
+ https://docs.oracle.com/javase/tutorial/reflect/member/fieldModifiers.html
+ https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.7.8

According to the JVM Spec: "A class member that does not appear in the source code must be marked using a Synthetic attribute." Also, "The Synthetic attribute was introduced in JDK release 1.1 to support nested classes and interfaces."

beiwei30: sounds reasonable

kimmking: class is also added synthetic checking?

Issue #1619, opened 2018-04-17
Exception during demo provider start


ralf0131:
When running the provider demo (com.alibaba.dubbo.demo.provider.Provider#main), there is an warn message with an exception stack trace as follows:

```sh
[17/04/18 04:30:35:035 CST] main  INFO config.AbstractConfig:  [DUBBO] The service ready on spring started. service: com.alibaba.dubbo.demo.DemoService, dubbo version: 2.0.0, current host: 30.5.96.22
[17/04/18 04:33:28:028 CST] main  WARN config.AbstractConfig:  [DUBBO] Address family not supported by protocol family (connect failed), dubbo version: 2.0.0, current host: 30.5.96.22
java.net.ConnectException: Address family not supported by protocol family (connect failed)
  at java.base/java.net.PlainSocketImpl.socketConnect(Native Method)
  at java.base/java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:400)
  at java.base/java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:243)
  at java.base/java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:225)
  at java.base/java.net.SocksSocketImpl.connect(SocksSocketImpl.java:402)
  at java.base/java.net.Socket.connect(Socket.java:591)
  at com.alibaba.dubbo.config.ServiceConfig.findConfigedHosts(ServiceConfig.java:577)
  at com.alibaba.dubbo.config.ServiceConfig.doExportUrlsFor1Protocol(ServiceConfig.java:469)
  at com.alibaba.dubbo.config.ServiceConfig.doExportUrls(ServiceConfig.java:357)
  at com.alibaba.dubbo.config.ServiceConfig.doExport(ServiceConfig.java:316)
  at com.alibaba.dubbo.config.ServiceConfig.export(ServiceConfig.java:215)
  at com.alibaba.dubbo.config.spring.ServiceBean.onApplicationEvent(ServiceBean.java:121)
  at com.alibaba.dubbo.config.spring.ServiceBean.onApplicationEvent(ServiceBean.java:50)
  at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:172)
  at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:165)
[...]

PR #1620, opened 2018-04-17
[Dubbo-1619] Fix exception stack trace during provider demo start.

ralf0131:

Fix exception stack trace during provider demo start.

## Brief changelog

When running the provider demo locally, the provider will try to connect to the ip specified in the registry url in order. In the demo code, a multicast address is specified, however, there is no need to connected to the multicast address since it will eventually failed. The failure of connecting to that address logs an warn message with exception trace. Although there is no harm, it will still make user confused.

This pull request just skip that connection if registry address is a multicast address.

## Verifying this change

Verified locally to confirm that the exception trace is no longer logged.

- [x] Make sure there is a [GITHUB_issue](https://github.com/alibaba/dubbo/issues) filed for the change (usually before you start working on it). Trivial changes like typos do not require a GITHUB issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue. 
- [x] Format the pull request title like `[Dubbo-XXX] Fix UnknownException when host config not exist`. Each commit in the pull request should have a meaningful subject line and body.
- [] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/alibaba/dubbo/tree/master/dubbo-test).
- [ ] Run `mvn clean install -DskipITs` to make sure unit-test pass. Run `mvn clean test-compile failsafe:integration-test`  to make sure integration-test pass.
- [ ] If this contribution is large, please file an [Apache Individual Contributor License Agreement](http://www.apache.org/licenses/#clas).

### DIFF
--- a/dubbo-common/src/main/java/com/alibaba/dubbo/common/Constants.java
+++ b/dubbo-common/src/main/java/com/alibaba/dubbo/common/Constants.java
@@ -633,6 +633,8 @@ public class Constants {
 
     public static final boolean DEFAULT_HESSIAN_OVERLOAD_METHOD = false;
 
+    public static final String MULTICAST = "multicast";
+
     /*
      * private Constants(){ }
      */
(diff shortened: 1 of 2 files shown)


ralf0131:
Related issue #1619

ralf0131:
I think this pull request needs to be included in the 2.6.2. The exception stack trace in the demo makes user confused, which IMO need to be fixed as soon as possible.

chickenlj:
Most users would not run into this branch, usually, the framework an get an ip from the hosts mapping in a former process.

This PR brings tiny change, i would merge this branch anyway.

PR #1647, opened 2018-04-21
[DUBBO-9]add apache-rat-plugin to check asf license header

lovepoem:

add apache-rat-plugin to check asf license header for https://issues.apache.org/jira/browse/DUBBO-9

As discussed in the [mail list](https://lists.apache.org/list.html?dev@dubbo.apache.org:2018-3:How%20to%20add%20Apache%20license%20head%20to%20the%20files%20of%20hessian-lite%20sub%20project),the hessian_lite sub-project is  excluded from check 

## Brief changelog

add missed asf license headers

## Verifying this change

CI pass


hessian-lite/src/test/java/com/alibaba/com/caucho/hessian/io/LocaleSerializerTest.java
```diff
@@ -1,3 +1,19 @@
+/*
```
htynkn: should hessian_lite be excluded from asf header check?

pom.xml
```diff
@@ -443,6 +444,56 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
htynkn: in checkstyle, there is one rule to check asf header. If team decide to use apache rat plugin, we may need remove checkstyle rule. 
Related files:
+ https://github.com/apache/incubator-dubbo/blob/master/codestyle/checkstyle-header.txt
+ https://github.com/apache/incubator-dubbo/blob/master/codestyle/checkstyle.xml#L10

hessian-lite/src/test/java/com/alibaba/com/caucho/hessian/io/LocaleSerializerTest.java
lovepoem: >should hessian_lite be excluded from asf header check?

Yes, this is discussed in the dev@a

pom.xml
lovepoem: ok

ralf0131:
Looks good to me!

PR #1658, opened 2018-04-23
[Dubbo-1637] Fix When using hibernate-validator, class 'org.hibernate.validator.internal.metadata.descriptor.ConstraintDescriptorImpl' is wrong with hessian deserialization.

### LINKED ISSUE #1637: Issue about the provider side parameter validation
**Background**
We use JSR303 validation protocol, and use hibernate as implementation of JSR303 protocol.

**Version**
Dubbo version 2.5.3
 
**Problem Description**
Recently, we want to use the validation on the provider side(No issue on the consumer side). But we can't get the validation errors on the consumer side after configured the validation="true" to the service on the provider side. We try to use the ExceptionHandler to parse the exception returned,but we can't get the validation errors. The exception returned is RpcException with RemotingException clause, but we need RpcException with ConstraintViolationException clause. Because we can get the failed property name and the validation error from ConstraintViolationException directly.

We have analysised the ValidationFilter, it throws new RpcException when validation failed.
```
} catch (Throwable t) {
                throw new RpcException(t.getMessage(), t);
            }
```

 And the exception will be wrapped with RemotingException when result status is not equal to OK and SERVER_TIMEOUT in another class.

The recent version of ValidationFilter has been upgrade to return RpcResult, but it seems not completely.
```
} catch (Throwable t) {
                return new RpcResult(t);
            }
```

**Candidate Solution**
To fix this issue, we add two more files to wrap the property name and validation error from the hibernate validator, and return to client. 

The code snippet:

```
Set<CustomConstraintViolation> set = rewriteFromConstraintViolationException();
return new RpcResult(new RpcException(new CustomConstraintViolationException(set)));

// Replace javax.validation.ConstraintViolation with CustomConstraintViolation
// Replace javax.validation.ConstraintViolationException with CustomConstraintViolationException

```

We have tried to return javax.validation.ConstraintViolationException directly
```
return new RpcResult(new RpcException(new ConstraintViolationException(set)));
```

But failed with ConstraintDescriptorImpl deserialization with some error likes "ConstraintDescriptorImpl could not be instantiated.". Because the ConstraintDescriptorImpl has no empty constructor and can't be instantiated when deserialization. So create above two custom classes to rewrite the exception.

Then the ExceptionHandler can handle the RpcException, and retrieve the property name & validation error from CustomConstraintViolationException.

ningyu1:
bugfix：When using hibernate-validator, class 'org.hibernate.validator.internal.metadata.descriptor.ConstraintDescriptorImpl' is wrong with hessian deserialization.

When jsr303 uses hibernate-validator，Dubbo serialization using hessian，Set validation=true on the provider side，Client cannot receive validation information，The error is as follows：com.alibaba.com.caucho.hessian.io.HessianProtocolException: 'org.hibernate.validator.internal.metadata.descriptor.ConstraintDescriptorImpl' could not be instantiated

## Brief changelog

1. add class: com.alibaba.dubbo.validation.support.jsr303.DubboConstraintViolation
2. add class: com.alibaba.dubbo.validation.support.jsr303.DubboConstraintViolationException
3. modify class: com.alibaba.dubbo.validation.filter.ValidationFilter

Fixes #1637


lovepoem:
Please remove the author info and  check the CI failure

ningyu1:
@lovepoem author info has been remove. CI failure is due to Dubbo's own unit test error

lovepoem:
The master branch is ci-passed

ningyu1:
Resubmit Pull Request

PR #1676, opened 2018-04-25
use mockito to replace easymock and jmockit

htynkn:

use mockito to replace easymock

## Brief changelog

Remove dependency for easymock and jmockit
Add dependency for mockito
Migrate easymock test to mockito test base on mockito doc (https://github.com/mockito/mockito/wiki/Mockito-vs-EasyMock)

## Verifying this change

CI Pass (https://travis-ci.org/diamondblack/incubator-dubbo/builds/371112027)

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -191,13 +191,6 @@ cache-api:
   * HOMEPAGE:
     * https://github.com/jsr107/jsr107spec
 
-easymock
-
-  * LICENSE:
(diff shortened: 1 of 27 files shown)


NOTICE
```diff
@@ -191,13 +191,6 @@ cache-api:
   * HOMEPAGE:
     * https://github.com/jsr107/jsr107spec
 
-easymock
```
ralf0131: NOTICE and LICENSE has been changed according to ASF policy. Generally you don't have change these two files.

PR #1707, opened 2018-04-26
add an ExpiryCache  for CacheFilter

ZhichX:
ExpiryCache - With the characteristic of expiration time


diecui1202:
@ZhichX Thanks for your PR. 

Please merge with master and fix the conflict. 

Another problem is that it's not good to register shutdown hook in ExpiryMap. Reference DubboBootstrap.java in master branch and do the stop action in DubboShutdownHook.

https://github.com/apache/incubator-dubbo/blob/master/dubbo-bootstrap/src/main/java/org/apache/dubbo/bootstrap/DubboBootstrap.java

ZhichX:
Can I add a destroy method to the interface cache ,And call the method in the DubboShutdownHook to clean up?

ZhichX:
I put the ExpiryMap resource release into the clear method, and use the cache method `destory` to release it

beiwei30:
@ZhichX I don't think we need `clear()` method for Cache and `destroy` method for CacheFactory. What you only need to do is to make sure the sweeping thread is daemon. Pls. also consider to make sure:

* change expiry to expiring
* consider not to use lock as much as possible.

dubbo-common/src/main/java/com/alibaba/dubbo/common/utils/LRUCache.java
```diff
@@ -46,8 +46,8 @@ protected boolean removeEldestEntry(java.util.Map.Entry<K, V> eldest) {
 
     @Override
     public boolean containsKey(Object key) {
+        lock.lock();
```
beiwei30: why move lock.lock() outside?

dubbo-filter/dubbo-filter-cache/src/main/java/com/alibaba/dubbo/cache/support/AbstractCacheFactory.java
```diff
@@ -22,6 +22,7 @@
 import com.alibaba.dubbo.common.URL;
 import com.alibaba.dubbo.rpc.Invocation;
 
+import java.util.Map;
```
beiwei30: looks like a useless import.

dubbo-filter/dubbo-filter-cache/src/main/java/com/alibaba/dubbo/cache/support/expiring/ExpiringCache.java
```diff
@@ -0,0 +1,50 @@
+/*
```
beiwei30: pls. use 'expire' as package name, but 'Expiring' in class name

dubbo-filter/dubbo-filter-cache/src/main/java/com/alibaba/dubbo/cache/support/expiring/ExpiringMap.java
```diff
@@ -0,0 +1,381 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.alibaba.dubbo.cache.support.expiring;
+
+import java.util.*;
```
beiwei30: pls. use explicit imports, instead of `*`

dubbo-filter/dubbo-filter-cache/src/test/java/com/alibaba/dubbo/cache/filter/CacheFilterTest.java
```diff
@@ -16,78 +16,47 @@
  */
 package com.alibaba.dubbo.cache.filter;
```
beiwei30: The change made on UT doesn't look right. The original UT can test against all cache types.

beiwei30:
@ZhichX I made more comments inlined your code, pls. check.

dubbo-common/src/main/java/com/alibaba/dubbo/common/utils/LRUCache.java
ZhichX: If according to the original version, when lock.lock() reports an error, it will also enter the finally release lock. According to the Lock.unlock() document, an unchecked exception will be thrown when the non-lock holding thread calls this method:

dubbo-filter/dubbo-filter-cache/src/main/java/com/alibaba/dubbo/cache/support/AbstractCacheFactory.java
ZhichX: removed

dubbo-filter/dubbo-filter-cache/src/main/java/com/alibaba/dubbo/cache/support/expiring/ExpiringCache.java
ZhichX: thanks, I have renamed the package name

dubbo-filter/dubbo-filter-cache/src/test/java/com/alibaba/dubbo/cache/filter/CacheFilterTest.java
beiwei30: you should rollback your change made against CacheFilterTest. This test uses `@RunWith(Parameterized.class)` to test against `lru`, `jcache`, `threadlocal`.

```java
    @Parameters		
     public static List<Object[]> cacheFactories() {		
         return Arrays.asList(new Object[][]{		
                 {"lru", new LruCacheFactory()},		
                 {"jcache", new JCacheFactory()},		
                 {"threadlocal", new ThreadLocalCacheFactory()}		
         });		
     }
```

you should consider how to add `expiring` into this test, or write a new test delegated to `expiring` only.

PR #1708, opened 2018-04-27
[Dubbo-1637] Fix When using hibernate-validator, class 'org.hibernate.validator.internal.metadata.descriptor.ConstraintDescriptorImpl' is wrong with hessian deserialization.

### LINKED ISSUE #1637: Issue about the provider side parameter validation

ningyu1:

When jsr303 uses hibernate-validator，Dubbo serialization using hessian，Set validation=true on the provider side，Client cannot receive validation information，The error is as follows：com.alibaba.com.caucho.hessian.io.HessianProtocolException: 'org.hibernate.validator.internal.metadata.descriptor.ConstraintDescriptorImpl' could not be instantiated

## Brief changelog

add class: com.alibaba.dubbo.validation.support.jsr303.DubboConstraintViolation
modify class: com.alibaba.dubbo.validation.filter.ValidationFilter

## Verifying this change

ValidationTest already covered

- [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/alibaba/dubbo/tree/master/dubbo-test).

Fixes #1637

### DIFF
--- a/dubbo-config/dubbo-config-api/src/test/java/com/alibaba/dubbo/config/validation/ValidationTest.java
+++ b/dubbo-config/dubbo-config-api/src/test/java/com/alibaba/dubbo/config/validation/ValidationTest.java
@@ -197,8 +197,10 @@ public class ValidationTest {
                     parameter = new ValidationParameter();
                     validationService.save(parameter);
                     Assert.fail();
-                } catch (RpcException e) {
-                    Assert.assertTrue(e.getMessage().contains("ConstraintViolation"));
+                } catch (ConstraintViolationException ve) {
+                    Set<ConstraintViolation<?>> violations = ve.getConstraintViolations();
+                    Assert.assertTrue(violations.size() == 3);
+                    Assert.assertNotNull(violations);
                 }
 
                 // Delete OK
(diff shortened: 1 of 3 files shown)


lovepoem:
@ningyu1 the commit comments also should written in English

ningyu1:
@lovepoem In addition to resubmit there are other ways to modify the commit comments?

lovepoem:
When the committer help to merge this pr , this can be modified by the committer.

chickenlj:
Thanks for you contribution. But i don't think dubbo needs a customized Exception for validation implementation.

ningyu1:
@chickenlj Can you suggest a better way?

chickenlj:
I would recommend avoid/solve this problem by extending Filter, a filter like ExceptionFilter may solve this problem, detailed violations don't need to send to consumer side.

chickenlj:
I will close this PR now.

ningyu1:
@chickenlj 
I don't think so. In actual development, the consumer side needs to know detailed violation information.
Especially in the case of front and back end separation, the consumer side is directly the user browser, and the user needs to know clear violation information to guide the user to improve the operation mode

luozujun:
这个pull request的代码怎么在2.7.4.1版本中又没有了？[ningyu1](https://github.com/ningyu1)

PR #1865, opened 2018-05-30
Better License

carryxyh:
sth to supplement apache 2.0 license


chickenlj:
Hi @carryxyh ,
Would you please replace the apache header with the netty one?
https://github.com/apache/incubator-dubbo/pull/1865/files#diff-771d58b7b783f77d80a3e99eaa16f998R10

And also, Dubbo currently put all sub-module license in LICENSE, i think there's no need to add an extra `LICENSE.netty.txt` file, put it in LICENSE directly.
https://github.com/apache/incubator-dubbo/blob/master/LICENSE#L266

chickenlj:
Which files should add netty license? I can tell is:
* InternalThreadLocal
* InternalThreadLocalMap

carryxyh:
Oh, I will submit a new pr as soon as possible
Thx

chickenlj:
Resolved with #1866, close this one.

PR #1876, opened 2018-06-01
[Dubbo-606] Optimize new async support for both consumer and provider side.

chickenlj:

Async support for provider and consumer.
* Consumer, support method with extra Future return type
```java
@AsyncFor(GreetingsServices.class)
public interface GreetingsServiceAsync extends GreetingsServices {
    CompletableFuture<String> sayHi(String name);
}
```
We also provide a tool to do that: [dubbo-async-processor](https://github.com/dubbo/dubbo-async-processor)  
* Provider, support async like what Servlet 3.0 async works.
```java
public class AsyncServiceImpl implements AsyncService {

        System.out.println("Main sayHello() method start.");
        final AsyncContext asyncContext = RpcContext.startAsync();
        new Thread(() -> {
            System.out.println("    -- Async start.");
            try {
                Thread.sleep(500);
            } catch (InterruptedException e) {
                e.printStackTrace();
            }
            asyncContext.write("Hello " + name + ", response from provider.");
            System.out.println("    -- Async end.");
        }).start();
        System.out.println("Main sayHello() method end.");
        return "hello, " + name;
    }
}
```

Drawbacks and following enhancements:
* For both consumer and provider async, response can't be processed with Filter chains anymore.
* Context still can't be restored between threads.

Thanks for @yuyijq for contributing code and ideas with #606.

## Verifying this change

This change has been deploy to central maven repository with a 2.6.3-SNAPSHOT version. You can see details of how to use from these two projects:
[dubbo-async-processor](https://github.com/dubbo/dubbo-async-processor)  
[dubbo-samples](https://github.com/dubbo/dubbo-samples/tree/samples_for_2.6.3-SNAPSHOT)  

- [ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/alibaba/dubbo/tree/master/dubbo-test).
#


.travis.yml
```diff
@@ -4,7 +4,6 @@ sudo: false # faster builds
 jdk:
     - oraclejdk9
     - oraclejdk8
-    - openjdk7
```
htynkn: if we decide remove java7 support. should we have a separate pr instead of mixing in this pr?

lovepoem: this pr should synchronize with the origin/master

chickenlj: I will synchronize with the origin/master branch.

chickenlj: > if we decide remove java7 support. should we have a separate pr instead of mixing in this pr?

If don't remove java7, travis CI would fail.

chickenlj:
I will provide enhancements with a new PR later:
* For both consumer and provider async, response can't be processed with Filter chains anymore. 
* Context still can't be restored between threads.

PR #1938, opened 2018-06-14
Fix a bug when the provider write a throwable into aysnc context

carryxyh:
* Fix a bug when the provider write a throwable into async context
* sth doc
* simple code


diecui1202:
Now I'm doing the repackage to org.apache. It's on the process. Please resend a PR after the repackage PR. Sorry.

carryxyh:
I will. thx

Issue #2023, opened 2018-07-04
ForkingClusterInvoker missing RpcContext and the attachments.


carryxyh:
- [x] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/wiki/FAQ) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.0-SNAPSHOT
* Operating System version: mac os
* Java version: 1.8

### Step to reproduce this issue

1. Export a service
2. Make cluster = "forking"
3. Set Attachment in RpcContext
4. Do rpc invoke

### Expected Result

The attachment in RpcContext has element.

### Actual Result

The attachment in RpcContext don't have element.

### The problem
I set attachment in the consumer side like this:
![image](https://user-images.githubusercontent.com/12047600/42265457-30483704-7fa6-11e8-8075-2eede7775b3a.png)

Then do rpc invoke,but the attachment is missing, (AbstractInvoker#invoke):
![image](https://user-images.githubusercontent.com/12047600/42265551-6b74d300-7fa6-11e8-8ad6-359cf091a3ac.png)

The problem is that ForkingClusterInvoker use threapool to do rpc invoke which will miss the threadlocal:
![image](https://user-images.githubusercontent.com/12047600/42265595-92dae95c-7fa6-11e8-8ae3-3332cc67e4d3.png)

carryxyh:
This issue has been fixed by #2024.
Close this one.

PR #2024, opened 2018-07-04
Binding attachment before a clusterInvoker invoke.

carryxyh:
* Binding attachments into invocation so that a rpc invoke have a same invocation
* Fix a bug that the ForkingClusterInvoker don't clear attachment when a rpc invoke is over.

### DIFF
--- a/dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/support/ForkingClusterInvoker.java
+++ b/dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/support/ForkingClusterInvoker.java
@@ -57,50 +57,55 @@ public class ForkingClusterInvoker<T> extends AbstractClusterInvoker<T> {
     @Override
     @SuppressWarnings({"unchecked", "rawtypes"})
     public Result doInvoke(final Invocation invocation, List<Invoker<T>> invokers, LoadBalance loadbalance) throws RpcException {
-        checkInvokers(invokers, invocation);
-        final List<Invoker<T>> selected;
-        final int forks = getUrl().getParameter(Constants.FORKS_KEY, Constants.DEFAULT_FORKS);
-        final int timeout = getUrl().getParameter(Constants.TIMEOUT_KEY, Constants.DEFAULT_TIMEOUT);
-        if (forks <= 0 || forks >= invokers.size()) {
-            selected = invokers;
-        } else {
-            selected = new ArrayList<Invoker<T>>();
-            for (int i = 0; i < forks; i++) {
-                // TODO. Add some comment here, refer chinese version for more details.
-                Invoker<T> invoker = select(loadbalance, invocation, invokers, selected);
-                if (!selected.contains(invoker)) {//Avoid add the same invoker several times.
-                    selected.add(invoker);
-                }
-            }
-        }
-        RpcContext.getContext().setInvokers((List) selected);
-        final AtomicInteger count = new AtomicInteger();
-        final BlockingQueue<Object> ref = new LinkedBlockingQueue<Object>();
-        for (final Invoker<T> invoker : selected) {
-            executor.execute(new Runnable() {
-                @Override
-                public void run() {
-                    try {
-                        Result result = invoker.invoke(invocation);
-                        ref.offer(result);
-                    } catch (Throwable e) {
-                        int value = count.incrementAndGet();
-                        if (value >= selected.size()) {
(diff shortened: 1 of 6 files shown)


carryxyh:
fix problem:
https://github.com/apache/incubator-dubbo/issues/1978

kimmking:
I think you can attach a testcase for this

carryxyh:
add some unit test to this case.

carryxyh:
Fix:
https://github.com/apache/incubator-dubbo/issues/2023

dubbo-cluster/src/test/java/com/alibaba/dubbo/rpc/cluster/support/ForkingClusterInvokerTest.java
```diff
@@ -120,8 +122,32 @@ public void testInvokeExceptoin() {
...
+        RpcContext.getContext().setAttachment(attachKey, attachValue);
+
+        Map<String, String> attachments = RpcContext.getContext().getAttachments();
+        Assert.assertTrue("set attachment failed!", attachments != null && attachments.size() == 1);
+        try {
+            invoker.invoke(invocation);
+            Assert.fail();
+        } catch (RpcException expected) {
+            Assert.assertTrue(expected.getMessage().contains("Failed to forking invoke provider"));
+            assertFalse(expected.getCause() instanceof RpcException);
```
kimmking: not the same code style
: )

carryxyh:
@kimmking 
fix code style
pls review

kimmking:
good job

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/support/ForkingClusterInvoker.java
```diff
@@ -57,50 +57,55 @@ public ForkingClusterInvoker(Directory<T> directory) {
...
-                    try {
-                        Result result = invoker.invoke(invocation);
-                        ref.offer(result);
-                    } catch (Throwable e) {
-                        int value = count.incrementAndGet();
-                        if (value >= selected.size()) {
-                            ref.offer(e);
+            RpcContext.getContext().setInvokers((List) selected);
+            final AtomicInteger count = new AtomicInteger();
+            final BlockingQueue<Object> ref = new LinkedBlockingQueue<Object>();
```
ningyu1: LinkedBlockingQueue You should give an initialization size, The default value should not be used: Integer.max_value

carryxyh: Is Integer.max_value has problems?

ningyu1: If the queue producer is faster than the consumer, and the queue is not fully blocked, memory consumption is high

carryxyh: 1. In this case, it will only be consumed once.
2.This list is a linked list so why the memory consumption is high?
3.The producer count is controllable and every producer only produce result once.

So it isn't a problem.

ningyu1: The number of producers is controllable, and that should be fine

carryxyh: OK, that make sence.
In this pr, it is to fix the clearing RpcContext.
Maybe next pr to fix the capacity problem :)

diecui1202:
LGTM, &READY-TO-MERGE&

diecui1202:
@carryxyh Does this PR need to merge into master? I think so.

carryxyh:
need another pr to master ?
@diecui1202

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/support/ForkingClusterInvoker.java
GuldanChen: GOOD JOB

chickenlj:
I will merge this PR. I think to wrap Cluster Invoker with Filter chain maybe another solution. Currently, the Filter chain only works with the underlying Invoker.

PR #2073, opened 2018-07-13
remove hessian-lite

lovepoem:

remove hessian-lite module of 2.6.x ,and use https://github.com/dubbo/hessian-lite

## Brief changelog

delete hessian-lite module

## Verifying this change
 CI pass


ralf0131:
Hi, I think we need to add hessian-lite dependency to `dependencies-bom/pom.xml`

lovepoem:
Now hessian-lite dependency id added to dependencies-bom/pom.xml

luyuanwan:
Hessian is dependent from dubbo ,is my understanding correct?

lovepoem:
>Hessian is dependent from dubbo ,is my understanding correct?
Your understanding is opposite

ralf0131:
@luyuanwan Dubbo depends Hessian.

PR #2074, opened 2018-07-13
remove hessian-lite and use https://github.com/dubbo/hessian-lite

lovepoem:

remove hessian-lite and use https://github.com/dubbo/hessian-lite

## Brief changelog

delete hessian-lite module

## Verifying this change
 CI pass


kimmking:
good job

Issue #1929, opened 2018-06-12
No such configuration property version


ulinkwo:
https://github.com/apache/incubator-dubbo/blob/5a4c0768fd8989c1f82710a50a288e26fa69215a/dubbo-config/dubbo-config-spring/src/test/resources/com/alibaba/dubbo/config/spring/annotation-version-consumer.xml#L26

as above, there is no configuration property `version` in dubbo.xsd, it is eally confused. i think it could be removed if no other purpose.

chickenlj:
Would you please send a pull request?

PR #2078, opened 2018-07-13
[dubbo-1929]: remove the verison property that is no use (#1929)

luyuanwan:
remove the verison property that is no use, you can see issue  #1929


ralf0131:
Looks good to me! Thanks for the pull request!

Issue #2011, opened 2018-07-01
ServiceAnnotationBeanPostProcessor的log信息级别有误


gy0801151351:
https://github.com/apache/incubator-dubbo/blob/master/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/beans/factory/annotation/ServiceAnnotationBeanPostProcessor.java

这个java类里面的private void registerServiceBean 这个方法, 下面这几行
if (scanner.checkCandidate(beanName, serviceBeanDefinition)) { // check duplicated candidate bean
            registry.registerBeanDefinition(beanName, serviceBeanDefinition);

            if (logger.isInfoEnabled()) {
                logger.warn("The BeanDefinition[" + serviceBeanDefinition +
                        "] of ServiceBean has been registered with name : " + beanName);
            }

        } else {

            if (logger.isWarnEnabled()) {
                logger.warn("The Duplicated BeanDefinition[" + serviceBeanDefinition +
                        "] of ServiceBean[ bean name : " + beanName +
                        "] was be found , Did @DubboComponentScan scan to same package in many times?");
            }

        }

请问, 为什么在检查没有重复的bean定义的时候, 还要打印一条警告信息? 不是应该是info信息吗?

ralf0131:
Hi, thanks for reporting, would you please send a pull request?

chickenlj:
@gy0801151351 This is a good start for contributing to Dubbo in code.

luyuanwan:
Why haven't  this pr fixed? see pr #2066 @lovepoem

kongwu-:
我也发现了，郁闷老半天，看源码才发现打成warning了

PR #2082, opened 2018-07-15
change log level from warn to info

luyuanwan:
change log level from warn to info ,see issue #2011


luyuanwan:
This is already a pull request fix this typo problem,see #2066,so I close it.

PR #2101, opened 2018-07-19
Docs in AsyncContext

carryxyh:
* Docs in AsyncContext
* Make the future final which is in the constructor of AsyncRpcResult because of using in a inner class.


ralf0131:
LGTM, thanks for the pull request!

PR #2114, opened 2018-07-22
Dubbo-2013

tswstarplanet:

Let user can config the threadpool of consumer !

## Brief changelog

add a tag "threadpool" to dubbo:consumer in dubbo.xsd add a property "threadpool" to ConsumerConfig class modify Client code to get the threadpool by user's config


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ConsumerConfig.java
```diff
@@ -31,6 +31,9 @@
     // networking framework client uses: netty, mina, etc.
     private String client;
 
+    // client thread pool type: cached, fixed, limit, eager
+    private String threadpool;
```
chickenlj: We may also need the following two parameters:
```xml
                <xsd:attribute name="threads" type="xsd:string" use="optional">
                    <xsd:annotation>
                        <xsd:documentation><![CDATA[ The thread pool size. ]]></xsd:documentation>
                    </xsd:annotation>
                </xsd:attribute>
                <xsd:attribute name="queues" type="xsd:string" use="optional">
                    <xsd:annotation>
                        <xsd:documentation><![CDATA[ The thread pool queue size. ]]></xsd:documentation>
                    </xsd:annotation>
                </xsd:attribute>
```

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/transport/AbstractClient.java
```diff
@@ -113,7 +111,12 @@ public AbstractClient(URL url, ChannelHandler handler) throws RemotingException
 
     protected static ChannelHandler wrapChannelHandler(URL url, ChannelHandler handler) {
         url = ExecutorUtil.setThreadName(url, CLIENT_THREAD_POOL_NAME);
-        url = url.addParameterIfAbsent(Constants.THREADPOOL_KEY, Constants.DEFAULT_CLIENT_THREADPOOL);
+        String key = Constants.DEFAULT_KEY + Constants.HIDE_KEY_PREFIX + Constants.THREADPOOL_KEY;
+        String threadpool = url.getParameter(key);
+        if (StringUtils.isNotEmpty(threadpool)) {
+            url = url.removeParameter(key);
+        }
+        url = url.addParameterIfAbsent(Constants.THREADPOOL_KEY, StringUtils.isNotEmpty(threadpool) ? threadpool : Constants.DEFAULT_CLIENT_THREADPOOL);
```
chickenlj: Do we need to change the code here?
I think we will have `default.threadpool` key in automatically.

tswstarplanet: Ok, because I found that when ReferenceConfig integrate the configuration of ConsumerConfig, it will add the prefix "default.", so the "threadpool" property of <dubbo:consumer ../> will change to "default.thread" parameter in the URL.
So, I think we have two ways to optimize the issue.
@chickenlj First, take some special process when integrate the "threadpool" property of ConsumerConfig to ReferenceConfig to wipe off the prefix "default."；
Second is just the method that this sumbit uses.
If the second way is not so good, I think we can take the second one.

chickenlj: Look at the `getParameter("threadpool")` logic in URL, it will: first check "threadpool" and then check "default.threadpool"

https://github.com/apache/incubator-dubbo/blob/0b6e6f3a375d82aab1a1f17b034e666c63f61103/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java#L421

chickenlj:
Would you please submit a patch for this PR according to the review comment now? I would suggest that we add this enhancement to V2.6.3,  this version is discussing in the community and is about to release soon.

Issue #944, opened 2017-11-28
AbstractClient reconnect() should be synchronized under concurrency scene


chenliuyang1989:
高并发重连出错，需要加锁，
eg：线程1关闭原有连接，重建连接
        并发线程2会关闭线程1新建的连接，重接新的连接。导致线程1发送数据失败
        并发线程3又会关闭线程2新建立的连接，重建新的连接，导致线程2发送数据失败。
问题：高并发重连时不加锁会出现频繁的连接关闭重建，

线上出错日志见附件
![default](https://user-images.githubusercontent.com/4150132/33326002-b4ee8f62-d48e-11e7-8b0b-70d7efa6c0db.jpg)

chenliuyang1989:
This is my ReImplementation of reconnect method ：
   public void reconnect() throws RemotingException {
        this.reConnectLock.lock();

        try {
            if(!this.isConnected()) {
                this.disconnect();
                this.resetConnectStatus();
                this.connect();
                return;
            }
        } finally {
            this.reConnectLock.unlock();
        }

    }

chenliuyang1989:
do you have plan when to fix the bug?

ralf0131:
Hi, would you mind send a pull request?

zonghaishang:
fixed, refer: https://github.com/apache/incubator-dubbo/pull/2135

zonghaishang:
@chenliuyang1989  Which version of dubbo you use?

PR #2135, opened 2018-07-25
Fix a bug when client reconnect

carryxyh:
Add reconnection lock to control only one thread can can reconnect method.
This will avoid problem in this case:
Thread A reconnecting, and success, then send msg.
Thread B reconnecting but invoke disconnect. Then the Thread A will send msg fail because of Thread B's disconnecting call.

The issue is here:
https://github.com/apache/incubator-dubbo/issues/944


zonghaishang:
It looks good to me

carryxyh:
There has sth wrong about my describe of this pr.
This is one optimize when two thread call reconnect.

In old way:
Thread A call reconnect.
Thread B call reconnect. This case maybe make twice connect and twice disconnect.

In new way:
Thread A call reconnect, Thread B wait for the lock.
After Thread A finishing reconnect, Thread B will not do connect and disconnect. Because of that Thread A finish connect, this client is connected, it dose not need to reconnect again.

Issue #2127, opened 2018-07-24
NullPointExexception  in RpcContextFilter


gaboolic:
- [ ] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.6.1 - newest

### Step to reproduce this issue

1.  
String value = null;
RpcContext.getContext().getAttachments().put("demokey", value)
2. 
then invoke one method will NullPointException

**The Exception reason:
In RpcContextFilter.java，public void filter(ClientRequestContext requestContext) throws IOException
for (Map.Entry<String, String> entry : RpcContext.getContext().getAttachments().entrySet()) {
            if (entry.getValue().contains(",") || entry.getValue().contains("=")

the entry.getValue not judge null!!**

gaboolic:
I hope com.alibaba.dubbo.rpc.protocol.rest.RpcContextFilter can judge entry.getValue()  null

carryxyh:
This is not a normal case that use put attachments.
I think there are only one case to use put(key,null) is to remove a key or value.
In this case, u can use remove to do the same thing.

carryxyh:
&READY-TO-CLOSE&

gaboolic:
String value = getStringFromXXX();
RpcContext.getContext().getAttachments().put("demokey", value)
这种场景做数据透传的时候会用。我并不需要校验value是否为空，我仅仅是从某个地方去获取几个字符串然后放到上下文里面透传到下一个服务。
然鹅这样会报错，而且报的错是服务提供方有空指针，实际上这个异常是dubbo自己的filter抛出的，但是被catch了。那个filter的作用仅仅是校验一下value是否有逗号和等于号，然后throw一个异常。如果Nul也不是预期的值，至少也该和逗号等于号一样，抛出一个异常来有一个友好提示吧？

gaboolic:
另外只有rest协议 会有这个RpcContextFilter。其他的协议似乎我可以想往RpcContext.getContext().getAttachments()里放什么就放什么吧，应该是可以放null的；就算我用rest协议，null   逗号 等于号 为什么不让我放呢，不能因为dubbo自己内部实现是用逗号和等于号就不让我用吧，其实完全可以做个转码啊。

[...]

PR #2147, opened 2018-07-27
Judge null for key and value in attachment in RpcContextFilter.

carryxyh:
Judge null for key and value in attachment in RpcContextFilter.
If not, NullPointerException will be thrown in RpcContextFilter in some case.

issue:
https://github.com/apache/incubator-dubbo/issues/2127#issuecomment-408344327


ralf0131:
Hi, do we have a unit test to cover this change?

carryxyh:
Sorry to rebase sth wrong into this branch, the new pr with unit test is here:

https://github.com/apache/incubator-dubbo/pull/2171

PR #2156, opened 2018-07-31
Optimize loadbalance init

carryxyh:
* optimize LoadBalance init.
* add needInitLoadBalance method to decide whether a cluseter invoker instance need LoadBalance.
* AvailableCluster、AvailableClusterInvoker、BroadcastClusterInvoker do not need to init a LoadBalance, so the needInitLoadBalance return false for them


diecui1202:
Get loadbalance instance is not waste time. Maybe AvailableCluster、AvailableClusterInvoker、BroadcastClusterInvoker will use it later.

Combine the loadbalance init logic together is a good idea.

carryxyh:
@diecui1202 
Ok, another pr will be sent soon.

Issue #904, opened 2017-11-22
LeastActiveLoadBalance负载均衡


wannshan:
74行  int offsetWeight = random.nextInt(totalWeight);//这里totalWeight是没有经过warmup降权的值的和
78    offsetWeight -= getWeight(invokers.get(leastIndex), invocation);//这个是经过warmup降权的。

当服务刚启动，经过warmup降权后的getWeight返回的值，可能经过leastCount次循环后，还是offsetWeight >0
造成选不到invoker
for (int i = 0; i < leastCount; i++) {
                int leastIndex = leastIndexs[i];
                offsetWeight -= getWeight(invokers.get(leastIndex), invocation);
                if (offsetWeight <= 0)
                    return invokers.get(leastIndex);
}

ralf0131:
Hi, could you please send a pull request?

carryxyh:
If the offsetWeight doesn't decrease lower than 0 after loop, it maybe mean that all the least active invokers are in warmup, then the LeastActiveLoadBalance will select a random one from these same active invokers.
Or, continue to loop until the value of offsetWeight drops below 0?

wannshan:
Change the line  
int weight = invoker.getUrl().getMethodParameter(invocation.getMethodName(), Constants.WEIGHT_KEY, Constants.DEFAULT_WEIGHT); // Weight
TO
int weight = getWeight(invoker, invocation); // have warm up  value

wannshan:
may be  taotalWeightWithWarmUp   will be better ,after all not after  but warming....

carryxyh:
That make sense.
Would u pls review this into my pr? And I will fix it soon.

carryxyh:
Hi, I submit a new issue here:
https://github.com/apache/incubator-dubbo/issues/2540
Focus on this issue, pls.

PR #2161, opened 2018-08-01
Optimize LeastActiveLoadBalance select.

carryxyh:
Now the select is select a random one when there are several least active invokers and all of them are in warm up.
After this pr, it will select also by weight and warm up.

ISSUE:
https://github.com/apache/incubator-dubbo/issues/904


ralf0131:
Hi, 

Would you please add some unit test for this change?

carryxyh:
Sorry for rebase sth wrong into current branch, the new pr is here:

https://github.com/apache/incubator-dubbo/pull/2172

PR #2171, opened 2018-08-02
Judge null for key and value in attachment in RpcContextFilter.

### LINKED ISSUE #2127: NullPointExexception  in RpcContextFilter

carryxyh:
Judge null for key and value in attachment in RpcContextFilter.
If not, NullPointerException will be thrown in RpcContextFilter in some case.

Fix #2127 : https://github.com/apache/incubator-dubbo/issues/2127#issuecomment-408344327

### DIFF
--- a/dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/RpcContextFilter.java
+++ b/dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/RpcContextFilter.java
@@ -18,7 +18,6 @@ package org.apache.dubbo.rpc.protocol.rest;
 
 import org.apache.dubbo.common.utils.StringUtils;
 import org.apache.dubbo.rpc.RpcContext;
-
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
 import javax.annotation.Priority;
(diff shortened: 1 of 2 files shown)


dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/RpcContextFilter.java
```diff
@@ -70,9 +69,17 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
...
+            String value = entry.getValue();
+            if (StringUtils.isNotEmpty(key)) {
+                if (illegalForRest(key)) {
+                    throw new IllegalArgumentException("The attachments of " + RpcContext.class.getSimpleName() + " must not contain ',' or '=' when using rest protocol");
+                }
+            }
+            if (StringUtils.isNotEmpty(value)) {
+                if (illegalForRest(value)) {
+                    throw new IllegalArgumentException("The attachments of " + RpcContext.class.getSimpleName() + " must not contain ',' or '=' when using rest protocol");
+                }
```
diecui1202: if ( (StringUtils.isNotEmpty(key) && illegalForRest(key)) || (StringUtils.isNotEmpty(value) && illegalForRest(value)) ) {
    throw new IllegalArgumentException("The attachments of " + RpcContext.class.getSimpleName() + " must not contain ',' or '=' when using rest protocol");
}

Is the above code is better ?

dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/RpcContextFilter.java
```diff
@@ -85,4 +92,14 @@ public void filter(ClientRequestContext requestContext) throws IOException {
...
     }
+
+    /**
+     * If a string value illegal for rest protocol(',' and '=' is illegal for rest protocol).
+     *
+     * @param v string value
+     * @return true for illegal
+     */
+    private boolean illegalForRest(String v) {
+        return v.contains(",") || v.contains("=");
```
diecui1202: Or judge null place here ?

carryxyh: Prefer judge null in 'illegalForRest'.
The meaning of this method is that if a value is not null, it should not contains ',' or '='.
I will fix this soon.
:)

diecui1202:
LGTM.

PR #2172, opened 2018-08-02
Optimize leastActiveSelect and weight test case

carryxyh:
Now the select is select a random one when there are several least active invokers and all of them are in warm up.
After this pr, it will select also by weight and warm up.

And fix a bug when two invoker's active is same and weight not same.

issue:
https://github.com/apache/incubator-dubbo/issues/904

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/LeastActiveLoadBalance.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/LeastActiveLoadBalance.java
@@ -37,8 +37,8 @@ public class LeastActiveLoadBalance extends AbstractLoadBalance {
         int leastActive = -1; // The least active value of all invokers
         int leastCount = 0; // The number of invokers having the same least active value (leastActive)
         int[] leastIndexes = new int[length]; // The index of invokers having the same least active value (leastActive)
-        int totalWeightAfterWarmUp = 0; // The sum of after warmup weights
-        int firstWeightAfterWarmUp = 0; // Initial value, used for comparision
+        int taotalWeightWithWarmUp = 0; // The sum of with warmup weights
+        int firstWeightWithWarmUp = 0; // Initial value, used for comparision
         boolean sameWeight = true; // Every invoker has the same weight value?
         for (int i = 0; i < length; i++) {
             Invoker<T> invoker = invokers.get(i);
@@ -48,15 +48,15 @@ public class LeastActiveLoadBalance extends AbstractLoadBalance {
                 leastActive = active; // Record the current least active value
                 leastCount = 1; // Reset leastCount, count again based on current leastCount
                 leastIndexes[0] = i; // Reset
-                totalWeightAfterWarmUp = afterWarmup; // Reset
-                firstWeightAfterWarmUp = afterWarmup; // Record the weight the first invoker
+                taotalWeightWithWarmUp = afterWarmup; // Reset
+                firstWeightWithWarmUp = afterWarmup; // Record the weight the first invoker
                 sameWeight = true; // Reset, every invoker has the same weight value?
             } else if (active == leastActive) { // If current invoker's active value equals with leaseActive, then accumulating.
                 leastIndexes[leastCount++] = i; // Record index number of this invoker
-                totalWeightAfterWarmUp += afterWarmup; // Add this invoker's after warmup weight to totalWeightAfterWarmUp.
+                taotalWeightWithWarmUp += afterWarmup; // Add this invoker's with warmup weight to totalWeightWithWarmUp.
                 // If every invoker has the same weight?
                 if (sameWeight && i > 0
-                        && afterWarmup != firstWeightAfterWarmUp) {
+                        && afterWarmup != firstWeightWithWarmUp) {
                     sameWeight = false;
                 }
             }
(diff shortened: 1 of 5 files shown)


whanice:
@carryxyh

I saw your pr,  I think the start warm up feature with the leastactive algorithm for loadbalance will be conflicted, because of the active for the new provider will be the least.

So is it necessary to calculate the weight for warm up?

Or maybe we can consider another implementation to supports warmup for leastactive loadbalance.

carryxyh:
This is not the problem that this PR has to solve. The problem that this PR has to solve is too many loops and a small bug.
And I think the problem you are talking about is not a big problem. Just need to prompt the user in the document: If the cluster uses LeastActive as the load balancing policy, then if the new service is in the warm-up period, then it will get more requests (because the new service itself is the least active).

carryxyh:
`So is it necessary to calculate the weight for warm up?`
Of course. 
If all the service is in warm-up. Or two least active service is in warm-up, if not calc weight after warm-up, u will get a wrong weight.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/LeastActiveLoadBalance.java
```diff
@@ -41,26 +39,26 @@
         int leastActive = -1; // The least active value of all invokers
         int leastCount = 0; // The number of invokers having the same least active value (leastActive)
         int[] leastIndexs = new int[length]; // The index of invokers having the same least active value (leastActive)
-        int totalWeight = 0; // The sum of weights
-        int firstWeight = 0; // Initial value, used for comparision
+        int taotalWeightWithWarmUp = 0; // The sum of with warmup weights
```
diecui1202: typo

carryxyh:
@whanice @chickenlj @zonghaishang 
Pls help to review this pr :)

carryxyh:
@diecui1202 
I think a same pr should be sent to branch 2.6.x.
How do u think about it?

carryxyh:
Ok, fix it soon.

carryxyh:
@zonghaishang @diecui1202 @chickenlj @whanice 
Hi, everyone.
Can you take the time to review this pr? This is a very simple bug, but it has been idle for a long time.

dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/loadbalance/LeastActiveBalanceTest.java
```diff
@@ -39,4 +48,119 @@ public void testLeastActiveLoadBalance_select() {
...
+            // never select invoker3 because it's active is more than invoker1 and invoker2
+            Assert.assertTrue("select is not the least active one", !selected.getUrl().getProtocol().equals("test3"));
+        }
+
+        // the sumInvoker1 : sumInvoker2 approximately equal to 1: 9
+        System.out.println(sumInvoker1);
+        System.out.println(sumInvoker2);
+    }
+
+    class MyLeastActiveLoadBalance extends AbstractLoadBalance {
```
zonghaishang: Could you reuse existing code? 

org.apache.dubbo.rpc.cluster.loadbalance.LeastActiveBalanceTest#testLeastActiveLoadBalance_select:
```
public void testLeastActiveLoadBalance_select() {
    ...
    Map<Invoker, AtomicLong> counter = getInvokeCounter(runs, LeastActiveLoadBalance.NAME);
    ...
```
Load your repaired code this way：

`ExtensionLoader.getExtensionLoader(LoadBalance.class).getExtension(LeastActiveLoadBalance.NAME)`

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/LeastActiveLoadBalance.java
```diff
@@ -70,7 +68,7 @@
         }
         if (!sameWeight && totalWeight > 0) {
             // If (not every invoker has the same weight & at least one invoker's weight>0), select randomly based on totalWeight.
-            int offsetWeight = ThreadLocalRandom.current().nextInt(totalWeight);
+            int offsetWeight = ThreadLocalRandom.current().nextInt(totalWeight) + 1;
```
zonghaishang: Why change `ThreadLocalRandom.current().nextInt(totalWeight)` to `ThreadLocalRandom.current().nextInt(totalWeight) + 1` ?

zonghaishang: Is it possible that +1 causes offset not equal or less 0?

carryxyh: If not change to `+1`, that will cause a bug.
U can look at my unit test for more detail.

carryxyh:
@zonghaishang 
I have optimize ut, pls check again.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/LeastActiveLoadBalance.java
code4wt: 最近正好在看 LeastActiveLoadBalance 的源码，我觉得 +1 这个逻辑有点突兀，不知道背景的同学可能不知道为什么要+1。更合理的方式，我觉得应该按照 RandomLoadBalance 逻辑去处理。将 if (offsetWeight <= 0) 改为 if (offsetWeight < 0)。这样两者的逻辑能够统一起来，只要大家能看懂 RandomLoadBalance 就一定能看懂 ，那么此处的代码也一样能看懂，而不用特地去思考为什么要 +1。这是一个小问题，改不改都行

Issue #2184, opened 2018-08-04
Return the hanging client requests ASAP. when connection is broken unexpectedly.


chickenlj:
I think we should solve the case 1 mentioned in #1968 .

PR #2185, opened 2018-08-04
Direct return when the server goes down unnormally.

carryxyh:
Keep the unfinished request in every channel.
When the server goes down un-normally, return the unfinished requests directly.
The current way is to wait until timeout.

issue:
https://github.com/apache/incubator-dubbo/issues/2184

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
@@ -48,9 +48,14 @@ public class DefaultFuture implements ResponseFuture {
 
     private static final Logger logger = LoggerFactory.getLogger(DefaultFuture.class);
 
-    private static final Map<Long, Channel> CHANNELS = new ConcurrentHashMap<Long, Channel>();
+    private static final Map<Long, Channel> CHANNELS = new ConcurrentHashMap<>();
 
-    private static final Map<Long, DefaultFuture> FUTURES = new ConcurrentHashMap<Long, DefaultFuture>();
+    private static final Map<Long, DefaultFuture> FUTURES = new ConcurrentHashMap<>();
+
+    /**
+     * keep the unfinished request for every channel.
+     */
+    private static final Map<Channel, Map<Long, Request>> UNFINISHED_REQUESTS = new ConcurrentHashMap<>();
 
     public static final Timer TIME_OUT_TIMER = new HashedWheelTimer(
             new NamedThreadFactory("dubbo-future-timeout", true),
@@ -77,6 +82,23 @@ public class DefaultFuture implements ResponseFuture {
         // put into waiting map.
         FUTURES.put(id, this);
         CHANNELS.put(id, channel);
+    }
+
+    /**
+     * init a DefaultFuture
+     * 1.init a DefaultFuture
+     * 2.keep it into UNFINISHED_REQUESTS
+     *
+     * @param channel channel
+     * @param request the request
+     * @param timeout timeout
+     * @return a new DefaultFuture
+     */
+    public static DefaultFuture initFuture(Channel channel, Request request, int timeout) {
+        DefaultFuture future = new DefaultFuture(channel, request, timeout);
+        Map<Long, Request> unfinishedRequests = UNFINISHED_REQUESTS.computeIfAbsent(channel, c -> new ConcurrentHashMap<>());
+        unfinishedRequests.put(future.getId(), future.getRequest());
+        return future;
     }
 
     /**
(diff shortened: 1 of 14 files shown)


diecui1202:
@carryxyh please check the travis ci failure.

carryxyh:
@diecui1202 
yeah. maybe still some problems. I will try to fix them later.

carryxyh:
@diecui1202 
Already fix unit test.

dubbo-rpc/dubbo-rpc-dubbo/src/test/java/org/apache/dubbo/rpc/protocol/dubbo/DubboInvokerAvilableTest.java
```diff
@@ -117,6 +117,7 @@ public void test_NoInvokers() throws Exception {
     }
 
     @Test
+    @Ignore
```
diecui1202: Why ignore this test case ?

carryxyh: I didn't ignore this.
Maybe it is a outdated content ?

dubbo-rpc/dubbo-rpc-dubbo/src/test/java/org/apache/dubbo/rpc/protocol/dubbo/ReferenceCountExchangeClientTest.java
```diff
@@ -27,10 +27,7 @@
 import org.apache.dubbo.rpc.Invoker;
 import org.apache.dubbo.rpc.ProxyFactory;
 
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.BeforeClass;
-import org.junit.Test;
+import org.junit.*;
```
diecui1202: It's not a good practice to import *

dubbo-rpc/dubbo-rpc-dubbo/src/test/java/org/apache/dubbo/rpc/protocol/dubbo/MultiThreadTest.java
```diff
@@ -25,28 +25,33 @@
...
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
-public class MultiThreadTest extends TestCase {
+public class MultiThreadTest {
 
     private Protocol protocol = ExtensionLoader.getExtensionLoader(Protocol.class).getAdaptiveExtension();
     private ProxyFactory proxy = ExtensionLoader.getExtensionLoader(ProxyFactory.class).getAdaptiveExtension();
 
+    @Ignore
```
kimmking: there is an ignore too.

dubbo-rpc/dubbo-rpc-dubbo/src/test/java/org/apache/dubbo/rpc/protocol/dubbo/ReferenceCountExchangeClientTest.java
kimmking: import * can't pass check-style when submitting a PR

dubbo-rpc/dubbo-rpc-dubbo/src/test/java/org/apache/dubbo/rpc/protocol/dubbo/MultiThreadTest.java
carryxyh: Fix it soon :)

zonghaishang:
I agree with Li Jun’s point of view.
Server graceful shutdown processing mechanism：

```
1. Unregister service from registry
2. Notify the client of the readonly event in real time（readonly event）
3. If the client settings timeout is not appropriate, the server will automatically close the connection, and the client will also receive notifications or timeouts.
```

The current elegant downtime work is relatively good.

carryxyh:
`Maybe keep the status of requests on DefaultFuture can be better?`
Ok, that is more clearly.
Fix soon :)

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
```diff
@@ -108,11 +129,34 @@ public static void sent(Channel channel, Request request) {
         }
     }
 
+    /**
+     * close a channel when a channel is inactive
+     * directly return the unfinished requests.
+     *
+     * @param channel
+     */
+    public static void closeChannel(Channel channel) {
```
beiwei30: I guess we may possibly avoid introducing `UNFINISHED_REQUESTS` by doing this:

```java
        for (long id : CHANNELS.keySet()) {
            if (channel.equals(CHANNELS.get(id))) {
                DefaultFuture future = getFuture(id);
                if (!future.isDone()) {
                    Response disconnectResponse = new Response(r.getId());
                    disconnectResponse.setStatus(Response.CHANNEL_INACTIVE);
                    disconnectResponse.setErrorMessage("Channel " + channel + " is inactive. Directly return the unFinished request.");
                    DefaultFuture.received(channel, disconnectResponse);
                }
            }
        }
```

carryxyh: Looks good to me! I’ll fix it soon.

PR #2197, opened 2018-08-07
[Dubbo-#2184] Return the hanging client requests immediately when connection is broken #2184

guangqianpeng:
Related issue： https://github.com/apache/incubator-dubbo/issues/2184
Return the hanging client requests immediately when connection is broken.

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
@@ -107,6 +107,13 @@ public class DefaultFuture implements ResponseFuture {
         }
     }
 
+    public static void disconnected(Channel channel) {
+        for (long id: FUTURES.keySet()) {
+            DefaultFuture future = FUTURES.get(id);
+            future.doDisconnected();
+        }
+    }
+
     @Override
     public Object get() throws RemotingException {
         return get(timeout);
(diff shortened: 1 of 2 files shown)


dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
```diff
@@ -107,6 +107,13 @@ public static void received(Channel channel, Response response) {
         }
     }
 
+    public static void disconnected(Channel channel) {
```
carryxyh: If the client connect to several server, DefaultFuture will keep all of the future which unfinished.
This method will disconnect all the futures which bind to different server?

carryxyh: Maybe it will wake up a DefaultFuture that is bind to a server that didn't close ?

guangqianpeng: Yes. Maybe we can check if future.channel==channel

carryxyh: That is.
If not, the future will be waken up incorrect and return a incorrect value to user.

guangqianpeng: If the future is waken up and the channel is still connected, the future will await again. We should adjust the timeout...

carryxyh: Yeah, another problem.
:)

chickenlj:
#2185 tries to solve the same problem.

PR #2206, opened 2018-08-07
[Dubbo-issue-2184] Fix Return the hanging client requests ASAP. when connection is broken unexpectedly. #2184

daryl666:

Fix Return the hanging client requests ASAP. when connection is broken unexpectedly. #2184

## Brief changelog

add the configuration of heartbeat reconnect, default true, if set false, when heartbeat timeout, consumer disconnect

## Verifying this change

org.apache.dubbo.common.Constants
org.apache.dubbo.remoting.exchange.support.header.HeartBeatTask

From Baiji, Term 268, Team 3, Issue #2184, Q3-5

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeartBeatTask.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeartBeatTask.java
@@ -17,6 +17,7 @@
 
 package org.apache.dubbo.remoting.exchange.support.header;
 
+import org.apache.dubbo.common.Constants;
 import org.apache.dubbo.common.Version;
 import org.apache.dubbo.common.logger.Logger;
 import org.apache.dubbo.common.logger.LoggerFactory;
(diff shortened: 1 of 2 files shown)


dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeartBeatTask.java
```diff
@@ -70,7 +71,8 @@ public void run() {
                     if (lastRead != null && now - lastRead > heartbeatTimeout) {
                         logger.warn("Close channel " + channel
                                 + ", because heartbeat read idle time out: " + heartbeatTimeout + "ms");
-                        if (channel instanceof Client) {
+
+                        if (channel instanceof Client && channel.getUrl().getParameter(Constants.HEARTBEAT_RECONNECT, Constants.DEFAULT_HEARTBEAT_RECONNECT)) {
```
carryxyh: This op can not make the hanging request return directly. It just control the reconnect.
I didn't think this pr can solve this problem.
Maybe u should fix or send another pr 
:)

chickenlj: Agree. This patch didn't solve the problem

PR #2209, opened 2018-08-07
[Dubbo-2131] Check if we can switch serialization/deser between IO and Dubbo Threads.

kid1994:

add ability to switch serialization/deser between IO and Dubbo Threads.

## Brief changelog

change dubbo xml schema
add one "decodeInIo" property to dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/ProtocolConfig.java
change "DECODE_IN_IO_THREAD_KEY " to "decodeinio"

## Verifying this change

write an unit-test and passed

### DIFF
--- a/dubbo-common/src/main/java/com/alibaba/dubbo/common/Constants.java
+++ b/dubbo-common/src/main/java/com/alibaba/dubbo/common/Constants.java
@@ -589,8 +589,7 @@ public class Constants {
 
     public static final long DEFAULT_TPS_LIMIT_INTERVAL = 60 * 1000;
 
-    //public static final String DECODE_IN_IO_THREAD_KEY = "decode.in.io";
-    public static final String DECODE_IN_IO_THREAD_KEY = "decodeinio";
+    public static final String DECODE_IN_IO_THREAD_KEY = "decode.in.io";
 
     public static final boolean DEFAULT_DECODE_IN_IO_THREAD = true;
 
--- a/dubbo-common/src/main/java/com/alibaba/dubbo/common/Constants.java
+++ b/dubbo-common/src/main/java/com/alibaba/dubbo/common/Constants.java
@@ -589,7 +589,8 @@ public class Constants {
 
     public static final long DEFAULT_TPS_LIMIT_INTERVAL = 60 * 1000;
 
-    public static final String DECODE_IN_IO_THREAD_KEY = "decode.in.io";
+    //public static final String DECODE_IN_IO_THREAD_KEY = "decode.in.io";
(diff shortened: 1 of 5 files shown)


dubbo-rpc/dubbo-rpc-dubbo/src/main/java/com/alibaba/dubbo/rpc/protocol/dubbo/DubboCodec.java
```diff
@@ -90,6 +90,7 @@ protected Object decodeBody(Channel channel, InputStream is, byte[] header) thro
                             result = new DecodeableRpcResult(channel, res, is,
                                     (Invocation) getRequestData(id), proto);
                             result.decode();
+                            System.out.println(Thread.currentThread().getName());
```
chickenlj: Remove this line

dubbo-common/src/main/java/com/alibaba/dubbo/common/Constants.java
```diff
@@ -589,7 +589,8 @@
 
     public static final long DEFAULT_TPS_LIMIT_INTERVAL = 60 * 1000;
 
-    public static final String DECODE_IN_IO_THREAD_KEY = "decode.in.io";
+    //public static final String DECODE_IN_IO_THREAD_KEY = "decode.in.io";
+    public static final String DECODE_IN_IO_THREAD_KEY = "decodeinio";
```
chickenlj: This change will cause the incompatible problem for users already used this property.

Use @Parameter annotation to solve this problem, see the comment below.

dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/ProtocolConfig.java
```diff
@@ -454,6 +456,14 @@ public void setExtension(String extension) {
         this.extension = extension;
     }
 
+    public Boolean getDecodeInIo() {
```
chickenlj: Add @Parameter(key = "decode.in.io")

kid1994: thank you for your repley.there is another group who use the similiar way,and it seems they  dont have to change the src code.I think their way is better.

chickenlj:
1. If we put this property on ProtocolConfig, then it's only available on Provider side.
2. Avoid Provider side configurations affect that on the Consumer side.

Currently, this demo works. If you fixed the above problems, feel free to reopen it. Thanks.

Issue #2187, opened 2018-08-06
Use CompletableFuture to rewrite the embedded ListenableFuture.


chickenlj:
We have some functionalities which were implemented with ListenableFuture, ListenableFuture is supplementary for Future introduced in Java 5. Now I think we can rewrite those codes with CompletableFuture provided by Java 8.

* The ListenableFuture related classes are defined here: https://github.com/apache/incubator-dubbo/tree/master/dubbo-common/src/main/java/org/apache/dubbo/common/concurrent
* Classes use ListenableFuture
  *  ZkClientWrapper. https://github.com/apache/incubator-dubbo/blob/master/dubbo-remoting/dubbo-remoting-zookeeper/src/main/java/org/apache/dubbo/remoting/zookeeper/zkclient/ZkClientWrapper.java
  * AbstractMonitorFactory. https://github.com/apache/incubator-dubbo/blob/master/dubbo-monitor/dubbo-monitor-api/src/main/java/org/apache/dubbo/monitor/support/AbstractMonitorFactory.java

Issue #2178, opened 2018-08-03
Please consider supporting Java 8 Date/Time type when serializing with Kryo


kennyliao1982:
### Environment

* Dubbo version: 2.6.2
* Java version: 1.8

I am using java.time.LocalDateTime and Dubbo complains "**class java.time.LocalDateTime has no zero-arg constructor and this will affect the serialization performance**", which is caused by Dubbo's wrapper class "**CompatibleKryo**"

Any thoughts ??

```
package com.alibaba.dubbo.common.serialize.kryo;

import com.alibaba.dubbo.common.logger.Logger;
import com.alibaba.dubbo.common.logger.LoggerFactory;
import com.alibaba.dubbo.common.serialize.kryo.utils.ReflectionUtils;

import com.esotericsoftware.kryo.Kryo;
import com.esotericsoftware.kryo.Serializer;
import com.esotericsoftware.kryo.serializers.JavaSerializer;

public class CompatibleKryo extends Kryo {

    private static final Logger logger = LoggerFactory.getLogger(CompatibleKryo.class);

    @Override
    public Serializer getDefaultSerializer(Class type) {
        if (type == null) {
            throw new IllegalArgumentException("type cannot be null.");
        }

        if (!type.isArray() && !type.isEnum() && !ReflectionUtils.checkZeroArgConstructor(type)) {
            if (logger.isWarnEnabled()) {
                logger.warn(type + " has no zero-arg constructor and this will affect the serialization performance");
            }
            return new JavaSerializer();
        }
        return super.getDefaultSerializer(type);
    }
}
```

jtanza:
hi, I haven't contributed before but would like to take this one on, is that okay? I am not actually sure of how to best address this one, but may get a better handle on things after taking a closer look at this portion of the code

jtanza:
Awesome, will do!

chickenlj:
I think this snippet works, only too much warn logs, is that right?
```java
if (!type.isArray() && !type.isEnum() && !ReflectionUtils.checkZeroArgConstructor(type)) {
            if (logger.isWarnEnabled()) {
                logger.warn(type + " has no zero-arg constructor and this will affect the serialization performance");
            }
            return new JavaSerializer();
        }
```

chickenlj:
It seems that Kryo 4.0.1 used by Dubbo already support Java 8 Date types. It's the CompatibleKryo short-circuited that logic by checking zero-arg constructor.

PR #2210, opened 2018-08-07
[Dubbo-issue-2178] Please consider supporting Java 8 Date/Time type when serializing with Kryo #2178

daryl666:

Please consider supporting Java 8 Date/Time type when serializing with Kryo #2178

## Brief changelog

add the if statement of java8 date/time when getDeclaredConstructor

## Verifying this change

org.apache.dubbo.common.serialize.kryo.CompatibleKryo

From Baiji, Term 268, Team 3, Issue #2187, Q3-4

### DIFF
--- a/dubbo-serialization/dubbo-serialization-kryo/src/main/java/org/apache/dubbo/common/serialize/kryo/CompatibleKryo.java
+++ b/dubbo-serialization/dubbo-serialization-kryo/src/main/java/org/apache/dubbo/common/serialize/kryo/CompatibleKryo.java
@@ -24,6 +24,10 @@ import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.Serializer;
 import com.esotericsoftware.kryo.serializers.JavaSerializer;
 
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+
 public class CompatibleKryo extends Kryo {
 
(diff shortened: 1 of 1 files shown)


dubbo-serialization/dubbo-serialization-kryo/src/main/java/org/apache/dubbo/common/serialize/kryo/CompatibleKryo.java
```diff
@@ -34,7 +38,7 @@ public Serializer getDefaultSerializer(Class type) {
             throw new IllegalArgumentException("type cannot be null.");
         }
 
-        if (!type.isArray() && !type.isEnum() && !ReflectionUtils.checkZeroArgConstructor(type)) {
+        if (!type.isArray() && !type.isEnum() && !type.equals(LocalDate.class) && !type.equals(LocalDateTime.class) && !type.equals(LocalTime.class) && !ReflectionUtils.checkZeroArgConstructor(type)) {
```
chickenlj: Simply check types will not solve the problem.

chickenlj: Sorry, I've checked that this works, and kryo has internally registered some Serializers for Java 8 Time types. 

But we still have a problem, we cannot list all possible classes here.

chickenlj:
See #2220

Issue #2136, opened 2018-07-26
Replace hard coded version number of hessian-lite with unified property variable.


chickenlj:
Notice, this work should be done on branch 2.6.x.

https://github.com/apache/incubator-dubbo/tree/2.6.x

chickenlj:
https://github.com/apache/incubator-dubbo/blob/69311abe3a84996ec6b1a38a381fd39f48bbfdf3/dubbo-serialization/dubbo-serialization-api/pom.xml#L41

https://github.com/apache/incubator-dubbo/blob/69311abe3a84996ec6b1a38a381fd39f48bbfdf3/dubbo-serialization/dubbo-serialization-hessian2/pom.xml#L41

PR #2211, opened 2018-08-07
[Dubbo-issue-2136] Replace hard coded version number of hessian-lite with unified property variable. #2136

daryl666:

Replace hard coded version number of hessian-lite with unified property variable. #2136

## Brief changelog

Replace hard coded version number of hessian-lite with unified property variable. 

## Verifying this change

dubbo-bom/pom.xml

From Baiji, Term 268, Team 3, Issue #2136, Q3-2


chickenlj:
Please focus on one issue on a single patch.

PR #2267, opened 2018-08-14
fix issue system property dubbo.service.delay invalid #1728

liangjingwen:
## Team_10: fix issue system property dubbo.service.delay invalid #1728 

Issue [https://github.com/apache/incubator-dubbo/issues/1728](url) has been fixed

## Brief changelog

Modify method export() and doExport(), add method publish() in ServiceConfig.java, 

## Verifying this change

The issue is due to bug in the class ServiceConfig.

When dubbo exports a service, there is a bean of ServiceConfig create by spring and the method export() is invoked. The bean of ServiceConfig is initialized by the configuration <dubbo:service /> in .xml. So in the method export(), the variable delay is initialized by <dubbo:service delay="5000"/> and the system property does not work. 

To fix this issue, I judge the variable delay in the method doExport() and the judgement is after the method appendProperties() which can check the overrides of the priorities and set the variables correct values.


diecui1202:
move the export to doExport method may not a good solution.

Issue #2321, opened 2018-08-19
Does provider "async = true" need to be remove from url when register to registry?


fingthinking:
- [ ] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version:  2.7.0-SNAPSHOT

### Step to reproduce this issue

1. https://github.com/apache/incubator-dubbo/blob/master/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java
```java
 private URL getRegisteredProviderUrl(final Invoker<?> originInvoker) {
        URL providerUrl = getProviderUrl(originInvoker);
        //The address you see at the registry
        return providerUrl.removeParameters(getFilteredKeys(providerUrl))
                .removeParameter(Constants.MONITOR_KEY)
                .removeParameter(Constants.BIND_IP_KEY)
                .removeParameter(Constants.BIND_PORT_KEY)
                .removeParameter(QOS_ENABLE)
                .removeParameter(QOS_PORT)
                .removeParameter(ACCEPT_FOREIGN_IP)
                .removeParameter(VALIDATION_KEY)
                .removeParameter(INTERFACES);
    }
TO
 private URL getRegisteredProviderUrl(final Invoker<?> originInvoker) {
        URL providerUrl = getProviderUrl(originInvoker);
       // 此处移除async标记，仅保留标记在provider端
        Set<String> keySet = providerUrl.getParameters().keySet();
        List<String> asyncKey = new ArrayList<>();
        for(String key : keySet) {
            if(key.endsWith("." + Constants.ASYNC_KEY) || key.equals(Constants.ASYNC_KEY)) {
                asyncKey.add(key);
            }
        }
        //The address you see at the registry
        return providerUrl.removeParameters(getFilteredKeys(providerUrl))
                .removeParameter(Constants.MONITOR_KEY)
                .removeParameter(Constants.BIND_IP_KEY)
                .removeParameter(Constants.BIND_PORT_KEY)
                .removeParameter(QOS_ENABLE)
                .removeParameter(QOS_PORT)
                .removeParameter(ACCEPT_FOREIGN_IP)
                .removeParameter(VALIDATION_KEY)
                .removeParameter(INTERFACES)
[...]

PR #2322, opened 2018-08-19
[DUBBO-2321]Provider "async = true" need to be remove from url when register to registry#2321

fingthinking:

provider "async = true" need to be remove from url when register to registry

## Brief changelog

Issue: [https://github.com/apache/incubator-dubbo/issues/2321](https://github.com/apache/incubator-dubbo/issues/2321)


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java
```diff
@@ -222,6 +223,14 @@ private URL getRegistryUrl(Invoker<?> originInvoker) {
      */
     private URL getRegisteredProviderUrl(final Invoker<?> originInvoker) {
         URL providerUrl = getProviderUrl(originInvoker);
+        // 此处移除async标记，仅保留标记在provider端
```
diecui1202: pls. translate the Chinese comment to English

diecui1202:
@chickenlj

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java
```diff
@@ -231,7 +240,8 @@ private URL getRegisteredProviderUrl(final Invoker<?> originInvoker) {
                 .removeParameter(QOS_PORT)
                 .removeParameter(ACCEPT_FOREIGN_IP)
                 .removeParameter(VALIDATION_KEY)
-                .removeParameter(INTERFACES);
+                .removeParameter(INTERFACES)
+                .removeParameters(asyncKey);
```
carryxyh: Hi, 
I saw your ISSUE, I think async should be registered to the registry so that we can see if the provider supports asynchronous on dubbo-ops.
IMO, the better way is that the Consumer side does not use the Provider-side async, which can be analogous to the timeout override strategy in Consumer side.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java
```diff
@@ -222,6 +223,14 @@ private URL getRegistryUrl(Invoker<?> originInvoker) {
      */
     private URL getRegisteredProviderUrl(final Invoker<?> originInvoker) {
         URL providerUrl = getProviderUrl(originInvoker);
+        // Remove the async tag here, leaving only the tag at the provider, and not register to registry.
+        Set<String> keySet = providerUrl.getParameters().keySet();
+        List<String> asyncKey = new ArrayList<String>();
+        for(String key : keySet) {
+            if(key.endsWith("." + Constants.ASYNC_KEY) || key.equals(Constants.ASYNC_KEY)) {
```
carryxyh: need to format your code
:)

chickenlj:
I agree with what carryxyh says in #2321, we should fix this problem on both Provider side and Consumer side:
1. Provider side, please focus on dev-meta for the changes made by @cvictory.
2. Consumer side, #2434 already merged.

chickenlj:
Please reopen if necessary.

PR #2343, opened 2018-08-24
Enhance TagRouter: modify tag router strategy

kimmking:

Enhance TagRouter: modify tag router strategy

## Brief changelog

Thanks kirito for making a new big feature at first time.
The key point is with current implement, an invocation with a tag can't return all invokers in TagRouter.
So I suggest here:
Add a forceTag flag(TAG_FORCE_KEY = "tag.force"), default value is **TRUE**.
1. When it's true, TagRouter strickly match invokers' tag. 
    - An invocation with "black" tag only match invokers with "black" tag. 
    - An invocation with empty tag only match invokers with empty tag.  
    - An invocation match no invokers will return no invokers.  
2. When it's false(default), TagRouter match invokers' tag by the most invokers strategy. 
    - An invocation with "black" tag match invokers with "black" tag. If no invokers with "black" tag, return all invokers.
    - An invocation with empty tag will match all invokers.
    - An invocation match no invokers will return all invokers.  
Then we can cover all conditions for using TagRouter.

## Verifying this change

UT passed.


lexburner:
> When forceTag is  true, TagRouter strickly match invokers' tag.
> An invocation with "black" tag only match invokers with "black" tag. 
I think empty invokers may be seen as a kind of special invoker，becasue there will be some old application don't want to use TagRouter. An invocation with 'black' tag can downgrade to empty invokers, while an invocation with empty tag can NOT upgrade to COLORFUL invokers.

No problem in other aspects，forceTag is a good idea to distinguish between two strategies.

kimmking:
@lexburner some old application don't want to use TagRouter, so force tag default is false, an old application with no tag will can't see all color, all invokers can be consider no tags.

lexburner:
`forceTag` may be a good idea determining the different intention of TagRouter, but this version's change conflicts with my original design about `TagRouter`.In my opinion, its behavior should looks like:

> An invocation with 'black' tag can downgrade to empty tag invokers, but can't downgrade to ‘red’ tag invokers. Only empty tag invokers can be seen as a secondary choice when `forceTag` is false, not all invokers.

Maybe a third participant(or more) is need to discuss this design, otherwise the origin design is enough（just works for me）.

kimmking:
can you merge it?

beiwei30:
@kimmking @lexburner have you guys reach the final agreement on this?

chickenlj:
Please focus on the new implementation, it has considered the `force` and `failover` strategy mentioned in here. 
https://github.com/apache/incubator-dubbo/blob/dev-metadata/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java

PR #2363, opened 2018-08-28
[Dubbo-issue1814][ Baiji-02] XML configuration to increase graceful downtime

a932846905:
增加优雅停机的xml配置


chickenlj:
I have tested and it works, but we'd better move this demo to samples.

PR #2368, opened 2018-08-28
[Dubbo-issue1054][Baiji-02]Support to send the Consumer application name to the Provider.

w1014804721:
### Get the application name of the Consumer by calling the getConsumerAppName method of RpcContext.


chickenlj:
I think it can work, but only for DubboProtocol, it would be better if consider other protocols, such as HessianProtocol, etc.

No need of extra app properties.

PR #2399, opened 2018-08-29
if to switch case

vaccywen:

to improve efficiency

## Brief changelog

org.apache.dubbo.common.compiler.support.ClassUtils

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/ClassUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/ClassUtils.java
@@ -79,40 +79,43 @@ public class ClassUtils {
     }
 
     public static Class<?> _forName(String className) throws ClassNotFoundException {
-        if ("boolean".equals(className))
-            return boolean.class;
-        if ("byte".equals(className))
-            return byte.class;
-        if ("char".equals(className))
-            return char.class;
-        if ("short".equals(className))
-            return short.class;
-        if ("int".equals(className))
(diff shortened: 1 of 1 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/ClassUtils.java
```diff
@@ -79,40 +79,43 @@ public static Object newInstance(String name) {
...
+                case "int[]":
+                    return int[].class;
+                case "long[]":
+                    return long[].class;
+                case "float[]":
+                    return float[].class;
+                case "double[]":
+                    return double[].class;
+                default:
+                    return arrayForName(className);
```
diecui1202: IMO, try catch block should only effect on default case.

zonghaishang:
I don't think this change will improve efficiency, but it helps readability.

vaccywen:
Speed: A switch statement might prove to be faster than ifs provided number of cases are good. If there are only few cases, it might not effect the speed in any case. Prefer switch if the number of cases are more than 5 otherwise, you may use if-else too.
 If a switch contains more than five items, it’s implemented using a lookup table or a hash list. This means that all items get the same access time, compared to a list of if:s where the last item takes much more time to reach as it has to evaluate every previous condition first.

https://www.geeksforgeeks.org/switch-vs-else/

diecui1202:
@vaccywen pls. check the review comments.

vaccywen:
@zonghaishang thank you.

PR #2433, opened 2018-09-04
[Dubbo-random enhance] replace Random with ThreadLocalRandom

lexburner:
## What is the purpose of the change

A little change, replace `Random` with `ThreadLocalRandom` .

Use of `ThreadLocalRandom` rather than shared `Random` objects in concurrent programs will typically encounter much less overhead and contention.

Use of `ThreadLocalRandom` is particularly appropriate when multiple tasks .

## Brief changelog

replace `Random` with `ThreadLocalRandom` in `LeastActiveLoadBalance` and `RandomLoadBalance`.

## Verifying this change

reuse the test case already exist.

Follow this checklist to help us incorporate your contribution quickly and easily:

- [x] Make sure there is a [GITHUB_issue](https://github.com/apache/incubator-dubbo/issues) filed for the change (usually before you start working on it). Trivial changes like typos do not require a GITHUB issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [x] Format the pull request title like `[Dubbo-XXX] Fix UnknownException when host config not exist #XXX`. Each commit in the pull request should have a meaningful subject line and body.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/apache/incubator-dubbo/tree/master/dubbo-test).
- [x] Run `mvn clean install -DskipTests` & `mvn clean test-compile failsafe:integration-test` to make sure unit-test and integration-test pass.
- [x] If this contribution is large, please follow the [Software Donation Guide](https://github.com/apache/incubator-dubbo/wiki/Software-donation-guide).

PR #2439, opened 2018-09-05
add JDK7 ThreadLocalRandom replace Random in concurrent env can promote performance

manzhizhen:

The performance of RandomLoadBalance and LeastActiveLoadBalance is improved by replacing Random with ThreadLocalRandom. In order to support JDK1.6, the implementation of ThreadLocalRandom is copied directly.

## Brief changelog

XXXXX

## Verifying this change

XXXXX


lexburner:
It is a little bit coincidental. 
see the pr `https://github.com/apache/incubator-dubbo/pull/2433` , i has already submit the same change a few days ago.

satansk:
And more confusing, why add a JDK `ThreadLocalRandom` source code to Dubbo, seems unnecessary.

dubbo-common/src/main/java/org/apache/dubbo/common/threadlocal/ThreadLocalRandom.java
```diff
@@ -0,0 +1,188 @@
+package org.apache.dubbo.common.threadlocal;
```
carryxyh: Without license.
Is this ThreadLocalRandom from Netty?
If yes, please add Netty license.

carryxyh:
Hi, can u provide a test case of performance?
If the difference is not big compared to the jdk version, I prefer to use the jdk version. Porting code takes more time to maintain the features of the new version

lexburner:
I notice that your `ThreadLocalRandom` was copied from netty3.

maybe more details need to take as consideration.

1. third-party dependency need use carefully.
2. bugs fix need to trace from upstream repository.
3. performance benchmark

```
@BenchmarkMode({Mode.AverageTime})
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@Warmup(iterations = 3, time = 5)
@Measurement(iterations = 3, time = 5)
@Threads(50)
@Fork(1)
@State(Scope.Benchmark)
public class RandomBenchmark {

    Random random = new Random();

    @Benchmark
    public int random() {
        return random.nextInt();
    }

    @Benchmark
    public int threadLocalRandom() {
        return ThreadLocalRandom.current().nextInt();
    }

    @Benchmark
    public int nettyThreadLocalRandom() {
        return io.netty.util.internal.ThreadLocalRandom.current().nextInt();
    }

    public static void main(String[] args) throws RunnerException {
        Options opt = new OptionsBuilder()
                .include(RandomBenchmark.class.getSimpleName())
                .build();

        new Runner(opt).run();
    }

}
```
and below is our benchmark result under 50 concurrency
```
Benchmark                               Mode  Cnt     Score      Error  Units
RandomBenchmark.nettyThreadLocalRandom  avgt    3   160.321 ±   83.518  ns/op
RandomBenchmark.random                  avgt    3  3281.972 ± 2093.187  ns/op
RandomBenchmark.threadLocalRandom       avgt    3    91.505 ±   39.702  ns/op
```
Maybe i did something wrong with `netty.ThreadLocalRandom`, but from the result we can know the truth
> JDK 's ThreadLocalRandom looks good.

carryxyh:
Agree with @lexburner 
Pls focus on https://github.com/apache/incubator-dubbo/pull/2433 .
Reopen if there are other optimization reasons

manzhizhen:
This ThreadLocalRandom is copy from JDK1.7，The reason for this is to be compatible with JDK1.6, otherwise Dubbo will not be able to run in JDK6.

PR #2457, opened 2018-09-06
Support multiple shared links

manzhizhen:

Dubbo By default, all service calls between a Provider and a Consumer instance share a tcp connection. In high concurrency situations, there will be a 40ms network delay, which is caused by the tcp throughput algorithm related restrictions (note, Dubbo has closed Nagel). So in order to avoid this network delay, dubbo supports separate tcp connection configuration for some services (for example, configure connections="2" in dubbo:reference to use two separate tcp connections for the service call). Why not change the default shared 1 tcp connection to dynamic configuration? Let's implement this now! !

## Brief changelog

XXXXX

## Verifying this change

XXXXX


diecui1202:
@manzhizhen Pls. check the ci failure.

```
[[1;31mERROR[m] Failed to execute goal [32morg.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:check[m [1m(checkstyle-validation)[m on project [36mdubbo-common[m: [1;31mFailed during checkstyle configuration[m: Exception was thrown while processing /home/travis/build/apache/incubator-dubbo/dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java: TokenStreamRecognitionException occurred while parsing file /home/travis/build/apache/incubator-dubbo/dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java. Unexpected character 0x60 in identifier -> [1m[Help 1][m
```

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
```diff
@@ -21,10 +21,7 @@
 import org.apache.dubbo.common.extension.ExtensionLoader;
 import org.apache.dubbo.common.serialize.support.SerializableClassRegistry;
 import org.apache.dubbo.common.serialize.support.SerializationOptimizer;
-import org.apache.dubbo.common.utils.ConcurrentHashSet;
-import org.apache.dubbo.common.utils.ConfigUtils;
-import org.apache.dubbo.common.utils.NetUtils;
-import org.apache.dubbo.common.utils.StringUtils;
+import org.apache.dubbo.common.utils.*;
```
diecui1202: DO NOT use import *

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
```diff
@@ -47,11 +44,7 @@
 import org.apache.dubbo.rpc.protocol.AbstractProtocol;
 
 import java.net.InetSocketAddress;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
diecui1202: the same problem

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -116,13 +116,15 @@
 
     public static final int DEFAULT_THREADS = 200;
 
-    public static final boolean DEFAULT_KEEP_ALIVE = true;
+    public static final boolean DEFAULT_KEEP_ALIVE = true;`````
```
diecui1202: What is this?

manzhizhen:
Here I share the pressure measurement data:
program execution time：30min
provider logic：direct new object return
provider thread pool size：800
consumer timeout：1s
provider timeout：default
service agreement：dubbo

result:

consumer concurrent | tcp links | total invoke times | invoke fail（timeout）times | service invoke success rate | service invoke QPS | service invoke avg cost time | invoke time more than 100ms rate | （consumer cost time - provider cost time） avg time
500 | 3links | 49287931 | 1771 | 99.9964% | 27271 | 18.1ms | 0.006% | 0.31ms
500 | 2links | 47601583 | 1141 | 99.9976% | 26464 | 18.9ms | 0.006% | 0.38ms
500 | 2links | 47871439 | 1146 | 99.9976% | 26919 | 18.7ms | 0.006% | 0.41ms
500 | 1links | 46258320 | 1051 | 99.9977% | 25702 | 18.8ms | 0.007% | 8.99ms
500 | 1links | 46421769 | 1005 | 99.9978% | 25910 | 18.7ms | 0.007% | 9.84ms
200 | 3links | 55025248 | 478 | 99.9991% | 31066 | 6.4ms | 0.004% | 0.16ms
200 | 1links | 47376374 | 255 | 99.9995% | 26362 | 7.1ms | 0.003% | 4.89ms
100 | 2links | 55366764 | 0 | 100% | 30640 | 3.2ms | 0.0002% | 0.13ms

We can see that in the case of concurrent 500 calls, the throughput of two tcp connection transmissions is 3.5% higher than that of one tcp connection transmission, and the throughput of three tcp connection transmissions is increased by 6.5% compared with the transmission of one tcp connection. In terms of network time consumption (consumer processing time minus provider processing time), 1 tcp connection per call on the network takes an average of 9.84ms, and 2 tcp or 3 tcp connections are called each time in the network. The average time spent on it is only 0.41ms.

We can see that in the case of concurrent 300 calls, the throughput of three tcp connection transmissions is 16.1% higher than that of one tcp connection transmission! ! This is an amazing number.

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
```diff
@@ -384,31 +379,37 @@ private void optimizeSerialization(URL url) throws RpcException {
...
-        if (client != null) {
-            if (!client.isClosed()) {
-                client.incrementAndGetCount();
-                return client;
-            } else {
-                referenceClientMap.remove(key);
-            }
+        List<ReferenceCountExchangeClient> clients = referenceClientMap.get(key);
+        if(clients == null) {
+            List<ReferenceCountExchangeClient> referenceCountExchangeClients = buildReferenceCountExchangeClientList(url, key, connectNum);
```
carryxyh: ReferenceCountExchangeClient is for sharing connections. I saw your pr. Your pr wants to create multiple connections instead of sharing connections by default. In this case, what is the purpose of using ReferenceCountExchangeClient?

manzhizhen: It's reasonable. I'll optimize it again.

carryxyh: Hello, if you just change the default share to multiple, then just change the default value of connections? How do u think about it?
:)

carryxyh:
Nice work!
The data for your stress test is very great!
Regarding your pr, I think that directly modifying the default values ​​of connections can achieve the same effect. U can just have try.
:)

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
manzhizhen: Is it convenient to leave a contact? My phone number: 18668225833

manzhizhen:
@carryxyh Can Dubbo now set the default shared connection number to more than one?

manzhizhen:
Here is the physical machine configuration that complements the performance test：

属性 | 值
cpu | 48核
内存 | 128G
磁盘 | HDD 4T
网卡带宽 | 10000Mb/s

manzhizhen:
@carryxyh  @diecui1202  两位大佬瞅瞅呗

kimmking:
👍一个

kimmking:
👍一个

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -122,7 +122,9 @@
 
     public static final int DEFAULT_ALIVE = 60 * 1000;
 
-    public static final int DEFAULT_CONNECTIONS = 0;
+    public static final String DEFAULT_CONNECTIONS = "1";
```
kimmking: int is better.

manzhizhen: I modified the DEFAULT_CONNECTIONS to see the code, which is not used before. The reason is changed to String type, because it is convenient with ConfigUtils.getProperty, haha

manzhizhen:
@beiwei30  By default, a jvm consumer and jvm provider instance share a TCP long connection. To solve the throughput bottleneck of a TCP long connection, dubbo can use the CONNECTIONS_KEY key to set TCP for a specific interface. Long connection. The new DEFAULT_CONNECTIONS_KEY key is an extension of the default sharing of one TCP long connection, so that all interfaces can enjoy the throughput of multiple TCP connections without having to separately call an interface. Configure the connection.

  Set CONNECTIONS_KEY="0" and set DEFAULT_CONNECTIONS_KEY="3". This is not the same as setting CONNECTION_KEY="3" directly. The former is the call sharing of all interfaces using 3 TCP connections, the latter is a setting [ Connections] The property's interface call uses a separate 3 TCP long connections, and the other interface calls still use the default shared 1 TCP long connection.

beiwei30:
> @beiwei30 By default, a jvm consumer and jvm provider instance share a TCP long connection. To solve the throughput bottleneck of a TCP long connection, dubbo can use the CONNECTIONS_KEY key to set TCP for a specific interface. Long connection. The new DEFAULT_CONNECTIONS_KEY key is an extension of the default sharing of one TCP long connection, so that all interfaces can enjoy the throughput of multiple TCP connections without having to separately call an interface. Configure the connection.
> 
> Set CONNECTIONS_KEY="0" and set DEFAULT_CONNECTIONS_KEY="3". This is not the same as setting CONNECTION_KEY="3" directly. The former is the call sharing of all interfaces using 3 TCP connections, the latter is a setting [connections] The property's interface call uses a separate 3 TCP long connections, and the other interface calls still use the default shared 1 TCP long connection.

Now I get the basic idea, I will give further comments on the current impl.

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
```diff
@@ -383,32 +388,57 @@ private void optimizeSerialization(URL url) throws RpcException {
...
         }
+
+        return clients;
+    }
+
+    private ReferenceCountExchangeClient buildReferenceCountExchangeClient(URL url, String key) {
+
+        ExchangeClient exchangeClient = initClient(url);
+
+        ReferenceCountExchangeClient client = new ReferenceCountExchangeClient(exchangeClient, ghostClientMap);
```
beiwei30: `ghostClientMap` is a mapping between URL and client, but `referenceClientMap` is a mapping between URL and a list of clients.

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
```diff
@@ -383,32 +388,57 @@ private void optimizeSerialization(URL url) throws RpcException {
 
     /**
      * Get shared connection
+     *
+     * @param url
+     * @param connectNum
      */
-    private ExchangeClient getSharedClient(URL url) {
+    private List<ReferenceCountExchangeClient> getSharedClient(URL url, int connectNum) {
```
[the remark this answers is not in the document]
manzhizhen: You are right, this does have a concurrency problem, I fix it, haha！！

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
manzhizhen: Delayed connections are used to reduce the number of long connections. Create a long connection when a call is initiated. <dubbo:protocol name="dubbo" lazy="true" />Note: This configuration only works for dubbo protocols that use long connections.
Later I read the code about the delayed connection and found that LazyConnectExchangeClient has two purposes:
1. Create a connection like the delay mentioned earlier, minimizing unnecessary long connection overhead.
2. A kind of protection when closing the connection resource. When the ReferenceCountExchangeClient is closed (multiple URLs refer to the same ReferenceCountExchangeClient, the last URL that will reference the counter to 0 can actually close the ReferenceCountExchangeClient) will be replaced with LazyConnectExchangeClient. When used again, the warn log will be printed. The put operation of ghostClientMap is only available here.

So, no matter how many ExchangeClients a URL corresponds to, it only needs a ghost connection in the ghostClientMap, which is a LazyConnectExchangeClient.

Also, only the shared ExchangeClient will use the ReferenceCountExchangeClient, so the default client used to share a TCP connection is the ReferenceCountExchangeClient, and the separate TCP connection set by connections does not need or use the ReferenceCountExchangeClient.

延迟连接用于减少长连接数。当有调用发起时,再创建长连接。<dubbo:protocol name="dubbo" lazy="true" />注意:该配置只对使用长连接的 dubbo 协议生效。 
后面我有仔细阅读了关于延迟连接的代码，发现LazyConnectExchangeClient有两个用途：
1. 像前面提到的延迟创建连接，尽可能的减少不必要的长连接开销。
2. 关闭连接资源时的一种保护，当ReferenceCountExchangeClient被关闭时（多个URL引用同一个ReferenceCountExchangeClient时，最后那个将引用计数器变为0的URL可以真正关闭该ReferenceCountExchangeClient），会被替换成LazyConnectExchangeClient，一旦再被使用，会打印warn日志。而ghostClientMap的put操作，只有在这里才有。

所以，不管一个URL对应多少个ExchangeClient，它在ghostClientMap中最多只需要一个幽灵连接，即一个LazyConnectExchangeClient。

并且，只有共享的ExchangeClient才会用到ReferenceCountExchangeClient，所以默认的共享1条TCP连接所使用的client就是ReferenceCountExchangeClient，而通过connections设置的单独TCP连接不需要也没有使用ReferenceCountExchangeClient。

manzhizhen: If I understand the mistake, please correct me, thank you

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -122,7 +122,9 @@
...
+
+    public static final String DEFAULT_CONNECTIONS_KEY = "default.connections.key";
```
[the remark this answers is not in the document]
manzhizhen: Good idea!

chickenlj:
Basically, I agree to support add `connections` on instance level, not for performance or resource concern (because I think keep-alive connection has really small costs) but only for easy to use concern.

```xml
<dubbo:consumer connections="3" />
```

I will try to dive into the impl now.

KamenRiderKuuga:
Since Nagel has been closed as you said, the 40ms delay should not exist, why do you need to provide this?
@manzhizhen 
https://github.com/apache/dubbo/blob/d42b93dfda471b448cf6ccea364ba402e58615ce/dubbo-remoting/dubbo-remoting-netty/src/main/java/org/apache/dubbo/remoting/transport/netty/NettyClient.java#L69

https://github.com/apache/dubbo/blob/d42b93dfda471b448cf6ccea364ba402e58615ce/dubbo-remoting/dubbo-remoting-netty/src/main/java/org/apache/dubbo/remoting/transport/netty/NettyServer.java#L83

PR #2505, opened 2018-09-14
Condition  applicationContext != null is always true

chenlushun:
Remove invalid judgement.


carryxyh:
I personally think that adding this check is more rigorous.

PR #2509, opened 2018-09-16
code promote: move constant field to Constants class

channgli:
to make to code more standard and clear

Just move constant field to Constants class and "/" will use Constants.PROTOCOL_SEPARATOR.

Thanks


carryxyh:
LGTM

PR #2529, opened 2018-09-19
Optimization unnecessary unboxing 'b.booleanValue()' and Remove invalid  throws java.lang.Exception

chenlushun:
1. Optimization unnecessary unboxing 'b.booleanValue()'
2. Remove invalid  throws java.lang.Exception


dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/ReferenceBean.java
```diff
@@ -79,13 +77,13 @@ public boolean isSingleton() {
 
     @Override
     @SuppressWarnings({"unchecked"})
-    public void afterPropertiesSet() throws Exception {
```
diecui1202: IMO, this throws Exception need to keep. Check org.springframework.beans.factory.InitializingBean#afterPropertiesSet

chenlushun: well

diecui1202:
looks good to me.

Issue #2427, opened 2018-09-02
Could not configure number of forks of ForkingCluster in XML


ralf0131:
If specify number of forks in dubbo-demo-consumer.xml

```xml
 <dubbo:reference id="demoService" check="false" interface="org.apache.dubbo.demo.DemoService" cluster="forking" forks="3" />
```

The following error will occur:

```
Exception in thread "main" org.springframework.beans.factory.xml.XmlBeanDefinitionStoreException: Line 33 in XML document from class path resource [META-INF/spring/dubbo-demo-consumer.xml] is invalid; nested exception is org.xml.sax.SAXParseException; lineNumber: 33; columnNumber: 127; cvc-complex-type.3.2.2: Attribute 'forks' is not allowed to appear in element 'dubbo:reference'.
	at org.springframework.beans.factory.xml.XmlBeanDefinitionReader.doLoadBeanDefinitions(XmlBeanDefinitionReader.java:399)
	at org.springframework.beans.factory.xml.XmlBeanDefinitionReader.loadBeanDefinitions(XmlBeanDefinitionReader.java:336)
	at org.springframework.beans.factory.xml.XmlBeanDefinitionReader.loadBeanDefinitions(XmlBeanDefinitionReader.java:304)
	at org.springframework.beans.factory.support.AbstractBeanDefinitionReader.loadBeanDefinitions(AbstractBeanDefinitionReader.java:181)
	at org.springframework.beans.factory.support.AbstractBeanDefinitionReader.loadBeanDefinitions(AbstractBeanDefinitionReader.java:217)
	at org.springframework.beans.factory.support.AbstractBeanDefinitionReader.loadBeanDefinitions(AbstractBeanDefinitionReader.java:188)
	at org.springframework.beans.factory.support.AbstractBeanDefinitionReader.loadBeanDefinitions(AbstractBeanDefinitionReader.java:252)
	at org.springframework.context.support.AbstractXmlApplicationContext.loadBeanDefinitions(AbstractXmlApplicationContext.java:127)
	at org.springframework.context.support.AbstractXmlApplicationContext.loadBeanDefinitions(AbstractXmlApplicationContext.java:93)
	at org.springframework.context.support.AbstractRefreshableApplicationContext.refreshBeanFactory(AbstractRefreshableApplicationContext.java:129)
	at org.springframework.context.support.AbstractApplicationContext.obtainFreshBeanFactory(AbstractApplicationContext.java:614)
[...]

PR #2568, opened 2018-09-27
Add 'forks' config into xml config

carryxyh:
issue:
https://github.com/apache/incubator-dubbo/issues/2427

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractMethodConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractMethodConfig.java
@@ -64,6 +64,19 @@ public abstract class AbstractMethodConfig extends AbstractConfig {
     // customized parameters
     protected Map<String, String> parameters;
 
+    /**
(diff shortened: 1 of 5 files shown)


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractMethodConfig.java
```diff
@@ -64,6 +64,19 @@
     // customized parameters
     protected Map<String, String> parameters;
 
+    /**
+     * forks for forking cluster
+     */
+    protected Integer forks;
```
htynkn: I think this is a new field? where do we use this value?

carryxyh: Yep, new field.
It acts on the forkingcluster and specifies the opening of forks for concurrent access. U can see the issue for more details.
:)

carryxyh:
@beiwei30 
Is this pr should be bring back to 2.6.x?
I think it should be.

PR #2620, opened 2018-10-09
Fix incorrect descriptions for dubbo-serialization module.

carryxyh:
Fix incorrect descriptions for dubbo-serialization module.
Simple fix which doesn't need test.


carryxyh:
Will bring back to 2.6.x.

PR #2622, opened 2018-10-09
Fixed 2.6.x branch a minor issue like #2595

manzhizhen:
Fixed 2.6.x branch a minor issue with doConnect not using getConnectTimeout() in NettyClient，like  #2595 
@carryxyh


carryxyh:
Nice work!
LGTM.

Issue #2030, opened 2018-07-05
[Proposal]Metadata(URL) problems and possible solutions


chickenlj:
Hello, community

We have previously roughly discussed the problems with Dubbo's metadata in GitHub issues and mailing list,  here when saying metadata I mainly refer to URL in Dubbo. Now I think it's time to bring the discussion onto the table and consider solving it.  Below is the proposal I've drafted:

**Problems** 

URL works as the carrier for Dubbo internal data transfer at runtime, and is also works as a protocol integrating with the external system: registration center, service query, service management, etc., resulting in serious coupling between the configuration and the registration center, and the data load pushed by the registration center is too large.

**Targets**

- Keep URL unchanged at runtime, and even enrich the runtime URL; Keep the immutability of the URL during a single invoke from start to end.
- Simplify Registry url as much as possible.
- Refactor the way of doing service configuration, rely on Registry currently, switch to third-party Config Center while keeping compatible with old Overrides protocol.
- Some data in URL are for service query purpose, we should register this part to another place instead of Registry. And it should be realized as extensible and users can decide to enable it or not.

**Solutions**

Here is a typical URL in Registry:
```
dubbo://10.122.111.22:20880/com.xxx.compose.ic.service.vas.ValueAddServiceCompose?anyhost=true&application=goods-compose&default.actives=400&default.delay=-1&default.dispatcher=all&default.group=online&default.loadbalance=leastactive&default.service.filter=-monitor&default.threads=400&default.timeout=2000&default.version=1.0&delay=1&dubbo=2.6.2&environment=product&interface=com.xxx.compose.ic.service.vas.ValueAddServiceCompose&logger=slf4j&methods=queryValueAddServiceEditVOs,auditPassByValueAddServiceIdList,getValueAddServiceVOByServiceIds,getValueAddServiceInfoByGoodsIdList,getRuleMatchVOByGoodsIdList,getValueAddServiceEditVOById,getValueAddServiceEditVOByIds,queryValueAddServiceVOs,disableValueAddService,auditRejectByValueAddServiceIdList,saveValueAddServiceEditVO,enableValueAddService&organization=someorg&owner=somebody&pid=54812&revision=1.18.0628.3&side=provider&timestamp=1530671441040
```

We can see this URL showed here is really long and redundant, here are my suggestions to refactor it step by step:

1. Only keep necessary part in Registry
   For example: 
```
[...]

PR #2626, opened 2018-10-09
Simplify registry data and add a new service data store seperated from registry #2030

cvictory:

For issue: #2030

URL works as the carrier for Dubbo internal data transfer at runtime, and is also works as a protocol integrating with the external system: registration center, service query, service management, etc., resulting in serious coupling between the configuration and the registration center, and the data load pushed by the registration center is too large.

Keep URL unchanged at runtime, and even enrich the runtime URL; Keep the immutability of the URL during a single invoke from start to end.
Simplify Registry url as much as possible.
Refactor the way of doing service configuration, rely on Registry currently, switch to third-party Config Center while keeping compatible with old Overrides protocol.
Some data in URL are for service query purpose, we should register this part to another place instead of Registry. And it should be realized as extensible and users can decide to enable it or not.

## Brief changelog

* Add unit test 
* Simplify the class info and store metadata into servicestore

## Verifying this change

XXXXX

- [x ] Format the pull request title like `[Dubbo-XXX] Fix UnknownException when host config not exist #XXX`. Each commit in the pull request should have a meaningful subject line and body.
- [x ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.


dubbo-servicedata/dubbo-servicedata-api/src/main/java/org/apache/dubbo/servicedata/metadata/builder/TypeDescriptorBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author cvictory ON 2018/9/18
```
lovepoem: please remove the @author info

dubbo-servicedata/dubbo-servicedata-api/src/test/java/org/apache/dubbo/servicedata/integration/ServiceStoreServiceTest.java
```diff
@@ -1,7 +1,184 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Supplier;
+
 /**
  * @author cvictory ON 2018/9/14
```
lovepoem: please remove the @author info

PR #2636, opened 2018-10-11
[Dubbo] move the classes of model from config to rpc-api for reusing

Jeff-Lv:

reusing those model in rpc module.

## Brief changelog

the refactor is very safe with unit test convering.


chickenlj:
If it prevents reuse when staying in the config module, then let's move these classes to the rpc module. 
This migration will have no effect on other functionalities but for telnet, please make sure telnet still works fine when doing this.

Issue #2570, opened 2018-09-27
 dubbo all in one fail to start from a tomcat server when spring framework is absent


yangfuhai:
错误如下：
```
27-Sep-2018 13:39:35.736 SEVERE [RMI TCP Connection(2)-127.0.0.1] org.apache.catalina.core.StandardContext.listenerStart Error configuring application listener of class org.springframework.web.context.ContextLoaderListener
 java.lang.ClassNotFoundException: org.springframework.web.context.ContextLoaderListener
	at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1333)
	at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1167)
	at org.apache.catalina.core.DefaultInstanceManager.loadClass(DefaultInstanceManager.java:520)
	at org.apache.catalina.core.DefaultInstanceManager.loadClassMaybePrivileged(DefaultInstanceManager.java:501)
	at org.apache.catalina.core.DefaultInstanceManager.newInstance(DefaultInstanceManager.java:120)
	at org.apache.catalina.core.StandardContext.listenerStart(StandardContext.java:4660)
	at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5181)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
	at org.apache.catalina.core.ContainerBase.addChildInternal(ContainerBase.java:725)
	at org.apache.catalina.core.ContainerBase.addChild(ContainerBase.java:701)
	at org.apache.catalina.core.StandardHost.addChild(StandardHost.java:717)
	at org.apache.catalina.startup.HostConfig.manageApp(HostConfig.java:1678)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at org.apache.tomcat.util.modeler.BaseModelMBean.invoke(BaseModelMBean.java:300)
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.invoke(DefaultMBeanServerInterceptor.java:819)
[...]

PR #2640, opened 2018-10-12
prototype for issue2570

beiwei30:
…pring framework is absent

a prototype for issue #2570

## Brief changelog

dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/initializer/DubboContextListener.java
dubbo-config/dubbo-config-spring/src/main/resources/META-INF/web-fragment.xml

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/initializer/DubboContextListener.java
+++ b/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/initializer/DubboContextListener.java
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.apache.dubbo.config.spring.initializer;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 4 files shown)


ralf0131:
Hi, 

This will solve the specific problem, but will break on another scenario. If user already specified `org.springframework.web.context.ContextLoaderListener` in web.xml, this code will add another Listener, which will result in failure.

beiwei30:
if user explicitly specify spring relevant listener, then it's user's responsibility to add spring dependency.

dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/initializer/DubboContextListener.java
```diff
@@ -0,0 +1,72 @@
...
+ * doesn't depend on spring framework.
+ */
+public class DubboContextListener implements ServletContextListener {
+    private static final Log logger = LogFactory.getLog(DubboContextListener.class);
+
+    private static final String SPRING_CONTEXT_LISTENER = "org.springframework.web.context.ContextLoaderListener";
+    private static final String SPRING_CONTEXT_ROOT = "org.springframework.web.context.WebApplicationContext.ROOT";
+
+    private ServletContextListener springContextListener;
+    private boolean executed = false;
```
diecui1202: what is "executed" used for?

diecui1202:
What's the initial order about DubboContextListener and spring ContextLoaderListener? ContextLoaderListener initializes first?

diecui1202:
sorry for closing it by mistake.

dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/initializer/DubboContextListener.java
beiwei30: it is a flag for contextDestroyed, this is necessary because there are probably multiple org.springframework.web.context.ContextLoaderListener registered, one from web.xml, explicitly configured by user, the other is org.apache.dubbo.config.spring.initializer.DubboContextListener, automatically wried in from web-fragment.xml

beiwei30:
according to servlet spec, the listeners configured in web.xml will be registered first, then other listeners from web-fragment.xml will be registered.

diecui1202:
It looks good to me.

carryxyh:
`private boolean executed = false;`
I have a question. Is `volatile` needed for this field like:
`private volatile boolean executed = false;`

beiwei30:
I don't think it's necessary. Servlet container will guarantee the code will only be executed once and only once.

PR #2643, opened 2018-10-14
Jacoco aggregate report

htynkn:

JaCoCo Maven goals used to work on single modules only: Tests are executed within the module and contributed coverage only to code within the same module.
Dubbo is a multi module project which combine by api and their implements, current configuration is not suitable for Dubbo. Coverage reports were created for each module separately. 

In order to get right test coverage, we can use new goal `maven goal jacoco:report-aggregate`. This report aggregates coverage data across Maven modules.

## Brief changelog

Change maven configuration. 

## Verifying this change

CI pass and coverage increase.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -474,10 +474,10 @@
                         </configuration>
                     </execution>
                     <execution>
-                        <id>jacoco-site</id>
-                        <phase>package</phase>
+                        <id>report-aggregate</id>
+                        <phase>verify</phase>
                         <goals>
-                            <goal>report</goal>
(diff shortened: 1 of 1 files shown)


diecui1202:
@htynkn need to back port to 2.6.x?

htynkn:
> @htynkn need to back port to 2.6.x?

@diecui1202 I don't think so. because test coverage is one thing to help we improving code quality.  2.6.x is a maintain branch, we don't have any plan and resource to force on test. Let's force this part on master branch

Issue #2578, opened 2018-09-28
The current RoundRobinLoadBalance implementation performance is not satisfactory


manzhizhen:
- [x] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: newest master and 2.5.3
* Operating System version: mac
* Java version: 1.8

### Steps to reproduce this issue
Our company currently uses dubbo-2.5.3, LoadBalance chooses roundrobin. After the weight adjustment is made, the distribution of traffic is not in line with expectations. So I saw the implementation of RoundRobinLoadBalance of 2.5.3, which logically led to the distribution of traffic. Not in line with expectations, so I read the latest dubbo version of the implementation, found that the latest version of RoundRobinLoadBalance implementation traffic distribution is in line with expectations, but the performance is not satisfactory.

Test Code：

```
package com.manzhizhen.study.loadbalance;

import com.alibaba.dubbo.common.utils.AtomicPositiveInteger;
import com.alibaba.fastjson.JSON;
import com.google.common.collect.Lists;

import java.util.ArrayList;
import java.util.LinkedHashMap;
import java.util.List;
import java.util.Map;
import java.util.concurrent.ConcurrentHashMap;
import java.util.concurrent.ConcurrentMap;
import java.util.concurrent.ThreadLocalRandom;

/**
 * Created by yizhenqiang on 18/9/26.
 */
public class DubboRoundRobinLoadBalance {

    /**
     * 假设该接口有10个可用的Invoker
     */
    private static final int INVOKER_SIZE = 10;
    private static final int[] INVOKER_WEIGHT_ARRAY = new int[]{100, 100, 200, 200, 300, 300, 400, 400, 500, 500};

    private static final String SERVICE_KEY = "com.test.Test.testMethod";

    private static final ConcurrentMap<String, AtomicPositiveInteger> sequences = new ConcurrentHashMap<String, AtomicPositiveInteger>();

    private static final ConcurrentMap<String, AtomicPositiveInteger> sequences1 = new ConcurrentHashMap<String, AtomicPositiveInteger>();
    private static final ConcurrentMap<String, AtomicPositiveInteger> weightSequences = new ConcurrentHashMap<String, AtomicPositiveInteger>();


[...]

PR #2647, opened 2018-10-16
[Dubbo-2578]Smooth Round Robin selection

jasonjoo2010:
The `Round Robin` balancer is not smooth with weights, such as:

A:B:C = 1:2:3

Will be:
[A, B, B, C, C, C,           A, B, B, C, C, C,         A, B, B, C, C, C,     ..........]

But we want it to be:
[A, B, C, B, C, C,           A, B, C, B, C, C,         A, B, C, B, C, C,     ..........]

which has been discuss before for performance: #2578 

## Brief changelog
reimplement `RoundRobinLoadBalance` and its unit test.

## Verifying this change
We will get more strictly balance result in unit test.

### DIFF
--- a/dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
+++ b/dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
@@ -17,87 +17,141 @@
 package com.alibaba.dubbo.rpc.cluster.loadbalance;
 
 import com.alibaba.dubbo.common.URL;
-import com.alibaba.dubbo.common.utils.AtomicPositiveInteger;
 import com.alibaba.dubbo.rpc.Invocation;
 import com.alibaba.dubbo.rpc.Invoker;
 
-import java.util.LinkedHashMap;
+import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * Round robin load balance.
+ * 
+ * Smoothly round robin's implementation @since 2.6.5 
+ * @author jason
  *
  */
 public class RoundRobinLoadBalance extends AbstractLoadBalance {
-
     public static final String NAME = "roundrobin";
+    
+    private static int RECYCLE_PERIOD = 60000;
+    
+    protected static class WeightedRoundRobin {
+        private int weight;
+        private AtomicLong current = new AtomicLong(0);
+        private long lastUpdate;
+        public int getWeight() {
+            return weight;
+        }
+        public void setWeight(int weight) {
+            this.weight = weight;
+            current.set(0);
+        }
+        public long increaseCurrent() {
+            return current.addAndGet(weight);
+        }
+        public void sel(int total) {
+            current.addAndGet(-1 * total);
+        }
+        public long getLastUpdate() {
+            return lastUpdate;
+        }
+        public void setLastUpdate(long lastUpdate) {
+            this.lastUpdate = lastUpdate;
+        }
+    }
 
-    private final ConcurrentMap<String, AtomicPositiveInteger> sequences = new ConcurrentHashMap<String, AtomicPositiveInteger>();
-
+    private ConcurrentMap<String, ConcurrentMap<String, WeightedRoundRobin>> methodWeightMap = new ConcurrentHashMap<String, ConcurrentMap<String, WeightedRoundRobin>>();
+    private AtomicBoolean updateLock = new AtomicBoolean();
+    
+    /**
+     * get invoker addr list cached for specified invocation
+     * <p>
+     * <b>for unit test only</b>
+     * 
+     * @param invokers
+     * @param invocation
+     * @return
+     */
+    protected <T> Collection<String> getInvokerAddrList(List<Invoker<T>> invokers, Invocation invocation) {
+        String key = invokers.get(0).getUrl().getServiceKey() + "." + invocation.getMethodName();
+        Map<String, WeightedRoundRobin> map = methodWeightMap.get(key);
+        if (map != null) {
+            return map.keySet();
+        }
+        return null;
+    }
+    
     @Override
     protected <T> Invoker<T> doSelect(List<Invoker<T>> invokers, URL url, Invocation invocation) {
         String key = invokers.get(0).getUrl().getServiceKey() + "." + invocation.getMethodName();
-        int length = invokers.size(); // Number of invokers
-        int maxWeight = 0; // The maximum weight
-        int minWeight = Integer.MAX_VALUE; // The minimum weight
-        final LinkedHashMap<Invoker<T>, IntegerWrapper> invokerToWeightMap = new LinkedHashMap<Invoker<T>, IntegerWrapper>();
-        int weightSum = 0;
-        for (int i = 0; i < length; i++) {
-            int weight = getWeight(invokers.get(i), invocation);
-            maxWeight = Math.max(maxWeight, weight); // Choose the maximum weight
-            minWeight = Math.min(minWeight, weight); // Choose the minimum weight
-            if (weight > 0) {
-                invokerToWeightMap.put(invokers.get(i), new IntegerWrapper(weight));
-                weightSum += weight;
+        ConcurrentMap<String, WeightedRoundRobin> map = methodWeightMap.get(key);
+        if (map == null) {
+            methodWeightMap.putIfAbsent(key, new ConcurrentHashMap<String, WeightedRoundRobin>());
+            map = methodWeightMap.get(key);
+        }
+        int totalWeight = 0;
+        long maxCurrent = Long.MIN_VALUE;
+        long now = System.currentTimeMillis();
+        Invoker<T> selectedInvoker = null;
+        WeightedRoundRobin selectedWRR = null;
+        for (Invoker<T> invoker : invokers) {
+            String identifyString = invoker.getUrl().toIdentityString();
+            WeightedRoundRobin weightedRoundRobin = map.get(identifyString);
+            int weight = getWeight(invoker, invocation);
+            if (weight < 0) {
+                weight = 0;
             }
+            if (weightedRoundRobin == null) {
+                weightedRoundRobin = new WeightedRoundRobin();
+                weightedRoundRobin.setWeight(weight);
+                map.putIfAbsent(identifyString, weightedRoundRobin);
+                weightedRoundRobin = map.get(identifyString);
+            }
+            if (weight != weightedRoundRobin.getWeight()) {
+                //weight changed
+                weightedRoundRobin.setWeight(weight);
+            }
+            long cur = weightedRoundRobin.increaseCurrent();
+            weightedRoundRobin.setLastUpdate(now);
+            if (cur > maxCurrent) {
+                maxCurrent = cur;
+                selectedInvoker = invoker;
+                selectedWRR = weightedRoundRobin;
+            }
+            totalWeight += weight;
         }
-        AtomicPositiveInteger sequence = sequences.get(key);
-        if (sequence == null) {
-            sequences.putIfAbsent(key, new AtomicPositiveInteger());
-            sequence = sequences.get(key);
-        }
-        int currentSequence = sequence.getAndIncrement();
-        if (maxWeight > 0 && minWeight < maxWeight) {
-            int mod = currentSequence % weightSum;
-            for (int i = 0; i < maxWeight; i++) {
-                for (Map.Entry<Invoker<T>, IntegerWrapper> each : invokerToWeightMap.entrySet()) {
-                    final Invoker<T> k = each.getKey();
-                    final IntegerWrapper v = each.getValue();
-                    if (mod == 0 && v.getValue() > 0) {
-                        return k;
+        if (!updateLock.get() && invokers.size() != map.size()) {
+            if (updateLock.compareAndSet(false, true)) {
+                try {
+                    // copy -> modify -> update reference
+                    ConcurrentMap<String, WeightedRoundRobin> newMap = new ConcurrentHashMap<String, WeightedRoundRobin>();
+                    newMap.putAll(map);
+                    Iterator<Entry<String, WeightedRoundRobin>> it = newMap.entrySet().iterator();
+                    while (it.hasNext()) {
+                        Entry<String, WeightedRoundRobin> item = it.next();
+                        if (now - item.getValue().getLastUpdate() > RECYCLE_PERIOD) {
+                            it.remove();
+                        }
                     }
-                    if (v.getValue() > 0) {
-                        v.decrement();
-                        mod--;
-                    }
+                    methodWeightMap.put(key, newMap);
+                } finally {
+                    updateLock.set(false);
                 }
             }
         }
-        // Round robin
-        return invokers.get(currentSequence % length);
-    }
-
-    private static final class IntegerWrapper {
-        private int value;
-
-        public IntegerWrapper(int value) {
-            this.value = value;
+        if (selectedInvoker != null) {
+            selectedWRR.sel(totalWeight);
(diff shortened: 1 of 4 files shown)


carryxyh:
Pls send pr to master first.

dubbo-cluster/src/test/java/com/alibaba/dubbo/rpc/cluster/loadbalance/LoadBalanceTest.java
```diff
@@ -120,80 +245,120 @@ public void testRoundRobinLoadBalance_select() {
...
-            if (selected.getUrl().getProtocol().equals("test3")) {
-                sumInvoker3++;
+        }
+        for (Entry<Invoker, InvokeResult> entry : resultMap.entrySet()) {
+            String protocol = entry.getKey().getUrl().getProtocol();
+            if (!protocol.equals("test3")) {
+                InvokeResult invokeResult = entry.getValue();
+                invokeResult.setTotalWeight(invokeResult.getTotalWeight() - test3Weight);
+                // less then 12% requests to expected
+                Assert.assertTrue(invokeResult.getDeltaPercentage(loop) < 12);
```
carryxyh: I don't think this is a good way to calc percentage of select.
It is not stable.

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
```diff
@@ -32,72 +34,131 @@
...
-                        return k;
-                    }
-                    if (v.getValue() > 0) {
-                        v.decrement();
-                        mod--;
+        synchronized (map) {
+            int totalWeight = 0;
+            int maxCurrent = Integer.MIN_VALUE;
+            long now = System.currentTimeMillis();
+            boolean needRecycle = false;
```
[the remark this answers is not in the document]
jasonjoo2010: Because nodes in list may change(after some nodes offline), some entries in WRR cache map will be skipped.

But if some nodes are offline permanent these entries are kinds of waste of memory.
Like `Cache` in guava the recycle will happen every `RECYCLE_TIME` and scan the map removing the entries which `lastupdate > RECYCLE_TIME`. `lastupdate` will be updated every time invoked.

dubbo-cluster/src/test/java/com/alibaba/dubbo/rpc/cluster/loadbalance/LoadBalanceTest.java
jasonjoo2010: > I don't think this is a good way to calc percentage of select.
> It is not stable.

The Unit Test before only covers the invokers without weight specified. So if we add weight to them the actually result will be indetermination in `Random` etc. . So i tried to use percentage to make it.

So is there other way to do that? The old logics just print the result.

dubbo-cluster/src/test/java/com/alibaba/dubbo/rpc/cluster/loadbalance/LoadBalanceTest.java
```diff
@@ -280,45 +429,4 @@ public void testLoadBalanceWarmup() {
                 .calculateWarmupWeight(20 * 60 * 1000, Constants.DEFAULT_WARMUP, Constants.DEFAULT_WEIGHT));
     }
 
-    /*------------------------------------test invokers for weight---------------------------------------*/
-
-    protected List<Invoker<LoadBalanceTest>> weightInvokers = new ArrayList<Invoker<LoadBalanceTest>>();
```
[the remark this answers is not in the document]
jasonjoo2010: > Why u move the code?

It's a setup logic, not a unit test, I move it to group the logic.

Actually I add weight to all test before someone add separate test for invokers with weight specified. Though i think tests with various weight are enough.

dubbo-cluster/src/test/java/com/alibaba/dubbo/rpc/cluster/loadbalance/LoadBalanceTest.java
```diff
@@ -120,80 +245,120 @@ public void testRoundRobinLoadBalance_select() {
...
+            Assert.assertNotEquals(weightInvokers.size(), lb.getInvokerAddrList(weightInvokers, weightTestInvocation).size());
+        } finally {
+            weightInvokers.remove(weightInvokerTmp);
+        }
+    }
+    
+    @Test
+    public void testRoundRobinLoadBalance_shouldRecycle() {
+        {
+            Field recycleTimeField = null;
```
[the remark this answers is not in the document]
jasonjoo2010: Yes, ur right, I will change it

dubbo-cluster/src/test/java/com/alibaba/dubbo/rpc/cluster/loadbalance/LoadBalanceTest.java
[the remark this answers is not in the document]
jasonjoo2010: And surely it's not a problem on master branch because the logics are split into separated class.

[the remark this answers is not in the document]
carryxyh: Pls keep your pr clean and don't put too much in the same pr.
I don't think this moving is necessary.
:)

dubbo-cluster/src/test/java/com/alibaba/dubbo/rpc/cluster/loadbalance/LoadBalanceTest.java
carryxyh: There is no better way yet. But using percentages is not a good way, because it is not stable, which causes us to spend a lot of time re-executing the entire ci.

jasonjoo2010: > There is no better way yet. But using percentages is not a good way, because it is not stable, which causes us to spend a lot of time re-executing the entire ci.

I gave a big threshold for those tests and actually there must be a problem if they fail under this condition.

Is it better then no assertion at all?

And Maybe we will optimize them in future to get a better threshold(eg. better Random).

dubbo-cluster/src/test/java/com/alibaba/dubbo/rpc/cluster/loadbalance/LoadBalanceTest.java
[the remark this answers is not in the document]
carryxyh: Hi, I don't think it is necessary.
Ci will fail if one of them is failed.
:)

dubbo-cluster/src/test/java/com/alibaba/dubbo/rpc/cluster/loadbalance/LoadBalanceTest.java
carryxyh: If the threshold is too large, we still can't find it if there is a problem. So how do we determine a suitable threshold?
I think this is difficult. And even if we find this threshold, ut can't be stable, so I don't think we should handle this.

How do u think about it?

jasonjoo2010: > If the threshold is too large, we still can't find it if there is a problem. So how do we determine a suitable threshold?
> I think this is difficult. And even if we find this threshold, ut can't be stable, so I don't think we should handle this.
> 
> How do u think about it?

Currently the threshold is 12% diff to expected requests should dispatch. Maybe there's some problem in Random for example, it will between 0~8% (I think the result should be stable when we simulate 10k request).

12% under weight is acceptable to me(more like a complete unit test and reliable after hundreds loop), but your opinion is reasonable too.

So i won't consist on it and we can make a decision whether to leave them as they were.

btw the CI failed and i am checking the problem(seems not because this unit test).

jasonjoo2010:
> Pls send pr to master first.

It's strange the CI failed twice with following error:

```
The command "bash install-jdk.sh -F 11 -L BCL --target $JAVA_HOME --workspace ${TRAVIS_HOME}/.cache/install-jdk" failed and exited with 8 during .
```

It seems a failure during downloading jdk because of the last line:

```
Downloading JDK from http://download.oracle.com/otn-pub/java/jdk/11+28/55eed80b163941c8885ad9298e6d786a/jdk-11_linux-x64_bin.tar.gz...
```

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
```diff
@@ -32,72 +34,131 @@
...
-                for (Map.Entry<Invoker<T>, IntegerWrapper> each : invokerToWeightMap.entrySet()) {
-                    final Invoker<T> k = each.getKey();
-                    final IntegerWrapper v = each.getValue();
-                    if (mod == 0 && v.getValue() > 0) {
-                        return k;
-                    }
-                    if (v.getValue() > 0) {
-                        v.decrement();
-                        mod--;
+        synchronized (map) {
```
kimmking: big lock???

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
```diff
@@ -32,72 +34,131 @@
...
+                }
+                if (cur > maxCurrent) {
+                    maxCurrent = cur;
+                    selectedInvoker = invoker;
+                    selectedWRR = weightedRoundRobin;
+                }
+                totalWeight += weight;
+            }
+            //recycle
+            if (needRecycle) {
```
kimmking: I suggestion use a out thread to check, not in this thread.

kimmking: Or ignore, cause not so much keys

beiwei30: Amazing implementation. 

But we should avoid `synchronized` as much as possible, besides this synchronization block is too large to effect performance:

> Synchronized lock is an auto-upgrade lock from light weight lock in object header (when no concurrent) to monitor lock. So if the code block can be executed fast enough we may get a light weight lock mostly.

This is not true in this scenario, since it is very possible that multiple threads execute `doSelect` at the same moment.

I am fine with other parts, but this synchronization block is my major concern.

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
```diff
@@ -17,87 +17,140 @@
...
-            int weight = getWeight(invokers.get(i), invocation);
-            maxWeight = Math.max(maxWeight, weight); // Choose the maximum weight
-            minWeight = Math.min(minWeight, weight); // Choose the minimum weight
-            if (weight > 0) {
-                invokerToWeightMap.put(invokers.get(i), new IntegerWrapper(weight));
-                weightSum += weight;
-            }
+        Map<String, WeightedRoundRobin> map = methodWeightMap.get(key);
+        if (map == null) {
+            methodWeightMap.putIfAbsent(key, new HashMap<String, WeightedRoundRobin>());
```
beiwei30: We should make this map ConcurrentMap too. It is possible that two threads manipulate the same map and even the same key.

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
```diff
@@ -17,87 +17,140 @@
...
+            String identifyString = invoker.getUrl().toIdentityString();
+            WeightedRoundRobin weightedRoundRobin = map.get(identifyString);
+            int weight = getWeight(invoker, invocation);
+            if (weight < 0) {
+                weight = 0;
+            }
+            if (weightedRoundRobin == null) {
+                weightedRoundRobin = new WeightedRoundRobin();
+                weightedRoundRobin.setWeight(weight);
+                map.put(identifyString, weightedRoundRobin);
```
beiwei30: for example: two thread may manipulate map here.

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
```diff
@@ -17,87 +17,140 @@
...
-                for (Map.Entry<Invoker<T>, IntegerWrapper> each : invokerToWeightMap.entrySet()) {
-                    final Invoker<T> k = each.getKey();
-                    final IntegerWrapper v = each.getValue();
-                    if (mod == 0 && v.getValue() > 0) {
-                        return k;
-                    }
-                    if (v.getValue() > 0) {
-                        v.decrement();
-                        mod--;
+        if (updateLock.get() == false && invokers.size() != map.size()) {
```
beiwei30: `updateLock.get() == false` is unnecessary.

jasonjoo2010: It's written for easily understood especially when i restructured it with optimistic lock.
Surely i don't insist on it.

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
jasonjoo2010: yeah you are right, i forgot to change it when i remove the synchronized

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
```diff
@@ -17,87 +17,140 @@
...
-            sequences.putIfAbsent(key, new AtomicPositiveInteger());
-            sequence = sequences.get(key);
+        int totalWeight = 0;
+        long maxCurrent = Long.MIN_VALUE;
+        long now = System.currentTimeMillis();
+        Invoker<T> selectedInvoker = null;
+        WeightedRoundRobin selectedWRR = null;
+        for (Invoker<T> invoker : invokers) {
+            String identifyString = invoker.getUrl().toIdentityString();
+            WeightedRoundRobin weightedRoundRobin = map.get(identifyString);
```
beiwei30: we should use `putIfAbsent` too here.

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
```diff
@@ -17,87 +17,140 @@
...
-            this.value = value;
-        }
-
-        public void decrement() {
-            this.value--;
+        if (selectedInvoker != null) {
+            selectedWRR.sel(totalWeight);
+            return selectedInvoker;
         }
+        return null;
```
beiwei30: one more thing, should we fall back to `random` here?

dubbo-cluster/src/main/java/com/alibaba/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
```diff
@@ -17,87 +17,142 @@
...
+            String identifyString = invoker.getUrl().toIdentityString();
+            WeightedRoundRobin weightedRoundRobin = map.get(identifyString);
+            int weight = getWeight(invoker, invocation);
+            if (weight < 0) {
+                weight = 0;
+            }
+            if (weightedRoundRobin == null) {
+                weightedRoundRobin = new WeightedRoundRobin();
+                weightedRoundRobin.setWeight(weight);
+                map.putIfAbsent(identifyString, weightedRoundRobin);
```
beiwei30: we cannot use Map's `putIfAbsent` here since it's a new API since 1.8. we need to make sure dubbo 1.6 works with java 1.6 and above.

Issue #2537, opened 2018-09-20
Increase the schedule frequency of HeartbeatTask.


chickenlj:
Here is how heartbeat works in Dubbo:
Heartbeat scheduler will schedule HeartbeatTask in every `heartbeat` time, when being scheduled, HeartbeatTask will internally continue to check the necessity of sending a heartbeat event (only send when it find that the idle interval `>heartbeat`).
Consider different combinations of these two intervals: schedule interval and idle interval, we may have a chance of idle for `2*heartbeat` on a wire without any request/response or heartbeat event being sent.

Though the timeout interval in Dubbo is `3*heartbeat`, and `2*heartbeat` idle would not mislead Dubbo to terminate the connection, we may need to increase the frequency of the scheduler to make it more accurate.

chickenlj:
https://github.com/apache/incubator-dubbo/blob/cd60e58e1a579749c89c6ac8ddb39f8e8e894ef5/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java#L186

carryxyh:
Hi, I will start a work on it.
I plan to use hashwheeltimer to make each client's heartbeat more frequent.
What's your opinion?

carryxyh:
Close via https://github.com/apache/incubator-dubbo/pull/2658.

PR #2658, opened 2018-10-18
Optimize heartbeat and reconnect task.

carryxyh:
Optimize heartbeat and reconnect task.
1.Use hashedWheelTimer.
2.Distinguish between reconnect and heartbeat.
3.Increase inspection cycle.

related issue:
https://github.com/apache/incubator-dubbo/issues/2537


carryxyh:
Hi, @zonghaishang 
Would u pls have review with this pr?
Thx.
:)

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
```diff
@@ -181,37 +177,40 @@ public boolean hasAttribute(String key) {
...
     }
 
+    /**
+     * Each interval cannot be less than 1000ms.
+     */
+    private long calcLeastTick(int time) {
+        if (time / Constants.HEARTBEAT_TICK <= 0) {
+            return Constants.LEAST_HEARTBEAT_TICK;
+        } else {
+            return time / Constants.HEARTBEAT_TICK;
```
beiwei30: I am little confused why `time/Constants.HEARTBEAT_TICK` is necessary.

carryxyh: I increased the rate of checks to make those free pipes checked as quickly as possible and send heartbeats as quickly as possible.
That is why I use time/Constants.HEARTBEAT_TICK.

PR #2679, opened 2018-10-23
issue2016: Consumer throws RpcException after RegistryDirectory notify in high QPS

beiwei30:

issue2016: Consumer throws RpcException after RegistryDirectory notify in high QPS

## Brief changelog

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java
@@ -126,10 +126,10 @@ public abstract class AbstractInvoker<T> implements Invoker<T> {
 
     @Override
     public Result invoke(Invocation inv) throws RpcException {
+        // if invoker is destroyed due to address refresh from registry, let's allow the current invoke to proceed
         if (destroyed.get()) {
-            throw new RpcException("Rpc invoker for service " + this + " on consumer " + NetUtils.getLocalHost()
-                    + " use dubbo version " + Version.getVersion()
-                    + " is DESTROYED, can not be invoked any more!");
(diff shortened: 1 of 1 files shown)


chickenlj:
It's a concurrent problem of Registry notification thread and RPC invoking thread.
I think your patch can reduce the chances of meeting this problem to some extent but makes no essential differences. If the `destroy` status of an Invoker was set, most likely, the remote connection and some other resources may have also been destroyed, which may still cause the invoke to fail. One side effect if the invoke fails later is that the user may finally get a raw connection exception which may further confuse them.

https://github.com/apache/incubator-dubbo/blob/6938d487a388cc8cd0a2bc8740fe9ee8d378767e/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboInvoker.java#L132

How about add a retry policy in AbstractClusterInvoker.invoke() for this scenario, and we can decide whether retry or not by recognize the RpcException with a special code?  I am still thinking of some other better solutions.

chickenlj:
```java
                for (ExchangeClient client : clients) {
                    try {
                        client.close(ConfigUtils.getServerShutdownTimeout());
                    } catch (Throwable t) {
                        logger.warn(t.getMessage(), t);
                    }
                }
```

I have double checked that `client.close();` will not actually close the client and any resources. Because the Client is a ReferenceCountExchangeClient and it will only close when there's no Invoker other than this one is using it, while `rerefer` must happens before `destroyUnusedInvokers`, so there will at least two invokers refering the client instance when close method is called.

```java
 @Override
    public void close(int timeout) {
        if (refenceCount.decrementAndGet() <= 0) {
            if (timeout == 0) {
                client.close();
            } else {
                client.close(timeout);
            }
            client = replaceWithLazyClient();
        }
    }
```

PR #2686, opened 2018-10-25
重构 ServiceBean 的 isDelay 方法，使其更符合语义

code4wt:
isDelay 方法目前返回值如下：
- true: delay == null || delay == -1，表示不延迟导出服务
- false: delay > 0，表示延迟导出服务

我觉得 isDelay  方法返回值如下时，更合理一些：
- true: 延迟导出服务
- false: 无需延迟导出服务

Fix the wrong meaning of the method isDelay

### DIFF
--- a/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/ServiceBean.java
+++ b/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/ServiceBean.java
@@ -107,7 +107,7 @@ public class ServiceBean<T> extends ServiceConfig<T> implements InitializingBean
 
     @Override
     public void onApplicationEvent(ContextRefreshedEvent event) {
-        if (isDelay() && !isExported() && !isUnexported()) {
+        if (!isDelay() && !isExported() && !isUnexported()) {
             if (logger.isInfoEnabled()) {
                 logger.info("The service ready on spring started. service: " + getInterface());
             }
@@ -121,7 +121,7 @@ public class ServiceBean<T> extends ServiceConfig<T> implements InitializingBean
         if (delay == null && provider != null) {
             delay = provider.getDelay();
         }
-        return supportedApplicationListener && (delay == null || delay == -1);
+        return delay != null && delay > 0;
     }
 
     @Override
@@ -251,7 +251,7 @@ public class ServiceBean<T> extends ServiceConfig<T> implements InitializingBean
                 setPath(beanName);
             }
         }
-        if (!isDelay()) {
+        if (!supportedApplicationListener || isDelay()) {
             export();
         }
     }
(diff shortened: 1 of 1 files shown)


beiwei30:
@code4wt In fact, I believe `isDelay()` is unnecessary any longer in ServiceBean. Here's the reason, the method `export` in ServiceConfig should have already covered this logic:

```java
    public synchronized void export() {
        if (provider != null) {
            if (export == null) {
                export = provider.getExport();
            }
            if (delay == null) {
                delay = provider.getDelay();
            }
        }
        if (export != null && !export) {
            return;
        }

        if (delay != null && delay > 0) {
            delayExportExecutor.schedule(new Runnable() {
                @Override
                public void run() {
                    doExport();
                }
            }, delay, TimeUnit.MILLISECONDS);
        } else {
            doExport();
        }
    }
```

But I agree we need to enhance ServiceBean since the current logic is confusing, what I propose is:

```diff
@@ -107,7 +107,7 @@ public class ServiceBean<T> extends ServiceConfig<T> implements InitializingBean
 
     @Override
     public void onApplicationEvent(ContextRefreshedEvent event) {
-        if (isDelay() && !isExported() && !isUnexported()) {
+        if (!isExported() && !isUnexported()) {^M
             if (logger.isInfoEnabled()) {
                 logger.info("The service ready on spring started. service: " + getInterface());
             }
@@ -115,15 +115,6 @@ public class ServiceBean<T> extends ServiceConfig<T> implements InitializingBean
         }
     }
 
-    private boolean isDelay() {
-        Integer delay = getDelay();
-        ProviderConfig provider = getProvider();
-        if (delay == null && provider != null) {
-            delay = provider.getDelay();
-        }
-        return supportedApplicationListener && (delay == null || delay == -1);
-    }
-
     @Override
     @SuppressWarnings({"unchecked", "deprecation"})
     public void afterPropertiesSet() throws Exception {
@@ -251,7 +242,7 @@ public class ServiceBean<T> extends ServiceConfig<T> implements InitializingBean
                 setPath(beanName);
             }
         }
-        if (!isDelay()) {
+        if (!supportedApplicationListener) {^M
             export();
         }
     }
```

Pls. take a look. If you agree with this, would you mind to revise your pull request?

beiwei30:
good job, thanks.

zonghaishang:
What if you want to expose the service 10 seconds after the spring container started ? 
The original intention `delay = 10000` represents execution after 10 seconds.

@beiwei30 @code4wt @chickenlj please help check that.

```java
public synchronized void export() {
    if (delay != null && delay > 0) {
        // 
        delayExportExecutor.schedule(new Runnable() {
            @Override
            public void run() {
                doExport();
            }
        }, delay, TimeUnit.MILLISECONDS);
    } else {
        doExport();
    }
}
```

beiwei30:
I think this behavior is expected. The reason is, the best timing for Dubbo to count down delay for exposing service is when Spring container is ready, then this timing should be the moment when spring context gets refreshed, since at this moment, all beans are settled, and are ready for wiring in their dependencies. If it happens earlier, say, at the moment after properties set, then Dubbo may fail to expose service because it depends on other spring bean. 

To conclude, Dubbo should start to expose services until spring context gets refreshed, no matter delay or no-delay is configured. Unfortunately in order to keep backward compatibility, considering not all spring versions support context refresh callback, we have to start to expose services after properties set if we find application listener is not supported, and hope Dubbo service can be exposed normally.

I hope this can clarify the behavior.

jarvisxiong:
赞，最近看这块有些别扭。

LuckyDevin:
very much like

574258039:
very much like

Alexbonbon:
看着真是别扭

iupikachu:
ok

Issue #2297, opened 2018-08-14
[Enhancement] Resolve the placeholders for sub-properties 


mercyblitz:
- [X] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [X] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Details

For now on, Dubbo does not support to resolve all placeholders from property value from any `PropertySource`, thus it will impact Dubbo Spring Boot's externalization configuration, like this issue : 
* [Can provider boot features external config placeholders in properties?](https://github.com/apache/incubator-dubbo-spring-boot-project/issues/249)

What's more, Dubbo's `PropertySourcesUtils` will be removed and replaced by [Alibaba's spring-context-support](https://github.com/alibaba/spring-context-support)'s [`PropertySourcesUtils`](https://github.com/alibaba/spring-context-support/blob/1.0.1/src/main/java/com/alibaba/spring/util/PropertySourcesUtils.java)

diecui1202:
@mercyblitz How about this issue? Ready for merge?

Issue #1755, opened 2018-05-08
[Bug] Bean name of @Service annotated class does not resolve placeholder


mercyblitz:
Code as following:

```java
@Service(
        version = "${dubbo.service.version}",
        application = "${dubbo.application.id}",
        protocol = "${dubbo.protocol.id}",
        registry = "${dubbo.registry.id}"
)
public class DefaultDemoService implements DemoService {

    public String sayHello(String name) {
        return "Hello, " + name + " (from Spring Boot)";
    }

}
```

The placeholder was not resolved from the bean name of `DefaultDemoService` , thus it's name will contain "${dubbo.service.version}"

diecui1202:
@mercyblitz Any progress?

PR #2687, opened 2018-10-25
Bug Fix & Enhancement

mercyblitz:
- [Bug] Bean name of @Service annotated class does not resolve placeholder #1755
- [Enhancement] Resolve the placeholders for sub-properties #2297


dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/context/annotation/DubboConfigBindingRegistrar.java
```diff
@@ -85,7 +86,7 @@ private void registerDubboConfigBeans(String prefix,
                                           boolean multiple,
                                           BeanDefinitionRegistry registry) {
 
-        Map<String, String> properties = getSubProperties(environment.getPropertySources(), prefix);
+        Map<String, Object> properties = getSubProperties(environment.getPropertySources(), prefix);
```
kimmking: why modify type?

mercyblitz: It's better for binding, `PropertySource` and `PropertyValue` both depend on `Map<String,Object>`

PR #2740, opened 2018-11-05
 Simplify the code of StringUtils to make it more graceful.

code4wt:

Simplify the code of StringUtils to make it more graceful.

## Brief changelog

Modified files:
dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
dubbo-common/src/test/java/org/apache/dubbo/common/utils/StringUtilsTest.java

New Files:
dubbo-common/src/main/java/org/apache/dubbo/common/utils/ArrayUtils.java
dubbo-common/src/test/java/org/apache/dubbo/common/utils/ArrayUtilsTest.java


beiwei30:
great work, thanks.

Issue #2757, opened 2018-11-08
ReferenceBean's getObject() called in toString()


beiwei30:
Intellij Idea's debugger renders object to string in object view (Preferences -> Build, Execution, Deployment -> Debugger -> Data Views -> Java -> Enable 'toString()' object view). This behavior is enabled by default and it leads some strange problem when debug in Dubbo, for examples: #2431, or user may observe a ReferenceConfig may init twice or ReferenceConfig#urls may have two same entries if the breakpoint happens to set in the relevant places. You may interested in looking into the details from http://t.cn/EAhta27, where @code4wt gives the excellent analysis.

carryxyh:
@beiwei30 
How do u think about that toString method print getObject's result.
IMO, toString method should not print this info because this is meaningless.
So I support @code4wt revision.

carryxyh:
Nice pr and analysis.
Thx

PR #2759, opened 2018-11-08
Fix the bug that ReferenceBean refers service more than once when debugging.

code4wt:

Fix issue [#2757](https://github.com/apache/incubator-dubbo/issues/2757) that ReferenceBean refers service more than once when debugging.
Please visit [http://t.cn/EAhta27](http://t.cn/EAhta27) for more detail

## Brief changelog

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractConfig.java


carryxyh:
LGTM.

PR #2763, opened 2018-11-08
Optimize retry for FailbackRegistry.

carryxyh:
Optimize retry for FailbackRegistry.
The retry operation splits into specific operations, such as subscriptions and registrations. This approach allows for very precise retry control.

related issue:
https://github.com/apache/incubator-dubbo/issues/2752


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/AbstractRetryTask.java
```diff
@@ -0,0 +1,84 @@
...
+        timer.newTimeout(timeout.task(), tick, TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    public void run(Timeout timeout) throws Exception {
+        if (timeout.isCancelled()) {
+            // other thread cancel this timeout.
+            return;
+        }
+        if (logger.isInfoEnabled()) {
```
zonghaishang: Please remove it , maybe print too many log.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/AbstractRetryTask.java
```diff
@@ -0,0 +1,84 @@
...
+            logger.info(taskName + " : " + url);
+        }
+        try {
+            doRetry(url, registry);
+        } catch (Throwable t) { // Ignore all the exceptions and wait for the next retry
+            logger.warn("Failed to " + taskName + url + ", waiting for again, cause:" + t.getMessage(), t);
```
zonghaishang: Rename to `"Failed to execute task " + taskName  + ", url: " + url + ": waiting for again, cause:" + t.getMessage()`

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/AbstractRetryTask.java
carryxyh: This info log exists in the previous old version. This log is still useful in the test environment. Do we need to optimize it

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/AbstractRetryTask.java
carryxyh: Ok, that is fine!

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/FailedNotifiedTask.java
```diff
@@ -0,0 +1,66 @@
...
+            return;
+        }
+        if (logger.isInfoEnabled()) {
+            logger.info(taskName + " : " + url);
+        }
+        try {
+            listener.notify(urls);
+            urls.clear();
+        } catch (Throwable t) { // Ignore all the exceptions and wait for the next retry
+            logger.warn("Failed to " + taskName + url + ", waiting for again, cause:" + t.getMessage(), t);
```
zonghaishang: again

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/AbstractRetryTask.java
```diff
@@ -0,0 +1,84 @@
...
+        Timer timer = timeout.timer();
+        if (timer.isStop() || timeout.isCancelled()) {
+            return;
+        }
+
+        timer.newTimeout(timeout.task(), tick, TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    public void run(Timeout timeout) throws Exception {
```
zonghaishang:
```
public void run(Timeout timeout) throws Exception {
    if (timeout.isCancelled() || timeout.timer().isStop() || isCancel()) {
        // other thread cancel this timeout or stop the timer or cancel the task.
        return;
    }

    try {
        doRun(url, registry);
    } catch (Throwable t) { // Ignore all the exceptions and wait for the next retry
        logger.xxx

        // reput this task when catch exception.
        reput(timeout, retryPeriod);
    }
}
```       
How to you think ?

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/FailedNotifiedTask.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        this.urls.addAll(urls);
+    }
+
+    public void removeRetryUrl(List<URL> urls) {
+        this.urls.removeAll(urls);
+    }
+
+    @Override
+    public void run(Timeout timeout) throws Exception {
```
zonghaishang: just override `doRetry` or `doRun` is enough. I think doRun is better(more general)

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/AbstractRetryTask.java
zonghaishang:
```
if (!recoverRegistered.isEmpty()) {
    if (logger.isInfoEnabled()) {
        logger.info("Recover register url " + recoverRegistered);
    }
}
```

But I found that the original log output is conditional, such as the collection is not empty

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/FailedNotifiedTask.java
carryxyh: This is my mistake. According to the previous logic, failedNotify needs to be retried until the URL-NotifyListener key-value pair is removed.

So the logic of failedNotify is to reput to timer and retry no matter what the situation. After reading your review, I found that I can guarantee the above logic by reputing after doRetry. I will fix it as soon as possible.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/AbstractRetryTask.java
carryxyh: In older versions, a log was printed if a failed collection was not empty. Now our logic is not in units of collections, but in the specific url of a failed. Entering this run method indicates that we are about to retry, so the log is printed here.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/AbstractRetryTask.java
carryxyh: Seems like the info logger is needed.
It is useful in test env.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/FailedNotifiedTask.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final String NAME = "retry subscribe";
+
+    private final NotifyListener listener;
+
+    private final List<URL> urls = new CopyOnWriteArrayList<>();
+
+    public FailedNotifiedTask(URL url, NotifyListener listener) {
+        super(url, null, NAME);
+        if (listener == null) {
+            throw new IllegalArgumentException();
```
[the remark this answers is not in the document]
carryxyh: never null. just a check.

beiwei30:
have we reached an agreement? can I merge it now? I noticed my comment has been solved.

PR #2788, opened 2018-11-15
Optimize getExtensionClass method.

carryxyh:
Optimize getExtensionClass method.
Just return the class instead of throwing exception.

org.apache.dubbo.common.extension.ExtensionLoader#getExtensionClass just has one caller that is hasExtension. So it will not effect other logic.


carryxyh:
@beiwei30 
Simple pr. Would u pls give a review?

Issue #2758, opened 2018-11-08
Test DubboProtocolTest.testDubboProtocolWithMina fails when run by itself


OrDTesters:
- [x] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.x (master)
* Operating System version: Ubuntu 16.04
* Java version: Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)

### Steps to reproduce this issue

1. `git checkout b0107e767651d066d68c3beaaca9736aed2292b8`
2. `mvn test -am -pl dubbo-rpc/dubbo-rpc-dubbo -Dtest=DubboProtocolTest#testDubboProtocolWithMina -DfailIfNoTests=false`

### Expected Result

The test should pass.

### Actual Result

The test fails. Output:
```
Running org.apache.dubbo.rpc.protocol.dubbo.DubboProtocolTest
2018-11-08 03:47:05,604 INFO [org.apache.dubbo.common.logger.LoggerFactory:?] - using logger: org.apache.dubbo.common.logger.log4j.Log4jLoggerAdapter
2018-11-08 03:47:06,034 INFO [org.apache.dubbo.rpc.protocol.dubbo.DubboProtocol:destroy] -  [DUBBO] Unexport service: dubbo://127.0.0.1:9010/org.apache.dubbo.rpc.protocol.dubbo.support.DemoService?server=mina, dubbo version: , current host: 172.17.0.12
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.657 sec <<< FAILURE! - in org.apache.dubbo.rpc.protocol.dubbo.DubboProtocolTest
testDubboProtocolWithMina(org.apache.dubbo.rpc.protocol.dubbo.DubboProtocolTest)  Time elapsed: 0.086 sec  <<< ERROR!
org.apache.dubbo.rpc.RpcException: Unsupported server type: mina, url: dubbo://127.0.0.1:9010/org.apache.dubbo.rpc.protocol.dubbo.support.DemoService?channel.readonly.sent=true&heartbeat=60000&server=mina
[...]

PR #2797, opened 2018-11-16
[Dubbo-2758] Adding dependency on dubbo-remoting-mina for tests in dubbo-rpc-dubbo

OrDTesters:

See Issue #2758 

DubboProtocolTest.testDubboProtocolWithMina fails when run by itself. The test tries to create a mina server, but mina is not supported. This pull request adds a dependency on dubbo-remoting-mina for the module so mina can be supported when the test runs.

## Brief changelog

dubbo-rpc/dubbo-rpc-dubbo/pom.xml b/dubbo-rpc/dubbo-rpc-dubbo/pom.xml

### DIFF
--- a/dubbo-rpc/dubbo-rpc-dubbo/pom.xml
+++ b/dubbo-rpc/dubbo-rpc-dubbo/pom.xml
@@ -61,6 +61,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.dubbo</groupId>
+            <artifactId>dubbo-remoting-mina</artifactId>
+            <version>${project.parent.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-all</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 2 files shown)


dubbo-rpc/dubbo-rpc-dubbo/pom.xml
```diff
@@ -60,6 +60,12 @@
             <version>${project.parent.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
carryxyh: It seems that only adding dependencies does not make the test cover mina.
We still need some modifications to the UT. Both the server and the client have a cache. The cached key is the address (127.0.0.1:9000). In order to create a different Client or Server instance, we need to modify the port number of the UT.

carryxyh: Personally think that different port numbers are better.

OrDTesters: Just to be clear, should I make the change to make every test have a unique port number in the URL before the PR can be merged? Right now, the PR only concerns testDubboProtocolWithMina, should changing port numbers for all other tests be put in a separate PR?

carryxyh: Since most of the ut is stable, I think it is not necessary to change every port.

PR #2807, opened 2018-11-20
Fixing test-order dependency in org.apache.dubbo.rpc.cluster.StickyTest

OrDTesters:

There are test-order dependencies in org.apache.dubbo.rpc.cluster.StickyTest. testStickyNoCheck fails when run after testStickyForceCheck. The tests depend on the same LoadBalance instance, which enters a different state after testStickyForceCheck runs, which then results in testStickNoCheck to fail when it runs.

The proposed fix introduces a method in ExtensionLoader to reset cached ExtensionLoader instances, for testing purposes only. StickyTest can then use this method to reset the ExtensionLoader for LoadBalance between test runs.

## Brief changelog

dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/StickyTest.java
dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java

### DIFF
--- a/dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/StickyTest.java
+++ b/dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/StickyTest.java
@@ -19,6 +19,7 @@ package org.apache.dubbo.rpc.cluster;
 
 import org.apache.dubbo.common.Constants;
 import org.apache.dubbo.common.URL;
+import org.apache.dubbo.common.extension.ExtensionLoader;
 import org.apache.dubbo.rpc.Invocation;
 import org.apache.dubbo.rpc.Invoker;
 import org.apache.dubbo.rpc.Result;
@@ -68,6 +69,8 @@ public class StickyTest {
         invokers.add(invoker2);
 
         clusterinvoker = new StickyClusterInvoker<StickyTest>(dic);
+
+        ExtensionLoader.resetExtensionLoader(LoadBalance.class);
     }
 
     @Test
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
@@ -124,6 +124,20 @@ public class ExtensionLoader<T> {
         return loader;
     }
 
+    // For testing purposes only
+    public static void resetExtensionLoader(Class type) {
+        ExtensionLoader loader = EXTENSION_LOADERS.get(type);
+        if (loader != null) {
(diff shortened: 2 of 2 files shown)


carryxyh:
I will have a check.

carryxyh:
I can't reproduce this problem on my mac.
I tried:
1. run testStickyForceCheck and then run testStickyNoCheck. UT passed.
2.run the whole test. UT passed.
3.move the test method testStickyForceCheck before testStickyNoCheck. UT passed.

Is my reproduce way is wrong?

OrDTesters:
The two tests have to be run in the same JVM. The order of the tests written in the file does not control the order that they are actually run by JUnit. To reproduce this issue, there is a way to force JUnit to run the tests in a specific order.

Add the annotation @org.junit.FixMethodOrder(org.junit.runners.MethodSorters.NAME_ASCENDING) to the test class. This will force JUnit to run the tests in the ascending alphabetical order of their names, so testStickyForceCheck will run before testStickyNoCheck.

Since without FixMethodOrder the order of the tests is not guaranteed, it is best to make sure the tests can pass when they are run in any order.

carryxyh:
I can reproduce this problem.
I am not sure if this modification is best. Maybe someone else can give a review.

dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/StickyTest.java
```diff
@@ -68,6 +69,8 @@ public void setUp() throws Exception {
         invokers.add(invoker2);
 
         clusterinvoker = new StickyClusterInvoker<StickyTest>(dic);
+
+        ExtensionLoader.resetExtensionLoader(LoadBalance.class);
```
beiwei30: I am not convinced either. Why we need to reset extension loader here?

OrDTesters: When either testStickyForceCheck or testStickyNoCheck runs, they eventually end up using an instance of RoundRobinLoadBalance, which is obtained and shared between the two tests through the ExtensionLoader, accessed using the LoadBalance.class as key. In between the two runs, one of the shared RoundRobinLoadBalance instance's internal map fields changes. The change is such that if testStickyNoCheck runs after testStickyForceCheck, the assertion for testStickyNoCheck fails. The goal of resetting is to make sure the two tests do not end up using the same RoundRobinLoadBalance instance when they run.

I would be happy to discuss any other ways of approaching this problem.

beiwei30: I think I understand what happens now, agree with your change.

Issue #2838, opened 2018-11-27
UT failed on windows with FileNetworkerTest


biyuhao:
- [x] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: master branch
* Operating System version: windows 10
* Java version: 1.8

### Steps to reproduce this issue

1. run FileNetworkerTest on windows platform
2. UT failed

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result
UT passed

What do you expected from the above steps？

### Actual Result
What actually happens?
exception thrown
ref [File_URI_scheme](https://en.wikipedia.org/wiki/File_URI_scheme)

If there is an exception, please attach the exception trace:

```
java.lang.NumberFormatException: For input string: "\Users\XXX\AppData\Local\Temp\junit7860990268511022074\junit2976714168777395029.tmp"

	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)
	at org.apache.dubbo.common.URL.valueOf(URL.java:249)
	at org.apache.dubbo.remoting.p2p.support.FileNetworkerTest.testJoin(FileNetworkerTest.java:57)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
[...]

PR #2839, opened 2018-11-27
[Dubbo-2838] Fix UT failed on windows with FileNetworkerTest

biyuhao:

Fix Dubbo-2838

## Brief changelog

Fix FileNetworkerTest failure

## Verifying this change

Run UT on windows platform

### DIFF
--- a/dubbo-remoting/dubbo-remoting-p2p/src/test/java/org/apache/dubbo/remoting/p2p/support/FileNetworkerTest.java
+++ b/dubbo-remoting/dubbo-remoting-p2p/src/test/java/org/apache/dubbo/remoting/p2p/support/FileNetworkerTest.java
@@ -51,7 +51,11 @@ public class FileNetworkerTest {
 
     @Test
     public void testJoin() throws RemotingException, InterruptedException, IOException {
-        final String groupURL = "file://" + folder.newFile();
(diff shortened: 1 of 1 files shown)


biyuhao:
Fix #2838

carryxyh:
I have seen another pr to fix this issue:
https://github.com/apache/incubator-dubbo/pull/2848

Personally think that his modification is simpler and effective.

What do u think about it?

biyuhao:
Please refer to https://en.wikipedia.org/wiki/File_URI_scheme
Actually an additional slash is only for Windows local file which hostname is omitted.
And for Unix like system remains the same. (`file:///tmp/junit2484075406115019206/junit2016855624112146186.tmp` ,three slashes in total including the root path one)

In fact the UT passed, so I think both pr are fine.  😄

carryxyh:
Hi,
We don't need to judge our system, we just need to add a '/'. The content after the third '/' will be considered the path of the URL (the current way of parsing dubbo). I know that this problem is caused by the separator of WINDOWS. But in the way of fixing, I think we only need to add a '/' to move closer to dubbo's parsing. 

Another thing is that that pr is simpler. So I personally prefer that one.  Thx for your pr, too. Very nice solution. 👍

Issue #2799, opened 2018-11-17
Router compareTo 方法实现引发异常


63914578:
- [ ] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.6.2
* Operating System version: Linux
* Java version: 8

### Steps to reproduce this issue

1. 添加路由
2. 指定相同的priority
3. 引发异常：java.lang.IllegalArgumentException: Comparison method violates its general contract! 
查看源码 ConditionRouter、ScriptRouter、TagRouter 的 compareTo 方法实现类似，会引发同类问题。

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result

What do you expected from the above steps？

### Actual Result

What actually happens?

If there is an exception, please attach the exception trace:

```
ERROR com.alibaba.dubbo.registry.zookeeper.ZookeeperRegistry  -  [DUBBO] Failed to notify for subscribe consumer://xx.xx.xx.xx/com.xxx.service.XxxService?application=xxx-service&category=providers,configurators,routers&default.check=true&dubbo=2.6.2&interface=com.xxx.service.XxxService&methods=getXxx,setXxx&owner=xxx&pid=160275&revision=3.5.9&side=consumer&timeout=10000×tamp=1542092651604, waiting for retry, cause: Comparison method violates its general contract!, dubbo version: 2.6.2, current host: xx.xx.xx
java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.util.ComparableTimSort.mergeLo(ComparableTimSort.java:740)
	at java.util.ComparableTimSort.mergeAt(ComparableTimSort.java:477)
	at java.util.ComparableTimSort.mergeCollapse(ComparableTimSort.java:402)
	at java.util.ComparableTimSort.sort(ComparableTimSort.java:209)
	at java.util.Arrays.sort(Arrays.java:1312)
	at java.util.Arrays.sort(Arrays.java:1506)
	at java.util.ArrayList.sort(ArrayList.java:1454)
	at java.util.Collections.sort(Collections.java:141)
	at com.alibaba.dubbo.rpc.cluster.directory.AbstractDirectory.setRouters(AbstractDirectory.java:110)
[...]

PR #2909, opened 2018-12-07
create AbstractRouter

Leishunyu:
https://github.com/apache/incubator-dubbo/issues/2799
change the jdk compatible

## Verifying this change

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/AbstractRouter.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/AbstractRouter.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2009-2018 Hangzhou FanDianEr Technology Co., Ltd. All rights reserved
+ */
+package org.apache.dubbo.rpc.cluster.router;
+
+import org.apache.dubbo.common.URL;
+import org.apache.dubbo.rpc.cluster.Router;
+
+/**
+ * AbstractRouter
+ *
+ * @since 2018-12-07
+ */
+public abstract class AbstractRouter implements Router {
+    protected int priority;
+    protected URL url;
+
+    protected AbstractRouter(int priority, URL url) {
+        this.priority = priority;
+        this.url = url;
+    }
+
+    protected AbstractRouter() {
+    }
+
+    @Override
+    public Integer getPriority() {
+        return priority;
+    }
+
+    @Override
+    public URL getUrl() {
+        return url;
+    }
+
+    @Override
+    public int compareTo(Router o) {
+        if (o == null) {
+            throw new IllegalArgumentException();
+        }
+        if (this.priority == o.getPriority()) {
+            if (o.getUrl() == null) {
+                return -1;
+            }
+            return url.toFullString().compareTo(o.getUrl().toFullString());
+        } else {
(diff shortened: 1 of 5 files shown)


carryxyh:
@ralf0131 
Would u pls have a look?

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/script/ScriptRouter.java
```diff
@@ -114,13 +113,24 @@ public URL getUrl() {
...
+    @Override
+    public Integer getPriority() {
+        return priority;
+    }
+
     @Override
     public int compareTo(Router o) {
-        if (o == null || o.getClass() != ScriptRouter.class) {
-            return 1;
+        if (o == null) {
```
beiwei30: shouldn't we call super first, and do extra logic when super return 0 instead?

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/AbstractRouter.java
```diff
@@ -0,0 +1,50 @@
+/*
```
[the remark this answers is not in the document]
Leishunyu: Your suggestion has been adopted, the default method of jdk8 is really better here, thank you

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/Router.java
```diff
@@ -51,4 +51,31 @@
      */
     <T> List<Invoker<T>> route(List<Invoker<T>> invokers, URL url, Invocation invocation) throws RpcException;
 
+    /**
+     * priority
+     *
+     * @return
+     */
+    Integer getPriority();
```
beiwei30: I think `int` is better, let's rely on auto-boxing.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/script/ScriptRouter.java
carryxyh: This does not match the old logic.
The old logic is: If the priorities are equal, compare the rules.

If you call super first and the priority is equal, continue to compare the url. If the url returns not 0, it will return directly.

Old logic doesn't compare to url and url would not effect result.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/Router.java
Leishunyu: Ok, modified, thank you

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/script/ScriptRouter.java
carryxyh: Simply,
Sometimes the url compare(super method) return 1. But the rule compare return -1.
If we can super first, it will return 1.
But in the old logic, it will return -1.

beiwei30:
I will merge it if it passes CI.

Leishunyu:
> 如果它通过CI，我将合并它。

3Q!

PR #2934, opened 2018-12-10
[Dubbo 2682]Optimize the logic of find port of one protocol

tswstarplanet:

Now the actual logic of get a port of a protocol is executed in the process of every service that use the same protocol. So we can optimize the logic so that it will only parse once actually.

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
@@ -633,6 +633,25 @@ public class ServiceConfig<T> extends AbstractServiceConfig {
      * @return
      */
     private Integer findConfigedPorts(ProtocolConfig protocolConfig, String name, Map<String, String> map) {
+        Integer portToBind = getPortFromProtocol(protocolConfig, name);
+
+        // save bind port, used as url's key later
+        map.put(Constants.BIND_PORT_KEY, String.valueOf(portToBind));
+
+        // registry port, not used as bind port by default
+        String portToRegistryStr = getValueFromConfig(protocolConfig, Constants.DUBBO_PORT_TO_REGISTRY);
+        Integer portToRegistry = parsePort(portToRegistryStr);
+        if (portToRegistry == null) {
+            portToRegistry = portToBind;
+        }
+
+        return portToRegistry;
+    }
+
+    private Integer getPortFromProtocol(ProtocolConfig protocolConfig, String name) {
+        if (protocolConfig.getPort() != null && protocolConfig.getPort() > 0) {
+            return protocolConfig.getPort();
+        }
         Integer portToBind = null;
 
         // parse bind port from environment
@@ -657,18 +676,8 @@ public class ServiceConfig<T> extends AbstractServiceConfig {
                 }
             }
         }
-
-        // save bind port, used as url's key later
-        map.put(Constants.BIND_PORT_KEY, String.valueOf(portToBind));
-
-        // registry port, not used as bind port by default
-        String portToRegistryStr = getValueFromConfig(protocolConfig, Constants.DUBBO_PORT_TO_REGISTRY);
-        Integer portToRegistry = parsePort(portToRegistryStr);
-        if (portToRegistry == null) {
-            portToRegistry = portToBind;
-        }
-
-        return portToRegistry;
+        protocolConfig.setPort(portToBind);
+        return portToBind;
     }
 
     private Integer parsePort(String configPort) {


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -633,6 +633,25 @@ private String findConfigedHosts(ProtocolConfig protocolConfig, List<URL> regist
      * @return
      */
     private Integer findConfigedPorts(ProtocolConfig protocolConfig, String name, Map<String, String> map) {
+        Integer portToBind = getPortFromProtocol(protocolConfig, name);
+
+        // save bind port, used as url's key later
+        map.put(Constants.BIND_PORT_KEY, String.valueOf(portToBind));
```
khanimteyaz: method name indicate a search operation based on method inputs, should we modify input object here?

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -657,18 +676,8 @@ private Integer findConfigedPorts(ProtocolConfig protocolConfig, String name, Ma
...
-
-        // registry port, not used as bind port by default
-        String portToRegistryStr = getValueFromConfig(protocolConfig, Constants.DUBBO_PORT_TO_REGISTRY);
-        Integer portToRegistry = parsePort(portToRegistryStr);
-        if (portToRegistry == null) {
-            portToRegistry = portToBind;
-        }
-
-        return portToRegistry;
+        protocolConfig.setPort(portToBind);
```
khanimteyaz: method name (getPortFromProtocol) indicate a get operation here, should we change the state of input parameter?

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -633,6 +633,25 @@ private String findConfigedHosts(ProtocolConfig protocolConfig, List<URL> regist
...
     private Integer findConfigedPorts(ProtocolConfig protocolConfig, String name, Map<String, String> map) {
+        Integer portToBind = getPortFromProtocol(protocolConfig, name);
+
+        // save bind port, used as url's key later
+        map.put(Constants.BIND_PORT_KEY, String.valueOf(portToBind));
+
+        // registry port, not used as bind port by default
+        String portToRegistryStr = getValueFromConfig(protocolConfig, Constants.DUBBO_PORT_TO_REGISTRY);
+        Integer portToRegistry = parsePort(portToRegistryStr);
+        if (portToRegistry == null) {
```
khanimteyaz: Can be replace by ternary operator
return portToRegistry !=null ? portToRegistry :portToBind;

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
tswstarplanet: The change of the state of the input parameter is used to cached the parsed port in protocol config bean, so that the subsequent accesses of port of a protocol can retrieve the port from the protocol config bean directly. Or the find port logic will be executed during the exporting process of every <dubbo:service>

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -633,6 +633,25 @@ private String findConfigedHosts(ProtocolConfig protocolConfig, List<URL> regist
...
+            portToRegistry = portToBind;
+        }
+
+        return portToRegistry;
+    }
+
+    private Integer getPortFromProtocol(ProtocolConfig protocolConfig, String name) {
```
carryxyh: I don't think this is a good method name.
When the port is missed in config, it will be found in system or protocol instance or somewhere.

tswstarplanet: I just think that current logic is just that it does duplicate work too many times. We can get available port from system properties, dubbo config properties or the config of xml. But it does not have to do the same thing in every dubbo service export process. It seems redundant

tswstarplanet: If the port is strongly related to the protocol, can we move the find port logic to the process of generate the protocol config bean ? It can not only reduce the duplicate execution of find port logic, but also can present a more clear logic.

tswstarplanet: I think current code is like that we have known the result of one computing problem after we solved it first time. But when we need the result, we still try to solve the problem once by once, but not to get the result directly. But the problem and the result does not change always.

carryxyh: Seems like most time we just have one protocol. So it is not 'once by once' like u said.
ProtocolConfig's port is set by user. U can fix it by other way, but not to change the user config.

tswstarplanet: 'Once by once' that I said means that when every <dubbo:service> that use the same protocol export, it will execute the same logic once.

carryxyh: I only have one problem. We can use serviceConfig and protocolConfig to publish the service right? If we accept your pr, the user does not set the port of protocolConfig before the release. After the release, protocolConfig has a port. How do we explain this to user that protocolConfig auto set a port?

tswstarplanet: I don't think this problem just from the user config side. I think if the user does not set the port of the protocol, it means that he does care the port and he just want the application to run correctly. So I think the port of protocol can regards as an inner state that the user can ignore as if the application can run with no problems.

tswstarplanet: I think common users of dubbo don’t care what class represents the <dubbo:protocol>, and they don’t have to know how dubbo run. If they don’t config a port clearly, they may or may not know dubbo will pick an available port for them. But I think it’s not important that the common users do know or do not know. They can only care about if dubbo run correctly. The inner running mechanism can be transparent to the common users.

tswstarplanet: If the user does not set the port, it of course means that he does not care about the port. It is common sense.

PR #2936, opened 2018-12-11
[Dubbo 2926] standardize semantics of all mergers,enhance mergeFactory and testcases

luchy0120:

This patch can standardize semantics of all mergers :  merge arguments can be null , and all the others should have same type . 

For ArrayMerger , as its argument type is  Object[]... items ,  so we can avoid checking whether item[i] is
an Array . Also , in order to avoid problems like 
```
        class Person{};
        class Teacher extends Person{};
        Person[] a= {new Person()};
        Teacher[] b= {new Teacher()};
        Object[] result = ArrayMerger.INSTANCE.merge(a,b);
```
we can require all elements to have the same type , al least the user should pass in the same type

For MergerFactory , if the required class is not in cache , we can throw an user friendly Exception 

For test case , I add some exception case , and fix an ArrayMerger case , also add some null argument
test case . 

## Brief changelog

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/*
dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/merger/ResultMergerTest

## Verifying this change

mvn clean install -DskipTests

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ArrayMerger.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ArrayMerger.java
@@ -16,6 +16,7 @@
  */
 package org.apache.dubbo.rpc.cluster.merger;
 
+import org.apache.dubbo.common.utils.ArrayUtils;
 import org.apache.dubbo.rpc.cluster.Merger;
 
 import java.lang.reflect.Array;
@@ -25,34 +26,48 @@ public class ArrayMerger implements Merger<Object[]> {
     public static final ArrayMerger INSTANCE = new ArrayMerger();
 
     @Override
-    public Object[] merge(Object[]... others) {
-        if (others.length == 0) {
-            return null;
+    public Object[] merge(Object[]... items) {
+        if (ArrayUtils.isEmpty(items)) {
+            return new Object[0];
         }
+
+        int i = 0;
+        while (i < items.length && items[i] == null) {
+            i++;
+        }
+
+        if (i == items.length) {
+            return new Object[0];
+        }
+
+        Class<?> type = items[i].getClass().getComponentType();
+
         int totalLen = 0;
-        for (int i = 0; i < others.length; i++) {
-            Object item = others[i];
-            if (item != null && item.getClass().isArray()) {
-                totalLen += Array.getLength(item);
-            } else {
-                throw new IllegalArgumentException((i + 1) + "th argument is not an array");
+        for (; i < items.length; i++) {
+            if (items[i] == null) {
+                continue;
             }
+            Class<?> itemType = items[i].getClass().getComponentType();
+            if (itemType != type) {
+                throw new IllegalArgumentException("Arguments' types are different");
+            }
+            totalLen += items[i].length;
         }
 
         if (totalLen == 0) {
-            return null;
+            return new Object[0];
         }
 
-        Class<?> type = others[0].getClass().getComponentType();
-
         Object result = Array.newInstance(type, totalLen);
+
         int index = 0;
-        for (Object array : others) {
-            for (int i = 0; i < Array.getLength(array); i++) {
-                Array.set(result, index++, Array.get(array, i));
+        for (Object[] array : items) {
+            if (array != null) {
+                for (int j = 0; j < array.length; j++) {
+                    Array.set(result, index++, array[j]);
+                }
             }
         }
         return (Object[]) result;
     }
-
 }
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ListMerger.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ListMerger.java
@@ -17,15 +17,20 @@
 
 package org.apache.dubbo.rpc.cluster.merger;
 
+import org.apache.dubbo.common.utils.ArrayUtils;
 import org.apache.dubbo.rpc.cluster.Merger;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 public class ListMerger implements Merger<List<?>> {
 
     @Override
     public List<Object> merge(List<?>... items) {
+        if (ArrayUtils.isEmpty(items)) {
+            return Collections.emptyList();
+        }
         List<Object> result = new ArrayList<Object>();
         for (List<?> item : items) {
             if (item != null) {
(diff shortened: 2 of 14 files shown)


beiwei30:
nice suggestion, I will take a look.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/MergerFactory.java
```diff
@@ -49,6 +53,11 @@
                 result = mergerCache.get(returnType);
             }
         }
+
+        if (result == null) {
```
beiwei30: I don't think we should throw an exception, let's just return null.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/MapMerger.java
```diff
@@ -25,8 +26,8 @@
 
     @Override
     public Map<?, ?> merge(Map<?, ?>... items) {
-        if (items.length == 0) {
-            return null;
+        if (ArrayUtils.isEmpty(items)) {
+            return new HashMap<Object, Object>();
```
beiwei30: pls. use Collections.emptyMap() instead.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ListMerger.java
```diff
@@ -26,6 +27,9 @@
 
     @Override
     public List<Object> merge(List<?>... items) {
+        if (ArrayUtils.isEmpty(items)) {
+            return new ArrayList<Object>();
```
beiwei30: Pls. use Collections.emptyList() instead.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/SetMerger.java
```diff
@@ -25,7 +26,9 @@
 
     @Override
     public Set<Object> merge(Set<?>... items) {
-
+        if (ArrayUtils.isEmpty(items)) {
+            return new HashSet<Object>();
```
beiwei30: pls. use Collections.emptySet()

beiwei30:
@luchy0120 could you pls. take a look?

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ArrayMerger.java
```diff
@@ -25,34 +26,48 @@
...
         Object result = Array.newInstance(type, totalLen);
+
         int index = 0;
-        for (Object array : others) {
-            for (int i = 0; i < Array.getLength(array); i++) {
-                Array.set(result, index++, Array.get(array, i));
+        for (Object[] array : items) {
+            if (array != null) {
+                for (int j = 0; j < array.length; j++) {
+                    Array.set(result, index++, array[j]);
```
kimmking: array[index++] =result?

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ArrayMerger.java
```diff
@@ -25,34 +26,48 @@
...
     @Override
-    public Object[] merge(Object[]... others) {
-        if (others.length == 0) {
-            return null;
+    public Object[] merge(Object[]... items) {
+        if (ArrayUtils.isEmpty(items)) {
+            return new Object[0];
         }
+
+        int i = 0;
```
kimmking: why check this?

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ArrayMerger.java
```diff
@@ -25,34 +26,48 @@
...
+
         int totalLen = 0;
-        for (int i = 0; i < others.length; i++) {
-            Object item = others[i];
-            if (item != null && item.getClass().isArray()) {
-                totalLen += Array.getLength(item);
-            } else {
-                throw new IllegalArgumentException((i + 1) + "th argument is not an array");
+        for (; i < items.length; i++) {
+            if (items[i] == null) {
```
kimmking: use if !=null then ...

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/SetMerger.java
luchy0120: Ok

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ListMerger.java
luchy0120: Ok

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/MapMerger.java
luchy0120: Ok

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/MergerFactory.java
luchy0120: May be we can add some comments about this NullPointerException

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ArrayMerger.java
luchy0120: This is not wrong . It means "result[index++] = array[j]"

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ArrayMerger.java
luchy0120: Because if the user pass in (   null ,null , new String[]{"abs","edf"} , null   ) ,  we need to find out the  String.class  , so we need to skip the null objects .

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/ArrayMerger.java
luchy0120: if I use "if != null ",  there may be too much {{}} ,  会形成嵌套的if语句

luchy0120:
@beiwei30, @kimmking,  I've updated code accroding to your review . I add some comments for mergeFactory and remove the unnecessary check

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/MergerFactory.java
beiwei30: I still think it is ok to return null.

beiwei30:
I think it's ready to go.

PR #2938, opened 2018-12-11
Improve several map iteration

biyuhao:

Improve several places of map iteration

In `JavaBeanSerializeUtil.java`, `Menu.java`, `DefaultFuture.java` and several Test cases,
we have some code similar to
```
for (Key key : map.KeySet()) {
  Value value = map.get(key);
  ......
}
``` 

This pr will replace it using `entrySet()` 

## Brief changelog

source code and test code updated

## Verifying this change

ut passwd

### DIFF
--- a/dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/loadbalance/ConsistentHashLoadBalanceTest.java
+++ b/dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/loadbalance/ConsistentHashLoadBalanceTest.java
@@ -31,8 +31,8 @@ public class ConsistentHashLoadBalanceTest extends LoadBalanceBaseTest {
     public void testConsistentHashLoadBalance() {
         int runs = 10000;
         Map<Invoker, AtomicLong> counter = getInvokeCounter(runs, ConsistentHashLoadBalance.NAME);
-        for (Invoker minvoker : counter.keySet()) {
-            Long count = counter.get(minvoker).get();
+        for (Map.Entry<Invoker, AtomicLong> entry : counter.entrySet()) {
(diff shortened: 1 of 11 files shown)


beiwei30:
@biyuhao, In my own opinion, it's nothing but code style. Both ways look good to me. Would you mind shine some light on me why your proposed way is better?

biyuhao:
@beiwei30 
I think this makes us a more unified code style and gain a better performance.
I did a little google search and got some betters answers than mine.

1. ["entrySet()" should be iterated when both the key and value are needed
](https://rules.sonarsource.com/java/tag/performance/RSPEC-2864)
> When only the keys from a map are needed in a loop, iterating the keySet makes sense. But when both the key and the value are needed, it's more efficient to iterate the entrySet, which will give access to both the key and value, instead.

2. [Performance Comparison of Different Ways to Iterate over HashMap](https://howtodoinjava.com/java/collections/hashmap/performance-comparison-of-different-ways-to-iterate-over-hashmap/)
>Using entrySet() in for-each loop : 50
Using keySet() in for-each loop : 76
Using entrySet() and iterator : 50
Using keySet() and iterator : 75

3. [performance-considerations-for-keyset-and-entryset-of-map](https://stackoverflow.com/questions/3870064/performance-considerations-for-keyset-and-entryset-of-map)
>Here you first need to get hold of the set of keys. This takes time proportional to the capacity of the map (as opposed to size for the LinkedHashMap). After this is done, you call get() once for each key. Sure, in the average case, with a good hashCode-implementation this takes constant time. However, it will inevitably require lots of .hashCode and .equals calls, which will obviously take more time than just doing a entry.value() call.

At last, I have not done such a profiling. So I think it's more about unifying code style here.
This is a trivial change, it's on your call. 😃

PR #2966, opened 2018-12-14
Add a Constants value and util method to getIDByName

Jeff-Lv:

Support the feature of alibaba inside's HSF which remoting protocol will be opensources to Dubbo.

## Brief changelog

add a constants value 
a new method for CodeSupporet

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
@@ -363,6 +363,8 @@ public class Constants {
 
     public static final String DUBBO_VERSION_KEY = "dubbo";
 
+    public static final String DEFAULT_VERSION = "0.0.0";
+
     public static final String HESSIAN_VERSION_KEY = "hessian.version";
 
     public static final String DISPATCHER_KEY = "dispatcher";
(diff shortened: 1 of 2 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -363,6 +363,8 @@
 
     public static final String DUBBO_VERSION_KEY = "dubbo";
 
+    public static final String DEFAULT_VERSION = "0.0.0";
```
beiwei30: why should we introduce this?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/transport/CodecSupport.java
```diff
@@ -36,6 +36,7 @@
     private static final Logger logger = LoggerFactory.getLogger(CodecSupport.class);
     private static Map<Byte, Serialization> ID_SERIALIZATION_MAP = new HashMap<Byte, Serialization>();
     private static Map<Byte, String> ID_SERIALIZATIONNAME_MAP = new HashMap<Byte, String>();
+    private static Map<String, Byte> Name_ID_MAP = new HashMap<String, Byte>();
```
beiwei30: **SERIALIZATIONNAME_ID_MAP** is better, in my opinion.

Jeff-Lv: good advice.

beiwei30:
this pull request not valid any longer, has talked to the author.

Issue #2981, opened 2018-12-17
dubbo2.6.3 服务提供者中使用泛化调用时，绑定attachments时interface错误


cadeeper:
- [*] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [*] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.6.3
* Operating System version: macos
* Java version: 1.8.0

### Steps to reproduce this issue

dubbo服务提供者中使用泛化调用时，绑定attachments时interface错误，导致thrift encode失败：Could not encode request, the specified interface may be incorrect.
1. 调用方调用 interfaceA.methodA方法
2. methodA中通过泛化调用interfaceB.methodB方法
3. attachements中的interface参数为interfaceA，而不是interfaceB

问题出现在AbstractClusterInvoker中的233行：
`
        // binding attachments into invocation.

        Map<String, String> contextAttachments = RpcContext.getContext().getAttachments();

        if (contextAttachments != null && contextAttachments.size() != 0) {

            ((RpcInvocation) invocation).addAttachments(contextAttachments);

        }

`

carryxyh:
Hi, can u provide a demo to reproduce this problem?
I'd like take a look.

cadeeper:
不好意思，之后上传demo，目前我查到的原因如下：
AbstractInvoker.java
![image](https://user-images.githubusercontent.com/5336857/50152042-86058180-02fd-11e9-8f6c-6cbcc70b5d4d.png)

carryxyh:
Seems like this problem will occur whether use thrift protocol or dubbo protocol right ?
Pls send a simple demo that can reproduce this problem.

cadeeper:
> Seems like this problem will occur whether use thrift protocol or dubbo protocol right ?
> Pls send a simple demo that can reproduce this problem.

@carryxyh  提供了重现demo：
https://github.com/cadeeper/dubbo-thrift-bug-demo/tree/master

carryxyh:
@cadeeper 
Ok, I'd like have a look.

carryxyh:
Hello, I don't know about the thrift protocol. I can try to reproduce this problem myself.

1. Consumer invoke interface1
2. in interface1's impl, generic invoke interface2

[...]

PR #3017, opened 2018-12-19
Refresh invocation's attachments in each invoke.

carryxyh:
Refresh invocation's attachments in each invoke.
Fix bug:
https://github.com/apache/incubator-dubbo/issues/2981

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java
@@ -150,9 +150,9 @@ public abstract class AbstractInvoker<T> implements Invoker<T> {
         }
         RpcUtils.attachInvocationIdIfAsync(getUrl(), invocation);
 
-        // If we have attachment, refresh the attachment to ensure that the last invoke's attach does not affect this call.
+        // If we have attachment, override the attachment to ensure that the last invoke's attach does not affect this call.
         // In most cases, attachments contains interface, timeout, group, and token.
-        // These info should be refreshed in each invoke.
+        // These info should be override in each invoke.
         //
         // See https://github.com/apache/incubator-dubbo/issues/2981
         if (attachment != null && attachment.size() > 0) {
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java
@@ -134,9 +134,7 @@ public abstract class AbstractInvoker<T> implements Invoker<T> {
         }
         RpcInvocation invocation = (RpcInvocation) inv;
         invocation.setInvoker(this);
-        if (attachment != null && attachment.size() > 0) {
-            invocation.addAttachmentsIfAbsent(attachment);
-        }
+
         Map<String, String> contextAttachments = RpcContext.getContext().getAttachments();
         if (contextAttachments != null && contextAttachments.size() != 0) {
             /**
@@ -152,6 +150,14 @@ public abstract class AbstractInvoker<T> implements Invoker<T> {
         }
         RpcUtils.attachInvocationIdIfAsync(getUrl(), invocation);
 
+        // If we have attachment, refresh the attachment to ensure that the last invoke's attach does not affect this call.
+        // In most cases, attachments contains interface, timeout, group, and token.
+        // These info should be refreshed in each invoke.
+        //
+        // See https://github.com/apache/incubator-dubbo/issues/2981
+        if (attachment != null && attachment.size() > 0) {
+            invocation.addAttachments(attachment);
+        }
 
         try {
             return doInvoke(invocation);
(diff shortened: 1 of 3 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java
```diff
@@ -151,6 +149,14 @@ public Result invoke(Invocation inv) throws RpcException {
         }
         RpcUtils.attachInvocationIdIfAsync(getUrl(), invocation);
 
+        // If we have attachment, refresh the attachment to ensure that the last invoke's attach does not affect this call.
+        // In most cases, attachments contains interface, timeout, group, and token.
+        // These info should be refreshed in each invoke.
+        //
+        // See https://github.com/apache/incubator-dubbo/issues/2981
+        if (attachment != null && attachment.size() > 0) {
+            invocation.addAttachments(attachment);
```
khanimteyaz: Just curious to know, as comment indicates refreshing timeout, group, and token. Do we need any place to be also refresh other than invocation?

carryxyh: Hey, maybe the description here is a bit problematic. We are not refreshing but covering. I will modify it later.

The global attachment here is information such as interface and timeout obtained from the URL. We want to make sure that the interface and timeout used for each call are in the url of the current invoker, not the other way. Other ways such as: RpcContext.getContext.setAttachment
So, before invoke, we should override these info.
:)

chickenlj: What if user has overridden the `timeout` or `token` value in RpcContext/Invocation, if you put replace `addAttachmentsIfAbsent ` with `addAttachments` and put it at the last place,  the user will never has a chance to override the keys listed here https://github.com/apache/incubator-dubbo/blob/e07038b00e057bb3cd6989252534b581039916f9/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java#L78

chickenlj: Cannot we just remove  or replace the INTERFACE_KEY with the right one?

chickenlj: I even find that the INTERFACE_KEY useless on the wire.  Instead `path` was used instead:
https://github.com/apache/incubator-dubbo/blob/e07038b00e057bb3cd6989252534b581039916f9/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java#L215

chickenlj:
A proposal, not strictly related to this issue, only a thought.

There are [Invocation.attachments](), [RpcContext.attachments](), [RpcContext.server_attachments](), we should have someone to figure out the relations, scopes and possible problems each kind of attachment.

Some problems I already know
1. 
```java
// `key1` will be passed though the RPC chain to the bottom: A -> B -> C -> D
RpcContext.getContext.setAttachment("key1", "value1");
```
2. 
```java
//  `server-key1` will be passed though the RPC chain to the top: A <- B <- C <- D
RpcContext.getServerContext.setAttachment("server-key1", "server-value1");
```

carryxyh:
yep, I'd like open an issue to track this.

What is your opinion on this pr?

If we just use the correct interface, but timeout, group, and token may still have the same problem as the interface. Should we consider this scenario?
Or should we modify this issue after adjusting the issue of attachment?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java
chickenlj: https://github.com/apache/incubator-dubbo/blob/33f1726713a4c5f2ce5880613acf2320ae2a4502/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/ContextFilter.java#L46

I am simply thinking of remove INTERFACE_KEY in here. GROUP_KEY, TOKEN_KEY and TIMEOUT_KEY would not have this problem because they've already been removed.

carryxyh: Ok!
i haven't notice it yet...

carryxyh: yes.
This problem does happen in the process of using Thrift.

carryxyh:
@chickenlj 
Already fix it.

chickenlj:
Good job, we can try to simplify the Dubbo Protocol further https://github.com/apache/incubator-dubbo/issues/3328

chickenlj:
I will try to update the Thrift protocol with another PR.

carryxyh:
Ok, thx for your review.
:)

PR #3018, opened 2018-12-19
[2.7.0] add a listener spi for application init

Jeff-Lv:

we need to do some stuff when application init.

## Brief changelog

Add a listner SPI and init method to applicationmodel. We can do some customized stuff when applicaiton init.


lovepoem:
Why so many changes?

Jeff-Lv:
> Why so many changes?

Sorry.small changes only andplease che ck again.

Issue #3026, opened 2018-12-20
AccessLogFilter simple date formate reduce instance creation


khanimteyaz:
AcceLogFilter message creation and last and current check was performing by each time simple date formate object creation. In this version of code I have use the ThreadLocal for each thread wise to create simple date format object instead of each call.

- [ ] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.0
* Operating System version: xxx
* Java version: xxx

### Steps to reproduce this issue

1. xxx
2. xxx
3. xxx

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result

What do you expected from the above steps？

### Actual Result

What actually happens?
AcceLogFilter message creation and last and current check was performing by each time simple date formate object creation. In this version of code I have use the ThreadLocal for each thread wise to create simple date format object instead of each call.
If there is an exception, please attach the exception trace:

```
Just put your stack trace here!
```

carryxyh:
Hi, I leave some comments, would u pls check it?
:)

khanimteyaz:
> Hi, I leave some comments, would u pls check it?
> :)

Thanks, I will.

PR #3027, opened 2018-12-20
Simple date format each time new object creation removed

khanimteyaz:

## Brief changelog

XXXXX

## Verifying this change

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
@@ -6,7 +6,7 @@
  * (the "License"); you may not use this file except in compliance with
  * the License.  You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licaenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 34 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -75,6 +75,20 @@
...
     private static final long LOG_OUTPUT_INTERVAL = 5000;
 
+    static ThreadLocal<SimpleDateFormat> FILE_DATE_FORMATTER= new ThreadLocal<SimpleDateFormat>() {
+        @Override
+        protected SimpleDateFormat initialValue() {
+            return new SimpleDateFormat(FILE_DATE_FORMAT);
+        }
+    };
+
+    static ThreadLocal<SimpleDateFormat> MESSAGE_DATE_FORMATTER= new ThreadLocal<SimpleDateFormat>() {
```
carryxyh: Hi, khan
We have integrated Netty's FastThreadLocal in dubbo. I think we can use it instead of ThreadLocal. I suggest you refer to RpcContext.LOCAL.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -113,7 +127,7 @@ public Result invoke(Invoker<?> invoker, Invocation inv) throws RpcException {
                 String version = invoker.getUrl().getParameter(Constants.VERSION_KEY);
                 String group = invoker.getUrl().getParameter(Constants.GROUP_KEY);
                 StringBuilder sn = new StringBuilder();
-                sn.append("[").append(new SimpleDateFormat(MESSAGE_DATE_FORMAT).format(new Date())).append("] ").append(context.getRemoteHost()).append(":").append(context.getRemotePort())
+                sn.append("[").append(MESSAGE_DATE_FORMATTER.get().format(new Date())).append("] ").append(context.getRemoteHost()).append(":").append(context.getRemotePort())
```
carryxyh: Another thing we should consider is that when we clear the threadlocal's value.
Should we keep them in memory all the time?

beiwei30:
If we can move all date time format into org.apache.dubbo.rpc.filter.AccessLogFilter.LogTask, then we can safely make the formatter static, since it is now running on a single thread.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: Thanks for suggesting this. I was looking into RpcContext code for LOCAL and SERVER implementation, I would like to go this approach.

khanimteyaz:
> If we can move all date time format into org.apache.dubbo.rpc.filter.AccessLogFilter.LogTask, then we can safely make the formatter static, since it is now running on a single thread.
@beiwei30 
I am planning for move FILE_DATE_FORMATTER to LogTask level (as you are suggesting, i also beleive it will make sense) as this is created once for each log thread(which is currently single threaded). as of now to simply intiale SimpleDateFormat once per LogTask thread instance as it is created (as of now it is not served from thread pool, when we have thread pool can change it accordingly) and MESSAGE_DATE_FORMATTER to RpcContext level. Would you correct me if this approach has problem. Any guidance will be appreciated.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: @carryxyh 
If the current thread which execute this **invoke** is from a thread pool then I think we should keep it because then it will be reusable and will avoid the time creating new date formatter object, otherwise we should remove it. 
What would you suggest to me?

khanimteyaz:
Hi all, would you care to review it.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: @carryxyh would you suggest is there any things I needs to change here?

khanimteyaz:
@beiwei30, would you suggest here me any thing ?

khanimteyaz:
> @khanimteyaz
> Hi,
> I have not found a LogTask in AccessLogFilter before. After I looked at the code, I personally thought that it would be fine to put all the log output into the LogTask.
It make sense writing to log through this thread, but just to confirm you mean writing to log only, not the log message creation part?
> 
> I think we can use the newSingleThreadScheduledExecutor to create a timed task. This way we can just create a SimpleDateFormat once (bind to a LogTask).
> How do u think?
Performing all write through single instance by **newSingleThreadScheduledExecutor ** is good approach and I am with you on this, only thing which is stoping me how in the case of writing pending log message in case of jvm shutdown. I think jdk provide a way for this and using the **scheduleWithFixedDelay** of **ScheduledExecutorService**. 

@carryxyh 
  Do you feel, the approach I am taking is right or would you suggest me something else here on the implementation?

carryxyh:
@khanimteyaz 
IMO, the log message will create and log in schedule thread.

> how in the case of writing pending log message in case of jvm shutdown

What I may understand about this sentence is not very accurate. What you mean is, how do you write the pending information to a file at the end of the JVM? I think we can let the scheduled thread daemon.

> Do you feel, the approach I am taking is right or would you suggest me something else here on the implementation?

I think we should change the `logQueue` to let us create log message in another thread.
Maybe like this: `ConcurrentMap<String, Set<LogData>> logQueue`. And then, our scheduled thread just check the queue regularly. This is my thought. After that, we can promise our log is print in only one thread and make the SimpleDateFormat single instance.

khanimteyaz:
@carryxyh 

> What I may understand about this sentence is not very accurate. What you mean is, how do you write the pending information to a file at the end of the JVM? I think we can let the scheduled thread daemon.

I am think I was wrong on the pending log message, thanks for correcting me and helping out here. Let the schedule thread to finish the pending task 😄 

Regarding 

> I think we should change the logQueue to let us create log message in another thread.
Maybe like this: ConcurrentMap<String, Set<LogData>> logQueue. And then, our scheduled thread just check the queue regularly. This is my thought. After that, we can promise our log is print in only one thread and make the SimpleDateFormat single instance.

I will make the changes accordingly.

khanimteyaz:
@carryxyh I have made the changes but builds failing due to travis log limit.

carryxyh:
Hi, seems like your pr has many commits of other persons.
Would u pls clear your pr first? I saw that this pr has changed about 600 lines code.

About the ci, I will have a look.

khanimteyaz:
> Hi, seems like your pr has many commits of other persons.
> Would u pls clear your pr first? I saw that this pr has changed about 600 lines code.
> 
> About the ci, I will have a look.

  Sure. on it.

khanimteyaz:
I will create a new branch where, would raised all the changes with all recommendation

khanimteyaz:
@carryxyh  created another PR for the same https://github.com/apache/incubator-dubbo/pull/3080

As I am not sure how to reduce number of commit from existing PR, so created a new one. Sorry for the inconvenience. 
Would request you to guide on this, how to reduce or clear other's commit from existing one, so that next time I can avoid creating separate PR.

PR #3035, opened 2018-12-21
[DUBBO-3023]: problem in ActiveLimitFilter

beiwei30:

make count work again when max <= 0

## Brief changelog

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/ActiveLimitFilter.java
dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/ExecuteLimitFilter.java

## Verifying this change

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
@@ -111,7 +111,7 @@ public class RpcStatus {
     public static boolean beginCount(URL url, String methodName, int max) {
         RpcStatus appStatus = getStatus(url);
         RpcStatus methodStatus = getStatus(url, methodName);
-        if (methodStatus.active.incrementAndGet() > max) {
+        if (max > 0 && methodStatus.active.incrementAndGet() > max) {
             methodStatus.active.decrementAndGet();
             return false;
         } else {
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
@@ -109,9 +109,10 @@ public class RpcStatus {
      * @param url
      */
     public static boolean beginCount(URL url, String methodName, int max) {
+        max = (max <= 0) ? Integer.MAX_VALUE : max;
         RpcStatus appStatus = getStatus(url);
         RpcStatus methodStatus = getStatus(url, methodName);
-        if (max > 0 && methodStatus.active.incrementAndGet() > max) {
+        if (methodStatus.active.incrementAndGet() > max) {
             methodStatus.active.decrementAndGet();
             return false;
         } else {
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/ActiveLimitFilter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/ActiveLimitFilter.java
@@ -47,7 +47,7 @@ public class ActiveLimitFilter implements Filter {
         String methodName = invocation.getMethodName();
         int max = invoker.getUrl().getMethodParameter(methodName, Constants.ACTIVES_KEY, 0);
         RpcStatus count = RpcStatus.getStatus(invoker.getUrl(), invocation.getMethodName());
-        if (max > 0 && !RpcStatus.beginCount(url, methodName, max)) {
+        if (!RpcStatus.beginCount(url, methodName, max)) {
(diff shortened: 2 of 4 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
```diff
@@ -111,7 +111,7 @@ public static void beginCount(URL url, String methodName) {
     public static boolean beginCount(URL url, String methodName, int max) {
         RpcStatus appStatus = getStatus(url);
         RpcStatus methodStatus = getStatus(url, methodName);
-        if (methodStatus.active.incrementAndGet() > max) {
+        if (max > 0 && methodStatus.active.incrementAndGet() > max) {
```
khanimteyaz: If we add max>0 in the condition there it will cause the issue because in finally block we always reducing the endCount irrespective of max value. 

e.g. 
1)max value is 0 so in RpcStatus.java's begin count we will not increment the counter but in ActiveLimitFilter's invoke's finally block we will be always reducing so the final value for single invocation will landed up -1 active count.

I think not  including max>0 and adding 
} finally {      
    if (max > 0) {
                RpcStatus.endCount(url, methodName, System.currentTimeMillis() - begin, isSuccess);
                synchronized (count) {
                    count.notifyAll();
                   }
               }
         }
In ActiveLimitFilter should also work, but I beleive you can rewrite this into more clear way then we suggest.


I think we might have another issue too (not because of your changes), in a scenario if max>0 (in RpcStatus.java) get satisfied (or lets max>0 condition is not even present in code as earlier version) and the `methodStatus.active.incrementAndGet() > max` return `false` then code of else block will be executing and again it will increment the count. So for one call we are eventually incrementing counter twice. Do you feel this might be also a another problem in future?Correct me if my understanding not correct.

khanimteyaz:
Would it possible to add UT for the same, or would like me to add once you done?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
beiwei30: first of all, `endCount` does more than limiting the traffic only, it is also responsible for running statistics, see below:

```java
    private static void endCount(RpcStatus status, long elapsed, boolean succeeded) {
        status.active.decrementAndGet();
        status.total.incrementAndGet();
        status.totalElapsed.addAndGet(elapsed);
        if (status.maxElapsed.get() < elapsed) {
            status.maxElapsed.set(elapsed);
        }
        if (succeeded) {
            if (status.succeededMaxElapsed.get() < elapsed) {
                status.succeededMaxElapsed.set(elapsed);
            }
        } else {
            status.failed.incrementAndGet();
            status.failedElapsed.addAndGet(elapsed);
            if (status.failedMaxElapsed.get() < elapsed) {
                status.failedMaxElapsed.set(elapsed);
            }
        }
    }
```

That's the reason we need to make sure `startCount` and `endCount` execute no matter `max` limit is set or not, but at the same time, we also need to make sure the filter will block incoming request if `max` limit exceeds.

beiwei30: In my current implementation:

```java
if (max > 0 && methodStatus.active.incrementAndGet() > max) {
   return false;
} else {
  return true;
}
```

it satisfies what I explained above. When limit exceeds, it will return false, and since it is atomic operation, it can make sure further incoming requests to wait. If limit doesn't exceeds or is not set, then it will return true and allow the requests to continue, instead of wait.

More important, if we decide to allow the request to proceed, we need to make sure 'startCount' and 'endCount' to be called in pair.

beiwei30:
> Would it possible to add UT for the same, or would you like me to add once you are done?

since you are in the middle of adding more unit test for this class, I'm prefer to leave it to you, what do you think :)

khanimteyaz:
> > Would it possible to add UT for the same, or would you like me to add once you are done?
> 
> since you are in the middle of adding more unit test for this class, I'm prefer to leave it to you, what do you think :)

Yes fine with me. Will take care of it once it is merge.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
beiwei30: can I consider this topic is resolved?

beiwei30: I see where the problem is. Pls. review the latest commit.

beiwei30:
more comments, @khanimteyaz

khanimteyaz:
> more comments? @khanimteyaz

You have made it such clean beautiful that I am comment less 😃 . I appreciate your effort. Thanks for taking care of cases.

PR looks good to me. 
As I don't have permission I can't merge it, some one have to merge it.

PR #3043, opened 2018-12-21
wrong event setting

CrazyHZM:

Event setting is wrong when the event is READONLY_EVENT.

## Brief changelog

XXXXX

## Verifying this change

XXXXX


dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
```diff
@@ -175,25 +172,35 @@ protected Object decodeBody(Channel channel, InputStream is, byte[] header) thro
             Request req = new Request(id);
             req.setVersion(Version.getProtocolVersion());
             req.setTwoWay((flag & FLAG_TWOWAY) != 0);
-            if ((flag & FLAG_EVENT) != 0) {
-                req.setEvent(Request.HEARTBEAT_EVENT);
-            }
+            Object data = null;
```
khanimteyaz: For the shake of mine understanding `if(flag & FLAG_REQUEST) == 0) ` true, then it is response scenario and if it is false means a request has been made to the service(which can be normal request, heartbeat or event received ). Is this understanding of mine correct?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
```diff
@@ -175,25 +172,35 @@ protected Object decodeBody(Channel channel, InputStream is, byte[] header) thro
...
                 }
-                req.setData(data);
             } catch (Throwable t) {
                 // bad request
                 req.setBroken(true);
                 req.setData(t);
             }
+            if (!req.isBroken()) {
+                if ((flag & FLAG_EVENT) != 0) {
+                    if (data != null && data.equals(Request.READONLY_EVENT)) {
```
khanimteyaz: `if(Request.READONLY_EVENT.equals(data)` would be more simplified i think.

CrazyHZM: If this is bad request，`req.setData(t);` will be executed，so I think `if (!req.isBroken())` is necessary.
About `if ((flag & FLAG_EVENT) != 0)` , I think that plus him is just for insurance .
for example value of data  is "R" but it is not an event.

khanimteyaz: Thanks for explaining this. but what I was saying is for line no 192, is 
instead of having `if (data != null && data.equals(Request.READONLY_EVENT))`  we could simplify this by just having
`if(Request.READONLY_EVENT.equals(data))`

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
```diff
@@ -175,25 +172,35 @@ protected Object decodeBody(Channel channel, InputStream is, byte[] header) thro
...
                     data = decodeRequestData(channel, in);
                 }
-                req.setData(data);
             } catch (Throwable t) {
                 // bad request
                 req.setBroken(true);
                 req.setData(t);
             }
+            if (!req.isBroken()) {
+                if ((flag & FLAG_EVENT) != 0) {
```
khanimteyaz: In case if the condition `if ((flag & FLAG_EVENT) != 0)` does not gets satisfied then, what will be the event value of request object? Is there any default event type?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
```diff
@@ -175,25 +172,35 @@ protected Object decodeBody(Channel channel, InputStream is, byte[] header) thro
...
                     data = decodeHeartbeatData(channel, in);
                 } else if (req.isEvent()) {
                     data = decodeEventData(channel, in);
                 } else {
                     data = decodeRequestData(channel, in);
                 }
-                req.setData(data);
             } catch (Throwable t) {
                 // bad request
                 req.setBroken(true);
```
khanimteyaz: Just for my curiosity wanted to know, in case if request is broken (I am guessing if we encounter exception and at the catch block means request is broken) then still we are proceeding and bulding the request object ?Is this understanding of my is correct?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
CrazyHZM: Thank you,It seems that you are right now, but if the value of READONLY_EVENT becomes empty in the future, this will be a hidden danger. Of course, this does not change much from the current point of view.
What are the advantages of removing the judgment?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
CrazyHZM: Yes, as far as I know, dubbo has a timeout retry mechanism

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
khanimteyaz: With the limited knowledege about this part (so @chickenlj  @carryxyh @beiwei30  correct me if I am not correct). I think we should either have a default event type if noting is passed in data(more informative way and declaring a default behaviour) or we should reject the request if no default behaviour is spported.  It will make the easy to debug and reason about the issue. 

If we provide some default behaviour (event type) and which is not communicated, for user it might be confusiong (because it is not documented or communicated to them).

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
khanimteyaz: It is always preferable to have more simplified code if can make it 😄 . In this case if READONLY_EVENT  value become empty which is big change because it might involve changes of user code so then we can still add the judgement back (in future we might even have to add more conditions to judgement).

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
CrazyHZM: It was used in handleRequest method of HeaderExchangeHandler.java

khanimteyaz: thanks

khanimteyaz:
@carryxyh @beiwei30 could you also have an look into this PR. If it looks good to you then we can merge it.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/Request.java
```diff
@@ -99,6 +99,10 @@ public void setEvent(String event) {
         mData = event;
     }
 
+    public void setEvent(boolean mEvent) {
```
beiwei30: looks like we should remove

```java
   public void setEvent(String event) {
        mEvent = true;
        mData = event;
    }
```

beiwei30:
@CrazyHZM, would you mind to take a look at my comment? I think the root cause is 

```java
            if ((flag & FLAG_EVENT) != 0) {
                res.setEvent(Response.HEARTBEAT_EVENT);
            }
```

once we change it to 

```java
            if ((flag & FLAG_EVENT) != 0) {
                res.setEvent(true);
            }
```

then we should not need the following logic any longer:

```java
if (!req.isBroken()) {
                if ((flag & FLAG_EVENT) != 0) {
                    if (data != null && data.equals(Request.READONLY_EVENT)) {
                        req.setEvent(Request.READONLY_EVENT);
                    } else {
                        req.setEvent(Request.HEARTBEAT_EVENT);
                    }
                }
                req.setData(data);
            } else {
                if ((flag & FLAG_EVENT) != 0) {
                    req.setEvent(true);
                }
            }
```

And, in your change, you consider request only, pls. take care response too.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/Request.java
beiwei30: now I see why the original setEvent() is necessary, pls. keep it. Sorry about this.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeServer.java
```diff
@@ -128,7 +128,8 @@ public void startClose() {
 
     private void sendChannelReadOnlyEvent() {
         Request request = new Request();
-        request.setEvent(Request.READONLY_EVENT);
+        request.setEvent(true);
```
beiwei30: the original `setEvent` is necessary to keep, so we will end up with two `setEvent` methods:

```java
void setEvent(bool);
void setEvent(String);
```

beiwei30:
@CrazyHZM I leave more comments, pls. take a look. I realize we need to keep the original `setEvent` method.

CrazyHZM:
> @CrazyHZM I leave more comments, pls. take a look. I realize we need to keep the original `setEvent` method.

I have modified it

CrazyHZM:
@beiwei30 I have modified , can you review it , pls.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/Request.java
khanimteyaz: Setting only the mEvent indicate that there an event but it does tell what kind of event it was. So should we set just set single value? I beleive setting only boolean may not provide complete information what do you say?

CrazyHZM: > Setting only the mEvent indicate that there an event but it does tell what kind of event it was. So should we set just set single value? I beleive setting only boolean may not provide complete information what do you say?

Because mData will carry the relevant value, so here is simply to mark whether the request is an event

beiwei30:
This change now looks good to me.

It looks like we should do the same thing for response, change

```java
            if ((flag & FLAG_EVENT) != 0) {
                res.setEvent(Response.HEARTBEAT_EVENT);
            }
```

to 

```java
            if ((flag & FLAG_EVENT) != 0) {
                res.setEvent(true);
            }
```

PR #3064, opened 2018-12-26
Review code of TypeDefinitionBuilder

carryxyh:
Review code of TypeDefinitionBuilder
1. use init method to init builds' list

@cvictory Would u pls have a look?

### DIFF
--- a/dubbo-metadata-report/dubbo-metadata-definition/src/main/java/org/apache/dubbo/metadata/definition/TypeDefinitionBuilder.java
+++ b/dubbo-metadata-report/dubbo-metadata-definition/src/main/java/org/apache/dubbo/metadata/definition/TypeDefinitionBuilder.java
@@ -35,20 +35,18 @@ import java.util.Map;
  */
 public class TypeDefinitionBuilder {
 
-    private static final ThreadLocal<ArrayList<TypeBuilder>> builders;
-
-    static {
-        builders = new ThreadLocal<ArrayList<TypeBuilder>>();
-        builders.set(new ArrayList<TypeBuilder>());
-        builders.get().add(new ArrayTypeBuilder());
-        builders.get().add(new CollectionTypeBuilder());
-        builders.get().add(new MapTypeBuilder());
-        builders.get().add(new EnumTypeBuilder());
-    }
+    private static final ThreadLocal<ArrayList<TypeBuilder>> builders = ThreadLocal.withInitial(() -> {
+        ArrayList<TypeBuilder> l = new ArrayList<>();
+        l.add(new ArrayTypeBuilder());
+        l.add(new CollectionTypeBuilder());
+        l.add(new MapTypeBuilder());
+        l.add(new EnumTypeBuilder());
+        return l;
+    });
 
     public static TypeDefinition build(Type type, Class<?> clazz, Map<Class<?>, TypeDefinition> typeCache) {
         TypeBuilder builder = getGenericTypeBuilder(type, clazz);
-        TypeDefinition td = null;
+        TypeDefinition td;
         if (builder != null) {
             td = builder.build(type, clazz, typeCache);
         } else {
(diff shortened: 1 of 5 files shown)


dubbo-metadata-report/dubbo-metadata-definition/src/main/java/org/apache/dubbo/metadata/definition/TypeDefinitionBuilder.java
```diff
@@ -35,20 +35,18 @@
...
-
-    static {
-        builders = new ThreadLocal<ArrayList<TypeBuilder>>();
-        builders.set(new ArrayList<TypeBuilder>());
-        builders.get().add(new ArrayTypeBuilder());
-        builders.get().add(new CollectionTypeBuilder());
-        builders.get().add(new MapTypeBuilder());
-        builders.get().add(new EnumTypeBuilder());
-    }
+    private static final ThreadLocal<ArrayList<TypeBuilder>> builders = ThreadLocal.withInitial(() -> {
```
khanimteyaz: Can we **InternalThreadLocal** instead of ThreadLocal

dubbo-metadata-report/dubbo-metadata-definition/src/main/java/org/apache/dubbo/metadata/definition/TypeDefinitionBuilder.java
```diff
@@ -35,20 +35,18 @@
...
+        ArrayList<TypeBuilder> l = new ArrayList<>();
```
khanimteyaz: is this suppose to be modified by any of the caller? If not then should we consider using a immutable list here?

dubbo-metadata-report/dubbo-metadata-definition/src/main/java/org/apache/dubbo/metadata/definition/TypeDefinitionBuilder.java
carryxyh: Hi, because of the InternalThreadLocal depend on InternalThread.
But these method is use from spring's thread or user thread.
So I don't use InternalThreadLocal..

dubbo-metadata-report/dubbo-metadata-definition/src/main/java/org/apache/dubbo/metadata/definition/TypeDefinitionBuilder.java
carryxyh: emmm...
Personally think immutable list maybe better.
I'd like fix this.

dubbo-metadata-report/dubbo-metadata-definition/src/main/java/org/apache/dubbo/metadata/definition/TypeDefinitionBuilder.java
khanimteyaz: ok. 

For the sake of my learning, would it be possible for explain it bit more so that I can understand the wisdom which lies in the combination of spring + threadlocal which you are using?

dubbo-metadata-report/dubbo-metadata-definition/src/main/java/org/apache/dubbo/metadata/definition/TypeDefinitionBuilder.java
```diff
@@ -35,20 +36,16 @@
...
 
-    static {
-        builders = new ThreadLocal<ArrayList<TypeBuilder>>();
-        builders.set(new ArrayList<TypeBuilder>());
-        builders.get().add(new ArrayTypeBuilder());
-        builders.get().add(new CollectionTypeBuilder());
-        builders.get().add(new MapTypeBuilder());
-        builders.get().add(new EnumTypeBuilder());
-    }
+    private static final List<TypeBuilder> BUILDERS = Arrays.asList(ARRAY_BUILDER, COLLECTION_BUILDER, MAP_BUILDER, ENUM_BUILDER);
```
ralf0131: Can you explain why you are removing `ThreadLocal`?

carryxyh: Hi, huxing. Sorry, I missed your message before.

I checked the implementation of the associated TypeBuilder and it seems that they are all thread-safe (no global variables). So we just need to use static final decoration.

PR #3078, opened 2018-12-27
import a pojo ServiceMeta which is used as service level info and  a addressListener to extends

Jeff-Lv:

If URL is used for data transferred before loadbalancer decides the remote peer, we may interator all list of URL with low efficence because  URL as the core model of dubbo is a level of peer remote. So we need a service level model like this.

add a addressListner for adding more info to servicemetada according to URLs updates when registry directory got updated addresses

## Brief changelog

A new pojo class and a listener class are suggested to import.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/ServiceMetada.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/ServiceMetada.java
@@ -1,4 +1,4 @@
-package org.apache.dubbo.common;/*
+/*
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
  * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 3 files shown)


chickenlj:
I think this patch should be merged after the release of 2.7.0.

Or maybe it's time to create a separate branch for v2.7.0 to avoid blocking the master now.

dubbo-common/src/main/java/org/apache/dubbo/common/ServiceMetada.java
```diff
@@ -0,0 +1,61 @@
+package org.apache.dubbo.common;/*
```
ralf0131: License info should appear first.

chickenlj:
URL is one core concept in Dubbo.
I agree with the defects you mentioned about the URL mechanism and I understand the benefits 
 `ServiceMetadata` can bring. But I think it's a big step which will result in massive refactor that should be considered as a whole with what Ian has proposed about 3.0 on the mailing list.

PR #3090, opened 2018-12-28
Acesslog dateformat enhancemnet

khanimteyaz:

In this PR 

1.  I have moved the access log creation and writing to file into a single thread and reusing a single instance of **SimpleDateFormat**
2. Refactored code to separate our and group related task in separate method and have enhance the readability by using 
         -  Method naming 
         - Reducing big methods to small.

This PR is a resurrection of old PR 
   1. 3027 (https://github.com/apache/incubator-dubbo/pull/3027) PR
   2.  3080 (https://github.com/apache/incubator-dubbo/pull/3080)
Sorry for the inconvenience 
## Brief changelog

XXXXX

## Verifying this change


khanimteyaz:
@carryxyh could you please review this.

beiwei30:
I didn't look into the change closely, but it looks like we are now moving the date formatter into one single thread to avoid contention completely, which I believe it is the right direction as I pointed out in the old pull request :)

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -67,139 +67,143 @@
...
-
     private static final int LOG_MAX_BUFFER = 5000;
 
     private static final long LOG_OUTPUT_INTERVAL = 5000;
 
-    private final ConcurrentMap<String, Set<String>> logQueue = new ConcurrentHashMap<String, Set<String>>();
+    private static final SimpleDateFormat FILE_DATE_FORMATTER = new SimpleDateFormat("yyyyMMdd");
 
-    private final ScheduledExecutorService logScheduled = Executors.newScheduledThreadPool(2, new NamedThreadFactory("Dubbo-Access-Log", true));
+    private static final SimpleDateFormat MESSAGE_DATE_FORMATTER = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
```
zonghaishang: not thread safe

khanimteyaz:
I think the whole log parsing and writing is happening in a single thread environment , irrespective of invoke. Would you provide us more details so that I can fix it if I can ?

khanimteyaz:
@zonghaishang could you please provide the details about the thread safety so that  I can work on this?

@beiwei30 Is there something to incorporate which you feel here?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -67,139 +67,143 @@
...
-    private void log(String accesslog, String logmessage) {
-        init();
-        Set<String> logSet = logQueue.get(accesslog);
-        if (logSet == null) {
-            logQueue.putIfAbsent(accesslog, new ConcurrentHashSet<String>());
-            logSet = logQueue.get(accesslog);
-        }
-        if (logSet.size() < LOG_MAX_BUFFER) {
-            logSet.add(logmessage);
+    private static final InternalThreadLocal<SimpleDateFormat> INLINE_MESSAGE_DATE_FORMATTER = new InternalThreadLocal<SimpleDateFormat>() {
```
carryxyh: Seems like this filed is never used.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -67,139 +67,143 @@
...
     private static final long LOG_OUTPUT_INTERVAL = 5000;
 
-    private final ConcurrentMap<String, Set<String>> logQueue = new ConcurrentHashMap<String, Set<String>>();
+    private static final SimpleDateFormat FILE_DATE_FORMATTER = new SimpleDateFormat("yyyyMMdd");
 
-    private final ScheduledExecutorService logScheduled = Executors.newScheduledThreadPool(2, new NamedThreadFactory("Dubbo-Access-Log", true));
+    private static final SimpleDateFormat MESSAGE_DATE_FORMATTER = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
 
-    private volatile ScheduledFuture<?> logFuture = null;
+    private final ConcurrentMap<String, Set<AccessLogData>> logQueue = new ConcurrentHashMap<String, Set<AccessLogData>>();
```
carryxyh: Can this field be static?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -67,139 +67,143 @@
...
 
+    /**
+     * This method logs the access log for service method invocation call.
+     *
+     * @param invoker service
+     * @param inv     Invocation service method.
+     * @return Result from service method.
+     * @throws RpcException
+     */
+
```
carryxyh: remove this line.
:)

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -67,139 +67,143 @@
...
-                                    writer.write("\r\n");
-                                }
-                                writer.flush();
-                            } finally {
-                                writer.close();
-                            }
-                        } catch (Exception e) {
-                            logger.error(e.getMessage(), e);
+    private void log(String accessLog, AccessLogData accessLogData) {
+        Set<AccessLogData> logSet = logQueue.get(accessLog);
```
carryxyh: Can we use computeIfAbsent to optimize this?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/AccessLogData.java
```diff
@@ -0,0 +1,255 @@
...
+/**
+ * AccessLogData is a container for log event data. In internally uses map and store each filed of log as value. It does not generate any
+ * dynamic value e.g. time stamp, local jmv machine host address etc. It does not allow any null or empty key.
+ */
+public final class AccessLogData {
+    /**
+     * Access log keys
+     */
+    private enum KEYS {
+        VERSION, GROUP, SERVICE, METHOD_NAME, INVOCATION_TIME, TYPES
```
carryxyh: Why not make this to be fileds like:
```
private String group;
private String version;
...
```

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: yes. I will remove it.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: I think it make sense now to make it static as it is now bound to single thread. 

In addition to this, I was thinking (correct me If I am wrong), instead of defining the logQueue as a member of this class, we can it get through a factory (although factory will supply single queue only), so that in future even if we have change it based on our defined we don't have to touch this class. May be we can do this as part of separate PR. What do you say?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: I think, I did not understand this one. Which line you wanted me to remove?

carryxyh: 105，seems like this line separate the doc and the code?
:)

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
carryxyh: emmm...
What is this factory produce? Maybe I haven't get your point~ Can u give more details?

khanimteyaz: it provide the storage object which its consumer object, 

e.g.   
   ```
Store store= InMemoryQueueStorageFactory.getInstance().getStore();
   store.add(accessLog);
```

something like above. The above example what I hvae given may not be very intuitive itself for explaination, but  for high level understanding it might help. If you want I might create a sample branch to demonestrate in more detail way by having some implementation. What do you say?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: ok. got it.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/AccessLogData.java
khanimteyaz: Done

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: Done.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: Done.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: done

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
carryxyh: Personally think that it is not necessary, because it will make the code more complicated. I think that the current structure is relatively simple, haha.

khanimteyaz: 😄

khanimteyaz:
@carryxyh I have made the changes, would you plz review it.

carryxyh:
@khanimteyaz 
nice work!
I will take a look.

khanimteyaz:
I have updated with comment, would you review my pr?

jasonjoo2010:
Why not introducing `FastDateFormat`?

I think it's more simple and make sense.

jasonjoo2010:
> @jasonjoo2010
> Is the class thread-safe?

Yes, it's thread-safe by implemented no sharing data and also including instances cache.

Generally we can just call `FastDataFormat.getInstance("yyyyMMdd")` as same as `new SimpleDateFormat("yyyyMMdd")`.

carryxyh:
@khanimteyaz 
Would pls check whether 'FastDataFormat' is better?
:)

Thx for your suggestion. @jasonjoo2010

jasonjoo2010:
> @khanimteyaz
> Would pls check whether 'FastDataFormat' is better?
> :)
> 
> Thx for your suggestion. @jasonjoo2010

But it's in `commons-lang` or `commons-lang3` library maybe we need introduce a new library to project. It's the only point.

lixiaojiee:
SimpleDateFormat is not thread-safe, which is why we have to create new objects every time before. I did some research. Java8 currently has A thread-safe date tool called LocalDateTime. After reviewing @jasonjoo2010 's recommendation, I did some testing and found that FastDateFormat performs much better than the other two. Therefore, I also recommend FastDateFormat, but I prefer to bring it into the dubbo-commom package as a common date tool for other modules to reuse.You can refer to the data in the table below. The first column is the number of samples, and the following data is the average time corresponding to each sample in ms.

  | SimpleDateFormat | FastDateFormat | LocalDateTime
-- | -- | -- | --
10 | 3 | 1 | 77
100 | 10 | 4 | 83
1000 | 54 | 26 | 137
10000 | 170 | 52 | 168
100000 | 470 | 153 | 341
1000000 | 1827 | 501 | 1024
10000000 | 11591 | 3473 | 6228

**Note** : For environmental reasons, the above data are for reference only

carryxyh:
Do we copy code into dubbo?
Seems like we can depend on common-langs package.

carryxyh:
Sure
We only use 1.8 and above.

lixiaojiee:
> Thanks @carryxyh @jasonjoo2010 @lixiaojiee for reviewing the PR and provding feedback. I have incorporated the changes and have included common-lang3 as well. Could you please have a look again.

lgtm

jasonjoo2010:
> Thanks @carryxyh @jasonjoo2010 @lixiaojiee for reviewing the PR and provding feedback. I have incorporated the changes and have included common-lang3 as well. Could you please have a look again.

It looks good now.

PR #3095, opened 2018-12-29
remove @Async from core framework.

chickenlj:

move @AsyncFor from core framework to side project.

## Brief changelog

remove @AsyncFor, call async counterpart directly.

## Verifying this change

see dubbo/dubbo-samples


khanimteyaz:
@chickenlj would be possible to explain for mine understanding purpose moving **@asynch** from core to side project?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/RpcUtils.java
```diff
@@ -183,14 +182,6 @@ public static boolean isAsync(URL url, Invocation inv) {
...
 
-    public static boolean isGeneratedFuture(Invocation inv) {
-        return Boolean.TRUE.toString().equals(inv.getAttachment(Constants.FUTURE_GENERATED_KEY));
-    }
-
-    public static boolean hasGeneratedFuture(Method method) {
-        return method.isAnnotationPresent(AsyncFor.class) && hasFutureReturnType(method);
-    }
-
     public static boolean isFutureReturnType(Invocation inv) {
```
khanimteyaz: Just an thought should we consider this name to ** isReturnTypeFuture** instead of isFutureReturnType

chickenlj:
As you may have noticed, we support a service definition with a CompletableFuture as the return type.
```java
public interface AsyncService {
    CompletableFuture<String> sayHello(String name);
}
``` 

But this is newly introduced in 2.7.0, which means, there would be many legacy services that can not leverage this feature. Because if they change the service definition (the interface definition) they will have to reimplement their service to meet the new definition.
For example, here is a typical service definition before 2.7.0
```java
public interface GreetingsService {
    String sayHi(String name);
}
```
To be able to call this service asynchronously, they have use `RpcContext` to get the `Future`, like this:
```java
// the invoke will return null immediately
greetingsService.findFoo(fooId);
// get the Future instance, the future will get notified once the result returns.
Future<Foo> fooFuture = RpcContext.getContext().getFuture();
```
Annotation @AsyncFor is a mechanism I introduced to simplify the process for those legacy services, you can refer to here for how it works.

But now, as long as I realized we have supported java 8, we can use the default method to avoid generating a new class. [Here](https://github.com/dubbo/dubbo-async-processor/blob/master/README.md) is the new way I have come up with, this is what mean by saying `move the @AsyncFor from core to side project '. This way, this would become a convenient solution we provide but not a mandatory solution provided in core.

khanimteyaz:
> @khanimteyaz Thanks for your attention, I will try to explain.
> 
> As you may have noticed, we support a service definition with a CompletableFuture as the return type.
> 
> ```java
> public interface AsyncService {
>     CompletableFuture<String> sayHello(String name);
> }
> ```
> 
> But this is newly introduced in 2.7.0, which means, there would be many legacy services that can not leverage this feature. Because if they change the service definition (the interface definition) they will have to reimplement their service to meet the new definition.
> For example, here is a typical service definition before 2.7.0
> 
> ```java
> public interface GreetingsService {
>     String sayHi(String name);
> }
> ```
> 
> To be able to call this service asynchronously, they have to use `RpcContext` to get the `Future`, like this:
> 
> ```java
> // the invoke will return null immediately
> greetingsService.sayHi("world");
> // get the Future instance, the future will get notified once the result returns.
> Future<String> future = RpcContext.getContext().getFuture();
> ```
> 
> Annotation @AsyncFor is a mechanism I introduced to simplify the process for those legacy services, you can refer to [here](https://github.com/dubbo/dubbo-async-processor/tree/deprecated-processer) for how it works.
> 
> But now, as long as I realized we have supported java 8, we can use the default method to avoid generating a new class. [Here](https://github.com/dubbo/dubbo-async-processor/blob/master/README.md) is the new way I have come up with, this is what mean by saying `move the @AsyncFor from core to side project '. This way, this would become a convenient solution we provide but not a mandatory solution provided in core.

  @chickenlj Thanks a lot for the details. Appreciate it.

  I will give a look into this PR.

khanimteyaz:
This looks to fine ok to me. 

Thinking loud here (please correct me if I am wrong here). As here we are enabling dubbo consumer to  make async call to service method, but to achieve this, service provider has to define the stub with CompletableFuture. Here whether a consumer can make a async call depends on whether service developer has enabled service interface with CompletableFuture or not in some form. 

The decision of communication mode (async or async) is consumer choice. So thinking, how can we make our provider and consumer independent or each other in this context. 

e.g. here an an analogy I am using to provide the approach
java hashmap is not synchronized but java collections come up with Collection.jave and provided a utility method `java.util.Collections.synchronizedMap()` which actually wrap hashmap with thread safety behavior. 

@chickenlj Do you is there something we can do now or in future?

chickenlj:
> The decision of communication mode (async or async) is consumer choice. So thinking, how can we make our provider and consumer independent or each other in this context.

This is a good point @khanimteyaz, I think I got what you want to express.  Maybe we can also provide a utility method or sth like that. I am on a new year's vacation now, I will think about it as soon as I come back to work.

chickenlj:
> The decision of communication mode (async or async) is consumer choice. So thinking, how can we make our provider and consumer independent or each other in this context.

This is a good point @khanimteyaz, I think I got what you want to express.  Maybe we can also provide a utility method or sth like that. I am on a new year's vacation now, I will think about it as soon as I come back to work.

chickenlj:
By the way, the implementation of the async feature has been an enhncent or patch effort based on the legacy code. I think it definitely deserves an refactor. 
For example, you can find that we have three types of Result, they are`AsyncRpcResult`, `RpcResult`, `SimpleAsyncRpcResult`. 
* `RpcResult` are for sync call  
* `AsyncRpcResult` and `SimpleAsyncRpcResult` are for different types of async calls  
* `AsyncRpcResult` inside composes two futures.
But I think we should unify them to one RpcResult, which should be an instance of `CompletationStage`.

Given that hou have put forward many constructive suggestions recently, which are of great help to us. It would be great if you've got some experience or you'd like to take a look at this part.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/RpcUtils.java
chickenlj: done.

chickenlj:
I will merge this PR to prepare for the current release.

khanimteyaz:
> By the way, the implementation of the async feature has been an enhncent or patch effort based on the legacy code. I think it definitely deserves an refactor.
> For example, you can find that we have three types of Result, they are`AsyncRpcResult`, `RpcResult`, `SimpleAsyncRpcResult`.
> 
> * `RpcResult` are for sync call
> * `AsyncRpcResult` and `SimpleAsyncRpcResult` are for different types of async calls
> * `AsyncRpcResult` inside composes two futures.
>   But I think we should unify them to one RpcResult, which should be an instance of `CompletationStage`.
> 
> Given that hou have put forward many constructive suggestions recently, which are of great help to us. It would be great if you've got some experience or you'd like to take a look at this part.

@chickenlj Thanks for your detail information. I will try my best to suggest something which if there is something good I can comeup. Thanks for your appreciation. 
For the same I have created an issue [3113](https://github.com/apache/incubator-dubbo/issues/3113)

chickenlj:
> For the same I have created an issue #3113

Looking forward to it.

PR #3131, opened 2019-01-03
style: using computeIfAbsent instead of checking manually

kezhenxu94:

This patch uses `Map#computeIfAbsent` to compute a value for a key instead of checking and putting explicitly.

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
@@ -32,6 +32,7 @@ import java.io.BufferedReader;
 import java.io.InputStreamReader;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
```diff
@@ -521,11 +511,15 @@ private T createExtension(String name) {
             throw findException(name);
         }
         try {
+            EXTENSION_INSTANCES.computeIfAbsent(clazz, c -> {
+                try {
+                    return c.newInstance();
+                } catch (Throwable t) {
+                    throw new IllegalStateException("Extension instance(name: " + name + ", class: " +
```
khanimteyaz: can we use this instead of string concatination.
`throw new IllegalStateException(String.format("Extension instance(name: %s,class: %s could not be instantiated: %s",name,type,t.getMessage),t);`

kezhenxu94: I don't think it's not very worthwhile because we are using Java 8 features, and in Java 8 the string concatenation is optimized in compile time, isn't it?

khanimteyaz: After looking at you reply I refesh my understanding, string concatination has better performance then String.format. . Only one point where String.format could be good chouce, if we want to throws localize message from resource bundle. But I think in this case it does not apply, so what I am on with you on this. No need of change for this. 
Thanks for helping out to refresh my understanding 😄

dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
```diff
@@ -521,11 +513,14 @@ private T createExtension(String name) {
...
-            T instance = (T) EXTENSION_INSTANCES.get(clazz);
-            if (instance == null) {
-                EXTENSION_INSTANCES.putIfAbsent(clazz, clazz.newInstance());
-                instance = (T) EXTENSION_INSTANCES.get(clazz);
-            }
+            T instance = (T) EXTENSION_INSTANCES.computeIfAbsent(clazz, c -> {
+                try {
+                    return c.newInstance();
+                } catch (Throwable t) {
+                    throw new IllegalStateException("Extension instance(name: " + name + ", class: " +
```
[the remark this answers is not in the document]
kezhenxu94: It is in the lambda block so we have to catch the exception within the lambda

PR #3135, opened 2019-01-04
avoid dup refresh for registry config

beiwei30:

XXXXX

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
@@ -228,7 +228,6 @@ public abstract class AbstractInterfaceConfig extends AbstractMethodConfig {
      */
     protected List<URL> loadRegistries(boolean provider) {
         // check && override if necessary
-        checkRegistry();
         List<URL> registryList = new ArrayList<URL>();
         if (registries != null && !registries.isEmpty()) {
             for (RegistryConfig config : registries) {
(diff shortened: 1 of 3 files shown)


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
```diff
@@ -478,7 +477,7 @@ private void convertRegistryIdsToRegistries() {
             }
             Arrays.stream(arr).forEach(id -> {
                 if (registries.stream().noneMatch(reg -> reg.getId().equals(id))) {
-                    RegistryConfig registryConfig = new RegistryConfig();
+                    RegistryConfig registryConfig = new RegistryConfig(RegistryConfig.NO_AVAILABLE);
```
chickenlj: I don't think we can give RegistryConfig a default value `RegistryConfig.NO_AVAILABLE`, instead we should leave it empty. We will try to `refresh` this RegistryConfig later, if the user fails to set the right item for RegistryConfig, it will receive an Exception then when checking `isValid`. But if we give it this default value, it will always pass `isValid` and does register URL to Registry Center which can be unexpected behaviour.

dubbo-config/dubbo-config-api/src/test/java/org/apache/dubbo/config/AbstractInterfaceConfigTest.java
```diff
@@ -118,6 +118,8 @@ public void checkApplication2() throws Exception {
     public void testLoadRegistries() throws Exception {
         System.setProperty("dubbo.registry.address", "addr1");
         InterfaceConfig interfaceConfig = new InterfaceConfig();
+        // FIXME: now we need to check first, then load
+        interfaceConfig.checkRegistry();
```
chickenlj: Maybe we should give `loadRegistries` another name that can better reflect what it does. It is misleading since it loads nothing except for converts RegistryConfigs to URLs.

PR #3174, opened 2019-01-09
Fix timeout filter not work in async way.

carryxyh:
Fix timeout filter not work in async way.

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
@@ -40,8 +40,8 @@ public class TimeoutFilter implements Filter {
 
     @Override
     public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
-        long start = System.currentTimeMillis();
         if (invocation.getAttachments() != null) {
+            long start = System.currentTimeMillis();
             invocation.getAttachments().put(TIMEOUT_FILTER_START_TIME, String.valueOf(start));
         }
         return invoker.invoke(invocation);
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
@@ -36,21 +36,34 @@ public class TimeoutFilter implements Filter {
 
     private static final Logger logger = LoggerFactory.getLogger(TimeoutFilter.class);
 
+    private static final String TIMEOUT_FILTER_START_TIME = "timeout_filter_start_time";
+
     @Override
     public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
         long start = System.currentTimeMillis();
-        Result result = invoker.invoke(invocation);
-        long elapsed = System.currentTimeMillis() - start;
-        if (invoker.getUrl() != null
-                && elapsed > invoker.getUrl().getMethodParameter(invocation.getMethodName(),
-                "timeout", Integer.MAX_VALUE)) {
-            if (logger.isWarnEnabled()) {
-                logger.warn("invoke time out. method: " + invocation.getMethodName()
-                        + " arguments: " + Arrays.toString(invocation.getArguments()) + " , url is "
-                        + invoker.getUrl() + ", invoke elapsed " + elapsed + " ms.");
+        if (invocation.getAttachments() != null) {
+            invocation.getAttachments().put(TIMEOUT_FILTER_START_TIME, String.valueOf(start));
+        }
+        return invoker.invoke(invocation);
+    }
+
+    @Override
+    public Result onResponse(Result result, Invoker<?> invoker, Invocation invocation) {
+        String startAttach = invocation.getAttachment(TIMEOUT_FILTER_START_TIME);
+        if (startAttach != null) {
+            long elapsed = System.currentTimeMillis() - Long.valueOf(startAttach);
+            if (invoker.getUrl() != null
+                    && elapsed > invoker.getUrl().getMethodParameter(invocation.getMethodName(),
+                    "timeout", Integer.MAX_VALUE)) {
+                if (logger.isWarnEnabled()) {
+                    logger.warn("invoke time out. method: " + invocation.getMethodName()
+                            + " arguments: " + Arrays.toString(invocation.getArguments()) + " , url is "
+                            + invoker.getUrl() + ", invoke elapsed " + elapsed + " ms.");
+                }
             }
+            // we should not change the attachments of the invocation.
+            invocation.getAttachments().remove(TIMEOUT_FILTER_START_TIME);
         }
         return result;
     }
-
 }
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
@@ -25,6 +25,7 @@ import org.apache.dubbo.rpc.Invocation;
 import org.apache.dubbo.rpc.Invoker;
 import org.apache.dubbo.rpc.Result;
 import org.apache.dubbo.rpc.RpcException;
+import org.apache.dubbo.rpc.RpcInvocation;
 
 import java.util.Arrays;
 
@@ -43,6 +44,12 @@ public class TimeoutFilter implements Filter {
         if (invocation.getAttachments() != null) {
             long start = System.currentTimeMillis();
             invocation.getAttachments().put(TIMEOUT_FILTER_START_TIME, String.valueOf(start));
+        } else {
+            if (invocation instanceof RpcInvocation) {
+                RpcInvocation invc = (RpcInvocation) invocation;
+                long start = System.currentTimeMillis();
+                invc.setAttachment(TIMEOUT_FILTER_START_TIME, String.valueOf(start));
+            }
         }
         return invoker.invoke(invocation);
     }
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
```diff
@@ -36,21 +36,34 @@
...
-        Result result = invoker.invoke(invocation);
-        long elapsed = System.currentTimeMillis() - start;
-        if (invoker.getUrl() != null
-                && elapsed > invoker.getUrl().getMethodParameter(invocation.getMethodName(),
-                "timeout", Integer.MAX_VALUE)) {
-            if (logger.isWarnEnabled()) {
-                logger.warn("invoke time out. method: " + invocation.getMethodName()
-                        + " arguments: " + Arrays.toString(invocation.getArguments()) + " , url is "
-                        + invoker.getUrl() + ", invoke elapsed " + elapsed + " ms.");
+        if (invocation.getAttachments() != null) {
```
khanimteyaz: If there is no attachement for invocation, to handle fix timeout should we  care to create a attachment and set TIMEOUT_FILTER_START_TIME ? What do you say?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
```diff
@@ -36,21 +36,34 @@
...
+                    && elapsed > invoker.getUrl().getMethodParameter(invocation.getMethodName(),
+                    "timeout", Integer.MAX_VALUE)) {
+                if (logger.isWarnEnabled()) {
+                    logger.warn("invoke time out. method: " + invocation.getMethodName()
+                            + " arguments: " + Arrays.toString(invocation.getArguments()) + " , url is "
+                            + invoker.getUrl() + ", invoke elapsed " + elapsed + " ms.");
+                }
             }
+            // we should not change the attachments of the invocation.
+            invocation.getAttachments().remove(TIMEOUT_FILTER_START_TIME);
```
khanimteyaz: I am not sure should we care about the scenario here, where in case of exception while executing _invoker.invoke(invocation)_ , should we remove TIMEOUT_FILTER_START_TIME from attacchment ? I think as this piece of code should take of removing TIMEOUT_FILTER_START_TIME even in case of exception. What do you say?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
carryxyh: Seems like you are right...
Great suggestion, I will fix it.
👍

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
carryxyh: Agree.
I will fix it.

carryxyh: Hi, @khanimteyaz 
After check the code, seems like it would throw any exception..
So I keep this code.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
carryxyh: I have fix this place.
Seems like only RpcInvocation can set attachments.

khanimteyaz: I was think to have the responsibility of assigning a default TIMEOUT_FILTER_START_TIME  in RpcInvocation, but we will be landed up mixing its Responsibility with TimerFilter.  

I am just thinking loud here (please bear with me 😄 ), should we create a FilterWrapper and can store these information here, it may give us better way of separating object information. e.g

```
Class FilterWrapper {
  private Invocation target;
  private Map<String,Object> properties = new HashMap<>();
   FilterWrapper (Invocation invocation)  {
      this.target = invocation;
  }
  public void addProperties(String key,Object value) {
   properties.put(key,value);
} 
   public Map<String,Object> getProperties() {
     return this.properties;
 }
   public Map<String,Object> getAttachment() {
      this.target.getAttachment();
  }
  ..........
_more getter and setter or deleter as per needs_
  ....
}
wrapper=FilterWrapper.of(invocation); 
wrapper.addProperties(TIMEOUT_FILTER_START_TIME, String.valueOf(start));
```

So even in future we need to keep add internal properties we may not be landed up overriding user propvided values.
What do you say?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
khanimteyaz: > Hi, @khanimteyaz
> After check the code, seems like it would throw any exception..
> So I keep this code.

In this case would we be missing the chance to remove TIMEOUT_FILTER_START_TIME  from attachment?

carryxyh: Won't miss remove TIMEOUT_FILTER_START_TIME unless there is a bug.
But in bug case, we need fix our code but not add a try catch..I think.
What do u  think about it?
:)

khanimteyaz: If there are bug we any how need to fix it :). 

I think it is fine, as I was worring about whether we are missing to remove it or not, but on our known path we are removing.

I have created a issue to write a UT for specific this scenario and will take it 
https://github.com/apache/incubator-dubbo/issues/3234

khanimteyaz:
@carryxyh in this PR are planning to address anything more?

carryxyh:
Nope.
That is all. @khanimteyaz  Thx for your review. 👍

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
```diff
@@ -36,21 +37,40 @@
 
     private static final Logger logger = LoggerFactory.getLogger(TimeoutFilter.class);
 
+    private static final String TIMEOUT_FILTER_START_TIME = "timeout_filter_start_time";
+
     @Override
     public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
-        long start = System.currentTimeMillis();
```
Kiddinglife: `System.currentTimeMillis();` is not good choice for interval measurement.
http://sangsoonam.github.io/2017/03/01/do-not-use-curenttimemillis-for-time-interval.html
https://github.com/openshift/jenkins-plugin/issues/76
http://pzemtsov.github.io/2017/07/23/the-slow-currenttimemillis.html

carryxyh: I know.
I have read an article and it seems that it is better to start another thread to get time regularly? But I think this is not a bottleneck right now.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
```diff
@@ -36,21 +37,40 @@
...
-        if (invoker.getUrl() != null
-                && elapsed > invoker.getUrl().getMethodParameter(invocation.getMethodName(),
-                "timeout", Integer.MAX_VALUE)) {
-            if (logger.isWarnEnabled()) {
-                logger.warn("invoke time out. method: " + invocation.getMethodName()
-                        + " arguments: " + Arrays.toString(invocation.getArguments()) + " , url is "
-                        + invoker.getUrl() + ", invoke elapsed " + elapsed + " ms.");
+        if (invocation.getAttachments() != null) {
+            long start = System.currentTimeMillis();
+            invocation.getAttachments().put(TIMEOUT_FILTER_START_TIME, String.valueOf(start));
```
chickenlj: Be careful that this attachment `TIMEOUT_FILTER_START_TIME ` will be passed throughout the RPC chain because of the drawback of RpcContext. #3017

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
chickenlj: @khanimteyaz That's interesting, but if we add an extra `properties` just like you did in FilterWrapper, we need to further figure out how to pass the `properties` onto the wire. But at present `invocation` is the only carrier, so everything needs to finally go into `invocation` to get transferred.

chickenlj:
My only concern with this PR is the accident transfer of TIMEOUT_FILTER_START_TIME along the RPC chain.  Though the real root is a drawback of the RpcContext.
Since TimeoutFilter will be included in every RPC call, it's a big thing and sth should be careful with.

carryxyh:
I changed a way, it seems that our Invocation is unchanged, maybe we can solve our problem with a global map.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
chickenlj: Let's keep it easy, functionality then performance.

chickenlj:
Oh, I just noticed that this filter is only executed on provider side `@Activate(group = Constants.PROVIDER)`. Maybe the problem I am worrying is not happening.

chickenlj:
I just tested locally, TimeoutFilter is behind ContextFilter and these two filters will only be triggered on the provider side, so any changes to Invocation in ContextFilter will not have the chance to affect RpcContext. 
My concern above is not going to happen.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/TimeoutFilter.java
chickenlj: Now I think the previous way you give is better, please see my comments below. 
So what do you think now? Would you please revert the code?

carryxyh: I have checked the code.
As you said, the timeoutfilter does not affect the RpcContext, nor does it affect the invocation downstream of the call chain. I have reverted the code.

PR #3184, opened 2019-01-10
Async enhancement

chickenlj:

* User does not have to set `async=true` on Provider side to use provider async feature anymore, only `AsyncContext.startAsync()` is necessary.
* stop tag from being transferred in the RPC chain

## Brief changelog

* AsyncContext related classes
* ContextFilter

## Verifying this change

run `samples-async`

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncContextImpl.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncContextImpl.java
@@ -68,7 +68,9 @@ public class AsyncContextImpl implements AsyncContext {
 
     @Override
     public void start() {
-        this.started.set(true);
+        if (this.started.compareAndSet(false, true)) {
+            setInternalFuture(new CompletableFuture<>());
+        }
     }
 
     @Override
@@ -78,8 +80,11 @@ public class AsyncContextImpl implements AsyncContext {
         // Restore any other contexts in here if necessary.
     }
 
-    @Override
-    public CompletableFuture getInternalFuture() {
+    public void setInternalFuture(CompletableFuture<Object> future) {
+        this.future = future;
+    }
+
+    public CompletableFuture<Object> getInternalFuture() {
         return future;
     }
 }
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncContextImpl.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncContextImpl.java
@@ -34,10 +34,6 @@ public class AsyncContextImpl implements AsyncContext {
     private RpcContext storedServerContext;
 
     public AsyncContextImpl() {
-    }
-
-    public AsyncContextImpl(CompletableFuture<Object> future) {
-        this.future = future;
         this.storedContext = RpcContext.getContext();
         this.storedServerContext = RpcContext.getServerContext();
     }
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcContext.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcContext.java
@@ -731,12 +731,11 @@ public class RpcContext {
     @SuppressWarnings("unchecked")
     public static AsyncContext startAsync() throws IllegalStateException {
         RpcContext currentContext = getContext();
-        if (currentContext.asyncContext != null) {
-            currentContext.asyncContext.start();
-            return currentContext.asyncContext;
-        } else {
-            throw new IllegalStateException("This service does not support asynchronous operations, you should open async explicitly before use.");
+        if (currentContext.asyncContext == null) {
+            currentContext.asyncContext = new AsyncContextImpl();
         }
+        currentContext.asyncContext.start();
+        return currentContext.asyncContext;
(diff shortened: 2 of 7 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncContextImpl.java
```diff
@@ -34,10 +34,6 @@
     private RpcContext storedServerContext;
 
     public AsyncContextImpl() {
-    }
-
-    public AsyncContextImpl(CompletableFuture<Object> future) {
-        this.future = future;
         this.storedContext = RpcContext.getContext();
```
khanimteyaz: as RpcContext.getContext() and RpcContext.getServerContext() is are per thread specific and can be available to caller using based on thread, so should we store them as separate member variable of this class? What do you say?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcContext.java
```diff
@@ -731,12 +731,11 @@ public void asyncCall(Runnable runnable) {
     @SuppressWarnings("unchecked")
     public static AsyncContext startAsync() throws IllegalStateException {
         RpcContext currentContext = getContext();
-        if (currentContext.asyncContext != null) {
-            currentContext.asyncContext.start();
-            return currentContext.asyncContext;
-        } else {
-            throw new IllegalStateException("This service does not support asynchronous operations, you should open async explicitly before use.");
+        if (currentContext.asyncContext == null) {
+            currentContext.asyncContext = new AsyncContextImpl();
```
khanimteyaz: As here we are setting asyncContext  if it is null and then starting. Where we have a separate public method to set context asyncContext. In the scenario if we started the context after creating an instance of AsyncContextImpl. Down the line if caller modify sets the asyncContext by **setAsyncContext** method of this class, then we will not be having any track of lost instance and we will not be able to release resource or store the context. 

What do you say?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/proxy/AbstractProxyInvoker.java
```diff
@@ -86,7 +87,7 @@ public Result invoke(Invocation invocation) throws RpcException {
             if (RpcUtils.isReturnTypeFuture(invocation)) {
                 return new AsyncRpcResult((CompletableFuture<Object>) obj);
             } else if (rpcContext.isAsyncStarted()) { // ignore obj in case of RpcContext.startAsync()? always rely on user to write back.
-                return new AsyncRpcResult(rpcContext.getAsyncContext().getInternalFuture());
+                return new AsyncRpcResult(((AsyncContextImpl)(rpcContext.getAsyncContext())).getInternalFuture());
```
khanimteyaz: There could me different implementation of AsyncContext as well ? In this case we might get class cast exception. What do you say? Am I missing something here?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncContextImpl.java
beiwei30:
```java
    public AsyncContextImpl() {
        this.storedContext = RpcContext.getContext();
        this.storedServerContext = RpcContext.getServerContext();
    }
```

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcContext.java
beiwei30: We should not allow user to set asynccontext(), like async servlet, we should only allow user to call 'startAsync()' to get a async context. 

I think we should remove setAsyncContext from RPCContext.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/proxy/AbstractProxyInvoker.java
beiwei30: @khanimteyaz I think its fine. There's only one impl class and will only have one impl. The main purpose for having to cast here is we don't want to expose 'internal future' to the end user on the interface.

beiwei30:
one more thing, should we override **setAsync(boolean)** on ProviderConfig to warn user this only works on client side?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncContextImpl.java
```diff
@@ -78,8 +76,11 @@ public void signalContextSwitch() {
         // Restore any other contexts in here if necessary.
     }
 
-    @Override
-    public CompletableFuture getInternalFuture() {
+    public void setInternalFuture(CompletableFuture<Object> future) {
```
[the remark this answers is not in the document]
chickenlj: Agree, will use 'this.future = future'  directly.

chickenlj:
> one more thing, should we override setAsync(boolean) on ProviderConfig to warn user this only works on client side?

I will check if `setAsync(boolean)` is newly added in 2.7.0 or not. If it is newly added, I think we should remove or override it.

zhangyl:
why stop tag from being transferred in the RPC chain?

PR #3185, opened 2019-01-10
[Dubbo-3169]Check future status before get(), return default value if not completed yet.

chickenlj:

* Check future status before get(), return the default value if not completed yet.
* Add comment to announce AsyncRpcResult as unstable and should only be used internally.

## Brief changelog

* AsyncRpcResult

## Verifying this change

run `dubbo-samples-async`

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncRpcResult.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncRpcResult.java
@@ -25,6 +25,34 @@ import java.util.concurrent.CompletionException;
 import java.util.function.Function;
 
 /**
+ * <b>NOTICE!!</b>
+ *
+ * <p>
+ * You should never rely on this class directly when using or extending Dubbo, the implementation of {@link AsyncRpcResult}
+ * is only a workaround for compatibility purpose. It may be changed or even get removed from the next version.
+ * Please only use {@link Result} or {@link RpcResult}.
+ *
+ * Extending the {@link Filter} is one typical use case:
+ * <pre>
+ * {@code
+ * public class YourFilter implements Filter {
+ *     @Override
+ *     public Result onResponse(Result result, Invoker<?> invoker, Invocation invocation) {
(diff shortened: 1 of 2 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncRpcResult.java
```diff
@@ -121,10 +149,14 @@ public void setResultFuture(CompletableFuture<Result> resultFuture) {
     public Result getRpcResult() {
```
khanimteyaz: I think the whole method can be simplified by 

```
public Result getRpcResult() {
        try {
            if (resultFuture.isDone()) {
                return resultFuture.get();
            }
        } catch (Exception e) {
            // This should never happen;
            logger.error("Got exception when trying to fetch the underlying result of AsyncRpcResult.", e);
        }
        return new RpcResult();
    }
```

what do you say?

beiwei30: +1

chickenlj: Well done. I think now I should spend some time to learn how to write clean neat code  😂.

carryxyh:
@chickenlj 
I found that the pr on this issue may cause an additional problem with `CacheFilter`:

```
String key = StringUtils.toArgumentString(invocation.getArguments());
Object value = cache.get(key);
if (value != null) {
    if (value instanceof ValueWrapper) {
        return new RpcResult(((ValueWrapper)value).get());
    } else {
        return new RpcResult(value);
    }
}
// In an asynchronous scenario, result.getValue() will return an empty RpcResult
Result result = invoker.invoke(invocation);
if (!result.hasException()) {
    // since the rpcResult#hasException will return false, this empty result will be cached.
    cache.put(key, new ValueWrapper(result.getValue()));
}
return result; 
```

In this scenario, CacheFilter will cache an empty `RpcResult`, which is incorrect.

PR #3224, opened 2019-01-14
prefer jax-rs annotation on impl and compatible with old version

Jeff-Lv:
As Luo's advice,

The jax-rs annoations should be preferred to implemation class as the reasons below
1) compatible to old dubbo version
2) Dubbo REST provider can be used without using of Dubbo Conumser for REST (User may access the rest API from JS client directly).
3) Interface should be kept clean without the pullation of REST annoatation.


chickenlj:
LGTM

Issue #3231, opened 2019-01-15
In dubbo-2.7 tag route, dubbo 2.6 client invoke 2.7 server work well but dubbo 2.7 client invoke 2.6 donot work well


cvictory:
在tag route下， dubbo2.6作为client调用dubbo2.7 server，路由规则已经起效果； dubbo2.7作为client调用dubbo2.6路由规则没有生效。（在dubbo2.7，ops没有配置路由规则的前提下，只通过应用url的tag打标）。
发现问题代码, org.apache.dubbo.rpc.cluster.router.tag.TagRouter：
```
if (CollectionUtils.isEmpty(invokers)) {
            return invokers;
        }

        if (tagRouterRule == null || !tagRouterRule.isValid() || !tagRouterRule.isEnabled()) {
            return invokers;
        }
```
没在ops里配置路由规则的时候，需要继续往下走。

----transplate here 

Under tag route, dubbo2.6 calls dubbo2.7 server as client, routing rules have been effective; dubbo2.7 as client calls dubbo2.6 routing rules have not taken effect. (In dubbo2.7, ops is not configured with routing rules, only by applying the url tag).
Found the problem code, org.apache.dubbo.rpc.cluster.router.tag.TagRouter:
```
if (CollectionUtils.isEmpty(invokers)) {
            return invokers;
        }

        if (tagRouterRule == null || !tagRouterRule.isValid() || !tagRouterRule.isEnabled()) {
            return invokers;
        }
```
When you do not configure routing rules in ops, you need to continue down.

cvictory:
In dubbo 2.6 , there are some key points in this class com.alibaba.dubbo.rpc.cluster.directory.AbstractDirectory.
 ```
    protected void setRouters(List<Router> routers) {
        // copy list
        routers = routers == null ? new ArrayList<Router>() : new ArrayList<Router>(routers);
        // append url router
        String routerkey = url.getParameter(Constants.ROUTER_KEY);
        if (routerkey != null && routerkey.length() > 0) {
            RouterFactory routerFactory = ExtensionLoader.getExtensionLoader(RouterFactory.class).getExtension(routerkey);
            routers.add(routerFactory.getRouter(url));
        }
        // append mock invoker selector
        routers.add(new MockInvokersSelector());
        Collections.sort(routers);
        this.routers = routers;
    }
```

so should we keep this logic code ?

chickenlj:
> ```java
[...]

PR #3065, opened 2018-12-26
[Dubbo - 2.6.x TagRouter Feature] Import old TagRouter into 2.6.x

lexburner:
## What is the purpose of the change

Import old TagRouter into 2.6.x

This feature is merged at 2.7.x

PR #3233, opened 2019-01-15
keep TagRouter consistent with 2.6.x

chickenlj:

interoperability with TagRouter in 2.6.x

## Brief changelog

add one new method in TagRouter

## Verifying this change

run Tag samples

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java
@@ -34,6 +34,7 @@ import org.apache.dubbo.rpc.cluster.router.AbstractRouter;
 import org.apache.dubbo.rpc.cluster.router.tag.model.TagRouterRule;
 import org.apache.dubbo.rpc.cluster.router.tag.model.TagRuleParser;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -88,7 +89,7 @@ public class TagRouter extends AbstractRouter implements Comparable<Router>, Con
         }
 
         if (tagRouterRule == null || !tagRouterRule.isValid() || !tagRouterRule.isEnabled()) {
-            return invokers;
+            return filterUsingStaticTag(invokers, url, invocation);
         }
 
         List<Invoker<T>> result = invokers;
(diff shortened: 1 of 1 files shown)


chickenlj:
https://github.com/apache/incubator-dubbo/issues/3231

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java
```diff
@@ -141,6 +141,37 @@ public URL getUrl() {
...
+     */
+    private <T> List<Invoker<T>> filterUsingStaticTag(List<Invoker<T>> invokers, URL url, Invocation invocation) {
+        List<Invoker<T>> result = invokers;
+        // Dynamic param
+        String tag = StringUtils.isEmpty(invocation.getAttachment(TAG_KEY)) ? url.getParameter(TAG_KEY) :
+                invocation.getAttachment(TAG_KEY);
+        // Tag request
+        if (!StringUtils.isEmpty(tag)) {
+            result = filterInvoker(invokers, invoker -> tag.equals(invoker.getUrl().getParameter(Constants.TAG_KEY)));
+            if (CollectionUtils.isEmpty(result) && !isForceUseTag(invocation)) {
```
khanimteyaz: Is this 
```
if (CollectionUtils.isEmpty(result) && !isForceUseTag(invocation)) {
                result = filterInvoker(invokers, invoker -> StringUtils.isEmpty(invoker.getUrl().getParameter(Constants.TAG_KEY)));
            }
```
should not be irrespective of whether **tag** present or not? I was just referring to 2.6.x implementaion where it seems to me this logic is outside of whether tag is present or not.

khanimteyaz:
Other than above doubt PR looks good to me.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java
chickenlj: Yes, I think you are right about the implementation in 2.6.x. 
But the difference here is 2.6.x does not have a deliberately defined `force` property, we can also think as it has a `force` key which is always `false`. So in 2.6.x, it will always go to the following branch at 
https://github.com/apache/incubator-dubbo/pull/3065/files#diff-179f8c2f0653380f5fd6240f6ce165b7R71
```java
f (result.isEmpty()) {
            // Only forceTag = true force match, otherwise downgrade
            String forceTag = RpcContext.getContext().getAttachment(Constants.FORCE_USE_TAG);
            if (StringUtils.isEmpty(forceTag) || "false".equals(forceTag)) {
                for (Invoker<T> invoker : invokers) {
                    if (StringUtils.isEmpty(invoker.getUrl().getParameter(Constants.TAG_KEY))) {
                        result.add(invoker);
                    }
                }
            }
        }
```
irrespective of whether tag present or not.

For 2.7, you can see that we have covered the branch in which tag is not present. 
```java
else {
            result = filterInvoker(invokers, invoker -> StringUtils.isEmpty(invoker.getUrl().getParameter(Constants.TAG_KEY)));
        }
```
But when the tag is empty, I think there's no need to check force, `force=true` only makes sense to the scenario when the tag is not empty.

khanimteyaz: Thanks for clearing my doubts. 👍

PR #3239, opened 2019-01-15
[Enhancement] Use ThreadLocalRandom and try-with-resource

satansk:

Small improvments, including:

* use `ThreadLocalRandom` instead of `Random`
* use try-with-resource to avoid explicit resource management
* fix duplicate code

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
@@ -117,7 +117,7 @@ public class NetUtils {
                 new InetSocketAddress(port) : new InetSocketAddress(host, port);
     }
 
-    static boolean isValidAddress(InetAddress address) {
+    static boolean isValidV4Address(InetAddress address) {
         if (address == null || address.isLoopbackAddress()) {
             return false;
         }
@@ -223,7 +223,7 @@ public class NetUtils {
                 return Optional.ofNullable(normalizeV6Address(v6Address));
             }
         }
-        if (isValidAddress(address)) {
+        if (isValidV4Address(address)) {
             return Optional.of(address);
         }
         return Optional.empty();
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
@@ -30,47 +30,41 @@ import java.net.ServerSocket;
 import java.net.UnknownHostException;
 import java.util.Enumeration;
 import java.util.Map;
-import java.util.Random;
+import java.util.Optional;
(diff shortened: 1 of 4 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
```diff
@@ -231,21 +216,31 @@ public static InetAddress getLocalAddress() {
...
     }
 
+    private static Optional<InetAddress> toValidAddress(InetAddress address) {
+        if (address instanceof Inet6Address) {
+            Inet6Address v6Address = (Inet6Address) address;
+            if (isValidV6Address(v6Address)) {
+                return Optional.ofNullable(normalizeV6Address(v6Address));
+            }
+        }
+        if (isValidAddress(address)) {
```
khanimteyaz: Better to refactor the method name isValidAddress to **isValidV4Address**. What do you say?

dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
```diff
@@ -231,21 +216,31 @@ public static InetAddress getLocalAddress() {
         return localAddress;
     }
 
+    private static Optional<InetAddress> toValidAddress(InetAddress address) {
```
satansk: `Optional` requires some costs regarding preformance, I'm not sure we should encourage it or not in Dubbo.

dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
```diff
@@ -30,47 +30,41 @@
...
-
     private static final Logger logger = LoggerFactory.getLogger(NetUtils.class);
-    private static final int RND_PORT_START = 30000;
 
+    // returned port range is [30000, 39999]
+    private static final int RND_PORT_START = 30000;
     private static final int RND_PORT_RANGE = 10000;
 
-    private static final Random RANDOM = new Random(System.currentTimeMillis());
+    // valid port range is (0, 65535]
```
CrazyHZM: The annotation of the attribute is best used
`/**
 * 
 */`

dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
carryxyh: This is yet to be decided. Personally think that you can use the judgment of null. As we have seen, most frameworks internally support 1.8, but rarely use optinal.

satansk: As we discussed in mail list, the community encourage us to use `Optional` :)

beiwei30:
LGTM too.

PR #3264, opened 2019-01-17
Mostly modification to satisfy more complexing support scenrio such as service Level model

Jeff-Lv:
- add a attributeMap to Invocation
-  rename the varibale name NAME_ID_MAP
-  add a method for class CodecSupport
- import servicemetada and associated it with models class

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/Invocation.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/Invocation.java
@@ -16,7 +16,6 @@
  */
 package org.apache.dubbo.rpc;
 
-import java.util.HashMap;
 import java.util.Map;
 
 /**
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/Invocation.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/Invocation.java
@@ -16,6 +16,7 @@
  */
 package org.apache.dubbo.rpc;
 
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -83,4 +84,7 @@ public interface Invocation {
      */
     Invoker<?> getInvoker();
 
+    Object put(Object key, Object value);
+
+    Object get(Object key);
 }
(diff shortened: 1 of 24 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/Invocation.java
```diff
@@ -83,4 +83,9 @@
      */
     Invoker<?> getInvoker();
 
+    Object put(Object key, Object value);
```
khanimteyaz: should we these interface level method documentation? what do you say?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/model/ConsumerMethodModel.java
```diff
@@ -53,6 +56,10 @@ public Method getMethod() {
         return method;
     }
 
+    public ConcurrentMap<String, Object> getAttributeMap() {
```
khanimteyaz: Returning the same copy of attributeMap may lead to programatic mistake as many different java file may modify it, we may not have the track of who is modifying what what modifying. 

Should we not have a put and get method of attribute map it self in this class. What do you say?

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/model/ProviderMethodModel.java
```diff
@@ -17,17 +17,18 @@
...
+import java.util.concurrent.ConcurrentMap;
 
 public class ProviderMethodModel {
-    private transient final Method method;
+    private final Method method;
     private final String methodName;
     private final String[] methodArgTypes;
-    private final String serviceName;
 
+    private final ConcurrentMap<String, Object> attributeMap = new ConcurrentHashMap<>();
```
khanimteyaz: `private final Map<String, Object> attributeMap = new ConcurrentHashMap<>();` would be better.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/model/ProviderMethodModel.java
```diff
@@ -44,8 +45,8 @@ public String getMethodName() {
         return methodArgTypes;
     }
 
-    public String getServiceName() {
-        return serviceName;
+    public ConcurrentMap<String, Object> getAttributeMap() {
+        return attributeMap;
```
khanimteyaz: return the mutable member variable ** attributeMap** may be harmful as any caller can have the reference and modify it. may be we can provide put and get method for attributeMap in this class it self.

PR #3269, opened 2019-01-18
sharedExecutor use lazy-init

scxwhite:

sharedExecutor use lazy-init

## Brief changelog

Shared Executor is not used in most cases. I think we should use lazy initialization instead of hungry initialization.

## Verifying this change

XXXXX


carryxyh:
It looks like trivial, an object does not take up much memory. There is no need to complicate our code for this.

scxwhite:
> It looks like trivial, an object does not take up much memory. There is no need to complicate our code for this.



> It looks like trivial, an object does not take up much memory. There is no need to complicate our code for this.

I don't agree with you. Many users are using dubbo, so you can't ignore an object because you think it takes up a small amount of memory. We should be responsible for these users.

scxwhite:
> I think this approach complicates the code, in which case we should choose memory and code complexity. Obviously, the increase of an object is not enough for us to add more judgments and locks. Of course, it does not seem that we are not responsible for our code.

ok.Learning.

PR #3274, opened 2019-01-18
Acesslog dateformat enhancemnet

khanimteyaz:

Previous PR I have closed as it was too old and due to recent major StringUtils changes 500+ files were showing as modified files and around 8000+ line of code changes, which will may create some confusion. To avoid this confusion have raised new PR.

In previous PR there was request change was asked by @zonghaishang , which (not thread safe)I beleive I have adrressed. 

In this PR 

1.  I have moved the access log creation and writing to file into a single thread and instead of each time creating a SimpleDateFormat using FastDateFormat formatter with new created DateUtil.
2. Refactored code to separate our and group related task in separate method and have enhance the readability by using 
         - Method renaming
         - Reducing big methods to small.
         - DateUtil utility using FastDateFormat

This PR is a resurrection of old PR 
   1. 3027 (https://github.com/apache/incubator-dubbo/pull/3027) PR
   2.  3080 (https://github.com/apache/incubator-dubbo/pull/3080)
   3.https://github.com/apache/incubator-dubbo/pull/3090
Sorry for the inconvenience 
## Brief changelog

XXXXX

## Verifying this change


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -68,141 +67,134 @@
...
         }
     }
 
-}
+    private void processWithAccessKeyLogger(Set<AccessLogData> logSet, File file) throws IOException {
+        FileWriter writer = new FileWriter(file, true);
+        try {
+            for (Iterator<AccessLogData> iterator = logSet.iterator();
+                 iterator.hasNext();
+                 iterator.remove()) {
```
satansk: need formation here.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -68,141 +67,134 @@
...
+                writer.write(iterator.next().getLogMessage());
+                writer.write("\r\n");
+            }
+            writer.flush();
+        } finally {
+            writer.close();
```
satansk: `FileWriter` can be used in try-with-resource, which is better.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: Here we have list of access log, which we are writting and then removing entry.

If you would provide more specific question then may be I can answer those.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -68,141 +67,134 @@
...
 
     private static final long LOG_OUTPUT_INTERVAL = 5000;
 
-    private final ConcurrentMap<String, Set<String>> logQueue = new ConcurrentHashMap<String, Set<String>>();
-
-    private final ScheduledExecutorService logScheduled = Executors.newScheduledThreadPool(2, new NamedThreadFactory("Dubbo-Access-Log", true));
+    private static final String FILE_DATE_FORMAT = "yyyyMMdd";
 
-    private volatile ScheduledFuture<?> logFuture = null;
+    private static final ConcurrentMap<String, Set<AccessLogData>> logQueue = new ConcurrentHashMap<String, Set<AccessLogData>>();
```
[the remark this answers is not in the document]
khanimteyaz: > We can use type inference which is better than explicit parameters.

Agree.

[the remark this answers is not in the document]
khanimteyaz: > And one more thing, why call it `logQueue`, but it's not a queue at all?

Yup. Changes to **logEntries**

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
khanimteyaz: done.

dubbo-common/src/main/java/org/apache/dubbo/common/utils/DateUtil.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     *  This method used to return a formatted string of a given date object.
+     * @param date Input data object
+     * @param format format of data.
+     * @return
+     */
+    public static String format(Date date,String format) {
+        Assert.notNull(date,"Given date can't be null");
+        Assert.notEmptyString(format,"Given date format can't be null or empty");
```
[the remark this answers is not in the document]
khanimteyaz: Done.

dubbo-common/pom.xml
```diff
@@ -39,6 +39,10 @@
             <groupId>commons-logging</groupId>
             <artifactId>commons-logging</artifactId>
         </dependency>
+        <dependency>
```
beiwei30: Let's avoid of introducing new dependency into the distribution. In this case, it looks like FastDateFormat doesn't help us much, since we are using the same instance in single thread to format all.

If I am wrong, try to include FastDataFormat's source code in Dubbo project, and update LICENSE file.

khanimteyaz: As FasDataFormat has better performance then other as a sample performance check was done by @lixiaojiee  [here](https://github.com/apache/incubator-dubbo/pull/3090#issuecomment-453394557). 

As you suggested we can add the source code of FastDataFormat in dubbo project it self, would you help me by guiding what is the approach of doing it.  Is that we can copy code of FastDataFormat and adding a license file? Any help would be appreciated.

carryxyh: @khanimteyaz 
I think u should copy the code of the class. I saw that the common-langs use apache lisence, so don't forget copy the license header (pls use FastDateFormtat's license header not dubbo's).

After copy the code, u should modify dubbo-parent file (pom file). U should add a apache-rat-plugin's exclude line like `<exclude>**/org/apache/dubbo/common/threadlocal/InternalThreadLocalMap.java</exclude>`. 

Finally, u should modify the LICENSE file. In the tail of the file, it can also find some examples.

khanimteyaz: @carryxyh and @beiwei30 thanks for guidance. I have made the requires changes. Could you have a look.

khanimteyaz:
@carryxyh @beiwei30 I have made the changes, could you plz have a look.

carryxyh:
Hi, Seems like the source code of FastDateFormat is very heavy...

There are many places in the source code that currently use SimpleDateFormat. If we introduce the source code, we should replace all the SimpleDateFormat and share the DateFormat.

@khanimteyaz 
Would you like to make an issue and optimize them? :)

beiwei30:
> Hi, Seems like the source code of FastDateFormat is very heavy...
> 
> There are many places in the source code that currently use SimpleDateFormat. If we introduce the source code, we should replace all the SimpleDateFormat and share the DateFormat.
> 
> @khanimteyaz
> Would you like to make an issue and optimize them? :)

Agree with @carryxyh, @khanimteyaz we should not rely on FastDateFormat. Would you mind to change it to rely on JDK's DateFormat?

khanimteyaz:
> > Hi, Seems like the source code of FastDateFormat is very heavy...
> > There are many places in the source code that currently use SimpleDateFormat. If we introduce the source code, we should replace all the SimpleDateFormat and share the DateFormat.
> > @khanimteyaz
> > Would you like to make an issue and optimize them? :)
> 
> Agree with @carryxyh, @khanimteyaz we should not rely on FastDateFormat. Would you mind to change it to rely on JDK's DateFormat?

@beiwei30 , you want me to revert the change and use DateFormat or we should us **FastDateFormat** and replace all existing **SimpleDateFormat** to use FastDateFormat?

khanimteyaz:
@beiwei30 could you provide our thought on this?

tswstarplanet:
We can consider using new api in  java.time.* packages to replace old java time apis.
The new apis provider immutable classes about time, so they are thread safe.

satansk:
> We can consider using new api in java.time.* packages to replace old java time apis.
> The new apis provider immutable classes about time, so they are thread safe.

+1, Yes, the new time API is much better.

beiwei30:
> > > Hi, Seems like the source code of FastDateFormat is very heavy...
> > > There are many places in the source code that currently use SimpleDateFormat. If we introduce the source code, we should replace all the SimpleDateFormat and share the DateFormat.
> > > @khanimteyaz
> > > Would you like to make an issue and optimize them? :)
> > 
> > 
> > Agree with @carryxyh, @khanimteyaz we should not rely on FastDateFormat. Would you mind to change it to rely on JDK's DateFormat?
> 
> @beiwei30 , you want me to revert the change and use DateFormat or we should us **FastDateFormat** and replace all existing **SimpleDateFormat** to use FastDateFormat using DateUtil? Sorry for being confuse here 😃

correct.

beiwei30:
@khanimteyaz I submitted a pull request https://github.com/khanimteyaz/incubator-dubbo/pull/11 in your repository. Would you mind to take a look?

My basic idea is: since we are now formatting date and time in single thread, it is safe to use a singleton formatter without any risk.

khanimteyaz:
@beiwei30  I have merged ur PR in branch to this branch. Could you please look into this. Thanks for your support.

khanimteyaz:
@satansk could you review, as you requested for changes.

PR #3276, opened 2019-01-19
Improve/heartbeat

lexburner:

remove the heartbeat on the server side

## Brief changelog

- Change the direction of the heartbeat request to a one-way request: client -> server.
- Added a timer to the server side to close the connection when the connection is idle.


lexburner:
@kexianjun In my opinion, idle check timer is responsible to a `Connection`, not a `Service`.

kexianjun:
> @kexianjun In my opinion, idle check timer is responsible to a `Connection`, not a `Service`.

目前的处理方式，一个HashedWheelTimer最多只处理两个任务(server端是一个，clien端是两个)，每个HashedWheelTimer都会启动一个线程检测，如果一个服务作为消费方调用服务比较多的情况下，就需要起很多线程做心跳检查，比如说我们的一个业务系统，调用其他不同系统的服务多达上十个，再加上线上集群部署，单是心跳检测线程就要起二三十个了。我觉得检测的任务和连接是对应的，定时器是可以共享的。

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/CloseTimerTask.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    protected void doTask(Channel channel) {
+        try {
+            Long lastRead = lastRead(channel);
+            Long lastWrite = lastWrite(channel);
+            Long now = now();
+            // check ping & pong at server
+            if ((lastRead != null && now - lastRead > idleTimeout)
```
khanimteyaz: I could be wrong here, but just wanted to understand and would love to gets wrong myself 😄 .  In the scenario where channel has received data but have not perform any write or vice versa should we consider as timeout  (as we have OR condition based check).

khanimteyaz: In **HeaderExchangeHandler.java** disconnect method we are setting KEY_READ_TIMESTAMP and KEY_WRITE_TIMESTAMP with the value of System.currentTimeMillis() , will these impact this detection here? Should we also consider whether a given channel has already been closed or not? What do you say?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
```diff
@@ -39,11 +39,10 @@
...
-    // heartbeat(ms), default value is 0 , won't execute a heartbeat.
     private int heartbeat;
-    private int heartbeatTimeout;
+    private int idleTimeout;
 
-    private HashedWheelTimer heartbeatTimer;
+    private HashedWheelTimer idleCheckTimer;
 
     public HeaderExchangeClient(Client client, boolean needHeartbeat) {
         if (client == null) {
```
khanimteyaz: Can be replace with 
`Assert.notNull(client, "Client can't be null");`

as already dubbo already have a util for this.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
```diff
@@ -55,16 +54,16 @@ public HeaderExchangeClient(Client client, boolean needHeartbeat) {
 
         this.heartbeat = client.getUrl().getParameter(Constants.HEARTBEAT_KEY, dubbo != null &&
                 dubbo.startsWith("1.0.") ? Constants.DEFAULT_HEARTBEAT : 0);
```
khanimteyaz: `private final int NO_HEARTBEAT = 0;`
can we replace **0** in case if **dubbo** does not start with **1.0** with NO_HEARTBEAT for better readability.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeServer.java
```diff
@@ -48,25 +48,24 @@
...
     private int heartbeat;
-    private int heartbeatTimeout;
+    private int idleTimeout;
     private AtomicBoolean closed = new AtomicBoolean(false);
 
-    private HashedWheelTimer heartbeatTimer;
+    private HashedWheelTimer idleCheckTimer;
 
     public HeaderExchangeServer(Server server) {
         if (server == null) {
```
khanimteyaz: can be replace with 

Can be replace with
Assert.notNull(server, "Server can't be null");

as already dubbo already have a util for this.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeServer.java
```diff
@@ -210,14 +209,14 @@ public void reset(URL url) {
                 int h = url.getParameter(Constants.HEARTBEAT_KEY, heartbeat);
```
khanimteyaz: **h** can be rename to **urlHeartBeatKey** what do you say?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeServer.java
```diff
@@ -210,14 +209,14 @@ public void reset(URL url) {
                 int h = url.getParameter(Constants.HEARTBEAT_KEY, heartbeat);
                 int t = url.getParameter(Constants.HEARTBEAT_TIMEOUT_KEY, h * 3);
```
khanimteyaz: can be renamed to **urlTimeOutKey**. What do you say?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
lexburner: > > @kexianjun In my opinion, idle check timer is responsible to a `Connection`, not a `Service`.
> 
> 目前的处理方式，一个HashedWheelTimer最多只处理两个任务(server端是一个，clien端是两个)，每个HashedWheelTimer都会启动一个线程检测，如果一个服务作为消费方调用服务比较多的情况下，就需要起很多线程做心跳检查，比如说我们的一个业务系统，调用其他不同系统的服务多达上十个，再加上线上集群部署，单是心跳检测线程就要起二三十个了。我觉得检测的任务和连接是对应的，定时器是可以共享的。

Got it. I will make this change.

lexburner:
> > @kexianjun In my opinion, idle check timer is responsible to a `Connection`, not a `Service`.
> 
> 目前的处理方式，一个HashedWheelTimer最多只处理两个任务(server端是一个，clien端是两个)，每个HashedWheelTimer都会启动一个线程检测，如果一个服务作为消费方调用服务比较多的情况下，就需要起很多线程做心跳检查，比如说我们的一个业务系统，调用其他不同系统的服务多达上十个，再加上线上集群部署，单是心跳检测线程就要起二三十个了。我觉得检测的任务和连接是对应的，定时器是可以共享的。

Got it. I will make this change.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/CloseTimerTask.java
lexburner: invocation can be divided into two kinds: normal, heartbeat. Both of them should update the  KEY_READ_TIMESTAMP and KEY_WRITE_TIMESTAMP. This is a deliberate design, which means IDLE CHECK.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
```diff
@@ -178,28 +174,20 @@ public boolean hasAttribute(String key) {
...
         HeartbeatTimerTask heartBeatTimerTask = new HeartbeatTimerTask(cp, heartbeatTick, heartbeat);
-        ReconnectTimerTask reconnectTimerTask = new ReconnectTimerTask(cp, heartbeatTimeoutTick, heartbeatTimeout);
+        ReconnectTimerTask reconnectTimerTask = new ReconnectTimerTask(cp, heartbeatTimeoutTick, idleTimeout);
 
         // init task and start timer.
-        heartbeatTimer.newTimeout(heartBeatTimerTask, heartbeatTick, TimeUnit.MILLISECONDS);
-        heartbeatTimer.newTimeout(reconnectTimerTask, heartbeatTimeoutTick, TimeUnit.MILLISECONDS);
-    }
-
-    private void stopHeartbeatTimer() {
```
beiwei30: why can we remove `stopHeartbeatTimer()`?

lexburner: The timer to send heartbeat was an instance of every `Connection` before this change, I have changed the scope of `Timer` to `static`, so we could not close the heartbeat timer by non-static method.

beiwei30: i see.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
```diff
@@ -39,32 +40,27 @@
 
     private final Client client;
     private final ExchangeChannel channel;
-    // heartbeat(ms), default value is 0 , won't execute a heartbeat.
     private int heartbeat;
-    private int heartbeatTimeout;
+    private int idleTimeout;
 
-    private HashedWheelTimer heartbeatTimer;
+    private static HashedWheelTimer idleCheckTimer = new HashedWheelTimer(new NamedThreadFactory("dubbo-client-idleCheck", true), 1,
```
carryxyh: I think we should establish a mapping relationship between client and task.

Since we are using the global Timer at the moment, when any client is closed, we should cancel the task corresponding to this client. In this way, we can guarantee that our tasks will not be too much.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeServer.java
lexburner: These variable names are named by other people. In my opinion, stay the same is also OK.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
lexburner: The magic number is indeed troublesome, but it isn't related to this pull request. Let us change them in other pr.

PR #3293, opened 2019-01-21
Fix typo in 2.7.0 CHANGES file

htynkn:
Change from `Dubbo` to `dubbo`


ralf0131:
Since it is trivial and 2.7.0 RC2 has been tagged, it will be fixed in the next release.

PR #3299, opened 2019-01-22
Optimize heartbeat

carryxyh:
We should cancel the timeout when the client or server is close to guarantee that our tasks will not be too much.

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
@@ -19,7 +19,9 @@ package org.apache.dubbo.remoting.exchange.support.header;
 import org.apache.dubbo.common.Constants;
 import org.apache.dubbo.common.URL;
 import org.apache.dubbo.common.timer.HashedWheelTimer;
+import org.apache.dubbo.common.timer.Timeout;
 import org.apache.dubbo.common.utils.Assert;
+import org.apache.dubbo.common.utils.CollectionUtils;
 import org.apache.dubbo.common.utils.NamedThreadFactory;
 import org.apache.dubbo.remoting.ChannelHandler;
 import org.apache.dubbo.remoting.Client;
@@ -30,7 +32,11 @@ import org.apache.dubbo.remoting.exchange.ExchangeHandler;
 import org.apache.dubbo.remoting.exchange.ResponseFuture;
 
 import java.net.InetSocketAddress;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -43,8 +49,10 @@ public class HeaderExchangeClient implements ExchangeClient {
     private int heartbeat;
     private int idleTimeout;
 
-    private static HashedWheelTimer idleCheckTimer = new HashedWheelTimer(new NamedThreadFactory("dubbo-client-idleCheck", true), 1,
+    private static final HashedWheelTimer IDLE_CHECK_TIMER = new HashedWheelTimer(new NamedThreadFactory("dubbo-client-idleCheck", true), 1,
             TimeUnit.SECONDS, Constants.TICKS_PER_WHEEL);
+
+    private static final Map<ExchangeClient, List<Timeout>> CLIENT_TASKS = new ConcurrentHashMap<>();
 
     public HeaderExchangeClient(Client client, boolean needHeartbeat) {
         Assert.notNull(client, "Client can't be null");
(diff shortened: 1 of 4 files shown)


carryxyh:
The task of the closed client will also remain in the timer, causing memory leaks.
U can check `AbstractTimerTask#run` for some info.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeServer.java
```diff
@@ -53,9 +57,11 @@
     private int idleTimeout;
     private AtomicBoolean closed = new AtomicBoolean(false);
 
-    private static HashedWheelTimer idleCheckTimer = new HashedWheelTimer(new NamedThreadFactory("dubbo-server-idleCheck", true), 1,
+    private static final HashedWheelTimer IDLE_CHECK_TIMER = new HashedWheelTimer(new NamedThreadFactory("dubbo-server-idleCheck", true), 1,
             TimeUnit.SECONDS, Constants.TICKS_PER_WHEEL);
 
+    private static final Map<ExchangeServer, List<Timeout>> SERVER_TASKS = new ConcurrentHashMap<>();
```
beiwei30: should we simply keep **Timeout** instance (instead of keeping it in a map)?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
```diff
@@ -43,9 +49,11 @@
     private int heartbeat;
     private int idleTimeout;
 
-    private static HashedWheelTimer idleCheckTimer = new HashedWheelTimer(new NamedThreadFactory("dubbo-client-idleCheck", true), 1,
+    private static final HashedWheelTimer IDLE_CHECK_TIMER = new HashedWheelTimer(new NamedThreadFactory("dubbo-client-idleCheck", true), 1,
             TimeUnit.SECONDS, Constants.TICKS_PER_WHEEL);
 
+    private static final Map<ExchangeClient, List<Timeout>> CLIENT_TASKS = new ConcurrentHashMap<>();
```
beiwei30: should we simply keep Timeout instance (instead of keeping it in a map)?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
```diff
@@ -43,9 +47,11 @@
     private int heartbeat;
     private int idleTimeout;
 
-    private static HashedWheelTimer idleCheckTimer = new HashedWheelTimer(new NamedThreadFactory("dubbo-client-idleCheck", true), 1,
+    private static final HashedWheelTimer IDLE_CHECK_TIMER = new HashedWheelTimer(new NamedThreadFactory("dubbo-client-idleCheck", true), 1,
             TimeUnit.SECONDS, Constants.TICKS_PER_WHEEL);
 
+    private List<Timeout> tasks = new ArrayList<>();
```
beiwei30: is an array list here a little bit over-kill?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeServer.java
```diff
@@ -53,9 +55,11 @@
     private int idleTimeout;
     private AtomicBoolean closed = new AtomicBoolean(false);
 
-    private static HashedWheelTimer idleCheckTimer = new HashedWheelTimer(new NamedThreadFactory("dubbo-server-idleCheck", true), 1,
+    private static final HashedWheelTimer IDLE_CHECK_TIMER = new HashedWheelTimer(new NamedThreadFactory("dubbo-server-idleCheck", true), 1,
             TimeUnit.SECONDS, Constants.TICKS_PER_WHEEL);
 
+    private List<Timeout> tasks = new ArrayList<>();
```
beiwei30: is an array list here a little bit over-kill?

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeClient.java
```diff
@@ -183,11 +189,20 @@ private void startIdleCheckTask() {
...
+        Timeout reconnectTimeout = IDLE_CHECK_TIMER.newTimeout(reconnectTimerTask, heartbeatTimeoutTick, TimeUnit.MILLISECONDS);
+
+        tasks.add(heartBeatTimeout);
+        tasks.add(reconnectTimeout);
     }
 
     private void doClose() {
+        if (CollectionUtils.isNotEmpty(tasks)) {
+            for (Timeout timeout : tasks) {
+                timeout.cancel();
```
kexianjun: timeout.cancel will never success when the task has expired once,because in this case it's status will be ST_EXPIRED , and in this case, based on org.apache.dubbo.common.timer.HashedWheelTimer.HashedWheelTimeout#cancel method code as following
```
public boolean cancel() {
            // only update the state it will be removed from HashedWheelBucket on next tick.
            if (!compareAndSetState(ST_INIT, ST_CANCELLED)) {
                return false;
            }
            // If a task should be canceled we put this to another queue which will be processed on each tick.
            // So this means that we will have a GC latency of max. 1 tick duration which is good enough. This way
            // we can make again use of our MpscLinkedQueue and so minimize the locking / overhead as much as possible.
            timer.cancelledTimeouts.add(this);
            return true;
        }
```
compareAndSetState will always reture false. Am I right?

carryxyh:
@kexianjun  @beiwei30 
my mistake. Pls check again...

Issue #3342, opened 2019-01-25
dubbo2.7.0 com.alibaba.com.caucho.hessian.io.HessianProtocolException: 'com.alibaba.dubbo.common.URL' could not be instantiated


qq925716471:
-  I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
-  I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.0
* Operating System version: Ubuntu18.04
* Java version: jdk8
 1. dubbo-ops with dubbo 2.7.0 and other appication use old version. 
 2. Copy MonitorService from old version to dubbo-ops.
 3. Other application send url call com.alibaba.dubbo.monitor.MonitorService#collect(URL statistics).

result:
 WARN dubbo.DecodeableRpcInvocation -  [DUBBO] Decode argument failed: 'com.alibaba.dubbo.common.URL' could not be instantiated, dubbo version: 2.7.0-SNAPSHOT, current host: 10.236.105.240
com.alibaba.com.caucho.hessian.io.HessianProtocolException: 'com.alibaba.dubbo.common.URL' could not be instantiated

[https://github.com/ebourg/hessian/blob/git-svn/src/com/caucho/hessian/io/JavaDeserializer.java](url)
`for (int i = 0; i < constructors.length; i++) {
      Class<?> []param = constructors[i].getParameterTypes();
      long cost = 0;

      for (int j = 0; j < param.length; j++) {
        cost = 4 * cost;

        if (Object.class.equals(param[j]))
          cost += 1;
        else if (String.class.equals(param[j]))
          cost += 2;
        else if (int.class.equals(param[j]))
          cost += 3;
        else if (long.class.equals(param[j]))
          cost += 4;
        else if (param[j].isPrimitive())
          cost += 5;
        else
          cost += 6;
      }

      if (cost < 0 || cost > (1 << 48))
        cost = 1 << 48;

      cost += (long) param.length << 48;

      if (cost < bestCost) {
        _constructor = constructors[i];
        bestCost = cost;
      }
    }
    if (_constructor != null) {
      _constructor.setAccessible(true);
      Class<?> []params = _constructor.getParameterTypes();
      _constructorArgs = new Object[params.length];
      for (int i = 0; i < params.length; i++) {
[...]

PR #3350, opened 2019-01-25
[Dubbo-3342] Fix NullPointerException when Hessian instantiate URL with JavaDeserializer #3342

qq925716471:

 #3342

## Verifying this change

### DIFF
--- a/dubbo-compatible/src/main/java/com/alibaba/dubbo/common/URL.java
+++ b/dubbo-compatible/src/main/java/com/alibaba/dubbo/common/URL.java
@@ -26,6 +26,10 @@ import java.util.Map;
 @Deprecated
 public class URL extends org.apache.dubbo.common.URL {
 
+    protected URL() {
+        super();
+    }
+
     public URL(org.apache.dubbo.common.URL url) {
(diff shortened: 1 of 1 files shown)


carryxyh:
Thx for your contribution.
I will deal this after ci passed.

dubbo-compatible/src/main/java/com/alibaba/dubbo/common/URL.java
```diff
@@ -26,6 +26,10 @@
 @Deprecated
 public class URL extends org.apache.dubbo.common.URL {
 
+    protected URL() {
```
carryxyh: seems like public is better.

chenlushun: @carryxyh  I don't think so. Because a good URL must have basic information, such as protocol, host, port.

carryxyh: emmm...
Maybe u are right. Seems like a protected constructor can fix this problem, too.

PR #3363, opened 2019-01-27
[Dubbo-900] Fix 通过 override 修改 hessian协议的提供者的配置 不生效 #900

CrazyHZM:
When modifying the configuration through override, you need to re-expose the newly modified service.
Compare the new url with the old url. If they are the same, they will be directly fetched from the collection. If they are different, re-create an export.

fix https://github.com/apache/incubator-dubbo/issues/900

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractProxyProtocol.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractProxyProtocol.java
@@ -28,6 +28,7 @@ import org.apache.dubbo.rpc.Result;
 import org.apache.dubbo.rpc.RpcException;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 /**
@@ -68,7 +69,7 @@ public abstract class AbstractProxyProtocol extends AbstractProtocol {
         if (exporter != null) {
             // When modifying the configuration through override, you need to re-expose the newly modified service.
             if ("hessian".equals(invoker.getUrl().getProtocol())) {
-                if (exporter.getInvoker().getUrl().equals(invoker.getUrl())) {
+                if (Objects.equals(exporter.getInvoker().getUrl(), invoker.getUrl())) {
                     return exporter;
                 }
             } else {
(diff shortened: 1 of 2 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractProxyProtocol.java
```diff
@@ -67,7 +67,11 @@ public void setProxyFactory(ProxyFactory proxyFactory) {
         Exporter<T> exporter = (Exporter<T>) exporterMap.get(uri);
         if (exporter != null) {
             // When modifying the configuration through override, you need to re-expose the newly modified service.
-            if (exporter.getInvoker().getUrl().equals(invoker.getUrl())) {
+            if ("hessian".equals(invoker.getUrl().getProtocol())) {
+                if (exporter.getInvoker().getUrl().equals(invoker.getUrl())) {
```
satansk: Could `getUrl` return null here? If so, `Objects.equals` might be better.

CrazyHZM: You are right, I will modify it.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractProxyProtocol.java
```diff
@@ -66,7 +67,14 @@ public void setProxyFactory(ProxyFactory proxyFactory) {
         final String uri = serviceKey(invoker.getUrl());
         Exporter<T> exporter = (Exporter<T>) exporterMap.get(uri);
         if (exporter != null) {
-            return exporter;
+            // When modifying the configuration through override, you need to re-expose the newly modified service.
+            if ("hessian".equals(invoker.getUrl().getProtocol())) {
```
beiwei30: We should not add special check for "hessian". It looks like we should compare URL for all proxy protocol when the cache is hit.

CrazyHZM:
> pls. check my comments, we should not add 'hessian' special check here.

I have modified it.

PR #3389, opened 2019-01-30
Enhancement/logger factory

kezhenxu94:

Polish/Enhance code to avoid many nested `try-catch` blocks

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/logger/LoggerFactory.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/logger/LoggerFactory.java
@@ -20,11 +20,14 @@ import org.apache.dubbo.common.extension.ExtensionLoader;
 import org.apache.dubbo.common.logger.jcl.JclLoggerAdapter;
 import org.apache.dubbo.common.logger.jdk.JdkLoggerAdapter;
 import org.apache.dubbo.common.logger.log4j.Log4jLoggerAdapter;
+import org.apache.dubbo.common.logger.log4j2.Log4j2Logger;
 import org.apache.dubbo.common.logger.log4j2.Log4j2LoggerAdapter;
 import org.apache.dubbo.common.logger.slf4j.Slf4jLoggerAdapter;
 import org.apache.dubbo.common.logger.support.FailsafeLogger;
 
 import java.io.File;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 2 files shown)


CrazyHZM:
Looks good to me.

dubbo-common/src/main/java/org/apache/dubbo/common/logger/LoggerFactory.java
```diff
@@ -34,40 +36,44 @@
...
-                        } catch (Throwable e4) {
-                            setLoggerAdapter(new JdkLoggerAdapter());
-                        }
+                        setLoggerAdapter(clazz.newInstance());
+                        break;
+                    } catch (Throwable ignored) {
                     }
                 }
-            }
+                break;
```
satansk: Seems no need to `break` in default ?

kezhenxu94: yes, just a personal habit, I've removed it

kezhenxu94:
@satansk unnecessary `break` in `switch` block has been removed and the ci tests failed on Java11, which seems unrelated to this PR, mind having a look?

satansk:
The CI don't work correctly sometimes, maybe we should provide some command like `rebuild` or `rerun` to deal with it, but for now, I think you can use an empty commit to trigger another CI :)

dubbo-common/src/main/java/org/apache/dubbo/common/logger/LoggerFactory.java
```diff
@@ -34,40 +36,43 @@
...
+                setLoggerAdapter(new JdkLoggerAdapter());
+                break;
+            case "log4j2":
+                setLoggerAdapter(new Log4j2LoggerAdapter());
+                break;
+            default:
+                List<Class<? extends LoggerAdapter>> candidates = Arrays.asList(
+                        Log4jLoggerAdapter.class,
+                        Slf4jLoggerAdapter.class,
+                        Log4j2LoggerAdapter.class,
```
cvictory: Is it need 3 other log class:  Log4j2LoggerAdapter, JclLoggerAdapter, JdkLoggerAdapter.

kezhenxu94: image
What do you mean, did I miss anything out?

dubbo-common/src/main/java/org/apache/dubbo/common/logger/LoggerFactory.java
```diff
@@ -103,12 +108,7 @@ public static void setLoggerAdapter(LoggerAdapter loggerAdapter) {
      * @return logger
      */
     public static Logger getLogger(Class<?> key) {
-        FailsafeLogger logger = LOGGERS.get(key.getName());
-        if (logger == null) {
-            LOGGERS.putIfAbsent(key.getName(), new FailsafeLogger(LOGGER_ADAPTER.getLogger(key)));
-            logger = LOGGERS.get(key.getName());
-        }
-        return logger;
+        return LOGGERS.computeIfAbsent(key.getName(), name -> new FailsafeLogger(LOGGER_ADAPTER.getLogger(key)));
```
ralf0131: I would suggest :

```java
return LOGGERS.computeIfAbsent(key.getName(), name -> new FailsafeLogger(LOGGER_ADAPTER.getLogger(name)));
```
for a better readability.
Note parameter for `getLogger` has been changed from `key` to `name`.

kezhenxu94: @ralf0131 If that's the case, why not just call the overloaded version `public static Logger getLogger(String key)`

ralf0131: When you are writing lambda x -> f(x) , `x` is just a placeholder, but in the function f(x) you treat  x as input parameters. 

Equivalently, in your case, it should be `name -> new FailsafeLogger(LOGGER_ADAPTER.getLogger(name))`, and `key` will be passed to the function at runtime.

You've implemented the same logic at Line 121.

When you are writing `name -> new FailsafeLogger(LOGGER_ADAPTER.getLogger(key))`, you are treating the function as zero parameter function, where `key` is a constant in this case.

kezhenxu94: I know what you meant, I'll update this patch to show what I meant

kezhenxu94: @ralf0131 Updated.

ralf0131: No, I didn't mean to call the overloaded method. Instead, I prefer your initial implementation. Call the overloaded method will introduce more overhead.

Your initial implementation is logically correct, what I mean is a minor change to improve the readability. You just need to replace `key` to `name`.

kezhenxu94: > You just need to replace key to name
That will also change the logic before this patch, it assumes that `org.apache.dubbo.common.logger.LoggerAdapter#getLogger(java.lang.Class<?>)` and `org.apache.dubbo.common.logger.LoggerAdapter#getLogger(java.lang.String)` return the same logger(which may be true in this case).

My thought was that `org.apache.dubbo.common.logger.LoggerFactory#getLogger(java.lang.Class<?>)` will get the logger via the key which is type of `Class` and, if I "replace key with name", the actual logger is got via name(which is type of `String`), anyway, it's the same in this situation, I'll update it soon

kezhenxu94: @ralf0131 updated

ralf0131:
Hi, it looks good to me!

I see the unit test for `org.apache.dubbo.common.logger.LoggerFactory#getLogger(java.lang.String)` is missing, would you mind add it as well?

kezhenxu94:
@ralf0131 added

PR #3398, opened 2019-01-30
move construction of ConfigChangeEvent outside the lambda expression

wanghbxxxx:

move construction of ConfigChangeEvent outside the lambda expression，
the same way as ZooKeeper configcenter do in CacheListener.java 

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-configcenter/dubbo-configcenter-apollo/src/main/java/org/apache/dubbo/configcenter/support/apollo/ApolloDynamicConfiguration.java
+++ b/dubbo-configcenter/dubbo-configcenter-apollo/src/main/java/org/apache/dubbo/configcenter/support/apollo/ApolloDynamicConfiguration.java
@@ -176,11 +176,8 @@ public class ApolloDynamicConfiguration implements DynamicConfiguration {
                     return;
                 }
 
-                listeners.forEach(listener -> {
-                            ConfigChangeEvent event = new ConfigChangeEvent(key, change.getNewValue(), getChangeType(change));
-                            listener.process(event);
-                        }
-                );
+                ConfigChangeEvent event = new ConfigChangeEvent(key, change.getNewValue(), getChangeType(change));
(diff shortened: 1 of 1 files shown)


carryxyh:
LGTM
Thx for your contribution.

carryxyh:
Tips: Objects used in inner classes are generally marked as `final`. 
This matter has nothing to do with logic right or wrong, but it is related to code specifications.

carryxyh:
Another thing, can we make the field of `ConfigChangeEvent` `final`?
It seems that if we want to share the `ConfigChangeEvent` instance, it is a better choice to make the `ConfigChangeEvent` field `final`.

wanghbxxxx:
or make the ConfigChangeEvent instance immutable?

carryxyh:
> or make the ConfigChangeEvent instance immutable?

This is not safe. If a listener modifies a property of `ConfigChangeEvent`, since the `ConfigChangeEvent` is shared, then the other listeners may not get the original data. This can lead to some unsafe factors.

wanghbxxxx:
> > or make the ConfigChangeEvent instance immutable?
> 
> This is not safe. If a listener modifies a property of `ConfigChangeEvent`, since the `ConfigChangeEvent` is shared, then the other listeners may not get the original data. This can lead to some unsafe factors.

sorry, i mean make ConfigChangeEvent unchangable, so no listener can modify it.
I'm trying to extract an Interface with the same name which have getters only.
and with a default implementation

carryxyh:
Pls commit your changes, I will have a look.
:)

satansk:
> Another thing, can we make the field of `ConfigChangeEvent` `final`?
> It seems that if we want to share the `ConfigChangeEvent` instance, it is a better choice to make the `ConfigChangeEvent` field `final`.

Agree, it's better to be final.

PR #3424, opened 2019-02-02
refactor javassist compiler: extract class CtClassBuilder

wanghbxxxx:

refactor javassist compiler: extract class CtClassBuilder
so the compile process is much more clearer

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/ClassUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/ClassUtils.java
@@ -433,9 +433,13 @@ public class ClassUtils {
     }
     
     /**
-     * get classname from qualified name
+     * get simple class name from qualified class name
      */
     public static String getSimpleClassName(String qualifiedName) {
+        if (null == qualifiedName) {
+            return null;
+        }
+        
         int i = qualifiedName.lastIndexOf('.');
         return i < 0 ? qualifiedName : qualifiedName.substring(i + 1);
     }
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/ClassUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/ClassUtils.java
@@ -431,5 +431,13 @@ public class ClassUtils {
         }
         return map;
     }
+    
+    /**
+     * get classname from qualified name
+     */
+    public static String getSimpleClassName(String qualifiedName) {
+        int i = qualifiedName.lastIndexOf('.');
+        return i < 0 ? qualifiedName : qualifiedName.substring(i + 1);
+    }
 
 }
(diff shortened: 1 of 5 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/ClassUtils.java
```diff
@@ -431,5 +431,13 @@ public static String getInitCode(Class<?> type) {
         }
         return map;
     }
+    
+    /**
+     * get classname from qualified name
+     */
+    public static String getSimpleClassName(String qualifiedName) {
+        int i = qualifiedName.lastIndexOf('.');
```
khanimteyaz: Better to check and have check for non empty or null string 

`org.apache.dubbo.common.utils.Assert.notEmptyString(qualifiedName, "Name can't be null or blank");`

dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/JavassistClassInfo.java
```diff
@@ -0,0 +1,160 @@
...
+import javassist.CtNewConstructor;
+import javassist.CtNewMethod;
+import javassist.NotFoundException;
+
+/**
+ * Javassist class info contains all the information used for JavassistCompiler, including:
+ * <p>
+ * class name, imported packages, super class name, implemented interfaces, constructors, fields, methods.
+ */
+public class JavassistClassInfo {
```
khanimteyaz: As once this class gets initialized then it is not getting modified, I would recommend to use builder for this and then have only getter or make it immutable. 

have addInterface, addConstructor keep it the door open for accidental bug. What do you say?

dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/ClassUtils.java
wanghbxxxx: thank you for your advice

wanghbxxxx: committed

dubbo-common/src/main/java/org/apache/dubbo/common/compiler/support/JavassistClassInfo.java
wanghbxxxx: could you give more detailed information what shall i do, could be some pseudo code or java code.
thank you

khanimteyaz: it was saying having setter for collections e.g. setImports(List<String> imports), setInterfaces(List<String> ifaces) .
but looking at again it seems you are doing more than just setter. In this case I think it is ok to have this.

wanghbxxxx: thanks

beiwei30:
LGTM.

Issue #3420, opened 2019-02-01
DO NOT USE org.apache.dubbo.common.utils.IOUtils.write() for your own project


yiding-he:
- [x] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.1-SNAPSHOT

`org.apache.dubbo.common.utils.IOUtils` 这个类看起来像是一个公共类，但实际上里面有坑，不要在你自己的项目里面使用这个类！

### 问题描述

该类有一个 write() 方法，内容如下：

```
public static long write(InputStream is, OutputStream os, int bufferSize) throws IOException {
    int read;
    long total = 0;
    byte[] buff = new byte[bufferSize];
    while (is.available() > 0) {   // !!!
        read = is.read(buff, 0, buff.length);
        if (read > 0) {
            os.write(buff, 0, read);
            total += read;
        }
    }
    return total;
}
```

其中的 while 循环使用了 `InputStream.available()` 的返回值来作为结束判断条件，这个方法在网络环境当中是不能作为读取结束的依据的，因为当网络传输出现延迟时，该方法立刻会返回 0，导致循环结束。测试代码如下：

```
@Test
public void testWriteURL() throws Exception {
    String url = "https://pan.baidu.com/static/images/16new/bg2.jpg";
    long expectedSize = 77_949L;

    InputStream inputStream = new URL(url).openStream();
    ByteArrayOutputStream bos = new ByteArrayOutputStream();
    long writeSize = org.apache.dubbo.common.utils.IOUtils.write(inputStream, bos);
    assertThat(writeSize, equalTo(expectedSize)); // org.hamcrest.MatcherAssert
}
```

这个测试通常会失败，`write()` 方法在读取了部分内容后就结束循环并返回了。

正确的写法请参考 `org.apache.commons.io.IOUtils#copy(java.io.InputStream, java.io.OutputStream)` 的实现，对应的测试方法如下：

```
public static void main(String[] args) throws IOException {
    String url = "https://pan.baidu.com/static/images/16new/bg2.jpg";
    long expectedSize = 77_949L;

    InputStream inputStream = new URL(url).openStream();
    ByteArrayOutputStream bos = new ByteArrayOutputStream();
[...]

PR #3427, opened 2019-02-02
BUG FIX: #3420

yiding-he:
1. `IOUtils.write(InputStream, OutputStream, int)` BUG#3420;
2. `IOUtils.write(Reader, Writer, int)` not using `bufferSize` parameter.

`org.apache.dubbo.common.io.Bytes#unzip()` is now using fixed `write()` method so it is worth running a `BytesTest#testZip()` test.


tswstarplanet:
I think your pr may have some problems. You don’t have to delete all the content of the origin file and then add all content including your fix to the file.
Please check your pr and resubmit a pr.

yiding-he:
@tswstarplanet It's just \r -> \r\n. open "Diff Settings" and check "Hide whitespace changes", you will see actual changes. I'll be careful next time.

PR #3435, opened 2019-02-03
Optimize the close and connect operations of AbstractClient

manzhizhen:

使AbstractClient的close操作和connect、disconnect操作一样，受到锁的保护，降低由于操作冲突而产生不确定性的概率；connect操作先判断client的close状态，如果client已经关闭，那么该client不需要重连。

The AbstractClient close operation is the same as the connect and disconnect operations, and is protected by the lock, which reduces the probability of uncertainty due to operation conflicts; the connect operation first determines the close state of the client. If the client is closed, the client does not need to be reconnected.


dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/transport/AbstractClient.java
```diff
@@ -179,6 +179,12 @@ protected void connect() throws RemotingException {
 
         try {
 
+            if (isClosed()) {
```
khanimteyaz: in case of reconnect would it create any issue? as reconnect call the connect method?

manzhizhen: The client can't be reused, it can only be closed once. So, if you decide to close a client, then when you close it, even if this is reconnecting the client, you can't make reconnect successful.

client其实是不能复用的，只能关闭一次，所以，如果决定关闭一个client，那么将其close时，哪怕这是正在对这个client进行reconnect，也不能让reconnect成功。

PR #3447, opened 2019-02-11
Add delay export  test case

LiZhenNet:

Add delay export test case #3447

## Brief changelog

Add delay export test case #3447

## Verifying this change

UT

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
@@ -27,6 +27,7 @@ import org.apache.dubbo.common.utils.CollectionUtils;
 import org.apache.dubbo.common.utils.ConfigUtils;
 import org.apache.dubbo.common.utils.NamedThreadFactory;
 import org.apache.dubbo.common.utils.StringUtils;
+import org.apache.dubbo.common.utils.CollectionUtils;
 import org.apache.dubbo.config.annotation.Service;
 import org.apache.dubbo.config.context.ConfigManager;
 import org.apache.dubbo.config.invoker.DelegateProviderMetaDataInvoker;
@@ -55,6 +56,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
 import static org.apache.dubbo.common.Constants.LOCALHOST_VALUE;
@@ -98,6 +101,11 @@ public class ServiceConfig<T> extends AbstractServiceConfig {
      * A random port cache, the different protocols who has no port specified have different random port
      */
     private static final Map<String, Integer> RANDOM_PORT_MAP = new HashMap<String, Integer>();
+
+    /**
+     * A delayed exposure service timer
+     */
+    private static final ScheduledExecutorService delayExportExecutor = Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory("DubboServiceDelayExporter", true));
 
     /**
      * The urls of the services exported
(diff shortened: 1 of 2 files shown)


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -341,7 +333,15 @@ public synchronized void export() {
         }
 
         if (delay != null && delay > 0) {
-            delayExportExecutor.schedule(this::doExport, delay, TimeUnit.MILLISECONDS);
```
scxwhite: can we call `delayExportExecutor.shutdown()` method after the delay task add work queue?

LiZhenNet: Yes,This is another way to fix this .

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -102,11 +99,6 @@
      */
     private static final Map<String, Integer> RANDOM_PORT_MAP = new HashMap<String, Integer>();
 
-    /**
-     * A delayed exposure service timer
-     */
-    private static final ScheduledExecutorService delayExportExecutor = Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory("DubboServiceDelayExporter", true));
```
carryxyh: It seems that this delayExportExecutor is a global delay exposed component. I don't think we can delete this field.

Since `ServiceConfig` exists, we can export the service at runtime, so we can't determine if `delayExportExecutor` will still be used. In addition, maintaining an object in memory does not have a big impact.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
carryxyh: Don't think so.
If I want to export the service at runtime(use `ServiceConfig`), turning off `delayExportExecutor` will cause problems.

LiZhenNet: > Don't think so.
> If I want to export the service at runtime(use `ServiceConfig`), turning off `delayExportExecutor` will cause problems.

you are right

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
LiZhenNet: Each ServerConfig instance holds its own delayExportExecutor and  delayExportExecutor only used for delay exporting,I feel it is unnecessary to use it.

kexianjun: > Each ServerConfig instance holds its own delayExportExecutor and delayExportExecutor only used for delay exporting,I feel it is unnecessary to use it.

agree with @carryxyh , all ServiceConfig holds the same delayExportExecutor

LiZhenNet: My fault,  I didn't notice that it was static.

LiZhenNet:
My fault, I have revert my change ，only keep the test case .
Thanks ,@carryxyh,@kexianjun

dubbo-config/dubbo-config-api/src/test/java/org/apache/dubbo/config/ServiceConfigTest.java
```diff
@@ -143,6 +151,15 @@ public void testProxy() throws Exception {
         assertEquals(2, TestProxyFactory.count); // local injvm and registry protocol, so expected is 2
     }
 
+
+    @Test
+    public void testDelayExport() throws Exception {
+        delayService.export();
+        //add 300ms to ensure that the delayService has been exported
```
lixiaojiee: It looks like it would be nice to add an assert here.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
scxwhite: > Don't think so.
> If I want to export the service at runtime(use `ServiceConfig`), turning off `delayExportExecutor` will cause problems.

yes.

PR #3459, opened 2019-02-13
extract new method to test if parameters already contained in the url

wanghbxxxx:

XXXXX

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
@@ -1029,33 +1029,27 @@ class URL implements Serializable {
      * @return A new URL
      */
     public URL addParameters(Map<String, String> parameters) {
-        if (CollectionUtils.isEmptyMap(parameters)) {
-            return this;
-        }
-
-        boolean hasAndEqual = true;
-        for (Map.Entry<String, String> entry : parameters.entrySet()) {
-            String value = getParameters().get(entry.getKey());
-            if (value == null) {
-                if (entry.getValue() != null) {
-                    hasAndEqual = false;
-                    break;
-                }
-            } else {
-                if (!value.equals(entry.getValue())) {
-                    hasAndEqual = false;
-                    break;
-                }
-            }
-        }
         // return immediately if there's no change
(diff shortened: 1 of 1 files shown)


lixiaojiee:
Why not put it in the pr #3458?

dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
```diff
@@ -1058,6 +1039,19 @@ public URL addParameters(Map<String, String> parameters) {
         return new URL(protocol, username, password, host, port, path, map);
     }
 
+    /**
+     * test if parameters are already contained in the url
+     */
+    private boolean contains(Map<String, String> parameters) {
```
khanimteyaz: instead of contains I think **isEqualParameter** would be better name from the task this method is doing. What do you say?

dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
```diff
@@ -1058,6 +1039,19 @@ public URL addParameters(Map<String, String> parameters) {
...
+     */
+    private boolean contains(Map<String, String> parameters) {
+        for (Map.Entry<String, String> entry : parameters.entrySet()) {
+            String value = getParameters().get(entry.getKey());
+            if (!StringUtils.isEquals(value, entry.getValue())) {
+                return false;
+            }
+        }
+        return true;
+    }
```
khanimteyaz: can we not use just 

`return parameters.equals(getParameters()));`

what do you say?

wanghbxxxx: i think meaning of the original code is contain: 
the parameters field may contains key that doesn't exists in the given parameters

wanghbxxxx:
> Why not put it in the pr #3458?

thanks

dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
wanghbxxxx: i have put it in the pr #3458

PR #3480, opened 2019-02-14
[Enhancement]: RestProtocol

kezhenxu94:
- use `Map#computeIfAbsent` to replace check-and-create
- use lambdas to replace annonymous inner types

XXXXX

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/RestProtocol.java
+++ b/dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/RestProtocol.java
@@ -28,16 +28,13 @@ import org.apache.dubbo.rpc.protocol.AbstractProxyProtocol;
 
 import org.apache.http.HeaderElement;
 import org.apache.http.HeaderElementIterator;
-import org.apache.http.HttpResponse;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.config.SocketConfig;
-import org.apache.http.conn.ConnectionKeepAliveStrategy;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
 import org.apache.http.message.BasicHeaderElementIterator;
 import org.apache.http.protocol.HTTP;
-import org.apache.http.protocol.HttpContext;
 import org.jboss.resteasy.client.jaxrs.ResteasyClient;
 import org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder;
 import org.jboss.resteasy.client.jaxrs.ResteasyWebTarget;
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/RestProtocol.java
```diff
@@ -91,12 +88,11 @@ public int getDefaultPort() {
     protected <T> Runnable doExport(T impl, Class<T> type, URL url) throws RpcException {
         String addr = getAddr(url);
         Class implClass = ApplicationModel.getProviderModel(url.getServiceKey()).getServiceInstance().getClass();
-        RestServer server = servers.get(addr);
-        if (server == null) {
-            server = serverFactory.createServer(url.getParameter(Constants.SERVER_KEY, DEFAULT_SERVER));
-            server.start(url);
-            servers.put(addr, server);
-        }
+        RestServer server = servers.computeIfAbsent(addr, addr0 -> {
```
CrazyHZM: It is recommended to change `addr0` to `restServer`

kezhenxu94: why？？

kezhenxu94: @CrazyHZM please refer to the JavaDoc of `computeIfAbsent`, note the second parameter.

CrazyHZM: Not a grammatical error, just add readability

scxwhite: I also think we  should replace addr0 into restServer.   the addr0 is not clear for people  to understand.

kezhenxu94: but I'm wondering why renaming a variable of type `String`  to `restServer` increases readability

CrazyHZM: This has nothing to do with whether it is a `String` type, but using `restServer` in the meaning description will be clearer than `addr0`.
This is also to let others see the source code, know that the second parameter of the `servers` is a `RestServer` type.

kezhenxu94: @scxwhite @CrazyHZM Since the changes is controversial, I've reverted the changes, as for 
> This is also to let others see the source code, know that the second parameter of the `servers` is a `RestServer` type.
I'd rather suggest renaming the `servers` to something like `serversKeyedByAddress` or anything else but what you said.

CrazyHZM: That part of the code does not need to reverted, I think you modified is good, just change `addr0` to `restServer`

kezhenxu94: Changes are made as requested

PR #3498, opened 2019-02-18
fix issue #1875: MergeableClusterInvoker 超时处理逻辑问题

beiwei30:

fix issue #1875: MergeableClusterInvoker 超时处理逻辑问题

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/MergeableClusterInvoker.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/MergeableClusterInvoker.java
@@ -41,7 +41,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/MergeableClusterInvoker.java
```diff
@@ -88,34 +88,38 @@ protected Result doInvoke(Invocation invocation, List<Invoker<T>> invokers, Load
...
             Future<Result> future = entry.getValue();
-            try {
-                Result r = future.get(timeout, TimeUnit.MILLISECONDS);
-                if (r.hasException()) {
-                    log.error("Invoke " + getGroupDescFromServiceKey(entry.getKey()) +
-                                    " failed: " + r.getException().getMessage(),
-                            r.getException());
-                } else {
-                    resultList.add(r);
+            executor.submit(() -> {
```
scxwhite: What do you think of replacing `submit` method with `execute` method?

chickenlj:
I am thinking of the possibility of making each remote call async: `invocation.setAttachment(Constants.ASYNC_KEY, Boolean.TRUE.toString());`

then, `invoker.invoke()` itself will return a Future like async result directly, if this result has started to count timeout from the moment its request data was sent to the remote server, we will not need to calculate timeout by ourself like what we do now for each Future.

chickenlj:
About the original code discussed in #1875:
```java
       for (Map.Entry<String, Future<Result>> entry : results.entrySet()) {
            Future<Result> future = entry.getValue();
            try {
                Result r = future.get(timeout, TimeUnit.MILLISECONDS);
                if (r.hasException()) {
                    log.error("Invoke " + getGroupDescFromServiceKey(entry.getKey()) +
                                    " failed: " + r.getException().getMessage(),
                            r.getException());
                } else {
                    resultList.add(r);
                }
            } catch (Exception e) {
                throw new RpcException("Failed to invoke service " + entry.getKey() + ": " + e.getMessage(), e);
            }
        }

```

I think if the global timeout is 5s, then each `invoker.invoke()` has been restricted to return a value in 5s (it can be a normal value or timeout exception), it will in turn complete the corresponding future, so all `future.get` will not block and will return immediatelly after 5s.

beiwei30:
after discuss with @chickenlj, we believe the original logic can handle the scenario the issue#1875 reports.

Issue #3494, opened 2019-02-18
URL produces too much garbage objects


kezhenxu94:
- [x] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

## PROBLEM
Most of the `setXxxx` and `addParameter*` in `URL` such as `org.apache.dubbo.common.URL#addParameter` and `setXxxx` will return a new `URL` object for chaining calls, and most of the *"middle objects"* are useless and unnecessary, example below:

> class `org.apache.dubbo.config.spring.SimpleRegistryExporter`

```java

    public static Exporter<RegistryService> export(int port, RegistryService registryService) {
        return protocol.export(proxyFactory.getInvoker(registryService, RegistryService.class,
                // HERE
                new URL("dubbo", NetUtils.getLocalHost(), port, RegistryService.class.getName())
                        .setPath(RegistryService.class.getName()) // useless
                        .addParameter(Constants.INTERFACE_KEY, RegistryService.class.getName()) // useless
                        .addParameter(Constants.CLUSTER_STICKY_KEY, "true") // useless
                        .addParameter(Constants.CALLBACK_INSTANCES_LIMIT_KEY, "1000") // useless
                        .addParameter("ondisconnect", "disconnect") // useless
                        .addParameter("subscribe.1.callback", "true") // useless
                        .addParameter("unsubscribe.1.callback", "false")));
    }
```

> class `org.apache.dubbo.config.ServiceConfig`

```java

    @SuppressWarnings({"unchecked", "rawtypes"})
    private void exportLocal(URL url) {
        if (!Constants.LOCAL_PROTOCOL.equalsIgnoreCase(url.getProtocol())) {
            // HERE
            URL local = URL.valueOf(url.toFullString())
                    .setProtocol(Constants.LOCAL_PROTOCOL) // useless
                    .setHost(LOCALHOST_VALUE) // useless
                    .setPort(0);
            Exporter<?> exporter = protocol.export(
                    proxyFactory.getInvoker(ref, (Class) interfaceClass, local));
            exporters.add(exporter);
            logger.info("Export dubbo service " + interfaceClass.getName() + " to local registry");
        }
    }
```

## SOLUTION

[...]

PR #3500, opened 2019-02-18
[DUBBO-3494]: Refactor URL to URLBuilder

kezhenxu94:

Refactor URL to URLBuilder. Related issue #3494 

## Brief changelog

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/file/FileRouterFactory.java
dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
dubbo-common/src/main/java/org/apache/dubbo/common/URLBuilder.java
dubbo-common/src/test/java/org/apache/dubbo/common/URLBuilderTest.java
dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
dubbo-config/dubbo-config-spring/src/test/java/org/apache/dubbo/config/spring/SimpleRegistryExporter.java
dubbo-metadata-report/dubbo-metadata-report-api/src/main/java/org/apache/dubbo/metadata/integration/MetadataReportService.java
dubbo-monitor/dubbo-monitor-default/src/main/java/org/apache/dubbo/monitor/dubbo/DubboMonitorFactory.java
dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java
dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistryFactory.java
dubbo-registry/dubbo-registry-default/src/main/java/org/apache/dubbo/registry/dubbo/DubboRegistryFactory.java
dubbo-registry/dubbo-registry-redis/src/main/java/org/apache/dubbo/registry/redis/RedisRegistry.java
dubbo-registry/dubbo-registry-zookeeper/src/main/java/org/apache/dubbo/registry/zookeeper/ZookeeperRegistry.java
dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/proxy/wrapper/StubProxyFactoryWrapper.java
dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/ReferenceCountExchangeClient.java
dubbo-rpc/dubbo-rpc-rest/src/test/java/org/apache/dubbo/rpc/protocol/rest/RestProtocolTest.java

## Verifying this change

XXXXX


kezhenxu94:
Can any one of the admins verify this PR? @chickenlj or any other admins?

dubbo-common/src/main/java/org/apache/dubbo/common/URLBuilder.java
```diff
@@ -0,0 +1,342 @@
...
+
+    public URLBuilder(URL url) {
+        protocol = url.getProtocol();
+        username = url.getUsername();
+        password = url.getPassword();
+        host = url.getHost();
+        port = url.getPort();
+        path = url.getPath();
+        parameters = new HashMap<>(url.getParameters());
+    }
```
chickenlj: I think providing a method like`URLBuilder.from(url)` would be better than using a constructor directly.

What do you think?

dubbo-common/src/main/java/org/apache/dubbo/common/URLBuilder.java
```diff
@@ -0,0 +1,342 @@
...
+
+    public String getParameter(String key) {
+        String value = parameters.get(key);
+        if (StringUtils.isEmpty(value)) {
+            value = parameters.get(Constants.DEFAULT_KEY_PREFIX + key);
+        }
+        return value;
+    }
+}
```
chickenlj: Does these two reading methods `hasParameter` `hasParameter` necessary for a Builder?

chickenlj:
All look good to me except for these above two minor problems.

dubbo-common/src/main/java/org/apache/dubbo/common/URLBuilder.java
kezhenxu94: The two methods are used inside the builder, and I just made them public because I think it's necessary because there are cases where we want to write code like this:
```java
URLBuilder builder = ...;
if (builder.hasParameter("serverSideProp")) {
    builder.addParameter("someServerSideProp1", v2);
    builder.addParameter("someServerSideProp2", v3);
}
```
without these methods we have to (again) create `URL`s just for calling the methods, what do you think

dubbo-common/src/main/java/org/apache/dubbo/common/URLBuilder.java
kezhenxu94: updated as requested

jasonjoo2010:
This issue i have also found when i worked on another PR. You exactly do what i want to. So i think it's a good improvement.

dubbo-common/src/main/java/org/apache/dubbo/common/URLBuilder.java
chickenlj: Got it.

dubbo-common/src/main/java/org/apache/dubbo/common/URLBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+        port = 0;
+        path = null;
+        parameters = new HashMap<>();
+    }
+
+    public URLBuilder(String protocol, String host, int port) {
+        this(protocol, null, null, host, port, null, null);
+    }
+
+    public URLBuilder(String protocol, String host, int port, String[] pairs) { // varargs ... conflict with the following path argument, use array instead.
```
beiwei30: this comments looks unnecessary to me.

dubbo-common/src/main/java/org/apache/dubbo/common/URLBuilder.java
```diff
@@ -0,0 +1,350 @@
...
+
+    public URLBuilder(String protocol, String host, int port, String path) {
+        this(protocol, null, null, host, port, path, null);
+    }
+
+    public URLBuilder(String protocol, String host, int port, String path, String... pairs) {
+        this(protocol, null, null, host, port, path, CollectionUtils.toStringMap(pairs));
+    }
+
+    public URLBuilder(String protocol, String host, int port, String path, Map<String, String> parameters) {
```
beiwei30: since it's builder, we should make all constructor **private**

kezhenxu94: > since it's builder, we should make all constructor **private**

I don't think the constructor of the builder itself should be `private`, it is the constructor of the class that the builder is to build(the `URL` class) should be `private` to prevent client code from creating the (`URL`) objects without using the builder. But for compatibility, the constructor of `URL` should not be private at this moment.

About the visibilities of the constructors, please refer to [this thread](https://stackoverflow.com/a/37028913)

> This looks a bit strange because usually a builder uses private constructor to totally avoid the creation of the instance by the client code without using the builder.

and [this post](https://dzone.com/articles/design-patterns-the-builder-pattern):

> The Pattern
The Builder pattern allows us to write readable, understandable code to set up complex objects. It is often implemented with a fluent interface, which you may have seen in tools like Apache Camel or Hamcrest. The builder will contain all of the fields that exist on the BankAccount class itself. We will configure all of the fields that we want on the builder, and then we'll use the builder to create accounts. At the same time, we'll remove the public constructor from the BankAccount class and replace it with a private constructor so that accounts can only be created via the builder.

kezhenxu94: @beiwei30 what do you think

Issue #3525, opened 2019-02-21
RpcContextFilter隐式传参取值问题


yuxinyin:
### Environment

* Dubbo version: 2.7.0
* Operating System version: Windowns 10
* Java version: 1.8

Class:  org.apache.dubbo.rpc.protocol.rest.RpcContextFilter  
Line:  73-74
Code:
String key = entry.getKey();
String value = entry.getKey();

导致隐式传参value值错误

carryxyh:
Thx. My bad. I will fix it.

carryxyh:
https://github.com/apache/incubator-dubbo/pull/3526

PR #3526, opened 2019-02-21
fix context filter's bug

carryxyh:
fix context filter's bug:
`String value = entry.getKey();` -> `String value = entry.getValue();`


carryxyh:
https://github.com/apache/incubator-dubbo/issues/3525

ralf0131:
Not sure what these changes will cause UT failure, would you please take a look?

carryxyh:
Of course.
I'd like have a look later..

Issue #3533, opened 2019-02-21
NPE occurred when the configuration was deleted. (...configcenter...zookeeper.CacheListener#childEvent)


hekrxe:
- [X] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.0
* Operating System version: MacOS
* Java version: 1.8

### Steps to reproduce this issue

When tag router is deleted by Dubbo Admin

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result

What do you expected from the above steps？

### Actual Result

What actually happens?

If there is an exception, please attach the exception trace:

```
java.lang.NullPointerException: null
	at java.lang.String.<init>(String.java:515) ~[na:1.8.0_161]
	at org.apache.dubbo.configcenter.support.zookeeper.CacheListener.childEvent(CacheListener.java:86) ~[dubbo-2.7.0.jar:2.7.0]
	at org.apache.curator.framework.recipes.cache.TreeCache$2.apply(TreeCache.java:685) [curator-recipes-2.8.0.jar:na]
	at org.apache.curator.framework.recipes.cache.TreeCache$2.apply(TreeCache.java:679) [curator-recipes-2.8.0.jar:na]
	at org.apache.curator.framework.listen.ListenerContainer$1.run(ListenerContainer.java:92) [curator-framework-2.8.0.jar:na]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [na:1.8.0_161]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [na:1.8.0_161]
	at java.lang.Thread.run(Thread.java:748) [na:1.8.0_161]
```

PR #3536, opened 2019-02-21
Dump TagRouterRule

carryxyh:
Dump TagRouterRule since the TagRouterRule can be changed to `null` by ConfigCenter.
Since his Issue[1] does not give a stack trace, I am not sure that his problem is caused by this.

[1] https://github.com/apache/incubator-dubbo/issues/3533

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java
@@ -87,12 +87,12 @@ public class TagRouter extends AbstractRouter implements Comparable<Router>, Con
             return invokers;
         }
 
-        if (tagRouterRule == null || !tagRouterRule.isValid() || !tagRouterRule.isEnabled()) {
+        // since the rule can be changed by config center, we should dump one to use.
+        TagRouterRule dump = tagRouterRule;
+        if (dump == null || !dump.isValid() || !dump.isEnabled()) {
             return filterUsingStaticTag(invokers, url, invocation);
         }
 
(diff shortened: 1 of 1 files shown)


chickenlj:
LGTM, I think your change is critical to guarantee consistent behavior when rule changes. 
But what's the possible relation with the issue mentioned here.

carryxyh:
When I saw his issue before, he did not give his exception stack. Later, the issuer added the stack. I looked at the code and found that this place also caused some problems, so I modified it. @chickenlj

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java
```diff
@@ -87,7 +87,9 @@ public URL getUrl() {
             return invokers;
         }
 
-        if (tagRouterRule == null || !tagRouterRule.isValid() || !tagRouterRule.isEnabled()) {
+        // since the rule can be changed by config center, we should dump one to use.
+        TagRouterRule dump = tagRouterRule;
```
ralf0131: Shall we mark it as final? Also I don't quite understand why it is named as dump?
If it is just a copy, will a name called `tagRouterRuleCopy` be better?

chickenlj: Agree, tagRouterRuleCopy or CopyOfXxx would be better.

carryxyh: For me, the semantics of `copy` or `refer` are good choices. So, I'd like chose `copy` in this case.

PR #3565, opened 2019-02-25
SpringContainer.getContext() 在服务刚暴露时，服务内部获取到的可能是null值

stoneapple:
SpringContainer.getContext() 在服务刚暴露时，服务内部获取到的可能是null值


ralf0131:
Please resolve the UT failure first. Generally we do not recommend sending pull request to 2.5.x branch, because I think it is close to End of life unless critical bug are found. (The community will discuss this.)

chickenlj:
Thanks. But I think this one is more reasonable https://github.com/apache/incubator-dubbo/pull/3600

Issue #3302, opened 2019-01-22
Optimize compareTo in Router to guarantee consistent behaviour.


chickenlj:
router1.compareTo(router2);
and
router2.compareTo(router1);

should always give the same order.

chickenlj:
compareTo in Router interface
```java
@Override
    default int compareTo(Router o) {
        if (o == null) {
            throw new IllegalArgumentException();
        }
        if (this.getPriority() == o.getPriority()) {
            if (o.getUrl() == null) {
                return 1;
            }
            if (getUrl() == null) {
                return -1;
            }
            return getUrl().toFullString().compareTo(o.getUrl().toFullString());
        } else {
            return getPriority() > o.getPriority() ? 1 : -1;
        }
    }
```


compareTo in ScriptRouter
```java
 @Override
    public int compareTo(Router o) {
        if (o == null || o.getClass() != ScriptRouter.class) {
            return 1;
        }
        ScriptRouter c = (ScriptRouter) o;
        return this.priority == c.priority ? rule.compareTo(c.rule) : (this.priority > c.priority ? 1 : -1);
    }
```

lexburner:
I will fix this.

PR #3566, opened 2019-02-25
Optimize compareTo in Router to guarantee consistent behaviour.

lexburner:
fix issue https://github.com/apache/incubator-dubbo/issues/3302

remove useless implement and the duplicate compareTo logic,use getPriotity instead.


lexburner:
changes are as below:

1. remove the duplicate interface `Comparable` since `Router` extends.
2. replace the compare logic from `compareTo` to `getPriority`. Router subclass should define its own priotity to decide its order on the chian.

chickenlj:
LGTM.

Issue #3289, opened 2019-01-21
Support wildcard ip for tag router rule.


chickenlj:

chickenlj:
@biyuhao +1

cvictory:
https://www.java2s.com/Code/Java/Network-Protocol/CheckifIPaddressmatchpattern.htm 
Maybe this meet this requirement.

cvictory:
https://stackoverflow.com/questions/2942299/converting-cidr-address-to-subnet-mask-and-network-address

cvictory:
https://github.com/cvictory/incubator-dubbo/tree/issue_3289

PR #3578, opened 2019-02-28
fix issue #3289. enhance tagRoute:  support ip expression match

cvictory:
https://github.com/apache/incubator-dubbo/issues/3289

It support both ipv4 and ipv6.  The supported format of ip expression as follow:
1. subnet . such as 192.168.1.192/26,  234e:0:4567::3d/64 
2. all ip named '* .* . *. * '
3. 192.168.1.*  and  234e:0:4567:0:0:0:3d:*
4. 192.168.1.1-23 and 234e:0:4567:0:0:0:3d:0-ea


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java
```diff
@@ -199,11 +201,26 @@ private boolean isForceUseTag(Invocation invocation) {
...
 
     private boolean addressNotMatches(URL url, List<String> addresses) {
-        return addresses == null || !addresses.contains(url.getAddress());
+        return addresses == null || !checkAddressMatch(addresses, url.getAddress());
+    }
+
+    private boolean checkAddressMatch(List<String> addresses, String targetAddress) {
+        for (String address : addresses) {
+            try {
+                if (NetUtils.matchIpExpression(address, targetAddress)) {
```
chickenlj: The address here is expected to have port included: `127.0.0.1:20880`, but `NetUtils.matchIpExpression` can only handle raw IP.

cvictory: I have fixed this one to support port configuration.

chickenlj:
Keeps outputting the following log:

[07/03/19 04:45:12:012 CST] main  INFO utils.NetUtils:  [DUBBO] The pattern port is not equals to port form param., dubbo version: 2.7.1-SNAPSHOT, current host:

PR #3579, opened 2019-02-28
fix heartbeat internal

carryxyh:
fix heartbeat internal

### DIFF
--- a/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyClient.java
+++ b/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyClient.java
@@ -78,7 +78,7 @@ public class NettyClient extends AbstractClient {
 
             @Override
             protected void initChannel(Channel ch) throws Exception {
-                int heartbeatInterval = UrlUtils.getIdleTimeout(getUrl());
+                int heartbeatInterval = UrlUtils.getHeartbeat(getUrl());
                 NettyCodecAdapter adapter = new NettyCodecAdapter(getCodec(), getUrl(), NettyClient.this);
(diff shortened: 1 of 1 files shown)


dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyClient.java
```diff
@@ -78,7 +78,7 @@ protected void doOpen() throws Throwable {
 
             @Override
             protected void initChannel(Channel ch) throws Exception {
-                int heartbeatInterval = UrlUtils.getIdleTimeout(getUrl());
+                int heartbeatInterval = UrlUtils.getHeartbeat(getUrl());
```
beiwei30: looks like we still should call `getIdleTimeout` instead, would you mind to check the old logic?

carryxyh: Since the current server does not return a heartbeat request, if you use getIdleTimeout(3 * heartbeat), the heartbeat send period will be 3 * heartbeat. Will this be a long time?

carryxyh:
@beiwei30 
Hi, beiwei, 
I still keep my opinion. It looks like `3 * heartbeat's` time to trigger a heartbeat is too long.
Please let me know if there is any error in my understanding. Thx.

dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyClient.java
chickenlj: I think @carryxyh is right. This value tells `IdleStateHandler` the interval of sending a heartbeat package.

beiwei30:
LGTM

PR #3596, opened 2019-03-05
Failover executor for failover request.

carryxyh:
Add a tag to the failover request. When the Provider receives such a request, it often means that the previous request is a failure/timeout. We need to open a new thread pool to execute these requests, preventing the failover request from filling our Provider side business thread pool.

If we accept this pr, there are still some things to decide:
- Whether to make the failover pool configurable
- The location of the failover pool
- Shell we set this request as a failover request in any exception situation?

### DIFF
--- a/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
+++ b/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
@@ -90,15 +90,47 @@ public class DubboProtocol extends AbstractProtocol {
 
         @Override
         public CompletableFuture<Object> reply(ExchangeChannel channel, Object message) throws RemotingException {
-
             if (!(message instanceof Invocation)) {
                 throw new RemotingException(channel, "Unsupported request: "
                         + (message == null ? null : (message.getClass().getName() + ": " + message))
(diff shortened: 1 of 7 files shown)


dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
```diff
@@ -90,15 +90,47 @@
...
 
-            Invocation inv = (Invocation) message;
-            Invoker<?> invoker = getInvoker(channel, inv);
+            final Invocation inv = (Invocation) message;
+            final Invoker<?> invoker = getInvoker(channel, inv);
+            if (isFailoverInvoke(inv)) {
+                // We execute these failover requests in the failover thread pool, as they may be wrong or timed out.
+                //
+                // Failure to do so may result in the business thread pool being quickly filled.
+                failoverExecutor.execute(() -> doService(channel, invoker, inv));
```
khanimteyaz: will there be any scenario where member variable of invocation can be shared across multle thread?

carryxyh: Hi, khan, I am very happy to receive your review again. 
:)

I also considered this problem, but at this time, the invoker has not yet started execution, so the variables unique to our thread have not yet taken effect, such as RpcContext.

PR #3601, opened 2019-03-06
Fix bug of ConfigManger#checkDuplicate

saltyx:

- The logic of the `ConfigManger#checkDuplicate `method is wrong
- `AbstractInterfaceConfig#setApplication` `AbstractInterfaceConfig#setModule` `AbstractInterfaceConfig#setMonitor` `AbstractInterfaceConfig#setConfigCenter`
first check config of ConfigManger class，then update it

## Brief changelog

调整`!oldOne.equals(newOne)`为`oldOne.equals(newOne)`

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
@@ -689,7 +689,10 @@ public abstract class AbstractInterfaceConfig extends AbstractMethodConfig {
     }
 
     public void setApplication(ApplicationConfig application) {
-        ConfigManager.getInstance().setApplication(application);
+        if (!ConfigManager.getInstance().getApplication().isPresent() ||
+                !ConfigManager.getInstance().getApplication().get().equals(application)) {
+            ConfigManager.getInstance().setApplication(application);
+        }
         this.application = application;
     }
 
@@ -714,7 +717,10 @@ public abstract class AbstractInterfaceConfig extends AbstractMethodConfig {
     }
 
     public void setModule(ModuleConfig module) {
-        ConfigManager.getInstance().setModule(module);
+        if (!ConfigManager.getInstance().getModule().isPresent()
+                || !ConfigManager.getInstance().getModule().get().equals(module)) {
+            ConfigManager.getInstance().setModule(module);
+        }
         this.module = module;
     }
 
@@ -756,7 +762,10 @@ public abstract class AbstractInterfaceConfig extends AbstractMethodConfig {
     }
 
     public void setMonitor(MonitorConfig monitor) {
-        ConfigManager.getInstance().setMonitor(monitor);
+        if (!ConfigManager.getInstance().getMonitor().isPresent()
+                || !ConfigManager.getInstance().getMonitor().get().equals(monitor)) {
+            ConfigManager.getInstance().setMonitor(monitor);
+        }
         this.monitor = monitor;
     }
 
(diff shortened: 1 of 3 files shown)


carryxyh:
LGTM

kezhenxu94:
@saltyx can you please also fix the failed CI build?

saltyx:
> @saltyx can you please also fix the failed CI build?

ok

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
```diff
@@ -689,7 +689,10 @@ public ApplicationConfig getApplication() {
     }
 
     public void setApplication(ApplicationConfig application) {
-        ConfigManager.getInstance().setApplication(application);
+        if (!ConfigManager.getInstance().getApplication().isPresent() ||
+                !ConfigManager.getInstance().getApplication().get().equals(application)) {
+            ConfigManager.getInstance().setApplication(application);
+        }
```
ralf0131: Why should we need this logic? The logic, including the null check and duplicate check have  implemented in org.apache.dubbo.config.context.ConfigManager#setApplication.  Why should be added it again?

So as the other change in this file.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/context/ConfigManager.java
```diff
@@ -310,7 +310,7 @@ public void refreshAll() {
     }
 
     private void checkDuplicate(AbstractConfig oldOne, AbstractConfig newOne) {
-        if (oldOne != null && !oldOne.equals(newOne)) {
+        if (oldOne != null && oldOne.equals(newOne)) {
```
ralf0131: I notice the UT is missing here, could you add a unit test for this change?

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
saltyx: such as AbstractInterfaceConfig#createApplicationIfAbsent , if `this.application` is null and the `ApplicationConfig` in `ConfigManager`  is not null , it will reset   `ApplicationConfig` variable，that always causes an error in `ConfigManager#checkDuplicate`

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/context/ConfigManager.java
saltyx: ok

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
ralf0131: Still do not quite get it. in AbstractInterfaceConfig#createApplicationIfAbsent , if `this.application` is null, it will directly return.
Can you write a unit test to reproduce the error?

saltyx: I thought about it again, this code seems to be doing this deliberately. my logic is duplicate check and block the IllegalStateException, i should remove them . 

And  because `ConfigManager#checkDuplicate` is logically wrong, it will cause some unit test errors which is  the purpose of changing this logic at the beginning...

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/context/ConfigManager.java
saltyx: please check `AbstractInterfaceConfigTest#checkDuplicateApplication`

chickenlj:
Two Configs with different are items are recognized as duplicate, that's why uses `!oldOne.equals(newOne)`.

Two Configs with identically the same items are thought to be the same and one should be ignored silently.

carryxyh:
`Two Configs with different are items are recognized as duplicate, that's why uses !oldOne.equals(newOne).`

Hi, Thank you for your explanation. 
This is my mistake, I did not understand the meaning of this part of the code. I will remove my previous comments.

chickenlj:
@carryxyh Haha, it's ok, please feel free to express your opinions.

chickenlj:
@saltyx Sorry, I will close this PR now.

PR #3664, opened 2019-03-14
[Dubbo-3662]fix ServerContext can not transform data in sync invoke

tswstarplanet:
Make RpcContext's ServerContext can transform data from provider to consumer

When user use sync invoke, the RpcContext's ServerContext cannot transform data from provider to consumer.

## Brief changelog

Make ContextFilter remove ServerContext after it copy data from ServerContext to result's attachments of the invoke when provider is responding consumer.

## Verifying this change

Have been tested.

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/ContextFilter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/ContextFilter.java
@@ -80,7 +80,6 @@ public class ContextFilter implements Filter {
         } finally {
             // IMPORTANT! For async scenario, we must remove context from current thread, so we always create a new RpcContext for the next invoke for the same thread.
             RpcContext.removeContext();
-            RpcContext.removeServerContext();
         }
     }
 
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/ContextFilter.java
```diff
@@ -80,14 +80,14 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
         } finally {
             // IMPORTANT! For async scenario, we must remove context from current thread, so we always create a new RpcContext for the next invoke for the same thread.
```
carryxyh: Didn't you see this comment?

We must remove the context at the end of the invoke. So I think your pr will cause new problems.

tswstarplanet: How about use a local variable to store the async identifier, and in the finally block, we can check if it is a async invoke, the servercontext will be removed. If not, we donnot remove it. Doesn't this impact the asyn invoke ?

PR #3693, opened 2019-03-19
improve:ProviderMethodModel add parameterClasses

qinliujie:

add more information for  ProviderMethodModel

## Brief changelog

ProviderMethodModel add parameterClasses and genericParameterTypes property


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/model/ProviderModel.java
```diff
@@ -74,6 +74,11 @@ public ProviderMethodModel getMethodModel(String methodName, String[] argTypes)
         return null;
     }
 
+    public List<ProviderMethodModel> getMethodModelList(String methodName) {
+        return methods.get(methodName);
```
ralf0131: This may return null value, are you expecting that? Typically when you want to return a List, user will expect an empty list rather than null.

PR #3731, opened 2019-03-26
Fix bug with CacheFilter when use async call.

carryxyh:
When use async call with cacheFilter, we will cache a empty result which is incorrect.


carryxyh:
related comment here:
https://github.com/apache/incubator-dubbo/pull/3185#issuecomment-476115059

carryxyh:
Related to https://github.com/apache/incubator-dubbo/pull/3997

PR #3748, opened 2019-03-27
[DUBBO-3746] Bugfix: infinite loop in AbstractRegistry when IOException. fix #3746

### LINKED ISSUE #3746: AbstractRegistry中doSaveProperties方法如果读取的文件权限不足会进入catch块，然后重试，死循环导致cpu占用率百分百

kezhenxu94:

resolve #3746 

When `IOException` is thrown in `org.apache.dubbo.registry.support.AbstractRegistry#doSaveProperties`, which is usually unrecoverable (permission denied, path not exists or no free disk space) it will retry infinitely, this patch tries to resolve this issue by stop retrying when `IOException` occurs.

## Brief changelog

org.apache.dubbo.registry.support.AbstractRegistry

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistry.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistry.java
@@ -185,6 +185,7 @@ public abstract class AbstractRegistry implements Registry {
                 return;
             }
             if (version < lastCacheChanged.get()) {
+                savePropertiesRetryTimes.set(0);
                 return;
             } else {
                 registryCacheExecutor.execute(new SaveProperties(lastCacheChanged.incrementAndGet()));
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistry.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistry.java
@@ -174,6 +174,10 @@ public abstract class AbstractRegistry implements Registry {
                 }
             }
         } catch (Throwable e) {
+            if (e instanceof IOException) { // may not be recoverable when IOException throws, give up retrying
+                logger.warn("Failed to save registry cache file, cause: " + e.getMessage(), e);
+                return;
+            }
             if (version < lastCacheChanged.get()) {
                 return;
             } else {
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistry.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistry.java
@@ -48,6 +48,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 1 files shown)


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistry.java
```diff
@@ -174,6 +174,10 @@ public void doSaveProperties(long version) {
                 }
             }
         } catch (Throwable e) {
+            if (e instanceof IOException) { // may not be recoverable when IOException throws, give up retrying
```
ralf0131: Why not catch IOException directly?

ralf0131:
What if other type of exception other than IOException is thrown, this might introduce the same issue. I think a safer way is to limit the number of time to retry.

kezhenxu94:
> What if other type of exception other than IOException is thrown, this might introduce the same issue. I think a safer way is to limit the number of time to retry.

@ralf0131 how many times to retry? I think 3 is enough, what's your opinion?

ralf0131:
> > What if other type of exception other than IOException is thrown, this might introduce the same issue. I think a safer way is to limit the number of time to retry.
> 
> @ralf0131 how many times to retry? I think 3 is enough, what's your opinion?

3 looks good to me.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistry.java
```diff
@@ -60,6 +61,8 @@
     private static final char URL_SEPARATOR = ' ';
     // URL address separated regular expression for parsing the service provider URL list in the file cache
     private static final String URL_SPLIT = "\\s+";
+    // Max times to retry to save properties to local cache file
+    private static final int MAX_RETRY_TIMES_SAVE_PROPERTIES = 3;
```
carryxyh: I have some doubts about the retry scenario. In what scenario do we need to try again when the save file fails? Or what scenario has the first save failure, the second time may be successful?

carryxyh: @ralf0131 
Would u pls also check this point? Thx
:)

ralf0131: Er.. I'd say I am not quite sure why it is designed like this. One scenario I can think is that multiple process try to write to the same file, if other process is holding the file lock and this process might failed.  Does that sound reasonable?

carryxyh: Hi,huxing
Thank you for your reply.

As far as I know, if multiple processes try to write data to the same file, `acquire FD` will be called at the bottom, and this operation is usually a lock operation, so I personally think that if this happens, actually will cause queuing instead of writing failure.

But I am not very sure about this point of view. We may need a local test to ensure the rigor of the logic? 
How do u think about it?
:)

ralf0131: Yes, you can try it.

kezhenxu94: Hi @carryxyh , do you have any conclusion? If not, I think I can have a try with it. :)

ralf0131:
Ping @carryxyh , what's your opinion on this pull request?

carryxyh:
Sorry for coming late. @kezhenxu94 @ralf0131 

I have no more comments on this pr. After my test, multiple threads read and write to the same file will not generate exceptions and will not be out of order. But we still have to consider some other special circumstances that cause the write to fail.

I think we can merge this pr first, I will conduct further testing and submit an extra pr if I think we need to optimize it.

carryxyh:
@kezhenxu94 
Pls help to check if `org.apache.dubbo.metadata.support.AbstractMetadataReport#doSaveProperties` has the same problem.

PR #3752, opened 2019-03-27
improve:change service metadata init order

qinliujie:

change service metadata init order, so that we can use service metadata when ServiceConfig/ReferenceConfig init 

## Brief changelog

change service metadata init order


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -1,1005 +1,1005 @@
...
-        this.protocols = convertProviderToProtocol(providers);
-    }
-
-    @Override
-    @Parameter(excluded = true)
-    public String getPrefix() {
-        return Constants.DUBBO + ".service." + interfaceName;
-    }
-}
+/*
```
kexianjun: what happened to this file ? pls make it clear what you have chenged to this file

qinliujie: sorry, I had make some conflict....I will make a new pr later

PR #3754, opened 2019-03-27
improve:change serviceMetadata order

qinliujie:

change service metadata init order, so that we can use service metadata when ServiceConfig/ReferenceConfig init 

## Brief changelog

change service metadata init order


ralf0131:
@Jeff-Lv You can merge it if you feel confident. Do you still want others to review?

PR #3825, opened 2019-04-08
[Dubbo-3829] support google pb generic invocation

vio-lin:

Currently, Generic Reference cannot support Google Protobuf. With bride usage of Google PB, we are going to support it via Generic Reference. See issue #3829

## Brief changelog
1.  add ProtobufUtil.class  to serialize /deserialize Google Protobuf Object
2.  modify GenericFilter.class

## feedback wanted
To support Google PB Testing, GenericFilter is part of my change, also Service Metadata have to support that. If the solution is ok, i will continue on the Service Metadata part.

- [x] Make sure there is a GITHUB_issue field for the change (usually before you start working on it). Trivial changes like typos do not require a GITHUB issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [x]   Format the pull request title like [Dubbo-XXX] Fix UnknownException when host config not exist #XXX. Each commit in the pull request should have a meaningful subject line and body.
- [x]   Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ]   Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in test module.
- [x]   Run mvn clean install -DskipTests=false & mvn clean test-compile failsafe:integration-test to make sure unit-test and integration-test pass.
- [x]   If this contribution is large, please follow the Software Donation Guide.


vio-lin:
can someone give me some feedback /cry

dubbo-common/pom.xml
```diff
@@ -77,5 +77,9 @@
             <groupId>de.ruedigermoeller</groupId>
             <artifactId>fst</artifactId>
         </dependency>
+        <dependency>
```
beiwei30: We should not introduce this dependency in the common. If only GenericFilter relies on it, pls. move it to "dubbo-rpc-api" module.

dubbo-common/src/main/java/org/apache/dubbo/common/utils/ProtobufUtils.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.google.protobuf.GeneratedMessageV3;
+import com.google.protobuf.GeneratedMessageV3.Builder;
+import com.google.protobuf.MessageOrBuilder;
+import com.google.protobuf.util.JsonFormat;
+import com.google.protobuf.util.JsonFormat.Parser;
+import com.google.protobuf.util.JsonFormat.Printer;
+
+/**
+ * 2019/4/4
```
beiwei30: pls. remove this header.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
```diff
@@ -107,6 +108,16 @@ public Result invoke(Invoker<?> invoker, Invocation inv) throws RpcException {
                                             args[i].getClass().getName());
                         }
                     }
+                }else if (ProtocolUtils.isProtobufGenericSerialization(generic)) {
```
beiwei30: pay more attention to code style. You should import https://github.com/apache/incubator-dubbo/tree/master/codestyle and go through your change thoroughly.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
```diff
@@ -125,7 +136,9 @@ public Result invoke(Invoker<?> invoker, Invocation inv) throws RpcException {
                     }
                 } else if (ProtocolUtils.isBeanGenericSerialization(generic)) {
                     return new RpcResult(JavaBeanSerializeUtil.serialize(result.getValue(), JavaBeanAccessor.METHOD));
-                } else {
+                } else if(ProtocolUtils.isProtobufGenericSerialization(generic)) {
```
beiwei30: Even move ProtobufUtils into dubbo-rpc-api, it is still a little bit heavier to let dubbo-rpc-api to depend on Protobuf. 

At least we need to file a new issue to track down this work, to refactor this filter to use extension loader to load generic serializations to avoid introduce unnecessary dependencies in this module.

dubbo-demo/dubbo-demo-protobuf/dubbo-demo-protobuf-consumer/pom.xml
```diff
@@ -0,0 +1,78 @@
+<!--
```
beiwei30: we should move this demo into [dubbo-samples](https://github.com/apache/incubator-dubbo-samples)

dubbo-common/src/main/java/org/apache/dubbo/common/utils/ProtobufUtils.java
vio-lin: Done

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
vio-lin: done， test by format another file,it works well

dubbo-demo/dubbo-demo-protobuf/dubbo-demo-protobuf-consumer/pom.xml
vio-lin: remove these demos.And these will be added in a pull request to dubbo-samples.

carryxyh:
I am not sure what pb means when using generalized calls.
In my mind, the use of generalized calls is limited and in these scenarios, performance is not something we are pursuing.

In addition, agree with beiwei's point of view, it seems that using pb is a very heavy dependency.

vio-lin:
> I am not sure what pb means when using generalized calls.
> In my mind, the use of generalized calls is limited and in these scenarios, performance is not something we are pursuing.
> 
> In addition, agree with beiwei's point of view, it seems that using pb is a very heavy dependency.

beiwei30:
@vio-lin why this issue starts to touch so many files? would you mind to close it and re-submit your pull request?

vio-lin:
> @vio-lin why this issue starts to touch so many files? would you mind to close it and re-submit your pull request?
OK  i will do it latter

PR #3833, opened 2019-04-09
metrics service

panxiaojun233:
exposing Metrics data by metrics service


ralf0131:
Hi, could you please resolve the conflict first?

dubbo-monitor/dubbo-monitor-default/src/main/java/org/apache/dubbo/monitor/dubbo/MetricsFilter.java
```diff
@@ -0,0 +1,251 @@
...
+                20880 : Integer.valueOf(invoker.getUrl().getParameter(Constants.METRICS_PORT));
+
+        this.protocolName = invoker.getUrl().getParameter(Constants.METRICS_PROTOCOL) == null ?
+                Constants.DEFAULT_PROTOCOL : invoker.getUrl().getParameter(Constants.METRICS_PROTOCOL);
+
+        if (exported.compareAndSet(false, true)) {
+            initMetricsInvoker();
+            Protocol protocol = ExtensionLoader.getExtensionLoader(Protocol.class).getExtension(protocolName);
+            try {
+                protocol.export(metricsInvoker);
```
chickenlj: `metricsInvoker` is only used in here,  it can be a local variable instead of defined as MetricsFilter's property.

dubbo-monitor/dubbo-monitor-default/src/main/java/org/apache/dubbo/monitor/dubbo/MetricsFilter.java
```diff
@@ -0,0 +1,251 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class MetricsFilter implements Filter {
+
+    private static final Logger logger = LoggerFactory.getLogger(MetricsFilter.class);
+    private static volatile AtomicBoolean exported = new AtomicBoolean(false);
+    private Integer port = 20880;
```
chickenlj: I find Protocol has a `getDefaultPort()` method to get each Protocol's default port. So maybe there's no need to define a default port here? How about following each protocol's default port. 

One thing to notice is that you may need to create a separate connection for metrics service, this can be helpful in isolating from normal RPC calls especially when they are working on the same port.

dubbo-monitor/dubbo-monitor-default/src/main/java/org/apache/dubbo/monitor/dubbo/MetricsFilter.java
```diff
@@ -0,0 +1,251 @@
...
+
+                List res = collector.build();
+                res.addAll(getThreadPoolMessage());
+                result.setValue(JSON.toJSONString(res));
+                return result;
+            }
+
+            @Override
+            public URL getUrl() {
+                return URL.valueOf(protocolName + "://" + NetUtils.getLocalAddress().getHostName() + ":" + port + "/" + MetricsService.class.getName());
```
chickenlj: `NetUtils.getLocalAddress().getHostName()`

The IP address used here may have a problem, we should think using a unified tool for getting outward IP, I remember there's one thread discussing the same topic.

chickenlj:
@nzomkxia Would you please help to review the Metrics collecting part?

nzomkxia:
> @nzomkxia Would you please help to review the Metrics collecting part?

metrics collecting part looks good. but  how to pass the metrics protocol and port to dubbo admin? they are not in the registry center under [simplified URL](http://dubbo.apache.org/zh-cn/docs/user/demos/simplify-registry-data.html) mode

nzomkxia:
the metrics stuff can be stored in metadata center

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ReferenceConfig.java
```diff
@@ -276,6 +282,7 @@ private void init() {
             }
         }
         map.put(Constants.INTERFACE_KEY, interfaceName);
+        appendParameters(map, metrics, Constants.METRICS_KEY);
```
chickenlj: `Constants.METRICS_KEY` should be removed

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -145,6 +145,11 @@
      */
     private ProviderConfig provider;
 
+    /**
+     * The metrics configuration
+     */
+    private MetricsConfig metrics;
```
chickenlj: Both ServiceConfig and ReferenceConfig have this property, how about moving it to one shared superclass.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -426,7 +431,9 @@ private void doExportUrlsFor1Protocol(ProtocolConfig protocolConfig, List<URL> r
 
         Map<String, String> map = new HashMap<String, String>();
         map.put(Constants.SIDE_KEY, Constants.PROVIDER_SIDE);
+
         appendRuntimeParameters(map);
+        appendParameters(map, metrics, Constants.METRICS_KEY);
```
chickenlj: Constants.METRICS_KEY should be removed.

chickenlj:
> the metrics stuff can be stored in metadata center

Seems to be the only way at present.

Issue #3802, opened 2019-04-03
当机器同时存在内网和外网IP时，无法发布内网IP服务


nevin9939:
### Environment

* Dubbo version: 2.7.1
* Operating System version: xxx
* Java version: 1.8.181

### Steps to reproduce this issue

1. 服务都是部署在阿里云的ECS上
2. hosts中修改机器名执行本机的内网IP
3. dubbo注册到zookeeper的服务IP是外网IP
4. 原整个环境的dubbo服务均提供内网服务，禁止外网的服务和请求进入，造成服务调不通

查看源代码发现在 request #3520 中调整为优先获取外网IP，造成了当机器同时存在内网IP和外网IP时，期望只提供内网IP服务时就无法做到，个人认为最好有个配置开关，有用户决定是提供内网服务还是外网服务

ralf0131:
+1 to you proposal, would you like to send a pull request to fix this?

nevin9939:
I can try it, not a lot of spare time.

lexburner:
@nevin9939 I send a pull request, could you pls review it and try it to see whether meet your requirement?  Thanks.

nevin9939:
I also finished coding  (*T_T*) . Looked at the difference between me and you mainly in the choice of preferred network card, synchronized with the operating system, this can prevent priority access to the virtual machine and Bluetooth bridge network card. 
```java
    private static InetAddress getLocalAddress0() {
        InetAddress candidateAddress = null;
        int interfacesIndex = 0;
        try {
            Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
            if (null != interfaces) {
                while (interfaces.hasMoreElements()) {
                    try {
                        NetworkInterface network = interfaces.nextElement();
                        // Remove loopback interface, subinterface, not running interface
                        if (network.isLoopback() || network.isVirtual() || !network.isUp()) {
                            continue;
                        }
                        Enumeration<InetAddress> addresses = network.getInetAddresses();
                        while (addresses.hasMoreElements()) {
                            try {
                                Optional<InetAddress> addressOp = toValidAddress(addresses.nextElement());
                                if (addressOp.isPresent()) {
                                    InetAddress tempAddress = addressOp.get();
                                    if (candidateAddress == null || interfacesIndex > network.getIndex()) {
                                        // Priority is given to the network card with a small index value,
                                        // which can exclude the IP of the self-built virtual machine.
                                        candidateAddress = tempAddress;
[...]

PR #3520, opened 2019-02-20
fix issue#538: 关于dubbo中获取本机IP的问题

beiwei30:
## What is the purpose of the change

fix issue #538: 关于dubbo中获取本机IP的问题

also allow register local ip to registry from both consumer and provider for better development experience, by default NetUtil#getLocalAddress will return a possible public IP.

## Brief changelog

XXXXX

## Verifying this change

XXXXX

Follow this checklist to help us incorporate your contribution quickly and easily:

- [x] Make sure there is a [GITHUB_issue](https://github.com/apache/incubator-dubbo/issues) field for the change (usually before you start working on it). Trivial changes like typos do not require a GITHUB issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [ ] Format the pull request title like `[Dubbo-XXX] Fix UnknownException when host config not exist #XXX`. Each commit in the pull request should have a meaningful subject line and body.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/apache/incubator-dubbo/tree/master/dubbo-test).
- [ ] Run `mvn clean install -DskipTests=false` & `mvn clean test-compile failsafe:integration-test` to make sure unit-test and integration-test pass.
- [ ] If this contribution is large, please follow the [Software Donation Guide](https://github.com/apache/incubator-dubbo/wiki/Software-donation-guide).

PR #3834, opened 2019-04-09
add a switch to control whether to prefer to register public address or site local address

lexburner:

fix the issue: https://github.com/apache/incubator-dubbo/issues/3802
improve the existing changes: https://github.com/apache/incubator-dubbo/pull/3520

## Brief changelog

- add a switch `dubbo.prefer.public.address`.
- change the strategy to achieve IP address.

steps to achieve IP:

1. judge the switch `dubbo.prefer.public.address`
2. search from hosts mapping using `InetAddress.getLocalHost()`
3. search from the network cards using `NetworkInterface.getNetworkInterfaces()`
4. if the address achieved in 2 & 3 not match the `dubbo.prefer.public.address`, downgrade to the reverse property of `dubbo.prefer.public.address`. In words, if no public address found, downgrade to site-local address; if no site-local address found, downgrade to public address;
5. if no address found, using loopback address 127.0.0.1.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
@@ -860,6 +860,8 @@ public class Constants {
 
     public static final int DEFAULT_GRPC_QUEUES = 300_0000;
 
+    public static final String PREFER_PUBLIC_ADDRESS = "dubbo.prefer.public.address";
+
     /**
      * metrics
      */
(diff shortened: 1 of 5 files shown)


ralf0131:
Hi, sorry but could you resolve the conflicts first?

lexburner:
@beiwei30 Conflicts are resolved, the logic rollbacks to reversion and adds the switch to control the public IP prefer.

lexburner:
I notice that the file style of `NetUtils` changes from CRLF to LF. This causes a miss-comparison between new and reversions. Which one of CRLF and LF prefers in Dubbo? or maybe it does not take any difference.

beiwei30:
> I notice that the file style of `NetUtils` changes from CRLF to LF. This causes a miss-comparison between new and reversions. Which one of CRLF and LF prefers in Dubbo? or maybe it does not take any difference.

make sure you set `autocrlf = false` in your gitconfig to avoid it.

initpassion:
最新的代码中并没有找到对应的开关和源码信息, 请问下高版本有保留该代码吗, 我使用的是2.7.15

PR #3862, opened 2019-04-12
bugfix/remove-parse-error-log

lexburner:

there is no need for any log in DubboBeanDefinitionParser since some class implements the interface: EnvironmentAware, ApplicationAware, etc. They only have setter methods, otherwise will cause the error log during application startup.

## Brief changelog

remove the error log


CrazyHZM:
I also submitted this part of the fix before, then I closed my submission. #3844

Issue #3952, opened 2019-04-29
Dubbo 2.7.1 ： delay export function doesn't work...


spiritme1984:
I set 'delay' label in dubbo config file as below:  

```xml
  <dubbo:provider timeout="${dubbo.timeout}" retries="${dubbo.retry}" accesslog="${dubbo.accesslog}" delay="30000" filter="dubboSofaTracrFilter"/> 
```

But when I start the project come across an error:
```
2019-04-29 17:12:21.525 [crm7] [ERROR] [,] [ContainerBackgroundProcessor[StandardEngine[Catalina]]] org.springframework.web.context.ContextLoader.initWebApplicationContext(ContextLoader.java:331) Context initialization failed
java.lang.NullPointerException
	at org.apache.dubbo.config.ServiceConfig.export(ServiceConfig.java:335)
	at org.apache.dubbo.config.spring.ServiceBean.export(ServiceBean.java:319)
	at org.apache.dubbo.config.spring.ServiceBean.onApplicationEvent(ServiceBean.java:113)
	at org.apache.dubbo.config.spring.ServiceBean.onApplicationEvent(ServiceBean.java:59)
        ....
```
And I debug the code and found in ServiceConfig.java, there is a method named 'export':
```java
    public synchronized void export() {
        checkAndUpdateSubConfigs();
        if (!shouldExport()) {
            return;
        }
        if (shouldDelay()) {
            delayExportExecutor.schedule(this::doExport, delay, TimeUnit.MILLISECONDS);
        } else {
            doExport();
        }
    }
```
In this method the 'delay' param is null , and I can't found  codes where set value for it.
Please correct me if I was wrong, thank you.

Leishunyu:
I have fixed it in this pr
https://github.com/apache/incubator-dubbo/pull/3957

ralf0131:
This bug has been fixed from 2.7.2 onwards.

PR #3957, opened 2019-04-30
fixed provider set delay NullPointerException

Leishunyu:

https://github.com/apache/incubator-dubbo/issues/3952

fixed this bug

## Brief changelog

provider's delay has not set in serviceconfig 

## Verifying this change

set provider's delay in  serviceconfig 

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
@@ -360,6 +360,7 @@ public class ServiceConfig<T> extends AbstractServiceConfig {
         if (delay == null && provider != null) {
             delay = provider.getDelay();
         }
+        this.delay = delay;
         return delay != null && delay > 0;
     }
 
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
@@ -335,10 +335,22 @@ public class ServiceConfig<T> extends AbstractServiceConfig {
         }
 
         if (shouldDelay()) {
-            delayExportExecutor.schedule(this::doExport, delay, TimeUnit.MILLISECONDS);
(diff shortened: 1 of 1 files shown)


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -360,6 +360,7 @@ private boolean shouldDelay() {
         if (delay == null && provider != null) {
             delay = provider.getDelay();
         }
+        this.delay = delay;
```
ralf0131: Hi, -1 to this change.

`shouldDelay` is a method that appears read-only. Modifying value within this method does not match with how this method should work.

I'd suggest to get the actual value of delay in `org.apache.dubbo.config.ServiceConfig#export`, or abstract it into a new method.

Leishunyu: ok , i change it

Leishunyu: i abstract it into a new method , can you help me see see

beiwei30: don't do this, the original change is good enough. The second change is unnecessary and make the code more complex.

beiwei30: pls. simply revert 95b394a6033e5a787e70fbdb1a3dd3b0dbc0aeb4. I will follow up.

beiwei30:
LGTM. I will follow up to make it better, since I also need to take care *shouldExport* too. Thanks.

ralf0131:
> LGTM. I will follow up to make it better, since I also need to take care _shouldExport_ too. Thanks.

This change does not look good to me. Why it is getting merged?
How would you planning to improve it?

beiwei30:
@ralf0131 @Leishunyu pls. review https://github.com/apache/incubator-dubbo/pull/3959

ralf0131:
The pull request looks good to me.
IMO, it would be better to let @Leishunyu to improve it.

Issue #3926, opened 2019-04-24
Serialization ContentTypeId conflict between avro protocol and protocoluff protocol


zhangfei9734:
- [x] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.2-SNAPSHOT
* Operating System version: OS X 10.14.4
* Java version: 1.8.0_202

### Steps to reproduce this issue

1. [DUBBO] Serialization extension org.apache.dubbo.common.serialize.protostuff.ProtostuffSerialization has duplicate id to Serialization extension org.apache.dubbo.common.serialize.avro.AvroSerialization, ignore this Serialization extension

```java
CodecSupport
static {
        Set<String> supportedExtensions = ExtensionLoader.getExtensionLoader(Serialization.class).getSupportedExtensions();
        for (String name : supportedExtensions) {
            Serialization serialization = ExtensionLoader.getExtensionLoader(Serialization.class).getExtension(name);
            byte idByte = serialization.getContentTypeId();
            if (ID_SERIALIZATION_MAP.containsKey(idByte)) {
                logger.error("Serialization extension " + serialization.getClass().getName()
                        + " has duplicate id to Serialization extension "
                        + ID_SERIALIZATION_MAP.get(idByte).getClass().getName()
                        + ", ignore this Serialization extension");
                continue;
            }
            ID_SERIALIZATION_MAP.put(idByte, serialization);
            ID_SERIALIZATIONNAME_MAP.put(idByte, name);
        }
    }

public class ProtostuffSerialization implements Serialization {
    @Override
    public byte getContentTypeId() {
        return 10;
    }
}

public class AvroSerialization implements Serialization {
    @Override
    public byte getContentTypeId() {
        return 10;
    }
}
```

zhangfei9734:
https://github.com/apache/incubator-dubbo/blob/9b11b7ac89c0d897220bc9b42b3e5e9c3eb708f2/dubbo-serialization/dubbo-serialization-avro/src/main/java/org/apache/dubbo/common/serialize/avro/AvroSerialization.java#L32
[...]

PR #3967, opened 2019-05-01
Fix ContentTypeId conflict

CrazyHZM:

Fix Serialization ContentTypeId conflict between avro protocol and protocoluff protocol #3926

## Brief changelog

XXXXX

## Verifying this change

XXXXX


dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -875,4 +875,18 @@
     public static final String METRICS_PROTOCOL = "metrics.protocol";
 
 
+    /**
+     * Serizlization ContentTypeId
+     */
+    public static final byte AVRO_SERIALIZATION_ID = 1;
```
beiwei30: this is wrong, avro is 10...

beiwei30: *public static final byte AVRO_SERIALIZATION_ID = 11* should be better

CrazyHZM: > _public static final byte AVRO_SERIALIZATION_ID = 11_ should be better

What are the uses of 1 and 5 here?

beiwei30: let's simply leave it as is. I believe the original intention is: 1, 2, 4, 8, 16...

CrazyHZM: I have modified it.

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -875,4 +875,18 @@
...
+    public static final byte JAVA_SERIALIZATION_ID = 3;
+    public static final byte COMPACTED_JAVA_SERIALIZATION_ID = 4;
+    public static final byte FASTJSON_SERIALIZATION_ID = 6;
+    public static final byte NATIVE_JAVA_SERIALIZATION_ID = 7;
+    public static final byte KRYO_SERIALIZATION_ID = 8;
+    public static final byte FST_SERIALIZATION_ID = 9;
+    public static final byte PROTOSTUFF_SERIALIZATION_ID = 10;
+    public static final byte AVRO_SERIALIZATION_ID = 11;
+    public static final byte GSON_SERIALIZATION_ID = 16;
+
```
cvictory: Maybe you can define a sub-constant class. What do you think?

satansk: +1 for a specific constants class

CrazyHZM: I think this suggestion is good, I will modify it later.

Issue #3137, opened 2019-01-04
[suggestion] Seperate Constants.java into some SubConstants Class


cvictory:
Currently we manage out constants values in org.apache.dubbo.common.Constants.java.  

This class is very bloated and this structure is not well enough.  Maybe we should categorize such constants into groups. In this way , we can add some XxxConstants, such as RegistryConstants,ConfigCenterConstants etc.   It'd be better if it is split by module of dubbo-core.

CrazyHZM:
@cvictory  
Is there a plan for this suggestion? If there is no plan, I plan to start doing this in the near future.

beiwei30:
@CrazyHZM pls. take a look at #4017. I've finished step1.

- step2, we need to create multiple **Constants** in dubbo-common module.
- step3, start to reference new **Constatns** one by one.
- step4, delete old **Constants**
- step5, examine each new **Constants** and evaluate if they are more suitable to move out from dubbo-common into its delegated module, say: org.apache.dubbo.common.serialize.Constants to dubbo-serialization-api

CrazyHZM:
@beiwei30 
I think this step is good, I am responsible for doing the remoting, rpc, config parts, I will start with remoting

beiwei30:
will you do it based on the current pull request, or wait until it's merged?

CrazyHZM:
Wait until after being merged

beiwei30:
@CrazyHZM #4017 is merged. pls. start to move constants into **constants** sub package first. In this step, let's don't touch org.apache.dubbo.common.Constants **untouched**.

beiwei30:
FYI, the first new constant class I created is: org.apache.dubbo.common.constants.CommonConstants

CrazyHZM:
I created 
org.apache.dubbo.common.constants.RemotingConstants、
org.apache.dubbo.common.constants.RpcConstants and org.apache.dubbo.common.constants.ConfigConstants.

beiwei30:
@CrazyHZM see #4019 4020

CrazyHZM:
@beiwei30 
step2 #4021

beiwei30:
> @beiwei30
> step2 #4021

I left my comment. Would you mind to review mine too?

CrazyHZM:
> @CrazyHZM see #4020

I will merge it when I pass ci

beiwei30:
I think we can now move to step3.

CrazyHZM:
I will start modifying the remoting later.

[...]

PR #4017, opened 2019-05-10
[DUBBO-3137]: Seperate Constants.java into some SubConstants Class - step 1

beiwei30:

issue #3137: step 1

## Brief changelog

XXXXX

## Verifying this change

XXXXX

- [ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add sample in [dubbo samples](https://github.com/apache/incubator-dubbo-samples) project.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
@@ -25,122 +25,50 @@ import java.util.regex.Pattern;
  */
 public class Constants {
 
+    // BEGIN common
     public static final String DUBBO = "dubbo";
 
     public static final String PROVIDER = "provider";
 
     public static final String CONSUMER = "consumer";
 
-    public static final String REGISTER = "register";
+    public static final String APPLICATION_KEY = "application";
 
-    public static final String UNREGISTER = "unregister";
-
-    public static final String SUBSCRIBE = "subscribe";
-
-    public static final String UNSUBSCRIBE = "unsubscribe";
-
-    public static final String CATEGORY_KEY = "category";
-
-    public static final String PROVIDERS_CATEGORY = "providers";
-
-    public static final String CONSUMERS_CATEGORY = "consumers";
-
-    public static final String ROUTERS_CATEGORY = "routers";
-
-    public static final String DYNAMIC_ROUTERS_CATEGORY = "dynamicrouters";
-
-    public static final String CONFIGURATORS_CATEGORY = "configurators";
-
-    public static final String DYNAMIC_CONFIGURATORS_CATEGORY = "dynamicconfigurators";
-    public static final String APP_DYNAMIC_CONFIGURATORS_CATEGORY = "appdynamicconfigurators";
-
-    public static final String CONFIGURATORS_SUFFIX = ".configurators";
-
-    public static final String ROUTERS_SUFFIX = ".routers";
-
-    public static final String CONFIG_CLUSTER_KEY = "config.cluster";
-    public static final String CONFIG_NAMESPACE_KEY = "config.namespace";
-    public static final String CONFIG_GROUP_KEY = "config.group";
-    public static final String CONFIG_CHECK_KEY = "config.check";
-    public static final String CONFIG_CONFIGFILE_KEY = "config.config-file";
-    public static final String CONFIG_ENABLE_KEY = "config.highest-priority";
-    public static final String CONFIG_TIMEOUT_KEY = "config.timeout";
-    public static final String CONFIG_APPNAME_KEY = "config.app-name";
-
-    public static final String DEFAULT_CATEGORY = PROVIDERS_CATEGORY;
+    public static final String REMOTE_APPLICATION_KEY = "remote.application";
 
     public static final String ENABLED_KEY = "enabled";
 
     public static final String DISABLED_KEY = "disabled";
 
-    public static final String VALIDATION_KEY = "validation";
-
-    public static final String CACHE_KEY = "cache";
-
-    public static final String DYNAMIC_KEY = "dynamic";
-
-    public static final String STATUS_KEY = "status";
-
-    public static final String CONTEXTPATH_KEY = "contextpath";
-
-    public static final String LISTENER_KEY = "listener";
-
-    public static final String LAYER_KEY = "layer";
-
     public static final String DUBBO_PROPERTIES_KEY = "dubbo.properties.file";
 
     public static final String DEFAULT_DUBBO_PROPERTIES = "dubbo.properties";
 
-    public static final String SENT_KEY = "sent";
+    public static final String ANY_VALUE = "*";
 
-    public static final boolean DEFAULT_SENT = false;
+    public static final String COMMA_SEPARATOR = ",";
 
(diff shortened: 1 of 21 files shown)


dubbo-serialization/dubbo-serialization-api/src/main/java/org/apache/dubbo/common/serialize/Constants.java
```diff
@@ -0,0 +1,16 @@
+package org.apache.dubbo.common.serialize;
+
+public interface Constants {
```
satansk: I have a doubt here, why put them in an interface, instead of a class, is there some specific consideration?

beiwei30: an interface is perfect place to introduce constants.

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -25,305 +25,256 @@
...
+    public static final String ACCEPTS_KEY = "accepts";
 
-    /**
-     * key for router type, for e.g., "script"/"file",  corresponding to ScriptRouterFactory.NAME, FileRouterFactory.NAME
-     */
-    public static final String ROUTER_KEY = "router";
+    public static final int DEFAULT_ACCEPTS = 0;
 
-    public static final String CLUSTER_KEY = "cluster";
+    public static final String SERIALIZATION_KEY = "serialization";
```
CrazyHZM: I think serialization can also be used as a constant alone. What do you think?

I think that a completely independent constant can be placed in its own module similar to the interface you added.
But constants that will be used by other modules can add SubConstants to the common module.
For example:
SERIALIZATION_KEY is used by other modules, but it belongs to the SERIALIZATION module. We can manage it by creating SerializationConstants in the common module, but the Serialization ID will only be used in the Serialization module. We can create Constants for management in the Serialization module as you do. 
What do you think?

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -25,305 +25,256 @@
...
 
-    public static final String REGISTRY_KEY = "registry";
+    public static final String DEFAULT_REMOTING_SERIALIZATION = "hessian2";
```
CrazyHZM: The same as above.

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -332,272 +283,304 @@
...
+    public static final boolean DEFAULT_HESSIAN_OVERLOAD_METHOD = false;
 
-    public static final String SERVER_KEY = "server";
+    public static final String DEFAULT_HTTP_CLIENT = "jdk";
 
-    public static final String CLIENT_KEY = "client";
+    public static final String DEFAULT_HTTP_SERVER = "servlet";
 
-    public static final String ID_KEY = "id";
+    public static final String DEFAULT_HTTP_SERIALIZATION = "json";
```
CrazyHZM: This is not used anymore, it is recommended to delete

dubbo-serialization/dubbo-serialization-api/src/main/java/org/apache/dubbo/common/serialize/Constants.java
satansk: I have some other thoughts: an interface is a type represents behaviour, if classes accidently implement `Constants`, the constants will be exposed and will pollute the their namespace. 

Just a noninstantiable class will avoid these problems and is better in my opinion, what do you think?

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -332,272 +283,304 @@
...
 
-    public static final String MOCK_PROTOCOL = "mock";
+    public static final String EXTENSION_KEY = "extension";
+    // END dubbo-rpc-rest
 
-    public static final String RETURN_PREFIX = "return ";
 
-    public static final String THROW_PREFIX = "throw";
+    // BEGIN dubbo-config-api
+    public static final String CLUSTER_KEY = "cluster";
```
CrazyHZM: This suggestion is put on dubbo-cluster

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -610,263 +593,248 @@
...
-     */
-    public static final int DEFAULT_CALLBACK_INSTANCES = 1;
+    public static final String ZOOKEEPER_PROTOCOL = "zookeeper";
 
-    public static final String CALLBACK_SERVICE_PROXY_KEY = "callback.service.proxy";
+    // FIXME: is this still useful?
+    public static final String SHUTDOWN_TIMEOUT_KEY = "shutdown.timeout";
 
-    public static final String IS_CALLBACK_SERVICE = "is_callback_service";
+    public static final int DEFAULT_SHUTDOWN_TIMEOUT = 1000 * 60 * 15;
```
CrazyHZM: This is not used, it is recommended to delete

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
```diff
@@ -877,20 +845,52 @@
...
+    // BEGIN dubbo-filter-cache
+    public static final String CACHE_KEY = "cache";
+    // END dubbo-filter-cache
+
+
+    // BEGIN dubbo-filter-validation
+    public static final String VALIDATION_KEY = "validation";
+    // END dubbo-filter-validation
+
+    public static final String DEFAULT_CLUSTER = "failover";
```
CrazyHZM: This is not used, it is recommended to delete, or put it in dubbo-cluster

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
beiwei30: let's do it later.

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
beiwei30: hesitate to move it into dubbo-cluster. It's only used in dubbo-config-api.

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
beiwei30: let's leave it as is, for now.

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
beiwei30: that's what I am thinking too. This is only step1, in step2, we will move them into separated classes, then in the later steps, we will evaluate if it is suitable to move into other module.

dubbo-common/src/main/java/org/apache/dubbo/common/Constants.java
beiwei30: let's do it later. will not do it in step1.

PR #4024, opened 2019-05-10
Fix Invocation attributes

qinliujie:

fix RpcInvocation can't pass attributes

## Brief changelog

change RpcInvocation and GenericFilter

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcInvocation.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcInvocation.java
@@ -91,9 +91,10 @@ public class RpcInvocation implements Invocation, Serializable {
         this(method.getName(), method.getParameterTypes(), arguments, null, null);
     }
 
-    public RpcInvocation(Method method, Object[] arguments, Map<String, String> attachment) {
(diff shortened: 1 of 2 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcInvocation.java
```diff
@@ -91,9 +91,10 @@ public RpcInvocation(Method method, Object[] arguments) {
         this(method.getName(), method.getParameterTypes(), arguments, null, null);
     }
 
-    public RpcInvocation(Method method, Object[] arguments, Map<String, String> attachment) {
+    public RpcInvocation(Method method, Object[] arguments, Map<String, String> attachment, Map<Object, Object> attributes) {
         this(method.getName(), method.getParameterTypes(), arguments, attachment, null);
         this.returnType = method.getReturnType();
+        this.attributes = attributes == null ? new HashMap<>() : attributes;
```
chickenlj: I am not sure if `attributes` will be passed downstream by the framework automatically.

I remember RpcInvocation is not encoded as a whole:
https://github.com/apache/incubator-dubbo/blob/ebe3e424579d5986bf4cbb518ca6a4ff28f9a257/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboCodec.java#L187

PR #4040, opened 2019-05-13
Shorten the life cycle of Timeout to get a better gc effect.

carryxyh:
Recently, some pressure tests were performed on the rpc inside the music163. I found that YGC did not perform well during the pressure measurement process. One of the reasons is that the lifetime of `TimeoutCheckTask` is relatively long (currently it will survive `timeout`).

I try to shorten its life cycle to get better GC performance. It seems that my modification is successful. Now `TimeoutCheckTask` will be removed when the request returns.

The purpose of this pr is to submit this optimization item to dubbo.

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
@@ -54,6 +54,8 @@ public class DefaultFuture implements ResponseFuture {
 
     private static final Map<Long, DefaultFuture> FUTURES = new ConcurrentHashMap<>();
 
+    private static final Map<Long, Timeout> PENDING_TASKS = new ConcurrentHashMap<>();
+
     public static final Timer TIME_OUT_TIMER = new HashedWheelTimer(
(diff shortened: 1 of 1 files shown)


chickenlj:
LGTM.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
```diff
@@ -86,7 +88,8 @@ private DefaultFuture(Channel channel, Request request, int timeout) {
      */
     private static void timeoutCheck(DefaultFuture future) {
         TimeoutCheckTask task = new TimeoutCheckTask(future);
-        TIME_OUT_TIMER.newTimeout(task, future.getTimeout(), TimeUnit.MILLISECONDS);
+        Timeout t = TIME_OUT_TIMER.newTimeout(task, future.getTimeout(), TimeUnit.MILLISECONDS);
+        PENDING_TASKS.put(future.getId(), t);
```
chickenlj: Can we avoid this `PENDING_TASKS` instance? I think `DefaultFuture` itself hosts the status of its timeouttask or `TIME_OUT_TIMER ` hosts the status of all timeout tasks can be better.

I need to merge this PR first to do some pressure test, please help to improve if possible.

PR #4076, opened 2019-05-16
optimize generic invoke

nzomkxia:
add a new generic type "raw.return" to reduce  unnecessary result convert. effective for hessian serialization only


chickenlj:
I think we need a precedent release of hessian-lite, right?

nzomkxia:
> I think we need a precedent release of hessian-lite, right?

hessian-lite 3.2.6 has been released

chickenlj:
https://sq.163yun.com/blog/article/185512233177817088

chickenlj:
This optimization only works for hessian2 serialization: 
1. Generic server serializes pojo directly into the pipeline (contains class description info, save the time of converting pojo to map and then serialize map).
2. Generic client deserializes raw bytes, hessian will fail not recognizing the class declared in the data and failover to MapDeserializer.

PR #4085, opened 2019-05-17
performance improve for DefaultFuture and AsyncToSyncInvoker

Jeff-Lv:

when I do stress testing, I found that the latest of version 3.0-dev with low performance for qps and rt, actually very impressive low compare to 2.7.1.

## Brief changelog

the main problem is about those two classes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ target/
 *.zip
 *.tar
 *.tar.gz
+.flattened-pom.xml
 
 # eclipse ignore
 .settings/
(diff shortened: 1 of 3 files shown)


carryxyh:
个人感觉，这两个pr合并在一起效果是最好的：
https://github.com/apache/incubator-dubbo/pull/4087

chickenlj:
> 个人感觉，这两个pr合并在一起效果是最好的：
#4087

Agree with you. I will merge your patch manually later.

Issue #4107, opened 2019-05-22
#PR:4109 Reference create proxy faild and retry create may lead to OOM


vio-lin:
- [x] I have searched the [issues](https://github.com/apache/incubator-dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/incubator-dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.0
* Operating System version: Windows7
* Java version: jdk8

### Steps to reproduce this issue

1. stop dubbo service
2. dubbo client set checked = true.  do get cudbbo client for times.
3. the size of list field [urls] in ReferenceConfig wil increase

Pls. provide [GitHub address] to reproduce this issue.
(project)[https://github.com/vio-lin/incubator-dubbo/tree/error-situation]  
run java file 
 incubator-dubbo\dubbo-demo\dubbo-demo-xml\dubbo-demo-xml-consumer\src\main\java\org\apache\dubbo\demo\consumer\Application.java
### Expected Result
the [urls] in ReferenceConfig will be updated.

### Actual Result
the avalidable registery urls be add to [urls] in ReferenceConfig

PR #4109, opened 2019-05-22
[Dubbo-2.7.2] fix registery urls increase forever when recreate reference proxy

vio-lin:
fix issue #4107 
when client retry init reference, clear registry urls to avoid OOM.
the registry url contains timestamp value, it can not change urls from list to set. 
## Brief changelog
do clear urls when in create remote proxy 

## Verifying this change
do clear urls when create remote proxy

- [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add sample in [dubbo samples](https://github.com/apache/incubator-dubbo-samples) project.

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ReferenceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ReferenceConfig.java
@@ -356,6 +356,7 @@ public class ReferenceConfig<T> extends AbstractReferenceConfig {
                 logger.info("Using injvm service " + interfaceClass.getName());
             }
         } else {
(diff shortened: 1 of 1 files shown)


vio-lin:
this urls 

> Hi, this line will change the logic of urls, we will lose contents of `urls` every time calling createProxy method . I believe it's not a good way to prevent infinite retries and OOM.
> 
> How do u think?
I think this createProxy() is a method to get a new client. It happended when last getProxy() failed and 
initialized = false. It isn't nessary to keep the old,failed urls of reference , when we getProxy next time. 
As the description of urls `The url of the reference service`，we should keep the last and success urls.

PR #4177, opened 2019-05-27
RpcContext删除remoteApplicationName字段，新增consumerApplicationName字段

manzhizhen:
1.The name of RpcContext#remoteApplicationName is not very accurate, it will make people think that it is the application of the provider when the consumer calls the provider.
2.The application name of the consumer is no longer passed through REMOTE_APPLICATION_KEY, because REMOTE_APPLICATION_KEY is used in the routing label, referring to the application name of the provider, using the new field CONSUMER_APPLICATION_KEY

1.之前RpcContext#remoteApplicationName的名称不太准确，会让人误以为是consumer调用provider时，provider的application。所以优化成RpcContext#consumerApplicationName。
2.不再通过REMOTE_APPLICATION_KEY来传递consumer的application name，因为REMOTE_APPLICATION_KEY在路由标签中有使用，指的是provider的application name，改用新的字段CONSUMER_APPLICATION_KEY


htynkn:
I think field `remoteApplicationName` is introduced recently and haven't release, if we agree to modify it from remoteApplicationName to `consumerApplicationName`, it's better to include in 2.7.2 release

@cvictory could you have a look at this one?

manzhizhen:
> I think field `remoteApplicationName` is introduced recently and haven't release, if we agree to modify it from remoteApplicationName to `consumerApplicationName`, it's better to include in 2.7.2 release
> 
> @cvictory could you have a look at this one?

Thanks for reminding, I have already returned the mail.

AlbumenJ:
Currently `remoteApplicationName` has been publish for long time. 
I think it might not be resonable to change it while `remoteApplicationName` works fine.

PR #4187, opened 2019-05-28
2.7.2 release note

cvictory:


chickenlj:
* Introduce `CompletableFuture<Object> $invokeAsync` for `GenericService`, now, for generic call, you can use:
   1. `$invoke` for sync method call with normal return type.
   2. `$invokeAsync` for async method call with `CompletableFuture<T>` signature.

* bugfix and enhancement  https://github.com/apache/dubbo/issues/3287

PR #4195, opened 2019-05-29
[WIP]Collect "cpu.user" metric

danielamorais:

Collect "cpu.user" for load balancer statistics
1. Creates a callback service: A thread inside this class is responsible for collecting CPU Usage. When the CPU Usage changes, the listener will receive a notification. 
2. Create unit test

## Verifying this change

Run CpuUsageServiceImplTest.java 


ralf0131:
UT failed, the reason is:
```
[ERROR] src/main/java/org/apache/dubbo/monitor/dubbo/CpuUsageGauge.java:[4,8] (imports) UnusedImports: Unused import - com.alibaba.metrics.Clock.
```

dubbo-monitor/dubbo-monitor-default/src/main/java/org/apache/dubbo/monitor/dubbo/MetricsFilter.java
```diff
@@ -123,6 +127,26 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
         }
     }
 
+    private void reportLbMetrics(Result result, Invoker<?> invoker, boolean isProvider) {
```
ralf0131: I think the logic of this method should be done in a dedicated thread, which will periodically update the cpu usage if necessary. 
What could be done here is just get the cpu.user data and put it into attachment.
I also think it is no necessary to create CpuUsageGauge, just use com.alibaba.metrics.os.linux.CpuUsageGaugeSet, it has done almost everything for you. You just need to make sure the com.alibaba.metrics.CachedMetricSet#refreshIfNecessary is called periodically so that you can always get the latest data.

ralf0131:
Hi,

To send data from provider to consumer, there are basically two ways. 

1) call setAttachment on invocation. The problem is that how do you know the consumer is needed or not, a solution will be if the consumer asks provider to send the data, then provide can send the data. But if consumer do not ask, you still have to start a thread to calculate the data, which is unnecsseary.

2) use server-side callback, the consumer add a listerner remotely to provider, in this call the consumer let proivder to know that consumer asks for server side data, the server then start a thread to periodically check the latest cpu, once there is a update, send cpu.user to the consumer use callback. see http://dubbo.apache.org/en-us/docs/user/demos/callback-parameter.html for more details.

Comparing with the 2 options, I think server-side call back option is better in terms of it enables the funcationality to collect server side data on demand, and the update will be sent to consumer immediately if there is a change.

dubbo-monitor/dubbo-monitor-default/src/test/java/org/apache/dubbo/monitor/dubbo/MetricsFilterTest.java
```diff
@@ -56,32 +59,32 @@
...
-        @Override
-        public Class<DemoService> getInterface() {
-            return DemoService.class;
-        }
+    private Invoker<DemoService> serviceInvoker;
 
-        public URL getUrl() {
-            return URL.valueOf("dubbo://" + NetUtils.getLocalHost() + ":20880/org.apache.dubbo.monitor.dubbo.service.DemoService");
-        }
+    @BeforeEach
```
danielamorais: @ralf0131 In the first PR, I made these changes to make it easier to create new unit tests (is using Mockito).

Should I remove this modification or open a new PR?

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/StatisticsLoadBalance.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class StatisticsLoadBalance extends AbstractLoadBalance {
+
+    private Map<String, Float> cpuUsage = new HashMap<>();
+
+    public StatisticsLoadBalance() {
+        CpuUsageService cpuUsageService = new CpuUsageServiceImpl(100L, 500L);
```
danielamorais: @ralf0131 Should I add these parameters as properties that could change in this PR?

dubbo-monitor/dubbo-monitor-default/src/main/java/org/apache/dubbo/monitor/dubbo/CpuUsageServiceImpl.java
```diff
@@ -0,0 +1,42 @@
+package org.apache.dubbo.monitor.dubbo;
```
danielamorais: I'm not sure if should be at package org.apache.dubbo.monitor.dubbo; or move to dubbo-cluster

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/StatisticsLoadBalance.java
ralf0131: Hi, I think they should be configurable.

dubbo-monitor/dubbo-monitor-default/src/main/java/org/apache/dubbo/monitor/dubbo/CpuUsageServiceImpl.java
ralf0131: Hi, I think put it under dubbo-cluster is ok.  Maybe you can create a new package called `org.apache.dubbo.rpc.cluster.loadbalance.statistics`, and put it there.

dubbo-monitor/dubbo-monitor-default/src/test/java/org/apache/dubbo/monitor/dubbo/MetricsFilterTest.java
ralf0131: If it is un-related, it would be better to create a separate one.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/StatisticsLoadBalance.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.dubbo.rpc.Invocation;
+import org.apache.dubbo.rpc.Invoker;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class StatisticsLoadBalance extends AbstractLoadBalance {
+
+    private Map<String, Float> cpuUsage = new HashMap<>();
```
guohao: There could be multi-threading invocations.
Choose an another thread-safe collection instead of HashMap.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/StatisticsLoadBalance.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public StatisticsLoadBalance() {
+        long timeToLive = Long.parseLong(ConfigurationUtils.getProperty("time.to.live"));
+        long collectCpuUsageInMill = Long.parseLong(ConfigurationUtils.getProperty("mill.to.collect.cpu.usage"));
+        CpuUsageService cpuUsageService = new CpuUsageServiceImpl(timeToLive, collectCpuUsageInMill);
```
guohao: CpuUsageService should be created by dubbo as  a remote stub instead of 'new' directly.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/statistics/CpuUsageListener.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.dubbo.rpc.cluster.loadbalance.statistics;
+
+public interface CpuUsageListener {
+    void cpuChanged(String ip, Float cpu);
```
guohao: Consider a scenario that two dubbo application running on different nodes but share  same IP( NAT ). 
Which cpu usage info will be used  here?  How to deal with this problem ?

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/StatisticsLoadBalance.java
danielamorais: What do you think about using SPI mechanism? (http://dubbo.apache.org/en-us/blog/introduction-to-dubbo-spi.html)

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/statistics/CpuUsageListener.java
danielamorais: Good catch, thanks. I think it should use IP and the port to avoid this.

danielamorais: @ralf0131 What do you think?

ralf0131: Hi, 

ip + port should be enough for this case.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/StatisticsLoadBalance.java
ralf0131: Hi, 

I did some local test, the proposed pull request could not work correctly. If you start two provider and a consumer, you can see that cpu usage is not being sent from provider to consumer.

I agree with @guohao that, we need to call the CpuUsageService remotely to add listener rather that `new` directly, and if there are multiple providers, we have to call the addListener method for each provider. 

The problem is the StatisticsLoadBalance is a bean managed by Spring, so it is impossible to get the provider ip addresses when initializing StatisticsLoadBalance class. So we have to delay the call the addListener to the `doSelect` method.

To obtain a proxy for remote service, you may checkout `org.apache.dubbo.monitor.dubbo.DubboMonitorFactory#createMonitor`, there are some example code.

Since it is lazy to obtain the cpu usage, the load balance may not working very well during the initial period of time (lack of cpu usage stat), so there should be some fall back strategies, once the cpu stat are collected, the stat based load balance can be activated.

ralf0131:
I did some further investigations. In order to establish connection between provider and consumer, we have to do several things:

* On the provider side: declare `CpuUsageService` on provider side, which expose the service to consumers.

```xml
    <dubbo:service interface="org.apache.dubbo.rpc.cluster.loadbalance.statistics.CpuUsageService" ref="cpuUsageService" connections="1" callbacks="1000">
        <dubbo:method name="addListener">
            <dubbo:argument index="1" callback="true" />
        </dubbo:method>
    </dubbo:service>
```

* On the consumer side, we have to programmably get the reference to `CpuUsageService`. Because in the dubbo-cluster module we cannot depend on a high level `ReferenceConfig` API in dubbo-config, so we have to use some low level APIs, which can be found at `org.apache.dubbo.monitor.dubbo.DubboMonitorFactory#createMonitor`.

The idea is that first randomly pick an invoker, and generate a proxy of `CpuUsageService` on that provider, then add the listener to that provider. Once we got enough stats, we can implement the load balance based on the stats.

```java

    @Override
    protected <T> Invoker<T> doSelect(List<Invoker<T>> invokers, URL url, Invocation invocation) {

        int index = (int)(invokers.size() * Math.random());
        Invoker<T> invoker = invokers.get(index);
        cpuUsage.computeIfAbsent(invoker.getUrl().getAddress(), value -> {
            URLBuilder urlBuilder = URLBuilder.from(invoker.getUrl());
            urlBuilder.setPath(CpuUsageService.class.getName());
            urlBuilder.addParameter(INTERFACE_KEY, CpuUsageService.class.getName());
            urlBuilder.addParameter("bean.name", CpuUsageService.class.getName());
            urlBuilder.addParameter(METHODS_KEY, "addListener,removeListener");
            urlBuilder.addParameters(CHECK_KEY, String.valueOf(false), REFERENCE_FILTER_KEY, "");
            urlBuilder.addParameter("addListener.1.callback", true);
            Invoker<CpuUsageService> cpuUsageInvoker = protocol.refer(CpuUsageService.class, urlBuilder.build());
            CpuUsageService cpuUsageService = proxyFactory.getProxy(cpuUsageInvoker);
            cpuUsageService.addListener("statisticsloadbalance", new CpuUsageListenerImpl());
            return 0.0f;
        });

        // TODO implement the load balance with statistics provided by the service provider.
        return invoker;
    }

    public void setProtocol(Protocol protocol) {
        this.protocol = protocol;
    }

    public void setProxyFactory(ProxyFactory proxyFactory) {
        this.proxyFactory = proxyFactory;
    }
```

Note that:

1. `protocol` and `proxyFactory` can be auto-wired, if you provider setters.
2. Not every parameters in URL matters, please explore it, however `addListener.1.callback` is important for the provider and consumer communication, it let them know that the second parameter of addListener method is a callback.
3. It is just a demo code, it is not mature, please improve it.
4. It better to use a dedicated class to implement `CpuUsageListener`, or there will be some issue during the serialization of provider-consumer communication.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/StatisticsLoadBalance.java
```diff
@@ -0,0 +1,69 @@
...
+        cpuUsageService.addListener("statisticsloadbalance", (ip, cpu) -> cpuUsage.put(ip, cpu));
+    }
+
+    @Override
+    protected <T> Invoker<T> doSelect(List<Invoker<T>> invokers, URL url, Invocation invocation) {
+        int index = (int)(invokers.size() * Math.random());
+        Invoker<T> invoker = invokers.get(index);
+        cpuUsage.computeIfAbsent(invoker.getUrl().getAddress(), value -> {
```
guohao: Listener will be  invalid when provider restart.  An address listener that  removes offline addresses may be needed.

danielamorais: Good catch, thanks.

danielamorais:
@ralf0131 

I made a test with 1 provider and 1 consumer. I found 2 problems in my tests:
1. After the consumer restarts, the provider doesn't collect CPU Usage anymore.
2. When is the provider side, RpcContext.getServerContext().getRemoteAddressString() returns "null:0". 

CPU usage data:
[ "null:0" -> value updated,  "some_ip:20880" -> 0.0]

* When the consumer starts,  a new invoker is added in Map<String, Float> cpuUsage. It is ok?

dubbo-consumer.xml

```xml
<dubbo:application name="demo-consumer"/>

<dubbo:registry address="multicast://224.5.6.7:1234"/>

<dubbo:protocol name="dubbo" port="20880"/>
 
<dubbo:reference id="demoService" check="false" interface="org.apache.dubbo.demo.DemoService" timeout="20000"/>
```

dubbo-provider.xml

```xml
<dubbo:application name="demo-provider"/>

<dubbo:registry address="multicast://224.5.6.7:1234" />

<dubbo:protocol name="dubbo"/>

<bean id="demoService" class="org.apache.dubbo.demo.provider.DemoServiceImpl"/>
<bean id="cpuUsageService" class="org.apache.dubbo.rpc.cluster.loadbalance.statistics.CpuUsageServiceImpl">
</bean>

<dubbo:service interface="org.apache.dubbo.demo.DemoService" ref="demoService" loadbalance="statistics"/>
<dubbo:service interface="org.apache.dubbo.rpc.cluster.loadbalance.statistics.CpuUsageService" ref="cpuUsageService" connections="1" callbacks="1000">
	<dubbo:method name="addListener">
	<dubbo:argument index="1" callback="true" />
	</dubbo:method>
</dubbo:service>
```

dubbo.properties (provider side)
```
dubbo.application.qos.port=22222
time.to.live=1000
mill.to.collect.cpu.usage=1000
```

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/statistics/DubboCpuUsageFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public CpuUsageService createCpuUsageService(URL url) {
+        URLBuilder urlBuilder = URLBuilder.from(url);
+        urlBuilder.setProtocol(url.getParameter(PROTOCOL_KEY, DUBBO_PROTOCOL));
+        urlBuilder.addParameters(CHECK_KEY, String.valueOf(false), REFERENCE_FILTER_KEY, "");
+        urlBuilder.addParameter("addListener.1.callback", true);
+        urlBuilder.setPath(CpuUsageService.class.getName());
+
+        Invoker<CpuUsageService> cpuUsageInvoker = protocol.refer(CpuUsageService.class, urlBuilder.build());
+        return proxyFactory.getProxy(cpuUsageInvoker);
```
danielamorais: @ralf0131 The class which implements CpuUsageService has a default constructor that schedule the thread to collect CPU Usage. 
Questions:
1) Should I call the default constructor? Does the proxyFactory do it?
2) In dubbo-provider.xml I added:
```xml
<bean id="cpuUsageService" class="org.apache.dubbo.rpc.cluster.loadbalance.statistics.CpuUsageServiceImpl"/>
```
Will Spring/Dubbo call the factory class (DubboCpuUsageService)?

ralf0131:
> I made a test with 1 provider and 1 consumer. 

You need to have at least 2 provider and 1 consumer in order to make a load balance to work. Please check the logic of `org.apache.dubbo.rpc.cluster.support.AbstractClusterInvoker#doSelect`

> I found 2 problems in my tests:

> After the consumer restarts, the provider doesn't collect CPU Usage anymore.

This is because you registered the listener to `org.apache.dubbo.rpc.cluster.loadbalance.statistics.CpuUsageServiceImpl#listeners`, but failed to remove it when consumer restart. 

When a provider failed to send cpu usage data to a consumer, it should throw an expcetion

> When is the provider side, RpcContext.getServerContext().getRemoteAddressString() returns "null:0".  

Because the thread has been switched when `org.apache.dubbo.rpc.cluster.loadbalance.statistics.CpuUsageServiceImpl#collectCpuUsage` is called, RpcContext.getServerContext() is a thread local object. If you switch thread you will lose it. You  can consider NetUtils.getLocalHost() instead.


There is a third issue, which is mentioned by @guohao , if the provider restart, the consumer should remove the statistics from the `org.apache.dubbo.rpc.cluster.loadbalance.StatisticsLoadBalance`

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/statistics/DubboCpuUsageFactory.java
ralf0131: Hi, why do you need to create CpuUsageFactory? I think right now it is ok to put the logic to `StatisticsLoadBalance` and refactor it later.

AlbumenJ:
@danielamorais hi, thanks for your contribution

Please merge the latest master branch to resolve conflict files.

AlbumenJ:

PR #4225, opened 2019-05-31
fix:make sure attribute exists

jcsxky:

make sure attribute exists

## Brief changelog

when dubug or call toString method of AbstractConfig,the field may not exist.For example, when getClass refer to ReferenceBean,it will call getObject() and init method in ReferenceConfig,when it is not initialized,interfaceName is null and throws an Exception, but initialized set to be true.this will lead to service bean cannot be created successfully.

## Verifying this change

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractConfig.java
@@ -590,6 +590,15 @@ public abstract class AbstractConfig implements Serializable {
                     if (MethodUtils.isGetter(method)) {
                         String name = method.getName();
                         String key = calculateAttributeFromGetter(name);
+                        /**
+                         * make sure attribute exists
+                         */
+                        try {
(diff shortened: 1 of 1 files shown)


chickenlj:
Thanks. Merged manually to avoid conflicts.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractConfig.java
```diff
@@ -590,6 +590,15 @@ public String toString() {
                     if (MethodUtils.isGetter(method)) {
                         String name = method.getName();
                         String key = calculateAttributeFromGetter(name);
+                        /**
+                         * make sure attribute exists
+                         */
+                        try {
+                            getClass().getField(key);
```
kylixs, line 597: Some attribute does not has a corresponding field, but has a pair of getter/setter methods, e.g. String getInterface()/setInterface(String) of ReferenceConfig. 
So it’s better to check if the attribute is writable, see aslo: https://github.com/apache/dubbo/pull/8289/

PR #4229, opened 2019-05-31
[maven] add maven plugin auto add and format license header

slievrly:
Signed-off-by: jimin.jm <slievrly@163.com>

1. add maven plugin auto add and format license header
2. format exist license header

[plugin]
license-maven-plugin    License: Apache License 2.0
http://code.mycila.com/license-maven-plugin/

             <plugin>
                <groupId>com.mycila</groupId>
                <artifactId>license-maven-plugin</artifactId>
                <version>3.0</version>
                <executions>
                    <execution>
                        <phase>generate-sources</phase>
                        <goals>
                            
                            <goal>format</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <quiet>true</quiet>
                    <header>${user.dir}/codestyle/copyright</header>
                    <includes>
                        <include>**/src/main/java/**</include>
                        <include>**/src/test/java/**</include>
                    </includes>
                    <strictCheck>true</strictCheck>
                    <mapping>
                        <java>SLASHSTAR_STYLE</java>
                    </mapping>
                </configuration>
            </plugin>

## Brief changelog

[maven] add maven plugin auto add and format license header

## Verifying this change

ci


dubbo-common/src/main/java/org/apache/dubbo/common/threadlocal/InternalThreadLocal.java
```diff
@@ -1,19 +1,19 @@
 /*
- * Copyright 2014 The Netty Project
```
slievrly: Why are there Netty headers here before? Please help confirm it.

htynkn:
+  Dubbo are using https://www.mojohaus.org/license-maven-plugin/index.html to do license check. Are we able to reuse existing plugin? (https://github.com/apache/dubbo/blob/cf1a09d3c381607585def4a7bb213496efea655b/pom.xml#L341)
+ If we can't reuse, why choose this one? accroding to release note, this plugin haven't release any new version for three years
```
 <plugin>
            <groupId>com.mycila</groupId>
            <artifactId>license-maven-plugin</artifactId>
</plugin>
```
+ we have rat plugin to check header, if we apply this plugin, do we still need that? or they are designed for different purpose?

dubbo-common/src/main/java/org/apache/dubbo/common/threadlocal/InternalThreadLocal.java
lovepoem: This is a special situation , because `InternalThreadLocal.java` is copied totally from `Netty`  project .  According to Apache's license rules, Netty's license header needs to be retained

slievrly:
At present, there are problems in the project that the license header is not uniform and missing. Need a way to help developers automatically add and format license header.
There are two major options to automatically add license headers to source files:
* Mojo License Maven plugin from codehaus
* Mycila Maven License plugin from mycila

1.  codehaus used in our projects is mainly used to check third-party dependencies license and defined in licenseCheck profile（not default build plugin）.It can be reused and updated license header，ref to [https://www.mojohaus.org/license-maven-plugin/update-file-header-mojo.html#projectName]. It automatically supports [cddl_v1, gpl_v3, mit, gpl_v2, gpl_v1, fdl_v1_3, epl_v2, epl_v1, epl_only_v2, epl_only_v1, agpl_v3, lgpl_v3, lgpl_v2_1, apache_v2, eupl_v1_1, bsd_3, bsd_2] license.
Here our apache_v2 header is different from our ASF header, so we can't use the existing one directly.
We need to define our header using licenseResolver, but here we need to define each module and fixed file name, ref to [https://www.mojohaus.org/license-maven-plugin/examples/example-add-license.html]. so I think that the codehaus plugin is difficult to use as an update license header.

2. Mycila is indeed almost three years away from the last update, but its corresponding github[https://github.com/mycila/license-maven-plugin] was updated 4 months ago. The current version of mycila is 3.0. I think its function is clear for a plugin. At present, the function of updating license header is undoubtedly satisfied.

3.  rat plugin[https://creadur.apache.org/rat/apache-rat-plugin/index.html] just check license header but not update and format license header.I hope there is a plugin that automatically helps developers format their license header.

I think it can be divided into two steps for the existing problems.
- Firstly, add or format the license header for the existing source file, but not include the plugin itself.
- Secondly, discuss how to use the plugin to help us format the license header in the project.

slievrly:
@htynkn @lovepoem @ralf0131

htynkn:
I agree with having a common and standard way to handle license header.

There are three thing need discuss:
1. what kind of format we agree to use (I see this PR has lots of line remove).
2. what benefit this kind of plugin can provide and what's the process after we introduce license plugin? should we run plugin before create PR?
3. does most of open source project use similar plugin? what's the difference between different kind of license plugin (mycila, Mojo etc)

I believe it's a good idea to use tool to automatic some manual stuff, but before we apply this change we need know the benefit and the possible impact for workflow (README, CONTRIBUTE doc may change too)

I suggest to start a discuss in maillist @slievrly and we may can get some help and information from community

slievrly:
> I agree with having a common and standard way to handle license header.
> 
> There are three thing need discuss:
> 
> 1. what kind of format we agree to use (I see this PR has lots of line remove).
> 2. what benefit this kind of plugin can provide and what's the process after we introduce license plugin? should we run plugin before create PR?
> 3. does most of open source project use similar plugin? what's the difference between different kind of license plugin (mycila, Mojo etc)
> 
> I believe it's a good idea to use tool to automatic some manual stuff, but before we apply this change we need know the benefit and the possible impact for workflow (README, CONTRIBUTE doc may change too)
> 
> I suggest to start a discuss in maillist @slievrly and we may can get some help and information from community

1.  The removed line is a blank line between the license header and the package. We need to specify whether we need to keep a blank line between the license header and the package. If you want to keep it, you need to define it in the template file, and it will also increase the blank line of many files. In short, we need to specify this format, all files will be strictly matched.

2. The mycila license plugin in this pr is used as the default bulid plugin, bound to the generate-sources phase of maven. So when we finish writing the code locally and do the maven compile check, it can automatically check the headers of all source files, add the headers that don't exist, and update the headers with incorrect formats.

3. For auto add and format licence headers, mojo is harder to use than mycila, and it needs to be defined separately for each module. Mojo is suitable for checking the licenses of third-party dependencies. mycila does not have this feature.

I will start a discussion with the community via mail group later.

dubbo-common/src/main/java/org/apache/dubbo/common/threadlocal/InternalThreadLocal.java
slievrly: @lovepoem revert all netty license header

htynkn:
referred discuss: https://lists.apache.org/thread.html/2393ac023cbec031146fec1ff037fd56f44fe225008f1af7bf24aed9%40%3Cdev.dubbo.apache.org%3E

This is no activity for a long time and I double check with mail list archives (it's really long time ago), I can't find a clear output.
so I'm going to close this PR. 
Thanks slievrly. If you still think this PR is necessary, please raise a new PR.

PR #4249, opened 2019-06-05
add excluded when get metadata

owen200008:

different platform compile the java file have the different methods sequence 
when get metadata, null will be set to false(like generic have getGeneric and isGeneric, getGeneric before isGeneric, the default value is set to false, it will be wrong)

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractConfig.java
@@ -493,8 +493,13 @@ public abstract class AbstractConfig implements Serializable {
                     String prop = calculateAttributeFromGetter(name);
                     String key;
                     Parameter parameter = method.getAnnotation(Parameter.class);
-                    if (parameter != null && parameter.key().length() > 0 && parameter.useKeyAsProperty()) {
-                        key = parameter.key();
+                    if (parameter != null) {
+                        if(parameter.excluded())
+                            continue;
+                        if(parameter.key().length() > 0 && parameter.useKeyAsProperty())
+                            key = parameter.key();
+                        else
+                            key = prop;
                     } else {
                         key = prop;
                     }


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractConfig.java
```diff
@@ -493,8 +493,13 @@ protected void appendAnnotation(Class<?> annotationClass, Object annotation) {
                     String prop = calculateAttributeFromGetter(name);
                     String key;
                     Parameter parameter = method.getAnnotation(Parameter.class);
-                    if (parameter != null && parameter.key().length() > 0 && parameter.useKeyAsProperty()) {
-                        key = parameter.key();
+                    if (parameter != null) {
+                        if(parameter.excluded())
```
owen200008: Parameter Annotation attitude need check excluded，when the Member variables is not Basic Data Types，default value is null，when the member function have isXXX and getXXX，the compile sequence will change the default value。
the other way to fix is make the MethodUtils.isGetter return false， so i think to change isXXX function name to other name，but this change have Too much influence，so need auth to check。

这边的Parameter标志去除应该包含进来，成员变量是对象性如果提供is和get函数的时候都符合，那么默认值就会跟编译获取methods的顺序相关

owen200008: i think appendAttributes also need add check excluded，now maybe it is Mutually exclusive（attitude and exclude），future it maybe not

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/AbstractConfig.java
```diff
@@ -493,8 +493,13 @@ protected void appendAnnotation(Class<?> annotationClass, Object annotation) {
...
+                            continue;
+                        if(parameter.key().length() > 0 && parameter.useKeyAsProperty())
+                            key = parameter.key();
+                        else
+                            key = prop;
                     } else {
```
owen200008: Parameter Annotation attitude need check excluded，when the Member variables is not Basic Data Types，default value is null，when the member function have isXXX and getXXX，the compile sequence will change the default value。
the other way to fix is make the MethodUtils.isGetter return false， so i think to change isXXX function name to other name，but this change have Too much influence，so need auth to check。
i think appendAttributes also need add check excluded，now maybe it is Mutually exclusive（attitude and exclude），future it maybe not。

这边的Parameter标志去除应该包含进来，成员变量是对象性如果提供is和get函数的时候都符合，那么默认值就会跟编译获取methods的顺序相关

PR #4291, opened 2019-06-12
fix org.apache.dubbo.common.extension.ExtensionLoader

zishinan:
fix org.apache.dubbo.common.extension.ExtensionLoader #when SPI config service names tag before "=" separated with ","


zishinan:
add commit error

PR #4292, opened 2019-06-12
fix-dubbo SPI config service when separated names with ","

zishinan:
just fix bug as title


zishinan:
add the test case code

zishinan:
重新提交一个

PR #4311, opened 2019-06-14
Improve/attachments

qinliujie:

support  attachment transfer any type, not only String

## Brief changelog

change attachment type from Map<String,String> to Map<String,Object>

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/mock/MockInvokersSelector.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/mock/MockInvokersSelector.java
@@ -50,7 +50,7 @@ public class MockInvokersSelector extends AbstractRouter {
         if (invocation.getAttachments() == null) {
             return getNormalInvokers(invokers);
         } else {
-            String value = invocation.getAttachments().get(Constants.INVOCATION_NEED_MOCK);
+            String value = (String) invocation.getAttachments().get(Constants.INVOCATION_NEED_MOCK);
             if (value == null) {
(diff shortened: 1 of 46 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/mock/MockInvokersSelector.java
```diff
@@ -50,7 +50,7 @@ public MockInvokersSelector() {
         if (invocation.getAttachments() == null) {
             return getNormalInvokers(invokers);
         } else {
-            String value = invocation.getAttachments().get(Constants.INVOCATION_NEED_MOCK);
+            String value = (String) invocation.getAttachments().get(Constants.INVOCATION_NEED_MOCK);
```
lixiaojiee: No casting is necessary here，attachments is a string mapping

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java
```diff
@@ -96,8 +96,8 @@ public URL getUrl() {
         }
 
         List<Invoker<T>> result = invokers;
-        String tag = StringUtils.isEmpty(invocation.getAttachment(TAG_KEY)) ? url.getParameter(TAG_KEY) :
-                invocation.getAttachment(TAG_KEY);
+        String tag = StringUtils.isEmpty((String) invocation.getAttachment(TAG_KEY)) ? url.getParameter(TAG_KEY) :
```
lixiaojiee: Same as above

PR #4339, opened 2019-06-18
fix bug about the generateReturnAndInvocation method of the class AdaptiveClassCodeGenerator

MarsMuse:

我发现在调用被代理的方法时，使用的参数名是该方法自身定义的旧的参数名，然而我们在生成代理类的过程中重载该方法时已经将参数名重新定义为arg为首，index为尾的参数名，例如arg0,arg1...。当使用旧的参数名调用被代理方法时，编译会出现错误，因为我们的局部变量之中已经不存在旧的参数名了，所以在调用被代理方法时应该使用新的参数名。
I found that when calling the proxy method, the parameter name used is the old parameter name defined by the method itself. However, when we overload the method in the process of generating the proxy class, we have redefined the parameter name as arg first and index tail, such as arg0, arg1.... When the proxy method is called with the old parameter name, compilation errors occur because the old parameter name does not exist in our local variables, so we should use the new parameter name when calling the proxy method.

## Brief changelog

XXXXX

## Verifying this change

XXXXX

- [ ] Make sure there is a [GITHUB_issue](https://github.com/apache/dubbo/issues) field for the change (usually before you start working on it). Trivial changes like typos do not require a GITHUB issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/extension/AdaptiveClassCodeGenerator.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/extension/AdaptiveClassCodeGenerator.java
@@ -63,7 +63,9 @@ public class AdaptiveClassCodeGenerator {
     
     
     private static final String CODE_EXTENSION_ASSIGNMENT = "%s extension = (%<s)%s.getExtensionLoader(%s.class).getExtension(extName);\n";
-    
+
+    private static final String CODE_EXTENSION_METHOD_INVOKE_ARGUMENT = "arg%d ";
+
     private final Class<?> type;
     
     private String defaultExtName;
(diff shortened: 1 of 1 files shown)


MarsMuse:
jdk version: 1.8.0_172 , compilation options : -parameters

beiwei30:
I think I understand what happens, you use 'javac -parameters' to compile. In this case, we should consider to enhance method signature generation instead, in my opinion.

beiwei30:
second thought, I went through the code again, there are lots of code based on 'arg%d' format. I agree your change is minimal. I will consider to merge.

tswstarplanet:
Sorry, I think the file committed is a little confused because it delete a file and add an almost same file. It' s difficult to identify the diff

PR #4368, opened 2019-06-21
fix zipkin compatible. #3728

### LINKED ISSUE #3728: examine backward compatibility for Zipkin and Skywalking on 2.7.2
see more from https://github.com/openzipkin/brave/issues/867, also see https://github.com/apache/incubator-dubbo/issues/3576

cvictory:
fix #3728

### DIFF
--- a/dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/protocol/dubbo/FutureAdapter.java
+++ b/dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/protocol/dubbo/FutureAdapter.java
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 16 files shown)


cvictory:
在zipkin的sample里测试，结果展示如下：


image

dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/protocol/dubbo/FutureAdapter.java
```diff
@@ -0,0 +1,114 @@
...
+                } else {
+                    if (appResponse.hasException()) {
+                        callback.caught(appResponse.getException());
+                    } else {
+                        callback.done((V) appResponse.getValue());
+                    }
+                }
+            }
+        };
+        futureAdapter.getAppResponseFuture().whenComplete(biConsumer);
```
chickenlj: `futureAdapter.getAppResponseFuture()` seems unnecessary, we can use CompletableFuture directly, there's no need to convert to `FutureAdapter` and use get it's internal `appResponseFuture`. 

What do you think？

PR #1843, opened 2018-05-24
[Dubbo-889-Attachments] support implicit delivery of attachments from provider to consumer

chickenlj:
## What is the purpose of the change
Fixes #889, support implicit delivery of attachments from provider to consumer  
* Separate protocol version with implementation version.
* Attachments on provider side automatically pass to consumer side.

## Brief changelog
* Version, ContextFilter, ConsumerContextFilter and Codec implementation of Dubbo protocol.

## Verifying this change
* Make sure changes compatible with lower dubbo version.
* Package dubbo, make sure log version is correct

Follow this checklist to help us incorporate your contribution quickly and easily:

- [x] Make sure there is a [GITHUB_issue](https://github.com/apache/incubator-dubbo/issues) filed for the change (usually before you start working on it). Trivial changes like typos do not require a GITHUB issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [x] Format the pull request title like `[Dubbo-XXX] Fix UnknownException when host config not exist #XXX`. Each commit in the pull request should have a meaningful subject line and body.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add integration-test in [test module](https://github.com/alibaba/dubbo/tree/master/dubbo-test).
- [x] Run `mvn clean install -DskipTests` & `mvn clean test-compile failsafe:integration-test` to make sure unit-test and integration-test pass.
- [x] If this contribution is large, please follow the [Software Donation Guide](https://github.com/apache/incubator-dubbo/wiki/Software-donation-guide).

PR #4396, opened 2019-06-25
Dubbo throws IoException with error message while decoding fail

qixiaobo:
have searched the issues of this repository and believe that this is not a duplicate.
  I have checked the FAQ of this repository and believe that this is not a duplicate.
Environment
Dubbo version: 2.6.6
Operating System version: *nix
Java version: 1.7+
Expected Result
throw new IOException("Unknown result flag, expect '0' '1' '2' '3' '4' '5', get " + flag);

Actual Result
throw new IOException("Unknown result flag, expect '0' '1' '2', get " + flag);

since dubbo add new type for class dubbocodec for issue #1843
we should also change the exception message .

For example, if dubbo provides new protocol version(higher than 2.0.2) .then the high version dubbo invoke the lower version dubbo[2.0.2] with no support codec ,then the lower version dubbo throws ioexception with incorrect message.
So we need fix it


qixiaobo:
https://github.com/apache/dubbo/issues/4395

PR #4413, opened 2019-06-27
Add unit tests for org.apache.dubbo.rpc.support.MockInvoker

Braavos96:
I've analysed your codebase and noticed that `org.apache.dubbo.rpc.support.MockInvoker` is not fully tested.
I've written some tests for the methods in this class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests will help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes that you consider important.

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/test/java/org/apache/dubbo/rpc/support/MockInvokerTest.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/test/java/org/apache/dubbo/rpc/support/MockInvokerTest.java
@@ -1,0 +1,139 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-api/src/test/java/org/apache/dubbo/rpc/support/MockInvokerTest.java
```diff
@@ -0,0 +1,138 @@
...
+ */
+package org.apache.dubbo.rpc.support;
+
+import java.io.Serializable;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+import org.apache.dubbo.common.URL;
+import org.apache.dubbo.rpc.*;
```
Moriadry: pls do not use `import *`

Braavos96: Apologies, I missed that. I have removed the wildcard

Issue #4400, opened 2019-06-26
dependency issue in dubbo bom


beiwei30:
In org.apache.dubbo:dubbo-dependencies-bom, there's no 8.5.31 for tomcat-embed-logging-juli in central repo:

```xml
      <dependency>
        <groupId>org.apache.tomcat.embed</groupId>
        <artifactId>tomcat-embed-logging-juli</artifactId>
        <version>${tomcat_embed_version}</version>
      </dependency>
```

cvictory:
<dependency>
                <groupId>org.apache.tomcat.embed</groupId>
                <artifactId>tomcat-embed-core</artifactId>
                <version>${tomcat_embed_version}</version>
            </dependency>
            <dependency>
                <groupId>org.apache.tomcat.embed</groupId>
                <artifactId>tomcat-embed-logging-juli</artifactId>
                <version>${tomcat_embed_version}</version>
            </dependency>

what are they used ?

cvictory:
I will try to delete this in next release

PR #4430, opened 2019-07-01
remove tomcat-embed-logging-juli #4400

cvictory:
#4400


chickenlj:
LGTM. 

The dependency removed is totally useless.

Issue #4393, opened 2019-06-25
tag rule should support anyhost


beiwei30:
The following rule doesn't work. We should support **0.0.0.0**.

```yaml
---
force: false
runtime: true
enabled: true
priority: 1
key: governance-tagrouter-provider
tags:
- name: tag1
  addresses: ["0.0.0.0:20881"]
- name: tag2
  addresses: ["0.0.0.0:20880"]
...
```

PR #4431, opened 2019-07-01
support all of ip #4393

cvictory:
#4393


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagRouter.java
```diff
@@ -212,6 +214,9 @@ private boolean checkAddressMatch(List<String> addresses, String host, int port)
                 if (NetUtils.matchIpExpression(address, host, port)) {
                     return true;
                 }
+                if ((ALL_IP_MATCH + ":" + port).equals(address)) {
```
chickenlj: Reuse CommonConstants.ANYHOST would be better.

cvictory: ok

PR #4523, opened 2019-07-10
fix #4125.   fix no info is thrown when zookeeper initialization

### LINKED ISSUE #4125: ZookeeperDynamicConfiguration 严重 BUG
### Environment

* Dubbo version: 2.7.1
* Operating System version: windows
* Java version: 8

当 zookeeper 地址配置错误时，`CountDownLatch` 会在下面位置锁死，**建议增加超时时间！！**

cvictory:
fix #4125

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/network/ConnectException.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/network/ConnectException.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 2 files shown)


chickenlj:
I suggest removing the CountdownLatch completely.

dubbo-common/src/main/java/org/apache/dubbo/common/network/ConnectException.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.dubbo.common.network;
+
+/**
+ * 2019-07-10
```
beiwei30: pls. remove this, it is useless.

beiwei30:
I think we should figure out why latch doesn't count down.

PR #4540, opened 2019-07-11
[Dubbo-4525] same as #4533, fix ClusterUtils.mergeUrl

chickenlj:


cvictory:
The project build failed in jdk1.7.   And ignore it temporary.

PR #4592, opened 2019-07-17
Optimize if-else block in ServiceConfig.java

wangnimaa:

i think it's to many if-else,I optimized it.

and i find this code:

                                        if (argument.getIndex() != -1) {
                                            xxxxxxx
                                        } else {
                                                xxxxxx
                                               if (argument.getIndex() != -1 && argument.getIndex() != j) {
                                                  throw new IllegalArgumentException("Argument config error : the 
                                                   index attribute and type attribute not match :index :" + 
                                                   argument.getIndex() + ", type:" + argument.getType());
                                                    }

i think it's never go to last if,so i delete it.


htynkn:
I think unit test for this part of code is missing (https://codecov.io/gh/apache/dubbo/src/master/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java#L494)

Is there any chance to add unit test for this change?I think it will make changes more safe

wangnimaa:
> I think unit test for this part of code is missing (https://codecov.io/gh/apache/dubbo/src/master/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java#L494)
> 
> Is there any chance to add unit test for this change?I think it will make changes more safe



> I think unit test for this part of code is missing (https://codecov.io/gh/apache/dubbo/src/master/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java#L494)
> 
> Is there any chance to add unit test for this change?I think it will make changes more safe

image
is this?

wangnimaa:
> UT start to fail:
> 
> ```
> [ERROR] Failures: 
> [ERROR]   ServiceConfigTest.testDelayExport:173 
> Expected: a collection with size <1>
>      but: collection size was <0>
> [ERROR] Errors: 
> [ERROR]   ServiceConfigTest.testExport:138 Â» NullPointer
> [ERROR]   ServiceConfigTest.testProxy:160 Â» NullPointer
> ```

fix

PR #4600, opened 2019-07-18
add a interface GraceFullShutDown to supoort customized action in shutdown

kele6413430:
Introduce:
In shutdown stage of a project ,gracefull shut down is necessary.
There is a application listener ,but it is not enough,some actions should be done with dubbo.
For example, Kafka producer should flush messages to the broker after protocol destroyed.
Another,MQ consumer should not poll  new messages any more after registries destroyed. 

Changelog
1. add a interface GraceFulShutDown
2.update SpringExtensionFactory.ShutdownHookListener to support customized actions when spring bean exists which implements GraceFullShutDown
3.add a unit test GraceFulShutDownTest


dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/extension/GraceFullShutDown.java
```diff
@@ -0,0 +1,27 @@
+package org.apache.dubbo.config.spring.extension;
+
+/**
+ * this interface is used to enable grace shutdown
+ * any class implements this interface should be a spring bean
+ * @author tiantian.yuan
+ * @date 2018年12月24日21:12:35
+ */
+public interface GraceFullShutDown {
```
lixiaojiee: Why not GracefulShutdown

dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/extension/GraceFullShutDown.java
```diff
@@ -0,0 +1,27 @@
+package org.apache.dubbo.config.spring.extension;
+
+/**
+ * this interface is used to enable grace shutdown
+ * any class implements this interface should be a spring bean
+ * @author tiantian.yuan
```
lixiaojiee: The author info should ⌫

lixiaojiee: The author and date info should not exist

dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/extension/GraceFullShutDown.java
kele6413430: already update

PR #4629, opened 2019-07-22
[dubbo 4458] doRefer in RestProtocol may cause OOM

finalcola:

https://github.com/apache/dubbo/issues/4458#issue-463641024

1. clear resouces when invoker(AbstructProxyInvoker) is not in use: 

```java
protected <T> Invoker<T> protocolBindingRefer(final Class<T> type, final URL url) throws RpcException {
        final Invoker<T> target = proxyFactory.getInvoker(doRefer(type, url), type, url);
        Invoker<T> invoker = new AbstractInvoker<T>(type, url) {
            @Override
            protected Result doInvoke(Invocation invocation) throws Throwable {
                // do invoke...
            }

            // invokers need clear unused invoker
            @Override
            public void destroy() {
                super.destroy();
                target.destroy();
                invokers.remove(this);
            }
        };
        invokers.add(invoker);
        return invoker;
    }
```

2. clear RestProtocol.cliens by using map(key is url, value is client)

## Brief changelog

XXXXX

## Verifying this change

XXXXX


beiwei30:
@finalcola I am not persuaded that the current solution is better than the original. The reason is, I believe 'connection manager' plus 'connection monitor' will reuse and recycle the client. The potential OOM issue may caused by 'connection manager' is not used when the client is created, which has in fact been fixed by you with another pull request: #4614

finalcola:
#4614 will clear HTTPConnection, but RestProtocol will keep too much client(ResteasyClient) reference which makes these ResteasyClient instance can't be GC.

finalcola:
@beiwei30

beiwei30:
I see. In your solution, when will ResteasyClient be recycled?

beiwei30:
> #4614 will clear HTTPConnection, but RestProtocol will keep too much client(ResteasyClient) reference which makes these ResteasyClient instance can't be GC.

a second thought: If I don't misunderstand, there's no way to recycle the clients, even if they are now saved in a map. This may lead to a potential problem, which is, the underlaying connection can be closed by connection manager (see connection monitor). If this happens, this will lead to invocation failure since the resteasy client maintained in the map doesn't have lifcecycle management.

Should we simply create Resteasy client every time and then drop it? I think it is not expensive since the connection will be reused, what do you think?

Or should we maintain the client in weak hash map instead?

finalcola:
The previous implementation was to use a List to collect clients, but add client only, never remove. Now clients are saved in map which means they are in use, and old client will be remove and close.    When doRefer happens,  will create a new client for this url.   @beiwei30

beiwei30:
> The previous implementation was to use a List to collect clients, but add client only, never remove. Now clients are saved in map which means they are in use, and old client will be remove and close. When doRefer happens, will create a new client for this url. @beiwei30

could you pls. point out when the client will be removed from the map?

finalcola:
it will not remove. My original idea was that each existing url would generate only one client（previous implementation using Sync List will add client everytime）, because It is not clear whether the client corresponding to the url is still in use.

maybe sync WeakHashMap can solve it, but it dose not pass the  travis-ci in JDK 11.
[https://travis-ci.org/apache/dubbo/builds/563891937](url)

@beiwei30

finalcola:
now, i use WeakRefence to hold client

finalcola:
@beiwei30 could you please have a look, if there are any problems

dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/RestProtocol.java
```diff
@@ -262,10 +272,19 @@ protected String getContextPath(URL url) {
 
     protected class ConnectionMonitor extends Thread {
         private volatile boolean shutdown;
-        private final List<PoolingHttpClientConnectionManager> connectionManagers = Collections.synchronizedList(new LinkedList<>());
+        private final Map<String, PoolingHttpClientConnectionManager> connectionManagers = new ConcurrentHashMap<>();
+
+        public void addConnectionManager(String key, PoolingHttpClientConnectionManager connectionManager) {
+            connectionManagers.put(key, connectionManager);
+//            connectionManagers.add(connectionManager);
```
xiantang: remove unuse comment

AlbumenJ:
@finalcola Hi, thanks for your contribution.

Please merge the latest master branch to resolve confilcts.

AlbumenJ:

Issue #4632, opened 2019-07-23
protobuf-json序列化方式bug


CodingSinger:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.2
* Operating System version: macOS 10.12.6
* Java version: 1.8

### Steps to reproduce this issue

1. 使用protobuf-json的序列化方式
2. 进行普通的dubbo RPC调用
### Expected Result

正常调用
### Actual Result

报错,序列化失败

If there is an exception, please attach the exception trace:

```java
Caused by: java.lang.IllegalArgumentException: This serialization only support google protobuf object, the object class is: java.lang.String
	at org.apache.dubbo.common.serialize.protobuf.support.GenericProtobufObjectOutput.writeObject(GenericProtobufObjectOutput.java:94)
	at org.apache.dubbo.common.serialize.protobuf.support.GenericProtobufObjectOutput.writeUTF(GenericProtobufObjectOutput.java:74)
	at org.apache.dubbo.rpc.protocol.dubbo.DubboCodec.encodeRequestData(DubboCodec.java:176)
	at org.apache.dubbo.remoting.exchange.codec.ExchangeCodec.encodeRequest(ExchangeCodec.java:238)
	at org.apache.dubbo.remoting.exchange.codec.ExchangeCodec.encode(ExchangeCodec.java:69)
	at org.apache.dubbo.rpc.protocol.dubbo.DubboCountCodec.encode(DubboCountCodec.java:40)
	at org.apache.dubbo.remoting.transport.netty4.NettyCodecAdapter$InternalEncoder.encode(NettyCodecAdapter.java:70)
	at io.netty.handler.codec.MessageToByteEncoder.write(MessageToByteEncoder.java:107)
	... 23 more

	at org.apache.dubbo.remoting.exchange.support.DefaultFuture.doReceived(DefaultFuture.java:191)
	at org.apache.dubbo.remoting.exchange.support.DefaultFuture.received(DefaultFuture.java:153)
	at org.apache.dubbo.remoting.exchange.support.DefaultFuture.received(DefaultFuture.java:141)
	at org.apache.dubbo.remoting.exchange.support.header.HeaderExchangeHandler.handleResponse(HeaderExchangeHandler.java:62)
[...]

PR #4638, opened 2019-07-23
Followup for #4632, use standard protobuf json format to write byte[].

chickenlj:
see #4632


CodingSinger:
LGTM

PR #4677, opened 2019-07-26
@Reference auto-wires the instance of generic interface #4594

CodingSinger:

@Reference auto-wires the instance of generic interface

see #4594

### DIFF
--- a/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/util/AnnotationUtils.java
+++ b/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/util/AnnotationUtils.java
@@ -18,7 +18,6 @@ package org.apache.dubbo.config.spring.util;
 
 import org.apache.dubbo.config.annotation.Reference;
 import org.apache.dubbo.config.annotation.Service;
-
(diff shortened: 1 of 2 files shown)


dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/util/AnnotationUtils.java
```diff
@@ -93,6 +92,14 @@ public static String resolveInterfaceName(Service service, Class<?> defaultInter
      * @throws IllegalStateException if interface name was not found
      */
     public static String resolveInterfaceName(AnnotationAttributes attributes, Class<?> defaultInterfaceClass) {
+        Boolean generic = getAttribute(attributes, "generic");
```
beiwei30: looks like this is unnecessary, since it works only for `@Service`

CodingSinger: No, it also use for `@Reference`. I think the method's comments is wrong.

beiwei30: nope, my mistake.

Issue #4470, opened 2019-07-05
同一个dubbo服务一个微服务内多个类引用时,有些类的此dubbo服务reference注解属性不起作用


jufeng98:
- [ ] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.6.2
* Operating System version:Windows10
* Java version: 1.8

### Steps to reproduce this issue

1. 同一个微服务多个类引用同样的dubbo服务
2. 每个类的reference除了version和group一样之外,其它属性均有不一样的配置
3. 项目运行起来后有些类的此dubbo服务reference注解属性不起作用

我看了源码,是这里的问题:
    /**
     * Generate a cache key of {@link ReferenceBean}
     *
     * @param reference {@link Reference}
     * @param beanClass {@link Class}
     * @return
     */
    private String generateReferenceBeanCacheKey(Reference reference, Class<?> beanClass) {

        String interfaceName = resolveInterfaceName(reference, beanClass);

        String key = reference.url() + "/" + interfaceName +
                "/" + reference.version() +
                "/" + reference.group();

        Environment environment = applicationContext.getEnvironment();

        key = environment.resolvePlaceholders(key);

        return key;

    }
这里会缓存dubbo service,但是key应该由所有reference属性组成才对

chickenlj:
Maybe we should count `reference id` as part of the key. 

```java
String key = reference.id() + "/" + interfaceName +
            "/" + reference.version() +
            "/" + reference.group();
```

jufeng98:
那是极好的，可以解决这个问题

mercyblitz:
This issue has been fixed, please update 2.6.6~ or refer https://github.com/apache/dubbo/blob/f9325699388f653ebbcd921cbdd0baf02d838715/dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/beans/factory/annotation/ReferenceAnnotationBeanPostProcessor.java#L178

jufeng98:
我看了2.6.6版本，问题并得到没有解决
![image](https://user-images.githubusercontent.com/26024578/60815841-0b358280-a1cb-11e9-83b4-2ddcc8dabbf2.png)
timeout和retries属性完全被忽略
[...]

Issue #4580, opened 2019-07-16
Consumer onconnect, ondisconnect is not working


codeimport:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.3
* Operating System version: windows10
* Java version: JDK1.8

### Steps to reproduce this issue

1. set onconnect,ondisconnect parameter on @Reference
2. write onconnect,ondisconnect callback on consumer

code:

    @Reference(onconnect = "onconnect")
    private HelloDubboService helloDubboService;

    public void onconnect(){
        System.out.println(">>>>>>>>>>>>>>>>>onconnect()");
    }

### Expected Result

it should invoke onconnect ,ondisconnect callback when the consumer connect the provider 

### Actual Result

it does not work

is it a bug?or these callback only be used on provider side?

chickenlj:
> is it a bug?or these callback only be used on provider side?

At present, it only works on the provider side.  But it would be a good complement if we can add support on the Consumer side.

beiwei30:
I will verify. I believe it starts working now after https://github.com/apache/dubbo/pull/4598 is merged.

beiwei30:
confirmed, it doesn't work. We need something like org.apache.dubbo.config.spring.schema.DubboBeanDefinitionParser#parseMethods when process `@Method`

tswstarplanet:
> > is it a bug?or these callback only be used on provider side?
> 
> At present, it only works on the provider side. But it would be a good complement if we can add support on the Consumer side.

In fact the consumer side can also trigger the connected and disconnected methods of ChannelHandler interface of dubbo.

cvictory:
从实现角度来看，provider和consumer的连接断开都应该可以增加一定的事件。
但是现在onconnect的配置是和接口绑定在一起，在后续的链路中会去找这个接口的对应实现。 consumer端是没有实现的，所以功能上暂时不能支持。

如果要在consumer端支持，应该需要调用一个独立的类和方法来完成。 暂时不建议增加。

----

From an implementation perspective, the disconnection between the provider and the consumer should be able to increase certain events.
But now the configuration of onconnect is bound to the interface, and the corresponding implementation of this interface will be found in subsequent links. The consumer side is not implemented, so it cannot be supported temporarily.

[...]

PR #4680, opened 2019-07-28
 #4470 fix referenceAnnotation config not work when a dubbo service wa…

yiranlau:

#4470 fix referenceAnnotation config not work when a dubbo service was referenced multi times

@mercyblitz have tried to fix this issue at 2.6.6 but not work. and I found the `referencedBeanName` was not really changed.


zhaozhiwei2017:
This solution will create the same number of spring beans as you use @Reference in your project

yiranlau:
> This solution will create the same number of spring beans as you use @reference in your project

yes, you are right. but I think it would be accepted if we want to implement  consumer side onconnect or ondisconnect event callback as notied at #4580,  which I still cannot figure out how to transport a bean proxy object which have injected bean field to DubboProtocol onconnect event or ondisconnect event.

zhaozhiwei2017:
> > This solution will create the same number of spring beans as you use @reference in your project
> 
> yes, you are right. but I think it would be accepted if we want to implement consumer side onconnect or ondisconnect event callback as notied at #4580, which I still cannot figure out how to transport a bean proxy object which have injected bean field to `DubboProtocol` onconnect event or ondisconnect event.

I can't figure out too, but I think this scenario is not enough gracefully, we may get another solution that gracefully than it

beiwei30:
this solution is not correct, close now. Pls. refer @mercyblitz 's comment at the original issue.

PR #4746, opened 2019-08-05
Remove unnecessary extend.

wayilau:
Remove Endpoint from Client, because Client extend Channel which extend Endpoint. so i think we can remove it.


wayilau:
no activity.

PR #4756, opened 2019-08-06
i think is better to remove the key ,when the corresponding listeners is empty in AbstractEventDispatcher

0704681032:
rt


0704681032:
@mercyblitz

PR #4758, opened 2019-08-07
I think it is better to remove the key ,when the corresponding listeners is empty in AbstractEventDispatcher.java

0704681032:
rt


dubbo-event/src/main/java/org/apache/dubbo/event/AbstractEventDispatcher.java
```diff
@@ -134,8 +134,11 @@ protected void doInListener(EventListener<?> listener, Consumer<Collection<Event
                 List<EventListener> listeners = listenersCache.computeIfAbsent(eventType, e -> new LinkedList<>());
                 // consume
                 consumer.accept(listeners);
-                // sort
-                sort(listeners);
+                if (listeners.size() == 0) {
+                    listenersCache.remove(eventType);
```
qixiaobo: Why this happens？
Why not remove all?

PR #4785, opened 2019-08-10
Remove fixed settings

brucelwl:
```
private AnnotatedMethodElement  --> public AnnotatedMethodElement  
add fields getter/ setter
```

 ``` 
 DubboConfigBindingBeanPostProcessor -- > remove fixed settings 
  private boolean ignoreUnknownFields = true;
  private boolean ignoreInvalidFields = true;```


ralf0131:
Hi, What's the purpose of this change?

brucelwl:
> Hi, What's the purpose of this change?

我需要通过AnnotatedMethodElement 获取里面的注解信息,但是这个类是私有的, 将他改成public,比较合理

Moriadry:
In what case, that you need to catch the private variables in `AnnotatedMethodElement`?
Can you post/attach an example that triggers this logic?

Issue #4784, opened 2019-08-09
Use generic invocation via API , lost "version" value


z529192557:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.3
* Operating System version: MAC OX 10.14.5
* Java version: 1.8

### Steps to reproduce this issue

1. Use GenericService by API,  assign the version is 1.0.0.test.qa.st
![image](https://user-images.githubusercontent.com/9245774/62794432-aab99e00-bb06-11e9-9422-8317c64374b1.png)

2. breakpoint at DubboInvoker.doInvoke
3. check the DubboInvoker.version = 0.0.0 
![image](https://user-images.githubusercontent.com/9245774/62794513-e2284a80-bb06-11e9-878a-ab54d420d57c.png)

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result
In the instance of DubboInvoker，the version should be  1.0.0.test.qa.st
### Actual Result
In the instance of DubboInvoker，the version’s value actully is  default value,0.0.0
### Cause
this method : "ClusterUtils.mergeUrl" at line 87 ,remove the version key, so lost the version's value

z529192557:
same as "group",  "ClusterUtils.mergeUrl" at line 84, remove the group key , so lost the group value

tswstarplanet:
Would you please submit a pr to fix it ?

z529192557:
> Would you please submit a pr to fix it ?
OK，I submit a pr to fix it，

z529192557:
the modify in the issues https://github.com/apache/dubbo/issues/4525 leads to this problem，I just fix GROUP_KEY and VERSION_KEY，
- when consumer has theirs value，and provider dose not have， use the consumer's value
- when consumer does not have theirs value，and provider has value， use the provider's value

beiwei30:
@z529192557 but I tried  it on main branch, but failed to reproduce.

beiwei30:
> @z529192557 but I tried it on main branch, but failed to reproduce.

I didn't realize your test is direct invocation, pls. ignore my previous comment.

z529192557:
> test

Yes，it‘s direct invocation，only in direct invocation via API， can reproduce.

z529192557:
[...]

PR #4787, opened 2019-08-10
Fix use generic invocation via API , lost "version" value [#4784]

z529192557:

fix use generic invocation via API , lost "version" value

## Brief changelog

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/ClusterUtils.java

## Issues
- https://github.com/apache/dubbo/issues/4784

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/ClusterUtils.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/ClusterUtils.java
@@ -81,10 +81,16 @@ public class ClusterUtils {
 
         if (localMap != null && localMap.size() > 0) {
             Map<String, String> copyOfLocalMap = new HashMap<>(localMap);
-            copyOfLocalMap.remove(GROUP_KEY);
+
+            if(map.containsKey(GROUP_KEY)){
+                copyOfLocalMap.remove(GROUP_KEY);
+            }
+            if(map.containsKey(VERSION_KEY)){
+                copyOfLocalMap.remove(VERSION_KEY);
+            }
+
             copyOfLocalMap.remove(RELEASE_KEY);
             copyOfLocalMap.remove(DUBBO_VERSION_KEY);
-            copyOfLocalMap.remove(VERSION_KEY);
             copyOfLocalMap.remove(METHODS_KEY);
             copyOfLocalMap.remove(TIMESTAMP_KEY);
             copyOfLocalMap.remove(TAG_KEY);
--- a/dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/support/ClusterUtilsTest.java
+++ b/dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/support/ClusterUtilsTest.java
@@ -133,8 +133,8 @@ public class ClusterUtilsTest {
         remoteURL = URL.valueOf("dubbo://localhost:20880/DemoService");
         mergedUrl = ClusterUtils.mergeUrl(remoteURL, localURL.getParameters());
 
-        Assertions.assertNull(mergedUrl.getParameter(VERSION_KEY));
-        Assertions.assertNull(mergedUrl.getParameter(GROUP_KEY));
+        Assertions.assertEquals(mergedUrl.getParameter(VERSION_KEY),localURL.getParameter(VERSION_KEY));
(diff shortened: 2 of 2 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/ClusterUtils.java
```diff
@@ -81,10 +81,16 @@ public static URL mergeUrl(URL remoteUrl, Map<String, String> localMap) {
...
             Map<String, String> copyOfLocalMap = new HashMap<>(localMap);
-            copyOfLocalMap.remove(GROUP_KEY);
+
+            if(map.containsKey(GROUP_KEY)){
+                copyOfLocalMap.remove(GROUP_KEY);
+            }
+            if(map.containsKey(VERSION_KEY)){
+                copyOfLocalMap.remove(VERSION_KEY);
+            }
+
```
tswstarplanet: This change means that if the provider set the group or version, the group or version set by the consumer will be override. I think it's not right, because it may cause wrong invoke.
I think the right solution is that just no remove group and version from localMap.

z529192557: OK,I once think should no remove from localMap,  But I check previous version of ClusterUtils, It always is  "if the provider set the version, the version set by the consumer will be override", I think this logic maybe use in RegistryDirectory.mergeUrl
image

tswstarplanet:
This pr can solve the problem, but I found the code is confusing.
If we use direct invoke, usually the direct link url doesn't have parameters(because we can define parameters in reference directly. So the remoteUrl in ClusterUtils.mergeUrl method is empty. So it doesn't have group and version key.

If we don't use direct invoke, and either group or version of ReferenceConfig and ServiceConfig is different, it will not go into the ClusterUtils.merge method.

Maybe we can split two method, one is used by direct invoke. The other is used by the indirect invoke.

z529192557:
I

> Details



> This pr can solve the problem, but I found the code is confusing.
> If we use direct invoke, usually the direct link url doesn't have parameters(because we can define parameters in reference directly. So the remoteUrl in ClusterUtils.mergeUrl method is empty. So it doesn't have group and version key.
> 
> If we don't use direct invoke, and either group or version of ReferenceConfig and ServiceConfig is different, it will not go into the ClusterUtils.merge method.
> 
> Maybe we can split the ClusterUtils.merge method into two methods, one is used by direct invoke. The other is used by the indirect invoke.

Yes，The best way to solve the problem is ClusterUtils.merge should be splited two methods

z529192557:
> This pr can solve the problem, but I found the code is confusing.
> If we use direct invoke, usually the direct link url doesn't have parameters(because we can define parameters in reference directly. So the remoteUrl in ClusterUtils.mergeUrl method is empty. So it doesn't have group and version key.
> 
> If we don't use direct invoke, and either group or version of ReferenceConfig and ServiceConfig is different, it will not go into the ClusterUtils.merge method.
> 
> Maybe we can split the ClusterUtils.merge method into two methods, one is used by direct invoke. The other is used by the indirect invoke.

I think we should discuss with the method's author ken.lj <ken.lj.hz@gmail.com> to decide how to fix

z529192557:
> This pr can solve the problem, but I found the code is confusing.
> If we use direct invoke, usually the direct link url doesn't have parameters(because we can define parameters in reference directly. So the remoteUrl in ClusterUtils.mergeUrl method is empty. So it doesn't have group and version key.
> 
> If we don't use direct invoke, and either group or version of ReferenceConfig and ServiceConfig is different, it will not go into the ClusterUtils.merge method.
> 
> Maybe we can split the ClusterUtils.merge method into two methods, one is used by direct invoke. The other is used by the indirect invoke.

I read the code ZookeeperRegistry.toUrlsWithEmpty, I found 
- if the ”group“ or” version“ is mismatching，the url will not be passed to next process
- so when invoke the RegistryDirectory.toInvokers, the group or version is equels between ReferenceConfig and ServiceConfig,
- then the method ClusterUtils.merge merge the provider's "gourp",or "version" value to consumer is redundant

So I think , when provider does not have "group" or "version", generally direct invoke ，does remove "group" or "version" key from local map，it can meet both direct invoke and from registry。

tswstarplanet:
LGTM

chickenlj:
Test if other parameters have have the same problem

zlba111:
2.7.4什么时候发布到mvn仓库？

PR #4811, opened 2019-08-13
remove duplicated references from ServiceConfig and ReferenceConfig, keep them in ConfigManager

mzorro:

Some global unique configuration references in ServiceConfig and ReferenceConfig is now duplicated from what's defined in ConfigManager, we can remove those duplicates.

## Brief changelog

Not much, see code for detail.

## Verifying this change

Some test case need to be updated: we should clear config in `@BeforeAll` instead of `@BeforeEach`


chickenlj:
I think this is a very good try and you've shown us the right way to go.
I am really sorry that this patch is currently severely conflicted with the code in the codebase. It would be appreciated if you could submit a new patch based on the latest codebase.

PR #4820, opened 2019-08-14
Optimizing and reusing code

yidadi:
Optimizing and reusing code

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/support/eager/EagerThreadPool.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/support/eager/EagerThreadPool.java
@@ -51,7 +51,7 @@ public class EagerThreadPool implements ThreadPool {
         int alive = url.getParameter(ALIVE_KEY, DEFAULT_ALIVE);
 
         // init queue and executor
-        TaskQueue<Runnable> taskQueue = new TaskQueue<Runnable>(queues <= 0 ? 1 : queues);
+        TaskQueue<Runnable> taskQueue = new TaskQueue<Runnable>(Math.max(1, queues));
         EagerThreadPoolExecutor executor = new EagerThreadPoolExecutor(cores,
                 threads,
                 alive,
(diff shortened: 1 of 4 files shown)


dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/ReferenceBean.java
```diff
@@ -181,19 +181,19 @@ public void afterPropertiesSet() throws Exception {
 
         if (getMetadataReportConfig() == null) {
             Map<String, MetadataReportConfig> metadataReportConfigMap = applicationContext == null ? null : BeanFactoryUtils.beansOfTypeIncludingAncestors(applicationContext, MetadataReportConfig.class, false, false);
-            if (metadataReportConfigMap != null && metadataReportConfigMap.size() == 1) {
+            if (CollectionUtils.mapSizeEq(metadataReportConfigMap, 1)) {
                 // first elements
                 super.setMetadataReportConfig(metadataReportConfigMap.values().iterator().next());
-            } else if (metadataReportConfigMap != null && metadataReportConfigMap.size() > 1) {
+            } else if (CollectionUtils.mapSizeGt(metadataReportConfigMap, 1)) {
```
ralf0131: I do not see the benefits of this change, actually it increase the difficulty to understand. I'd rather keep the original logic.

dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/support/eager/EagerThreadPool.java
```diff
@@ -51,7 +51,7 @@ public Executor getExecutor(URL url) {
         int alive = url.getParameter(ALIVE_KEY, DEFAULT_ALIVE);
 
         // init queue and executor
-        TaskQueue<Runnable> taskQueue = new TaskQueue<Runnable>(queues <= 0 ? 1 : queues);
+        TaskQueue<Runnable> taskQueue = new TaskQueue<Runnable>(Math.max(1, queues));
```
Moriadry: Both of two lines got the same meaning, I believe it's unnecessary to change this.

chickenlj:
As commented by the reviewers.

Issue #4765, opened 2019-08-07
run dubbo-demo-api error


SparkandScala:
- [ ] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.4
* Operating System version: Mac
* Java version: 1.8

### Steps to reproduce this issue

1. import the source code of dubbo.
2. at root of the dubbo, execute mvn install
3. run zookeeper at localhost:2181
4. cd to dubbo-demo/dubbo-demo-api/dubbo-demo-api-provide/src/......  run main()
5. has error No such extension org.apache.dubbo.configcenter.DynamicConfigurationFactory by name zookeeper

i want to run this demo at local.

tswstarplanet:
You should add the maven dependency related to ZookeeperDynamicConfigurationFactory

SparkandScala:
can you show me the dependency, thank you,

SparkandScala:
<dependency>
            <groupId>org.apache.dubbo</groupId>
            <artifactId>dubbo-remoting-zookeeper</artifactId>
            <version>${project.parent.version}</version>
        </dependency>
is it this?

SparkandScala:
<dependency>
            <groupId>org.apache.dubbo</groupId>
            <artifactId>dubbo-remoting-zookeeper</artifactId>
            <version>${project.parent.version}</version>
        </dependency>

SparkandScala:
Exception in thread "main" java.lang.IllegalStateException: No such extension org.apache.dubbo.configcenter.DynamicConfigurationFactory by name zookeeper
	at org.apache.dubbo.common.extension.ExtensionLoader.findException(ExtensionLoader.java:517)
	at org.apache.dubbo.common.extension.ExtensionLoader.createExtension(ExtensionLoader.java:524)
	at org.apache.dubbo.common.extension.ExtensionLoader.getExtension(ExtensionLoader.java:348)
	at org.apache.dubbo.config.AbstractInterfaceConfig.getDynamicConfiguration(AbstractInterfaceConfig.java:313)
	at org.apache.dubbo.config.AbstractInterfaceConfig.prepareEnvironment(AbstractInterfaceConfig.java:289)
[...]

PR #4832, opened 2019-08-15
[Dubbo-4765] fix demo dependencies

mzorro:
fix issue #4765


dubbo-demo/dubbo-demo-api/dubbo-demo-api-consumer/pom.xml
```diff
@@ -40,7 +40,11 @@
         </dependency>
         <dependency>
             <groupId>org.apache.dubbo</groupId>
-            <artifactId>dubbo-registry-multicast</artifactId>
```
ralf0131: Please do not remove the the multicast dependency.  We need to keep the convenience for People to switch between multicast and zookeeper.

Issue #4838, opened 2019-08-15
CuratorZookeeperClient中isConnected判断逻辑


hengyoush:
目前的判断逻辑:
```java
    public boolean isConnected() {
        return client.getZookeeperClient().isConnected();
    }
```
这里存在一个问题, 就是在client的实现`org.apache.curator.framework.imps.CuratorFrameworkImpl`中,close方法实现简要如下:
```java
if ( state.compareAndSet(CuratorFrameworkState.STARTED, CuratorFrameworkState.STOPPED) ) {
...
client.close(); // 此处的client即client.getZookeeperClient()的返回值
}
```
所以原先的isConnected实现可能会出现问题, 可以将其修改为同时判断client.getState()和client.getZookeeperClient().isConnected()

wavesZh:
`client.close()`会将`isConnected.set(false);`。如果close了，`CuratorZookeeperClient#isConnected()`的确会返回false，看起来没什么问题。你指的场景是什么？

CrazyHZM:
No feedback for a long time, please close the issue temporarily. If there is still a problem, you can reopen it.

PR #4839, opened 2019-08-15
[fix 4838] 优化CuratorZookeeperClient中isConnected逻辑判断

hengyoush:
fix issue #4838


chickenlj:
I don't see the necessity of doing this, reopen if necessary.

PR #4870, opened 2019-08-19
Fix warm up issue when provider's timestamp is bigger than local machine's timestamp.

aftersss:
The provider's timestamp may be bigger than local machine's timestamp, if this happens, uptime will be negative, then weight will use the default value configured, it breaks the warm up.

This PR tries to fix this.

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/AbstractLoadBalance.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/AbstractLoadBalance.java
@@ -24,9 +24,9 @@ import org.apache.dubbo.rpc.cluster.LoadBalance;
 
 import java.util.List;
 
+import static org.apache.dubbo.common.constants.CommonConstants.TIMESTAMP_KEY;
 import static org.apache.dubbo.rpc.cluster.Constants.DEFAULT_WARMUP;
 import static org.apache.dubbo.rpc.cluster.Constants.DEFAULT_WEIGHT;
-import static org.apache.dubbo.rpc.cluster.Constants.REMOTE_TIMESTAMP_KEY;
 import static org.apache.dubbo.rpc.cluster.Constants.WARMUP_KEY;
 import static org.apache.dubbo.rpc.cluster.Constants.WEIGHT_KEY;
 
@@ -73,7 +73,7 @@ public abstract class AbstractLoadBalance implements LoadBalance {
     protected int getWeight(Invoker<?> invoker, Invocation invocation) {
         int weight = invoker.getUrl().getMethodParameter(invocation.getMethodName(), WEIGHT_KEY, DEFAULT_WEIGHT);
         if (weight > 0) {
-            long timestamp = invoker.getUrl().getParameter(REMOTE_TIMESTAMP_KEY, 0L);
+            long timestamp = invoker.getUrl().getParameter(TIMESTAMP_KEY, 0L);
             if (timestamp > 0L) {
                 long uptime = System.currentTimeMillis() - timestamp;
                 if (uptime >= Integer.MAX_VALUE) {
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/AbstractLoadBalance.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/AbstractLoadBalance.java
@@ -75,10 +75,16 @@ public abstract class AbstractLoadBalance implements LoadBalance {
         if (weight > 0) {
             long timestamp = invoker.getUrl().getParameter(REMOTE_TIMESTAMP_KEY, 0L);
             if (timestamp > 0L) {
-                int uptime = (int) (System.currentTimeMillis() - timestamp);
+                long uptime = System.currentTimeMillis() - timestamp;
+                if (uptime >= Integer.MAX_VALUE) {
+                    return weight;
+                }
+                else if (uptime < 0) {
+                    return 1;
+                }
                 int warmup = invoker.getUrl().getParameter(WARMUP_KEY, DEFAULT_WARMUP);
                 if (uptime > 0 && uptime < warmup) {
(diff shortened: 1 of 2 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/AbstractLoadBalance.java
```diff
@@ -75,10 +75,16 @@ protected int getWeight(Invoker<?> invoker, Invocation invocation) {
         if (weight > 0) {
             long timestamp = invoker.getUrl().getParameter(REMOTE_TIMESTAMP_KEY, 0L);
```
beiwei30: pls. change `REMOTE_TIMESTAMP_KEY` to `TIMESTAMP_KEY`. Pls. also remove org.apache.dubbo.rpc.cluster.Constants#REMOTE_TIMESTAMP_KEY since it's now useless.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/AbstractLoadBalance.java
```diff
@@ -75,10 +75,16 @@ protected int getWeight(Invoker<?> invoker, Invocation invocation) {
         if (weight > 0) {
             long timestamp = invoker.getUrl().getParameter(REMOTE_TIMESTAMP_KEY, 0L);
             if (timestamp > 0L) {
-                int uptime = (int) (System.currentTimeMillis() - timestamp);
+                long uptime = System.currentTimeMillis() - timestamp;
```
beiwei30: should we change this method to

```java
    protected int getWeight(Invoker<?> invoker, Invocation invocation) {
        int weight = invoker.getUrl().getMethodParameter(invocation.getMethodName(), WEIGHT_KEY, DEFAULT_WEIGHT);
        if (weight > 0) {
            long timestamp = invoker.getUrl().getParameter(TIMESTAMP_KEY, 0L);
            if (timestamp > 0L) {
                int uptime = (int) Math.max(System.currentTimeMillis() - timestamp, 0);
                int warmup = invoker.getUrl().getParameter(WARMUP_KEY, DEFAULT_WARMUP);
                if (uptime >= 0 && uptime < warmup) {
                    weight = calculateWarmupWeight(uptime, warmup, weight);
                }
            }
        }
        return weight >= 0 ? weight : 0;
    }
```

beiwei30: the diff is

```diff
index ed6f3c4e9..14640d4fb 100644
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/AbstractLoadBalance.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/AbstractLoadBalance.java
@@ -24,6 +24,7 @@ import org.apache.dubbo.rpc.cluster.LoadBalance;
 
 import java.util.List;
 
+import static org.apache.dubbo.common.constants.CommonConstants.TIMESTAMP_KEY;^M
 import static org.apache.dubbo.rpc.cluster.Constants.DEFAULT_WARMUP;
 import static org.apache.dubbo.rpc.cluster.Constants.DEFAULT_WEIGHT;
 import static org.apache.dubbo.rpc.cluster.Constants.REMOTE_TIMESTAMP_KEY;
@@ -73,11 +74,11 @@ public abstract class AbstractLoadBalance implements LoadBalance {
     protected int getWeight(Invoker<?> invoker, Invocation invocation) {
         int weight = invoker.getUrl().getMethodParameter(invocation.getMethodName(), WEIGHT_KEY, DEFAULT_WEIGHT);
         if (weight > 0) {
-            long timestamp = invoker.getUrl().getParameter(REMOTE_TIMESTAMP_KEY, 0L);
+            long timestamp = invoker.getUrl().getParameter(TIMESTAMP_KEY, 0L);^M
             if (timestamp > 0L) {
-                int uptime = (int) (System.currentTimeMillis() - timestamp);
+                int uptime = (int) Math.max(System.currentTimeMillis() - timestamp, 0);^M
                 int warmup = invoker.getUrl().getParameter(WARMUP_KEY, DEFAULT_WARMUP);
-                if (uptime > 0 && uptime < warmup) {
+                if (uptime >= 0 && uptime < warmup) {^M
                     weight = calculateWarmupWeight(uptime, warmup, weight);
                 }
             }
```

aftersss: I don't think so,  if the provider is started 24 days ago, then `System.currentTimeMillis() - timestamp` will be bigger than `Integer.MAX_VALUE`.

Consider that `System.currentTimeMillis() = 1566209746000`(2019-08-19 18:15:46), and `timestamp = 1561914711000`(2019-07-01 01:11:51),  then `int uptime = (int) Math.max(System.currentTimeMillis() - timestamp, 0)` is `67704`, this will result in a smaller weight.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/AbstractLoadBalance.java
```diff
@@ -73,12 +73,18 @@ static int calculateWarmupWeight(int uptime, int warmup, int weight) {
...
         if (weight > 0) {
-            long timestamp = invoker.getUrl().getParameter(REMOTE_TIMESTAMP_KEY, 0L);
+            long timestamp = invoker.getUrl().getParameter(TIMESTAMP_KEY, 0L);
             if (timestamp > 0L) {
-                int uptime = (int) (System.currentTimeMillis() - timestamp);
+                long uptime = System.currentTimeMillis() - timestamp;
+                if (uptime >= Integer.MAX_VALUE) {
+                    return weight;
+                }
+                else if (uptime < 0) {
```
FinalVersion, line 82: uptime == 0 case is not correctly handled

PR #4877, opened 2019-08-19
Speed up registry destroy.

aftersss:
When we shutdown our program, we found that it costs too much time at destroying registry from zookeeper. Consider that if we have 200 urls registered to zookeeper, and every unregister will cost 100ms, it will cost 200*100 = 20000ms to shutdown the program.

This PR uses a thread pool to reduce the shutdown time.

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistry.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/AbstractRegistry.java
@@ -45,11 +45,16 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


beiwei30:
> When we shutdown our program, we found that it costs too much time at destroying registry from zookeeper. Consider that if we have 200 urls registered to zookeeper, and every unregister will cost 100ms, it will cost 200*100 = 20000ms to shutdown the program.
> 
> This PR uses a thread pool to reduce the shutdown time.

@aftersss, with your solution, how much time can it save from the shutdown process?

aftersss:
We have `153` providers In one of our applications, when we shutdown this application, `AbstractRegistry.destroy` costs `5394ms`  (I have tested it several times, this is the average time).

If we change `AbstractRegistry.destroy` to let unregister happens concurrently,  `AbstractRegistry.destroy` costs only `213ms`.

chickenlj:
How about the batch operation of zookeeper, have you ever tried?

chickenlj:
Please solve the conflicts and reopen.

More importantly, I am not sure if there are any side effects if we make the unregister process parallel and async.

chickenlj:
I tend to keep it as it is now and wait for more feedback.

PR #4879, opened 2019-08-19
dubbo zookeeper registry too slow

qixiaobo:
ref to https://github.com/apache/dubbo/issues/4828
I just add concurrentHashSet to avoid checking persistent node exist


qixiaobo:
Sorry for auto format

qixiaobo:
image
Sorry for not committing as is

Issue #4828, opened 2019-08-14
dubbo zookeeper registry too slow


qixiaobo:
- [ ] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.6.x
* Operating System version: *nix
* Java version: 1.7

### Steps to reproduce this issue

1. xxx
2. xxx
3. xxx

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result

What do you expected from the above steps？

### Actual Result
![image](https://user-images.githubusercontent.com/2881751/63033889-f3989a80-beea-11e9-8d7a-7b54021aa7cd.png)

What actually happens?

We can see code like this

```java
   @Override
    public void create(String path, boolean ephemeral) {
        if (!ephemeral) {
            if (checkExists(path)) {
                return;
            }
        }
        int i = path.lastIndexOf('/');
        if (i > 0) {
            create(path.substring(0, i), false);
        }
        if (ephemeral) {
            createEphemeral(path);
        } else {
            createPersistent(path);
        }
    }
```
You can see recursive here.
If we have a lot of dubbo reference or service.
This method would execute too many times and cost a large time.
In fact , if dubbo group is same then the path may no need to check every time

qixiaobo:
![image](https://user-images.githubusercontent.com/2881751/63034254-89342a00-beeb-11e9-9543-3c2f838a624b.png)

qixiaobo:
2019-08-15 14:17:54,187 [myid:] - INFO  [ProcessThread(sid:0 cport:2181)::PrepRequestProcessor@653] - Got user-level KeeperException when processing sessionid:0x1004228f12d5154 type:create cxid:0x495 zxid:0x7c8c87 txntype:-1 reqpath:n/a Error Path:/f6-test/XXX/providers Error:KeeperErrorCode = NodeExists for /f6-test/XXX/providers

qixiaobo:
tail -n 10000 zookeeper.out |grep  NodeExists|wc -l
9732
We can see almost 97% NodeExistsException

PR #4880, opened 2019-08-19
dubbo zookeeper registry too slow

qixiaobo:
ref to #4828
I just add concurrentHashSet to avoid checking persistent node exist

### DIFF
--- a/dubbo-remoting/dubbo-remoting-zookeeper/src/main/java/org/apache/dubbo/remoting/zookeeper/support/AbstractZookeeperClient.java
+++ b/dubbo-remoting/dubbo-remoting-zookeeper/src/main/java/org/apache/dubbo/remoting/zookeeper/support/AbstractZookeeperClient.java
@@ -174,6 +174,9 @@ public abstract class AbstractZookeeperClient<TargetDataListener, TargetChildLis
     public void create(String path, String content, boolean ephemeral) {
         if (checkExists(path)) {
             delete(path);
+            if(!ephemeral){
+                persistentExistNodePath.remove(path);
+            }
         }
         int i = path.lastIndexOf('/');
         if (i > 0) {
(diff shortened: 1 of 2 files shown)


dubbo-remoting/dubbo-remoting-zookeeper/src/main/java/org/apache/dubbo/remoting/zookeeper/support/AbstractZookeeperClient.java
```diff
@@ -45,6 +46,8 @@
 
     private volatile boolean closed = false;
 
+    private final Set<String>  persistentExistNodePath = Sets.newConcurrentHashSet();
+
```
tswstarplanet: If a path is deleted, the path should be removed from the set.

qixiaobo: > If a path is deleted, the path should be removed from the set.
From me , I think we delete only the non persistent node. I will have a look with the code

qixiaobo:
@tswstarplanet  You are right, I find the metaDataReport delete the persist node

dubbo-remoting/dubbo-remoting-zookeeper/src/main/java/org/apache/dubbo/remoting/zookeeper/support/AbstractZookeeperClient.java
beiwei30: pls. avoid of using guava. It's introduced from

dubbo-remoting/dubbo-remoting-zookeeper/src/main/java/org/apache/dubbo/remoting/zookeeper/support/AbstractZookeeperClient.java
```diff
@@ -16,6 +16,7 @@
  */
 package org.apache.dubbo.remoting.zookeeper.support;
 
+import com.google.common.collect.Sets;
```
beiwei30: pls. avoid of using guava directly.

qixiaobo: OK, I will modify the code

Issue #4693, opened 2019-07-30
@EnableDubbo 不搜索scanBasePackages指定目录


w548544375:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.3
* Operating System version: MacOS Catalina 10.15 Beta
* Java version: 1.8
### project struct
![image](https://user-images.githubusercontent.com/832040/62096991-240bf200-b2b8-11e9-8d4e-c9cee2802c06.png)
### DubboConfiguration
```
@Configuration
@EnableDubbo(scanBasePackages = "com.weiguang.www.dubboprovider.service")
@PropertySource(value = "classpath:dubbo-provider.properties")
public class DubboConfiguration {

    @Bean
    public ProviderConfig providerConfig(){
        ProviderConfig providerConfig = new ProviderConfig();
        providerConfig.setTimeout(3000);
        return providerConfig;
    }

}

```
### log
```
[DUBBO] No Spring Bean annotating Dubbo's @Service was found under package[com.weiguang.www.dubboconsumer.config], dubbo version: 2.7.3, current host: 192.168.1.111
```
搜索的是config文件所在的目录，并不是 ``` @EnableDubbo(scanBasePackages = "com.weiguang.www.dubboprovider.service") ``` 指定的目录。

导致在zookeeper中,provider内容为空，
```
[zk: 127.0.0.1:2181(CONNECTED) 0] ls /
[dubbo, zookeeper]
[zk: 127.0.0.1:2181(CONNECTED) 1] ls /dubbo
[com.weiguang.www.dubbointerface.service.HelloService]
[zk: 127.0.0.1:2181(CONNECTED) 2] ls /dubbo/com.weiguang.www.dubbointerface.service.HelloService
[configurators, consumers, providers, routers]
[zk: 127.0.0.1:2181(CONNECTED) 3] ls /dubbo/com.weiguang.www.dubbointerface.service.HelloService/providers
[]
[zk: 127.0.0.1:2181(CONNECTED) 4] 
```
最终导致consumer 调用时出现 No provider availabe 错误:
``` 
    "timestamp": "2019-07-30T02:39:07.740+0000",
    "status": 500,
    "error": "Internal Server Error",
[...]

PR #4895, opened 2019-08-20
[Dubbo-4693] fix EnableDubbo's alias problem (no effect though declared)

jasonjoo2010:

fix the issued prompted in #4693 

## Brief changelog

Use `AnnotatedElementUtils.findMergedAnnotation` to make inherited annotations' `@AliasFor` take effects.

## Verifying this change

@EnableDubbo(scanBasePackages = "") should bring the property to `@EnableDubboConfig`

## Reference
https://docs.spring.io/spring/docs/4.2.0.RC3_to_4.2.0.RELEASE/Spring%20Framework%204.2.0.RELEASE/org/springframework/core/annotation/AliasFor.html

```
Similarly, AnnotatedElementUtils supports explicit meta-annotation attribute overrides when @AliasFor is used within an annotation hierarchy. 
```


jasonjoo2010:
> @jasonjoo2010 would you mind to take a look at [apache/dubbo-samples@7e25d4e](https://github.com/apache/dubbo-samples/commit/7e25d4e)? I cannot reproduce the issue.

Hi, luo

Sorry for the latency.

I think problem will occur when using spring-boot to initialize the project.
See `https://github.com/jasonjoo2010/dubbo-samples/tree/EnableDubbo` for sample project `dubbo-samples-annotation-springboot`.

But surely there is a limitation for this PR. `AnnotatedElementUtils.findMergedAnnotationAttributes` will only available since spring `4.2`. I don't know whether it's accepted.

jasonjoo2010:
@beiwei30 

You can reproduce it by add a breakpoint on `DubboComponentScanRegistrar.getPackagesToScan` (L102) before running `AnnotationProviderBootstrap`. Break at that point and check the variable `basePackages` and will find it is empty. It's different to the result in `AnnotationConfigApplicationContext`.

jasonjoo2010:
Update:

Recently I just found some ways which will cause the inheriting alias fail but didn't dig the root cause (It maybe not important because we can't make everyone not to do it like this way):

1. Put a `@ComponentScan` in your bootable class `XXX.class` (It may be referenced as `SpringApplication.run(XXXX.class, argv)` )
2. Put a @EnableDubbo in where you like (in the same class or a separated class)

Then the information will not be brought to referenced annotations. And `@ComponentScan` can be also introduced by introducing `@SpringBootApplication`.

It will also be reproduced by the pair of @Configuration + @EnableAutoConfiguration.
Generally I think if the tag is injected by scanning but not imported manually the issue will occur.

So I think it's a kind of SpringBoot issue. Just like the strange flow of asynchronized servlet in SpringBoot compared SpringMVC in container(standalone tomcat)[1].

Maybe we can't unify them into the same processing flow but to make some changes adapting for the special environment.

You can make more tests in the repository provided in previous comment.

[1] https://github.com/alibaba/Sentinel/pull/310

mercyblitz:
Sorry, this PR will not be merged into, thus `@org.apache.dubbo.config.spring.context.annotation.DubboComponentScan` has to be compatible with Spring Framework 3.2 +, however, `@AliasFor` is dependent on  Spring 4.2 and above.

jasonjoo2010:
> Sorry, this PR will not be merged into, thus `@org.apache.dubbo.config.spring.context.annotation.DubboComponentScan` has to be compatible with Spring Framework 3.2 +, however, `@AliasFor` is dependent on Spring 4.2 and above.

I am curious that is this treated a bug or just let users ignore it? (LOST of scanBasePackages)
After all compatibility issue what you just mentioned has many alternative solutions.

Issue #4882, opened 2019-08-19
在使用配置中心动态更新配置的时候,无法更新多层配置


ZhouZiHe:
使用apollo 作为配置中心的时候,目前无法进行配置map属性
![image](https://user-images.githubusercontent.com/20511319/63263546-85255500-c2bb-11e9-954c-4cf24aec90f4.png)
在源码中判断set方法的时候进行自动去除这个方法
![image](https://user-images.githubusercontent.com/20511319/63263499-60c97880-c2bb-11e9-94ec-173bddee0078.png)

ZhouZiHe:
类似于parameters这类参数应该是无法通过配置中心进行动态配置的

mzorro:
The other day I was just considering this.
If we are gonna support overriding of `Map` properties, we should consider it's format in config center, I was thinking something like `a=1&b=2&c=3`, then we can parse it with `URL.valueOf("?"+value).getParameters()`.
I'll try to make a PR this week.

PR #4899, opened 2019-08-21
[Dubbo-4882] Add support for overriding Map properties in AbstractConfig.refresh

mzorro:
Solution for issue #4882 
Add support for overriding Map properties in AbstractConfig.refresh, for example if we have configuration `dubbo.application.parameters=a=1&b=2`, then `ApplicationConfig.refresh` will do `parameters.put("a", "1");parameters.put("b", "2")`.

**The format has been changed to `dubbo.application.parameters=[{a:1}, {b:2}}]`**


beiwei30:
@mzorro would you mind take a look at https://github.com/mzorro/dubbo/pull/1? My idea is we should try to limit changes in AbstractConfig by splitting the logics into smaller methods, instead of moving them into other places. 

The reason is the current implementation in AbstractConfig is not good enough. It should help us to make it better if we put all relevant code in one single class.

mzorro:
> @mzorro pls. check my comment.

thanks for the help, I have merged your PR

chickenlj:
change format to `dubbo.application.parameters=[{a:1}, {b:2}]`

mzorro:
> change format to `dubbo.application.parameters=[{a:1}, {b:2}]`

@chickenlj why do you think this format is more readable? It's not even standard JSON.

PR #4921, opened 2019-08-23
simplifed and beautify code

leechor:

simplifed and beautify code.not change behaviour.
pass.


beiwei30:
@leechor would you mind to take look into CI? there are unit tests failing.

leechor:
> @leechor would you mind to take look into CI? there are unit tests fa

misunderstand a method's function, and the unit test of this method not contains this case, have fix it.

leechor:
> @leechor would you mind to take look into CI? there are unit tests failing.

misunderstand a method's function, and the unit test of this method not contains this case, have fix it.

PR #4935, opened 2019-08-25
considering methodStatus.active == Integer.MAX_VALUE

sunchuanleihit:

in case of methodStatus.active is already equals Integer.MAX_VALUE, methodStatus.active.incrementAndGet() will become Integer.MIN_VALUE

## Brief changelog

avoid potential value out of bounds

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
@@ -112,6 +112,9 @@ public class RpcStatus {
         max = (max <= 0) ? Integer.MAX_VALUE : max;
         RpcStatus appStatus = getStatus(url);
         RpcStatus methodStatus = getStatus(url, methodName);
+        if (methodStatus.active.get() == Integer.MAX_VALUE) {
+            return false;
+        }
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
```diff
@@ -112,6 +112,9 @@ public static boolean beginCount(URL url, String methodName, int max) {
         max = (max <= 0) ? Integer.MAX_VALUE : max;
         RpcStatus appStatus = getStatus(url);
         RpcStatus methodStatus = getStatus(url, methodName);
+        if (methodStatus.active.get() >= max) {
```
beiwei30: should we change this to a particluar check, otherwise it would be confusing:

```java
if (methodStatus.active.get() == Integer.MAX_VALUE) {
```

It would be nice if you put one comment above to explain why we need this.

sunchuanleihit:
> @sunchuanleihit pls. check my comment.

you are right

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java
sunchuanleihit: you are right, I have changed, pls verify

PR #4978, opened 2019-09-01
Simplified code : replace explicit type with <>

freekry:
Simplified code : replace explicit type with <>


freekry:
I don't know why it can be compiled in JDK8, but not in JDK11. I only did the same for the submitted file.

PR #4988, opened 2019-09-03
[Dubbo-compatible with dubbox]

632657122:

兼容dubbo和dubbox

## Brief changelog

dubbox修改了编码格式使得无法与现行dubbo版本互通，所以增加了与dubbox互相调用的兼容性

## Verifying this change

已测试可以互相调用

- [✔️] Make sure there is a [GITHUB_issue](https://github.com/apache/dubbo/issues) field for the change (usually before you start working on it). Trivial changes like typos do not require a GITHUB issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [ ✔️] Format the pull request title like `[Dubbo-XXX] Fix UnknownException when host config not exist #XXX`. Each commit in the pull request should have a meaningful subject line and body.
- [ ✔️] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [✔️ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add sample in [dubbo samples](https://github.com/apache/dubbo-samples) project.
- [✔️ ] Run `mvn clean install -DskipTests=false` & `mvn clean test-compile failsafe:integration-test` to make sure unit-test and integration-test pass.
- [ ✔️] If this contribution is large, please follow the [Software Donation Guide](https://github.com/apache/dubbo/wiki/Software-donation-guide).

### DIFF
--- a/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DecodeableRpcInvocation.java
+++ b/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DecodeableRpcInvocation.java
@@ -101,6 +101,15 @@ public class DecodeableRpcInvocation extends RpcInvocation implements Codec, Dec
         setAttachment(VERSION_KEY, in.readUTF());
 
         setMethodName(in.readUTF());
+        
+        /**
+         * 兼容dubbox 和dubbo
+         * 如果发现consumer是dubbox 则采取dubbox的编码格式
+         */
+        if (!Version.DEFAULT_DUBBO_PROTOCOL_VERSION.equals(dubboVersion)) {
+            in.readInt();
+        }
+        
         try {
             Object[] args;
             Class<?>[] pts;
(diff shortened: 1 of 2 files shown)


632657122:
> We cannot accept this change, @632657122
> 
> Would you mind to take look into org.apache.dubbo.common.Version#isSupportResponseAttachment:
> 
> ```java
>     public static boolean isSupportResponseAttachment(String version) {
>         if (StringUtils.isEmpty(version)) {
>             return false;
>         }
>         // for previous dubbo version(2.0.10/020010~2.6.2/020602), this version is the jar's version, so they need to
>         // be ignore
>         int iVersion = getIntVersion(version);
>         if (iVersion >= 2001000 && iVersion < 2060300) {
>             return false;
>         }
> 
>         // 2.8.x is reserved for dubbox
>         if (iVersion >= 2080000 && iVersion < 2090000) {
>             return false;
>         }
> 
>         return iVersion >= LOWEST_VERSION_FOR_RESPONSE_ATTACHMENT;
> ```
> 
> I believe we can come up with a better solution if a "isDubboX()" method is introduced.

where is the isDubbox() method supposed to be used? I think the main reason why dubbox cannot work with dubbo is their way to encode/decode request is different, so we just need to judge them when encode/decode request. maybe i ignored something,could you please just give some advice?

chickenlj:
I would suggest using multi-protocol mechanism of Dubbo Framework for migrating from dubbox to dubbo.

PR #4998, opened 2019-09-04
[refactor] use java 8 grammar to simplify some code

caidezhi:
refactor some code by using java 8 grammar


caidezhi:
@kezhenxu94 many thanks for reviewing. I have resolved conflicts.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
```diff
@@ -1,141 +1,142 @@
-/*
```
kezhenxu94: so many unexpected changes

PR #5007, opened 2019-09-05
avoid resize

fanghuabing:

The threshold of an HashMap is the product of current capacity and load factor.

Threshold = (Current Capacity) * (Load Factor)

For example, if the HashMap is created with initial capacity of 16 and load factor of 0.75f, then threshold will be,

Threshold = 16 * 0.75 = 12

That means, the capacity of the HashMap is increased from 16 to 32 after the 12th element (key-value pair) is added into the HashMap.


fanghuabing:
https://javaconceptoftheday.com/initial-capacity-and-load-factor-of-hashmap-in-java/

fanghuabing:
- 

-

PR #5053, opened 2019-09-11
add BroadcastCluster's NAME

thewindkee:

add BroadcastCluster's NAME

## Brief changelog

add BroadcastCluster's NAME

## Verifying this change

no need.


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/BroadcastCluster.java
```diff
@@ -27,6 +27,8 @@
  */
 public class BroadcastCluster implements Cluster {
 
+    public static final String NAME = "broadcast";
```
sonymoon: ^_^, each `Cluster` has a name except this one.

Issue #5049, opened 2019-09-11
support specify protocol for Service annotation


haiyang1985:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.3
* Operating System version: macOs sierra
* Java version: 1.8

### Steps to reproduce this issue

1. define two dubbo protocols with different port, like below.
```
@Configuration
public class DubboConfiguration {

  @Bean
  public ProtocolConfig config1() {
    ProtocolConfig config = new ProtocolConfig();
    config.setId("dubbo1");
    config.setName("dubbo");
    config.setPort(20880);
    return config;
  }

  @Bean
  public ProtocolConfig config2() {
    ProtocolConfig config = new ProtocolConfig();
    config.setId("dubbo2");
    config.setName("dubbo");
    config.setPort(20770);
    return config;
  }
}
```
2. specify dubbo bean id "dubbo1" as Service annotation protocol, but not spring bean name "config1".
```
@Service(protocol = "dubbo1")
```
3. startup the provider and see output log, and see application failed to start without dubbo1 protocol.

```
***************************
APPLICATION FAILED TO START
***************************

Description:

A component required a bean named 'dubbo1' that could not be found.
```

If we are specify config1 as the protocol, the application can start up succeed. When ServiceBean initialize, it will get reference bean from spring, cannot recognize dubbo properties.
@Service(protocol = "config1")

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result

The application start up succeed.

### Actual Result

The application start up failed without dubbo1 protocol.

PR #5063, opened 2019-09-12
[Dubbo-5049]support @Bean on Config class to config id as alias

sonymoon:
as #5049 mentioned, when `@Service(protocol = "dubbo1")` is annotated on a class, the provider starts failed, `dubbo1` is the `id` property of `ProtocolConfig`.
```java
@Bean
  public ProtocolConfig config1() {
    ProtocolConfig config = new ProtocolConfig();
    config.setId("dubbo1");
    config.setName("dubbo");
    config.setPort(20880);
    return config;
  }
```
The failure cause is `dubbo1` is not an name or alias of `ProtocolConfig` bean.

When using xml way:
```xml
<dubbo:application name="demo-provider"/>

    <dubbo:registry address="zookeeper://127.0.0.1:2181"/>

    <dubbo:protocol name="dubbo" id="dubbo1"/>

    <bean id="demoService" class="org.apache.dubbo.demo.provider.DemoServiceImpl"/>

    <dubbo:service interface="org.apache.dubbo.demo.DemoService" ref="demoService" protocol="dubbo1"/>
```
xml config with` id` property assigned with `dubbo1`, the provider starts successfully.

I think the annotation way should work fine as well.

### My suggestion approach
Use spring extension interfaces, such as:
1. BeanDefinitionRegistry: can do the registry alias stuff
2. BeanPostProcessor: can do `1`  after config beans's  initialization by implementing method `postProcessAfterInitialization`, which is already implemented by class `DubboConfigBindingBeanPostProcessor` with an empty method.

Actually, @haiyang1985 has a PR #5051 solving this issue:
But I think using `getBean` in `postProcessBeanDefinitionRegistry` method cause the initialization of ProtocolConfig's bean too early (bean registry phase in spring) as @mercyblitz mentioned

### DIFF
--- a/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/beans/factory/annotation/DubboConfigBindingBeanPostProcessor.java
+++ b/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/beans/factory/annotation/DubboConfigBindingBeanPostProcessor.java
@@ -17,6 +17,7 @@
 package org.apache.dubbo.config.spring.beans.factory.annotation;
 
 import org.apache.dubbo.common.utils.Assert;
+import org.apache.dubbo.common.utils.StringUtils;
 import org.apache.dubbo.config.AbstractConfig;
 import org.apache.dubbo.config.spring.context.annotation.DubboConfigBindingRegistrar;
 import org.apache.dubbo.config.spring.context.annotation.EnableDubboConfigBinding;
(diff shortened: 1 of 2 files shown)


haiyang1985:
The PR have not solved my issue. When user defined ProtocolConfig as a spring bean , service annotation cannot get the dubbo bean id. 

```
@Service(protocol = "dubbo1")
public class DubboDemoServiceImpl implements DubboDemoService {
}
```

```
@Configuration
public class DubboConfiguration {
  @Bean
  public ProtocolConfig getConfig() {
    ProtocolConfig config = new ProtocolConfig();
    config.setId("dubbo1");
    config.setName("dubbo");
    config.setPort(20770);
    return config;
  }
}
```

I've tried your solution in my local, still not able to get dubbo bean id.
```
***************************
APPLICATION FAILED TO START
***************************

Description:

A component required a bean named 'dubbo1' that could not be found.

```

sonymoon:
Can you provider a GitHub address which can reproduce this?

PR #5098, opened 2019-09-21
Fixes a bug in protobuf_json's deserialization feature

goodjava:
+ Fixes a bug in protobuf_json's deserialization feature
+ 参考:Fixes a bug in gson's deserialization feature #5099


chickenlj:
Thanks, but this has been fixed.

PR #5107, opened 2019-09-24
Simplify dubbo-common transitive dependencies

chickenlj:

remove unuseful dependencies.

## Brief changelog

dubbo-common/pom.xml

## Verifying this change

mvn clean compile


chickenlj:
fixes https://github.com/apache/dubbo/issues/5106

PR #5134, opened 2019-10-04
[Dubbo-5004] Performance tuning: Reduce object allocation

LinShunKang:

Reduce object allocation when get `Method` returnTypes.
The original code allocate 2 `Optional` objects, 2 `Consumer` objects and 1 `Iterator` object each time it gets `Method` returnTypes, it will increase GC pressure.

## Brief changelog

* Use ` != null` instead of  `Optional.ifPresent(Consumer)`
* Use `Map<String, List<MethodModel>>`  instead of  `Map<String, Set<MethodModel>>` in `ServiceModel`

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcInvocation.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcInvocation.java
@@ -19,6 +19,8 @@ package org.apache.dubbo.rpc;
 import org.apache.dubbo.common.URL;
 import org.apache.dubbo.common.utils.StringUtils;
 import org.apache.dubbo.rpc.model.ApplicationModel;
+import org.apache.dubbo.rpc.model.MethodModel;
+import org.apache.dubbo.rpc.model.ServiceModel;
 
 import java.io.Serializable;
 import java.lang.reflect.Method;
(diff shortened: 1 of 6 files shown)


htynkn:
`Optional` is more readable, but it may cause performance issue. 

Some related info:

https://stackoverflow.com/questions/54443944/gc-overhead-of-optionalt-in-java
https://stackoverflow.com/questions/34696884/performance-of-java-optional

I also did a search in github, spring and hibernate use `optional`
https://github.com/spring-projects/spring-boot/blob/21f1b3a79d1c2168aa0088e45368f3f75957b22c/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/CompositeHandlerAdapter.java#L56
https://github.com/spring-projects/spring-framework/blob/ad6231ad2989ca5428bb51afdf48a9112710ce1d/spring-core/src/main/java/org/springframework/core/annotation/MissingMergedAnnotation.java#L103
https://github.com/hibernate/hibernate-orm/blob/79a8f43ba57d9e4a1269c5027dfd063b18d4f7d9/hibernate-core/src/main/java/org/hibernate/bytecode/enhance/internal/bytebuddy/ByteBuddyEnhancementContext.java#L88

It's more like a trade off.

And also if we choose not use Optional, I suggest add related rule in checkstyle check

LinShunKang:
Yeah, it's a trade off. We can use `Optional` in non-critical paths to improve code readability. But, using `Optional` in critical paths increases the pressure on the GC.

Spring, Hibernate and Dubbo are different types of frameworks, and their workloads are very different. Dubbo carries thousands of requests per second, which requires higher performance stability, so maybe `! = null` is a better choice.

LinShunKang:
what？

chickenlj:
I agree with @LinShunKang on all the improvements made in this PR, I think we should prefer performance instead of readability in this case. 

The performance-2.7.x branch has been merged into master, could you please open a new one

LinShunKang:
@chickenlj I was already open new PR in [here](https://github.com/apache/dubbo/pull/5362).

PR #5201, opened 2019-10-18
[Dubbo-5200] fix javadoc for Serialization.getContentTypeId

### LINKED ISSUE #5200: javadoc of Serialization class need to correct
The javadoc of Serialization.getContentTypeId needs to correct, the content type id should less than ExchangeCodes.SERIALIZATION_MASK.

I will create a pr.

RojerAlone:
Fix #5200

### DIFF
--- a/dubbo-serialization/dubbo-serialization-api/src/main/java/org/apache/dubbo/common/serialize/Serialization.java
+++ b/dubbo-serialization/dubbo-serialization-api/src/main/java/org/apache/dubbo/common/serialize/Serialization.java
@@ -37,7 +37,8 @@ public interface Serialization {
 
     /**
      * Get content type unique id, recommended that custom implementations use values different with
-     * any value of {@link Constants} and don't greater than ExchangeCodes.SERIALIZATION_MASK (31)
+     * any value of {@link Constants} and don't greater than ExchangeCodec.SERIALIZATION_MASK (31) 
+     * because dubbo protocol use 5 bits to record serialization ID in header.
      *
      * @return content type id
      */
--- a/dubbo-serialization/dubbo-serialization-api/src/main/java/org/apache/dubbo/common/serialize/Serialization.java
+++ b/dubbo-serialization/dubbo-serialization-api/src/main/java/org/apache/dubbo/common/serialize/Serialization.java
@@ -36,7 +36,8 @@ import java.io.OutputStream;
 public interface Serialization {
 
     /**
-     * Get content type unique id, recommended that custom implementations use values greater than 20.
+     * Get content type unique id, recommended that custom implementations use values different with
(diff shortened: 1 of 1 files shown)


RojerAlone:
@htynkn is there something wrong? can review the code?

htynkn:
@RojerAlone 
sorry for late reply. Master branch is not stable at this moment, that's why PR is delay. but for this one, it's for javadoc document, it's ok to ignore CI result.

Just one question

what do you mean by ```and don't greater than ExchangeCodes.SERIALIZATION_MASK (31)```?
are you referring to https://github.com/apache/dubbo/blob/master/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java#L59

RojerAlone:
@htynkn yes, the serialization module's javadoc can't reference other module which it does't import, so I don't use `@link` doc tag.
Your meaning is I should describe why it's 31?

htynkn:
so I think there is a typo it should be  ExchangeCodec not ExchangeCodes.

another question is why it can't greater than 31？ the reason for it may not include in javadoc if it's too long. just make sure it coming from correct  source

RojerAlone:
I had used a very big number, and some exceptions are throwed, then I read the code, found that the serialization ID is stored in the dubbo protocol response header, it would be used to find which serialization protocol to use, code is https://github.com/apache/dubbo/blob/master/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java#L138.

In this line, dubbo use and function with SERIALIZATION_MASK and serialization tag, the root reason is dubbo use 5 bits to record serializtion id in protocol header, official document is https://dubbo.apache.org/zh-cn/blog/dubbo-protocol.html

So, if the developer use a big number bigger than SERIALIZATION_MASK, unexpected things would happened (can't find serialization protocol or use incorrect protocol to deserialize response etc.)

dubbo-serialization/dubbo-serialization-api/src/main/java/org/apache/dubbo/common/serialize/Serialization.java
```diff
@@ -36,7 +36,8 @@
 public interface Serialization {
 
     /**
-     * Get content type unique id, recommended that custom implementations use values greater than 20.
+     * Get content type unique id, recommended that custom implementations use values different with
+     * any value of {@link Constants} and don't greater than ExchangeCodes.SERIALIZATION_MASK (31)
```
RojerAlone: how about updating it to :
```
and don't greater than 31 because dubbo protocol use 5 bits to record serialization ID in header.
```

htynkn:
thanks for clarifying. I'm going to merge please ignore CI fail

PR #5319, opened 2019-11-13
Support specify service name when subscribing using Service Discovery mechainism.

chickenlj:
check sample dubbo-demo/dubbo-call-sc

works like below:
```xml
<dubbo:registry address="consul://127.0.0.1:8500">
        <dubbo:parameter key="registry-type" value="service"/>
    </dubbo:registry>

    <dubbo:reference provided-by="spring-cloud-provider-for-dubbo" id="userService" check="false" protocol="rest"
                     interface="org.apache.dubbo.samples.microservices.sc.rest.UserService"/>
```


cvictory:
单元测试没通过。

PR #5346, opened 2019-11-18
fix the consul always pull service list

CodingSinger:


chickenlj:
-1 means we want to pull the latest service instance list, and that's what we want to do.

chickenlj:
Sorry, maybe I am wrong and you are right. https://www.consul.io/api/features/blocking.html

PR #5367, opened 2019-11-20
Solve the compilation error of org.apache.dubbo.remoting.http.tomcat.tomcathttpbindertest ා shouldablehandlerequestfortomcatbinder method

liujianbigboss:

Solve the compilation error of org.apache.dubbo.remoting.http.tomcat.tomcathttpbindertest ා shouldablehandlerequestfortomcatbinder method

## Brief changelog

Introducing httpclient package needed for compilation into POM file under Dubbo remoting HTTP module


chickenlj:
As the following dependency gives transitive dependency to httpclient, I don't think it's necessary.
```xml
      <dependency>
            <groupId>org.apache.httpcomponents</groupId>
            <artifactId>fluent-hc</artifactId>
            <version>4.5.5</version>
            <scope>test</scope>
        </dependency>
```

chickenlj:
Reopen if necessary.

PR #5403, opened 2019-11-29
improve InvokerInvocationHandler#invoke

hengyunabc:


htynkn:
This change is quite clean to me. but I'm a litte concern about code readability.

Original code is check method name first and then check params, it's straightforward to understand.
Refactored code will check params first and then check method name.

I would like to wait sometime to see other opinion.

lkj41110:
> This change is quite clean to me. but I'm a litte concern about code readability.
> 
> Original code is check method name first and then check params, it's straightforward to understand.
> Refactored code will check params first and then check method name.
> 
> I would like to wait sometime to see other opinion.

`Original code is check method name first and then check params`，我也比较同意这个。
用`&&`时，当前面的表达式不成立的时候不会对参数个数进行判断了，而这个修改先判断参数个数，理论上增加了判断的次数

hengyunabc:
先判断 length，是因为这个最高效，远比判断多次String equals要快。

chickenlj:
I think we can sacrifice readability for performant in this case.

PR #5430, opened 2019-12-04
[Dubbo-Optimize] Simple performance optimization

LinShunKang:

* Reduce object allocation:
   - Reduce `Iterator` instance allocation when get `MethodDescriptor` in `ServiceDescriptor`.
   - Reduce `StringBuilder` instance allocation  and memory copy in `DecodeableRpcResult.decode()` and `DubboMonitor.send()`.

* Optimizing `String` usage for `URL` and `UrlUtils`.

## Brief changelog

* Use `Map<String, List<MethodDescriptor>>` rather than `Map<String, Set<MethodDescriptor>>` to store `methods`
* Add `Logger.isDebugEnabled()` before invoke `Logger.debug()` in `DecodeableRpcResult.decode()`
* Use `String.indexOf(char)` instead of `String.indexOf(String)` for single character string.
* Use `String.indexOf(char)` instead of `String.contains(String)` for single character string.

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
@@ -513,7 +513,7 @@ public final class StringUtils {
     }
 
     public static boolean isNotContains(String str, char ch) {
-        return isEmpty(str) || str.indexOf(ch) < 0;
+        return !isContains(str, ch);
     }
 
     /**
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
@@ -508,6 +508,14 @@ public final class StringUtils {
         return isNotEmpty(values) && isContains(COMMA_SPLIT_PATTERN.split(values), value);
     }
 
+    public static boolean isContains(String str, char ch) {
+        return isNotEmpty(str) && str.indexOf(ch) >= 0;
+    }
+
+    public static boolean isNotContains(String str, char ch) {
+        return isEmpty(str) || str.indexOf(ch) < 0;
+    }
+
     /**
      * @param values
      * @param value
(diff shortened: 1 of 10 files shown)


chickenlj:
LGTM.

dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
```diff
@@ -508,6 +508,14 @@ public static boolean isContains(String values, String value) {
         return isNotEmpty(values) && isContains(COMMA_SPLIT_PATTERN.split(values), value);
     }
 
+    public static boolean isContains(String str, char ch) {
```
satansk: `contains` seems to be a better name to me

dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
```diff
@@ -508,6 +508,14 @@ public static boolean isContains(String values, String value) {
         return isNotEmpty(values) && isContains(COMMA_SPLIT_PATTERN.split(values), value);
     }
 
+    public static boolean isContains(String str, char ch) {
+        return isNotEmpty(str) && str.indexOf(ch) >= 0;
+    }
+
+    public static boolean isNotContains(String str, char ch) {
```
satansk: Can be implemented by the above method.

chickenlj: I think what satansk suggests is quite reasonable, could you please take a look ?

dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
LinShunKang: @satansk I use `isContains` because of there has already two methods named `isContains` in `StringUtils` and `isXXX` is seems to be a better name for `XXXUtils`

dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
LinShunKang: OK，I will rewrite it

PR #5465, opened 2019-12-12
Fix getColonSeparatedKey does not work in 'URL simplified' mode.

chickenlj:
Fixes https://github.com/apache/dubbo/issues/4621

```xml
   public String getColonSeparatedKey() {
        StringBuilder serviceNameBuilder = new StringBuilder();
        serviceNameBuilder.append(this.getServiceInterface());
        append(serviceNameBuilder, VERSION_KEY, false);
        append(serviceNameBuilder, GROUP_KEY, false);
        return serviceNameBuilder.toString();
    }
```


chickenlj:
Unstable ut failed the ci, merge directly.

PR #5469, opened 2019-12-12
fix:eureka InstanceInfo.InstanceStatus should be OUT_OF_SERVICE when unregister

MI-cool:

fix, eureka InstanceInfo.InstanceStatus should be OUT_OF_SERVICE when unregister

## Brief changelog

change InstanceInfo.InstanceStatus from UP to OUT_OF_SERVICE when unregister


chickenlj:
It's weird that the CI fails.

Issue #4310, opened 2019-06-14
dubbo spi 


huifer:
- [*] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [*] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.2
* Operating System version: windows10(18362.175)
* Java version: jdk12.0.1

### Steps to reproduce this issue

1. create interface
```
@SPI
public interface Robot {
    void sayHello();
}
```
2. implements Roobot
```
public class Bumblebee implements Robot {

    @Override
    public void sayHello() {
        System.out.println("Hello, I am Bumblebee.");
    }
}

public class OptimusPrime implements Robot {

    @Override
    public void sayHello() {
        System.out.println("Hello, I am Optimus Prime.");
    }
}
```
3. create a folder
`/MEAT-INF/dubbo/`
```
optimusPrime=com.huifer.dubbo.client.spi.OptimusPrime
bumblebee=com.huifer.dubbo.client.spi.Bumblebee

```
4. run demo 
```
public class SpiDemo {

    public static void main(String[] args) {

        ExtensionLoader<Robot> extensionLoader =
                ExtensionLoader.getExtensionLoader(Robot.class);
        Robot optimusPrime = extensionLoader.getExtension("optimusPrime");
        optimusPrime.sayHello();
        Robot bumblebee = extensionLoader.getExtension("bumblebee");
        bumblebee.sayHello();

    }

}
```

Pls. provide [GitHub address](https://github.com/huifer/javaBook-src/tree/master/distributed/dubbo/dubbo-client/src/main/java/com/huifer/dubbo/client/spi) to reproduce this issue.

### Expected Result
Console output
```
Hello, I am Optimus Prime.
Hello, I am Bumblebee.
```

### Actual Result

- It has an exception and does not find optimusPrime

```
Exception in thread "main" java.lang.IllegalStateException: No such extension com.huifer.dubbo.client.spi.Robot by name optimusPrime
	at org.apache.dubbo.common.extension.ExtensionLoader.findException(ExtensionLoader.java:520)
	at org.apache.dubbo.common.extension.ExtensionLoader.createExtension(ExtensionLoader.java:527)
	at org.apache.dubbo.common.extension.ExtensionLoader.getExtension(ExtensionLoader.java:351)
[...]

PR #5491, opened 2019-12-16
fix potential performance drop of URL operations introduced in #4310

chickenlj:
Avoid map instance recreation.


chickenlj:
Thanks @zhaixiaoxiang , I will merge it now.

PR #5516, opened 2019-12-20
add auto-generated Regsitry and Protocol to ConfigManager

chickenlj:
When user does not specify any ProtocolConfig or RegistryConfig, the auto-generated Regsitry and Protocol should be added to ConfigManager


CodingSinger:
LGTM

PR #5521, opened 2019-12-21
add epoll support for netty4

chickenlj:
Referenced from
https://github.com/apache/dubbo/pull/5024/
https://github.com/apache/dubbo/pull/4493/
https://github.com/apache/dubbo/pull/4012/
https://github.com/apache/dubbo/pull/2738/

Thank you all very much.

### DIFF
--- a/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyClient.java
+++ b/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyClient.java
@@ -34,7 +34,12 @@ import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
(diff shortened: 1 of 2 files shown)


chickenlj:
need to add a switch and enable some more Epoll specific properties supported by Netty.

chickenlj:
https://stackoverflow.com/questions/23465401/why-native-epoll-support-is-introduced-in-netty

dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyClient.java
```diff
@@ -54,7 +59,17 @@
     /**
      * netty client bootstrap
      */
-    private static final NioEventLoopGroup nioEventLoopGroup = new NioEventLoopGroup(Constants.DEFAULT_IO_THREADS, new DefaultThreadFactory("NettyClientWorker", true));
+    private static EventLoopGroup nioEventLoopGroup;
```
lkj41110: i think `EventLoopGroup` is better  than `nioEventLoopGroup`.

PR #5529, opened 2019-12-23
[Dubbo] refactoring to remove feature envy

klebertarcisio:
I found a refactoring opportunity on Dubbo project.
I believe the online method of the Online class can be improved. This method calls 8 other methods from other classes.

Besides that, the online method calls many times ProviderModel.RegisterStatedURL inner class. In my view, a part of the online method should be extracted to a new method, and this new method should be moved to RegisterStatedURL. What do you think?

The same logic is applied to the offline method.

The advantages of theses changes are:
Remove a feature envy (this method has 4 calls to the ProviderModel.RegisterStatedURL).

Regards


chickenlj:
Sorry, but the change cannot be applied.

ProviderModel is a basic class in the common module, it cannot depend on Registry and RegistryFactory.

chickenlj:
We may need to consider move RegisterStatedURL out of the common module to somewhere else.

PR #5550, opened 2019-12-27
Add a function:check if consumer reference new service methods is not…

HughYiXuan:
Add a function:check if consumer reference new service methods is not registered in the registry when the consumer starts

This situation occurs during actual use. In the SDK used this time, there is an additional method under the same service, but it was not noticed during the online process. When the method is actually called, an error occurs. 
1、When the consumer starts, it can not only detect the existence of the provider, but also whether the provider's methods meet the current use of the server
2、Check whether the check is configured or not, the difference is that if check is true, the program will be interrupted
## Brief changelog
Added a new method to compare the method used when the consumer is started with the existing method on the registry

## Verifying this change
Use dubbo-demo-annotation test case.
Verification method: At present, there are two services in this  case. First delete one service and then start the provider. When the deleted service is restored, and the consumer is started, it will find that the consumer starts to report an error.


gule123:
1、In fact, my change is to add method-level verification when the consumer starts, because the online environment does have this problem.
2、I have a question: When the method-level verification fails, should I start the consumer or just record a log, my current approach is to record a log
3、My suggestion is to interrupt the container startup

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
```diff
@@ -660,13 +663,47 @@ public boolean isAvailable() {
         if (localUrlInvokerMap != null && localUrlInvokerMap.size() > 0) {
             for (Invoker<T> invoker : new ArrayList<>(localUrlInvokerMap.values())) {
                 if (invoker.isAvailable()) {
+                    isConMethodLessThanPro(invoker);
```
cvictory: boolean返回值有什么用？ 看到整个方法打印了一段日志，对逻辑本身没有影响？

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
```diff
@@ -660,13 +663,47 @@ public boolean isAvailable() {
...
+     *
+     * @param invoker invoker
+     * @return true:consumer <= provider ,false:consumer > provider
+     */
+    private boolean isConMethodLessThanPro(Invoker<T> invoker) {
+        String consumerMethod = queryMap.get(METHODS_KEY);
+        String providerMethod = invoker.getUrl().getParameter(METHODS_KEY);
+        List<String> alertMethodList = new ArrayList<>();
+        if (StringUtils.isNotEmpty(consumerMethod)) {
+            String[] newMethods = consumerMethod.split(",");
```
cvictory: 这个old和new变量名，有点奇怪。

cvictory:
> 1、In fact, my change is to add method-level verification when the consumer starts, because the online environment does have this problem.
> 2、I have a question: When the method-level verification fails, should I start the consumer or just record a log, my current approach is to record a log

I think it cannot stop the process.  If you want to stop the process, you can add a configuration to open 
this.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
HughYiXuan: 一开始我想改成类似强校验的，如果有缺少的方法，直接返回false，后面想了下，如果配置里面check打开的话，感觉校验的太严格了，然后就单纯的打印了个日志。我把这个方法的返回值改下，确实不严谨

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
HughYiXuan: 嗯，我调整下命名

HughYiXuan:
后面感觉进行到方法级别的校验是不是太严格了呀，所以就仅仅写了个警告的日志，你感觉有必要单独加个配置去打开这块吗@cvictory ，我想听下你的想法，学习下

cvictory:
@huangdx0726 HU 我觉得这个PR的意义不是很大，你要从provider端里拿到方法数据，主要目的是为了提早发现问题，从技术上来说，减少一次网络请求。但是一般来说，这个是需要业务方去规避，删除方法的这个逻辑的。

还有一点，现在注册到注册中心的服务数据内容太多了，很多定制的业务，都期望减少provider端写入到注册中心的数据。这个provider的方法数据，往往会不写入到注册中心，consumer也就无法拿到了。

cvictory:
@huangdx0726 hu 你先看下，我建议这个pr可以关闭。有疑问，可以直接回复我。

huangdx0726:
@cvictory  是不是艾特错了

cvictory:
> @cvictory 是不是艾特错了

sorry..

cvictory:
@HughYiXuan 
看下我上面回复。

cvictory:
暂时先关闭，有疑问的时候，重新reopen。

PR #5597, opened 2020-01-07
Optimize code in dubbo-common

OLPMO:
in java8,new Method[0] is better than new Method[methods.size()],you can learn about this from the below url:
https://shipilev.net/blog/2016/arrays-wisdom-ancients/

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/threadlocal/InternalThreadLocal.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/threadlocal/InternalThreadLocal.java
@@ -58,7 +58,7 @@ public class InternalThreadLocal<V> {
             if (v != null && v != InternalThreadLocalMap.UNSET) {
                 Set<InternalThreadLocal<?>> variablesToRemove = (Set<InternalThreadLocal<?>>) v;
                 InternalThreadLocal<?>[] variablesToRemoveArray =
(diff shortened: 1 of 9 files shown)


OLPMO:
I have watched the log of ci,but I did not know how to fix this error.If anyone could tell me what should I  do ,I would be very grateful.

lzf971107:
> I have watched the log of ci,but I did not know how to fix this error.If anyone could tell me what should I do ,I would be very grateful.

Fortunately, he performed correctly in my computer, or you can manually close the pr, reopen the pr, or submit a merge request to reactivate the ci check and try.

lndj:
> in java8,new Method[0] is better than new Method[methods.size()],you can learn about this from the below url:
> https://shipilev.net/blog/2016/arrays-wisdom-ancients/

I have read this article, are you sure new Method[0] is better than new Method[methods.size()]? Maybe you should read it again.

OLPMO:
You can read the following sentences in the conclusion part. @lndj 
Bottom line: **toArray(new T[0]) seems faster, safer, and contractually cleaner, and therefore should be the default choice now.** Future VM optimizations may close this performance gap for toArray(new T[size]), rendering the current "believed to be optimal" usages on par with an actually optimal one. Further improvements in toArray APIs would follow the same logic as toArray(new T[0]) — the collection itself should create the appropriate storage.

PR #5602, opened 2020-01-08
add PropertiesConfiguration constructor test

zhuliangcai:

add junit test


chickenlj:
Thanks. But the unit test does not work properly.

PR #5623, opened 2020-01-09
backward compatible with ListenableFilter

### LINKED ISSUE #5619: Fix the bug can't backward compatible with ListenableFilter

CodingSinger:
fix #5619 
This pr is to fix the bug that is not compatible with the previous listenable filter in version 2.7.5, but if this pr is used, it will not be compatible with the filter created by the  listener declared in 2.7.5. I think this should be regarded as an urgent bug-fix.
@chickenlj What do you think?  Is there any other way to solve this problem perfectly?

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/Filter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/Filter.java
@@ -48,7 +48,7 @@ public interface Filter {
     Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException;
 
     /**
-     * Please use {@link Listener#onMessage(Result, Invoker, Invocation)} instead.
+     * Please use {@link Listener2#onMessage(Result, Invoker, Invocation)} instead.
      * This method is kept only for compatibility and may get removed at any version in the future.
      *
      * @param appResponse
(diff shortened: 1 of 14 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/Filter.java
```diff
@@ -60,11 +60,19 @@ default Result onResponse(Result appResponse, Invoker<?> invoker, Invocation inv
...
     }
 
+    @Deprecated
     interface Listener {
 
+        void onResponse(Result appResponse, Invoker<?> invoker, Invocation invocation);
+
+        void onError(Throwable t, Invoker<?> invoker, Invocation invocation);
+    }
+    interface Listener2 {
```
RojerAlone: I think keep two methods in Listener is better:
```
interface Listener {
        @Deprecated
        void onResponse(Result appResponse, Invoker<?> invoker, Invocation invocation);

        default void onMessage(Result appResponse, Invoker<?> invoker, Invocation invocation) {
                onResponse(appResponse, invoker, invocation);
        }

        void onError(Throwable t, Invoker<?> invoker, Invocation invocation);
    }
```
Then call the `onMessage` method in `ProtocolFilterWrapper`, so user can implement eithor `onMessage` or `onResponse` method.

PR #5637, opened 2020-01-10
 Remove @adaptive annotation from bind method of ServletHttpBinder class.

xiaoheng1:
i use dubbo 2.7.5

ServletHttpBind class info：

public class ServletHttpBinder implements HttpBinder {

    @Override
    @Adaptive()
    public HttpServer bind(URL url, HttpHandler handler) {
        return new ServletHttpServer(url, handler);
    }

}

I analyzed JettyHttpBinder and TomcatHttpBinder and found that ServletHttpBinder is using this servlet request. In other words, if a service is deployed on the server, then a request is initiated, and then it is performed in a servlet container.
deal with. The Adaptive annotation here has no effect on the implementation of the function, so it is recommended to remove this annotation on the method.

modify：

public class ServletHttpBinder implements HttpBinder {

    @Override
    public HttpServer bind(URL url, HttpHandler handler) {
        return new ServletHttpServer(url, handler);
    }

}


NOTICE
```diff
@@ -1,5 +1,9 @@
 Apache Dubbo
+<<<<<<< HEAD
+Copyright 2018-2020 The Apache Software Foundation
+=======
```
AlbumenJ, line 4: pls fix merge confictions

AlbumenJ (changes requested):

NOTICE
xiaoheng1: ok，i will do it

xiaoheng1: I am already re-merging the dubbo/master branch in my own branch. Is there anything else I need to do here?

PR #5639, opened 2020-01-12
optimize #5623 : optimize compatible solution of filter listener

RojerAlone:

optimize #5623


RojerAlone:
@chickenlj

chickenlj:
I would suggest removing `onMessage` completely in the next version.

2.7.5 is not a stable version in our release plan, as stated here.

chickenlj:
https://github.com/apache/dubbo/issues/5669

chickenlj:
Also, we need to remove the `default` implementation of onResponse once onMessage gets removed.

RojerAlone:
> I would suggest removing `onMessage` completely in the next version.

The `onMessage` method was added in 2.7.5, the change means to rename `onResponse` method, why removing it ?



> Also, we need to remove the `default` implementation of onResponse once onMessage gets removed.

Got it, the target of this pr is to keep compatible with 2.7.5, and I add a `TODO` on `onMessage` method about changing it to non-default .

chickenlj:
https://github.com/apache/dubbo/pull/5731

PR #5646, opened 2020-01-13
optimize getSharedExecutorService

lkj41110:
## Verifying this change

1.There is no need to call method `getExecutor` before calling method `getSharedExecutorService`
2.optimize doc


chickenlj:
getExecutor is called very often, most of the time, getExecutor will return successfully.

PR #5649, opened 2020-01-14
[Dubbo-2.7.5]Add arguments router to ConditionRouter

furaul:
Add arguments router logic to ConditionRouter.

Add arguments router to ConditionRouter.
Compatible to current ConditionRouter logic.

## Brief changelog

Add arguments router to ConditionRouter

## Verifying this change

Add arguments router to ConditionRouter

## Examples
In some cases, we need to choose the right provider while the arguments of the method is setted.
The condition rule is like

> arguments[1].str!=b  => host = 1.1.1.1

While the str property of the second argument is equal to b, the providers whose host is equal to 1.1.1.1 are to be choosed。
We could use this feature to route requests according to business params. 
The feature is compatible to current ConditionRouter logic.

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
@@ -26,8 +26,10 @@ import org.apache.dubbo.common.utils.UrlUtils;
 import org.apache.dubbo.rpc.Invocation;
 import org.apache.dubbo.rpc.Invoker;
 import org.apache.dubbo.rpc.RpcException;
+import org.apache.dubbo.rpc.cluster.Constants;
 import org.apache.dubbo.rpc.cluster.router.AbstractRouter;
 
+import java.lang.reflect.Field;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


chickenlj:
I think it would be more reasonable if we add a new independent Router dedicated to handling arguments.

furaul:
> dedicated

Could you explain to me why we need a new independent Router for arguments?
你好，请帮忙解释下需要一个单独参数Router的原因呢？

furaul:
> I think it would be more reasonable if we add a new independent Router dedicated to handling arguments.



> I think it would be more reasonable if we add a new independent Router dedicated to handling arguments.

Could you explain to me why we need a new independent Router for arguments?
你好，请帮忙解释下需要一个单独参数Router的原因呢？

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
```diff
@@ -1,301 +1,352 @@
...
-                        return true;
-                    }
-                }
-                return false;
-            }
-            return false;
-        }
-    }
-}
+/*
```
AlbumenJ, line 1: pls recover end-line format of this file

AlbumenJ (changes requested):

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
furaul: I have fix it in https://github.com/apache/dubbo/pull/7378, please refer to the new PR.

furaul:
The format problem has been fixed in https://github.com/apache/dubbo/pull/7378, please refer to the new one. This PR would be closed later.

AlbumenJ:
close this and continues discussion  in #7378

PR #5659, opened 2020-01-16
2.7.5 release

dqqzj:

org.apache.dubbo.common.extension.ExtensionLoader#cacheDefaultExtensionName
Remove unnecessary logical judgments,i think there is a default extension name on extension ,it should't split with pattern.

## Verifying this change
```java
 private void cacheDefaultExtensionName() {
         final SPI defaultAnnotation = type.getAnnotation(SPI.class);
         if (defaultAnnotation == null) {
             return;
         }
         String value = defaultAnnotation.value();
         if ((value = value.trim()).length() > 0) {
             cachedDefaultName = value;
            /* String[] names = NAME_SEPARATOR.split(value);
             if (names.length > 1) {
             }
             if (names.length == 1) {
                 cachedDefaultName = names[0];
             }*/
         }
     }
```


dqqzj:
operate error

PR #5660, opened 2020-01-16
Remove unnecessary logical judgments

dqqzj:

org.apache.dubbo.common.extension.ExtensionLoader#cacheDefaultExtensionName
Remove unnecessary logical judgments,i think there is a default extension name on extension ,it should't split with pattern.

## Verifying this change
```java
 private void cacheDefaultExtensionName() {
         final SPI defaultAnnotation = type.getAnnotation(SPI.class);
         if (defaultAnnotation == null) {
             return;
         }

         String value = defaultAnnotation.value();
         if ((value = value.trim()).length() > 0) {
             cachedDefaultName = value;
            /* String[] names = NAME_SEPARATOR.split(value);
             if (names.length > 1) {
             }
             if (names.length == 1) {
                 cachedDefaultName = names[0];
             }*/
         }
     }
```

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
@@ -747,14 +747,15 @@ public class ExtensionLoader<T> {
 
         String value = defaultAnnotation.value();
         if ((value = value.trim()).length() > 0) {
-            String[] names = NAME_SEPARATOR.split(value);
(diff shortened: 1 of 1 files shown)


dqqzj:
If there are multiple type names separated by commas, an error will be directly reported below, that is, multiple types cannot be displayed, so why do we need to use separators? Think of it as a type name

chickenlj:
check is used for the following scenario.

```java
@SPI(name="name1, alias2")
class Ext {
}
```

PR #5661, opened 2020-01-16
Extension loader load extension classes

dqqzj:
Check if type name starts with org.apache

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
@@ -727,10 +727,12 @@ public class ExtensionLoader<T> {
         cacheDefaultExtensionName();
 
         Map<String, Class<?>> extensionClasses = new HashMap<>();
-
+        String typeName = type.getName();
         for (LoadingStrategy strategy : strategies) {
-            loadDirectory(extensionClasses, strategy.directory(), type.getName(), strategy.preferExtensionClassLoader(), strategy.excludedPackages());
-            loadDirectory(extensionClasses, strategy.directory(), type.getName().replace("org.apache", "com.alibaba"), strategy.preferExtensionClassLoader(), strategy.excludedPackages());
+            loadDirectory(extensionClasses, strategy.directory(), typeName, strategy.preferExtensionClassLoader(), strategy.excludedPackages());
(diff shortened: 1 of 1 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
```diff
@@ -747,14 +749,15 @@ private void cacheDefaultExtensionName() {
...
+            cachedDefaultName = value;
+           /* String[] names = NAME_SEPARATOR.split(value);
             if (names.length > 1) {
                 throw new IllegalStateException("More than 1 default extension name on extension " + type.getName()
                         + ": " + Arrays.toString(names));
             }
             if (names.length == 1) {
                 cachedDefaultName = names[0];
-            }
+            }*/
```
chickenlj: Why comment this snippet?

dqqzj: If there are multiple type names separated by commas, an error will be directly reported below, that is, multiple types cannot be displayed, so why do we need to use separators? Think of it as a type name

dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
```diff
@@ -727,10 +727,12 @@ private boolean isSetter(Method method) {
         cacheDefaultExtensionName();
 
         Map<String, Class<?>> extensionClasses = new HashMap<>();
-
+        String typeName = type.getName();
         for (LoadingStrategy strategy : strategies) {
-            loadDirectory(extensionClasses, strategy.directory(), type.getName(), strategy.preferExtensionClassLoader(), strategy.excludedPackages());
-            loadDirectory(extensionClasses, strategy.directory(), type.getName().replace("org.apache", "com.alibaba"), strategy.preferExtensionClassLoader(), strategy.excludedPackages());
```
dqqzj: If someone uses a lower version, org.apache will not be matched, and it is useless logic to perform the replacement.

AlbumenJ: In current, `org.apache` will be matched always, would this be roundabout?

AlbumenJ:

PR #5701, opened 2020-01-30
Make THREAD_NAME_KEY work

manzhizhen:
我们知道可以给Provider的业务处理线程池设置类型和size（见ThreadPool扩展点），在org.apache.dubbo.common.constants.CommonConstants中我们定义了THREADPOOL_KEY、THREAD_NAME_KEY、THREADS_KEY等，但唯独THREAD_NAME_KEY的值threadname在dubbo.xsd中没有设定，我想应该是漏了，导致我们无法给线程池设置想要的名字。所以此PR是为了解决这个问题。

We know that you can set the type and size of the Provider's business processing thread pool (see the ThreadPool extension point). In org.apache.dubbo.common.constants.CommonConstants we define THREADPOOL_KEY, THREAD_NAME_KEY, THREADS_KEY, etc., but only the value of THREAD_NAME_KEY The threadname is not set in dubbo.xsd, I think it should be missed, which prevents us from setting the desired name for the thread pool. So this PR is to solve this problem.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/ProtocolConfig.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/ProtocolConfig.java
@@ -60,6 +60,11 @@ public class ProtocolConfig extends AbstractConfig {
     private String threadpool;
 
     /**
+     * Thread pool name
+     */
+    private String threadname;
+
+    /**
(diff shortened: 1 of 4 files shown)


chickenlj:
LGTM.

PR #5709, opened 2020-02-04
Refactor, configuration override does not work properly.

chickenlj:
Fixes https://github.com/apache/dubbo/issues/5224


chickenlj:
reviewed by issue reporters

Issue #5297, opened 2019-11-09
Only one of the multiple registration centers using nacos can register


liunancun:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.4.1
* Operating System version: Window7
* Java version: JDK8

### Steps to reproduce this issue

1. Use nacos as a registration center
2. The configuration of application.properties is as follows
dubbo.registries.one.address=nacos://192.168.99.237:8848
dubbo.registries.two.address=nacos://192.168.99.237:8848?namespace=9f0825ac-0a6a-4907-b5ee-34530ce828b1
3. Register the service under the different namespaces of nacos

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result

Services exist in different namespaces

### Actual Result

The service only exists in the first registered namespace

Is it because ip is the same as the port, but the namespace is different, is it considered to be caused by repeated registration?

chickenlj:
which version of nacos dis you use?

chickenlj:
Nacos version 0.8+ support namespace feature.

PR #5756, opened 2020-02-18
[#5297] Only one of the multiple registration centers using nacos can register 

mercyblitz:


dubbo-common/src/main/java/org/apache/dubbo/config/RegistryConfig.java
```diff
@@ -213,7 +213,7 @@ public void setAddress(String address) {
                 URL url = URL.valueOf(address);
                 setUsername(url.getUsername());
                 setPassword(url.getPassword());
-                updateIdIfAbsent(url.getProtocol());
+//                updateIdIfAbsent(url.getProtocol());
```
mercyblitz: Please confirm this change

PR #5856, opened 2020-03-12
void potential shutdown race condition.

chickenlj:


tswstarplanet:
I think it is right. But the destroyProtocols() method of DubboBootstrap is duplicate to destroyProtocols of DubboShutdownHook. I think we can delete the former to the later. And add a lock to the later one.

chickenlj:
> But the destroyProtocols() method of DubboBootstrap is duplicate to destroyProtocols of DubboShutdownHook

Done, now only DubboShutdownHook.destroyProtocols is kept.

PR #5881, opened 2020-03-17
To fix concurrency problems in RpcStatus and RoundRobinLoadBalance

trytocatch:
to fix some concurrency problems

### RpcStatus.java
In some extreme cases, incrementAndGet and decrementAndGet will cause the method 
"beginCount" to returen incorrectly
if max is 10, active is 10 now，here are 3 thread, thread-b calls the metod 'endCount' to decrease the active
```
thread-a	thread-b	thread-c	active's value
increase	--------	--------	11
--------	decrease	--------	10
--------	--------	increase	11
decrease	--------	--------	10
--------	--------	decrease	9
 failed 	--------	 failed 	
```
both a and c failed, but active's value is 9 now, it's less than max

### RoundRobinLoadBalance.java
let's look at this situation:
there are two thread A and B, A copied the old map to an new map, B put a new WeightedRoundRobin into the old map, and go on to try to get lock, it will failed becuase A hold the lock now, A replace the old map with new map. So the new WeightedRoundRobin object was lost!

ConcurrentHashMap is thread-safe, includes its iterator, it's OK to remove items in original map.

## Brief changelog

- replace increase and decrease with CAS loop

- use 'computeIfAbsent' to replace 'get-create-putIfAbsent-get'

- remove 'updateLock'

## Verifying this change

it's obvious

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/RoundRobinLoadBalance.java
@@ -25,7 +25,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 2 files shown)


chickenlj:
LGTM.

PR #5910, opened 2020-03-24
[LICENSE] Adding the checkstyle module for checking the Apache 2.0 License

mercyblitz:


htynkn:
I thought rat plugin are doing this check at this moment?

slievrly:
It is best to automatically add and format copyright header when executing maven goals. I recommend use the `com.mycila:license-maven-plugin` plugin.
image

PR #5920, opened 2020-03-26
fix bad taste

qiaolin-li:
fix bad taste 

## issue url
https://github.com/apache/dubbo/issues/5919

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
@@ -576,14 +576,10 @@ public class ExtensionLoader<T> {
             }
         }
         StringBuilder buf = new StringBuilder("No such extension " + type.getName() + " by name " + name);
-
+        buf.append(", possible causes: ");
 
         int i = 1;
         for (Map.Entry<String, IllegalStateException> entry : exceptions.entrySet()) {
-            if (i == 1) {
-                buf.append(", possible causes: ");
-            }
-
             buf.append("\r\n(");
             buf.append(i++);
             buf.append(") ");


htynkn:
@qiaolin-li  thanks for your PR. I just double check this change and notice that the previous code rely on exceptions is not empty. so which means this change is not equivalent to previous code.

qiaolin-li:
I don't understand what you mean..

htynkn:
sorry @qiaolin-li , I think it's my mistake when I first time look at the issue you raised.
I test this change in my local, It's possible if one stream open fail, so exceptions map can be empty, so we can't print `, possible causes:` in error message

dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
```diff
@@ -576,14 +576,10 @@ private IllegalStateException findException(String name) {
             }
         }
         StringBuilder buf = new StringBuilder("No such extension " + type.getName() + " by name " + name);
-
+        buf.append(", possible causes: ");
 
         int i = 1;
         for (Map.Entry<String, IllegalStateException> entry : exceptions.entrySet()) {
```
htynkn: if exceptions is empty, we should not print `, possible causes: `

qiaolin-li: Yes, let's add a check when exceptions. EntrySet ().size() is greater than 0

htynkn: but my idea is we already have `i` in here, so we don't need calculate exceptions size again.
it's more like a trade off, so I would like to keep original code.

qiaolin-li: Ok, I see

qiaolin-li: Do I close this PR now

htynkn: apologize again for the mistake when the first time I check the issue your raised.
I'm going to close this PR. thanks again

PR #5925, opened 2020-03-27
remove useless code in MergeableClusterInvoker.java

wolfdriver:


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/MergeableClusterInvoker.java
```diff
@@ -54,7 +54,6 @@
 public class MergeableClusterInvoker<T> extends AbstractClusterInvoker<T> {
```
htynkn: Could you please also check if we can remove some unused import related this change?
for example

```
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
```

wolfdriver: done

Issue #5930, opened 2020-03-27
Recovery problem after zk reconnect


bigbang555:
Three configuration scenarios:
Scenario 1: Configured in /dubbo/xxx.xxx.XxxService/configurators in the ZK directory
Scenario 2: Configured and configured in /dubbo/config/dubbo/application-name.configurators in the ZK directory
Scenario 3: Configured in /dubbo/config/dubbo/org.apache.dubbo.demo.XxxService::.configurators in the ZK directory

When the service is started, if ZK is disconnected halfway, org.apache.dubbo.registry.integration.RegistryProtocol.OverrideListener # doOverrideIfNecessary will cause reExport, remove the old registered URL in reExport, delete ZK Nodes, but unable to communicate with ZK, will cause this notification to fail, and re-notification will not be able to sense the url change, so it will not be reExported. After subsequent reconnection to ZK, the service will be lost if the recover operation is re-registered.

PR #5931, opened 2020-03-27
fix-[#5930] fix the bug of recover 

bigbang555:
fix-[#5930]

## Brief changelog
Modify the order of operating memory and ZK in the unregister method


chickenlj:
I think this is an issue that deserves further investigation.

bigbang555:
it can recur stably in that case which I mention at [#5930]

chickenlj:
Please also check #5959

bigbang555:
yep.I will check it tomorrow.😁

chickenlj:
@bigbang555 Any update on this issue?

bigbang555:
no,I will append in #5959 after check.

PR #5952, opened 2020-03-31
Fix default config does not work as expected.

chickenlj:
fixes https://github.com/apache/dubbo/issues/5947

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/ReflectUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/ReflectUtils.java
@@ -1260,15 +1260,15 @@ public final class ReflectUtils {
     }
 
     /**
-     * Get the value from the specified bean and its property.
+     * Get the value from the specified bean and its getter method.
      *
-     * @param bean         the bean instance
-     * @param propertyName the name of property
-     * @param <T>          the type of property value
+     * @param bean       the bean instance
+     * @param methodName the name of getter
+     * @param <T>        the type of property value
      * @return
      * @since 2.7.5
      */
-    public static <T> T getProperty(Object bean, String propertyName) {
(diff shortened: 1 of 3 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/utils/ReflectUtils.java
```diff
@@ -1273,16 +1272,18 @@ public static boolean isPublicInstanceField(Field field) {
         Class<?> beanClass = bean.getClass();
         BeanInfo beanInfo = null;
         T propertyValue = null;
+
+        beanClass.getDeclaredFields();
```
zonghaishang, line 1276: Why invoke the getDeclaredFields method here ?

dubbo-common/src/main/java/org/apache/dubbo/common/utils/ReflectUtils.java
```diff
@@ -1273,16 +1272,18 @@ public static boolean isPublicInstanceField(Field field) {
...
         try {
             beanInfo = Introspector.getBeanInfo(beanClass);
-            propertyValue = (T) Stream.of(beanInfo.getPropertyDescriptors())
-                    .filter(propertyDescriptor -> propertyName.equals(propertyDescriptor.getName()))
-                    .map(PropertyDescriptor::getReadMethod)
```
zonghaishang, line 1280: May be use PropertyDescriptor::getReadMethod, but consider the getXXX and isXXX scenarios more better ?

dubbo-common/src/main/java/org/apache/dubbo/common/utils/ReflectUtils.java
chickenlj: It's useless, will remove it now.

PR #5967, opened 2020-04-04
[Dubbo-5961] Fix disorder problem of notification

laddcn:

Fix the bug of notification order mentioned in https://github.com/apache/dubbo/issues/5961 

## Brief changelog

1. Fix disorder problem of notification
2. Add version control of notification data  and avoid useless notify by version comparation
3. Fix bugs with FailedNotifiedTask
4. Add test cases for retry notify

## Verifying this change

Test cases run with success for for retry notify


wavesZh:
If using version  comparison, where version should be apply from, consumer or provider? If consumer perceive provider's changes by pull mode, eg Zookeeper or others, with network latency. Maybe i'm sensitive

laddcn:
Yes, you are right. The best way is to tag version when data are produced in provider side(or dubbo admin which will produce config info). However, it is difficult to to implement:
1.	Since providers and dubbo admin may have many instances, we need import a distributed id generator (such as leaf https://tech.meituan.com/2017/04/21/mt-leaf.html) to generate version id. 
2.	Dubbo version compatibility is also very difficult. Application with lower dubbo version will not produce data with version id. And it is almost impossible to update all applications in company.
3.	Codes with data production distribution in various projects. And it is also difficult to modify these codes to tag data version.

Notification data change no often. And it occurs rarer in a very short time which may lead disorder problem by network latency. So the problem you mentioned above occur rarely.

In my opinion, it is not worth to tag version in provider side. I have produce tag in consumer side now.

chickenlj:
Does FailedNotifiedTask need to keep the address list it used when fail? What if we change FailedNotifiedTask to only stands for one event and notify the latest address list.

chickenlj:
Now, I doubt the design of the notification retry. In what case does a notification need to be retried, every exception in every step is predictable and they will happen again if they happened for the first time.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/FailedNotifiedTask.java
```diff
@@ -58,10 +63,31 @@ public void removeRetryUrl(List<URL> urls) {
...
+            try {
+                registry.doNotify(url, listener, urls);
+
+                //When notify with success, clear urls and stop this task
+                urls.clear();
+                this.cancel();
+                logger.info("Retry to notify with success for subscribe " + url);
+            } catch (Exception e) {
+                //When notify with failure, retry again
+                reput(timeout, retryPeriod);
```
chickenlj, line 88: There's reput logic in the parent class.

chickenlj:
Closing

PR #5985, opened 2020-04-08
Fix: Cluster get wrong extension name

AlbumenJ:

After the pr of distinguish usage of getUrl and getConsumerUrl (#5686), the adaptive extension of `Cluster` still use `getUrl` to get consumer's url which will cause properties of `Reference` lose when refer cluster. (#5966)

## Brief changelog

Add custom `AdaptiveCluster` of `Cluster` which will use `getConsumerUrl` prefer when 
get consumer's url.


chickenlj:
This issue has been fixed with #6343

Issue #6046, opened 2020-04-23
Http protocol unable to pass implicit parameters properly


AlbumenJ:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.6
* Operating System version:  Windows 10
* Java version: JDK 1.8

### Steps to reproduce this issue

- Configure provider's protocol to http and using `RpcContext.getContext().setAttachment()` to pass implicit parameters to provider.
- Provider unable to get the attachment.

### Possible Reason

There is any wrong wit the code used to pass implicit parameters in http protocol.

PR #6047, opened 2020-04-23
[Enhancement]: Make http protocol support implicit parameters

AlbumenJ:

Make http protocol support implicit parameters
related issue #6046 

## Brief changelog

- Override `JsonProxyFactoryBean.invoke` to make implicit parameters in request headers and perfect the code `InternalHandler` to decode implicit parameters.
- Remove unused `RemoteInvocationFactory` override which take no effect in http protocol with `jsonrpc4j`
- Add unit test of implicit parameters.


avalon5666:
Now I also need this feature, I'm confused how to implement this function. 
There is a problem of implementation, the AbstractProxyProtocol.doRefer only executes once at startup, so the extraHttpHeaders of JsonProxyFactoryBean will share with every request, it will make mistakes in concurrency.

PR #6068, opened 2020-04-27
[Enhancement] Adding the overriding rule for ExtensionLoader

mercyblitz:
`ExtensionLoader`  is a core API to extend the features of Dubbo components, it's based on the Dubbo's Service Provider mechanism that will search three classpaths in order :

1. `META-INF/dubbo/internal/`
2. `META-INF/dubbo/`
3. `META-INF/services/`

However, Dubbo framework just loads only one SPI file if found, does not support the overriding rule, for example : 

The implementation in `META-INF/services/`  can override `META-INF/dubbo/`, 
`META-INF/dubbo/` can override `META-INF/dubbo/internal/` .


qinliujie:
I don't think it's necessary, and it make ExtensionLoader more  complicated

mercyblitz:
> I don't think it's necessary, and it make ExtensionLoader more complicated

Or change to be simpler, three classpaths to be one:

- META-INF/dubbo/internal/
- META-INF/dubbo/
- META-INF/services/

Just Leaving the first path will be better.

chickenlj:
Notice that `LoadingStrategy` is defined as Java standard SPI, which means `ExtensionLoader`, an enhanced version of SPI now relies on Java SPI to load.

PR #6111, opened 2020-05-07
put default version&group from provider and consumer into service metadata.

chickenlj:
override getVersion and getGroup.


qinliujie:
LGTM

PR #6146, opened 2020-05-13
Registry perf

chickenlj:


chickenlj:
https://github.com/apache/dubbo/pull/5613

PR #6151, opened 2020-05-13
Refactoring (move method) to AccessLogData class

klebertarcisio:
Hi everyone,

I discovered an opportunity to improve Dubbo. I believe that the buildAccessLogData method should be in the AccessLogData class because it uses only operations from that class.


AlbumenJ (approved):
LGTM.

chickenlj (approved):
LGTM

PR #6153, opened 2020-05-14
fix protocol version compatibility

chickenlj:
Fix: version numbers published by developers that have higher numbers than the official versions are considered as new protocol compatible, which is not always true.


qinliujie:
so many magic code......

PR #6159, opened 2020-05-15
Fix multi registry subscription related issues.

chickenlj:
* Introduce ClusterInvoker
* Introduce a way to disable SPI wrapper mechanism


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/MockInvoker.java
```diff
@@ -177,12 +178,27 @@ public static Throwable getThrowable(String throwstr) {
...
+            if (!isDefault) {// does not check Spring bean if it is default config.
+                ExtensionFactory extensionFactory =
+                        ExtensionLoader.getExtensionLoader(ExtensionFactory.class).getAdaptiveExtension();
+                Object obj = extensionFactory.getExtension(serviceType, mockService);
+                if (obj != null) {
+                    return obj;
+                }
+            }
+            throw new IllegalStateException("Did not find mock class or instance " + mockService + ", please check" +
+                    " if there's mock class or instance implementing interface " + serviceType.getName(), e);
```
qinliujie, line 199: ` please check" + " if there's mock class or instance `   put it on line would be better

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java
```diff
@@ -449,14 +444,12 @@ private String getCacheKey(final Invoker<?> originInvoker) {
...
-                return doRefer(getMergeableCluster(), registry, type, url);
+                return doRefer(Cluster.getCluster("mergeable"), registry, type, url);
             }
         }
-        return doRefer(cluster, registry, type, url);
-    }
 
-    private Cluster getMergeableCluster() {
-        return ExtensionLoader.getExtensionLoader(Cluster.class).getExtension("mergeable");
+        Cluster cluster = Cluster.getCluster(qs.get(CLUSTER_KEY));
```
imsunv, line 451: use a default value will be better?

PR #6179, opened 2020-05-18
Group variables into a common class

klebertarcisio:
Hi,

I saw that both ServiceConfig and ReferenceConfig use similar variables. I believe that a good solution is to move these variables to a common class.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/ServiceConfigBase.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/ServiceConfigBase.java
@@ -43,10 +43,7 @@ public abstract class ServiceConfigBase<T> extends AbstractServiceConfig {
 
     private static final long serialVersionUID = 3033787999037024738L;
 
-    /**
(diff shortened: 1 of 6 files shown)


chickenlj:
LGTM.

AlbumenJ (approved):
LGTM.

dubbo-common/src/main/java/org/apache/dubbo/config/ServiceConfigBase.java
```diff
@@ -403,15 +390,7 @@ public String getUniqueServiceName() {
         return URL.buildKey(interfaceName, getGroup(), getVersion());
     }
 
-    @Override
-    public String getGroup() {
-        return StringUtils.isEmpty(this.group) ? (provider != null ? provider.getGroup() : this.group) : this.group;
-    }
 
-    @Override
```
wuwen5, line 411: https://github.com/apache/dubbo/issues/7529

wuwen5:
#7529\ #7575

PR #6201, opened 2020-05-21
To prevent nodes that did not succeed at one time, the calculated estimateResponse is the lowest

yiyanghua:

To prevent nodes that did not succeed at one time, the calculated estimateResponse is the lowest

## Brief changelog
dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/ShortestResponseLoadBalance.java


Augustvic:
If like that, the nodes that have not been selected will never be selected, because all nodes initialled with zero succeeded average elapsed. Maybe we need to find a more effective way.

PR #6217, opened 2020-05-25
improve execution efficiency in some cases

fakefrog:

The purpose is to **improve execution efficiency** in some cases.

Using Strategy Pattern to improve efficiency.When the count of clients is is the power of two,using '&' to choose next client instead of '%',which is also used in **netty** to improve efficiency.

As for ThriftInvoker, which is deprecated, there's no need to apply changes on it.


chickenlj:
I think improving the performance by checking `single` `power of two` is a good shot.

But:
1. It does not need to be so complex, for example, no need to implement it as SPI
2. please provide the performance statistics.

AlbumenJ:
How about just simply it to several lines or exact into a single method in `DubboInvoker` without creating such three class?

AlbumenJ:

PR #6268, opened 2020-06-04
fix overrideDirectory not effect by configurators

imsunv:
overrideDirectoryUrl  not effect by configurators 

such as 
change `mock` dynamic configuration will not take effect.


chickenlj:
Fixed

PR #6311, opened 2020-06-11
change the comments of Annotation Class [org.apache.dubbo.config.spring.context.annotation.EnableDubboConfig]

mygodmele:
In Annotation Class org.apache.dubbo.config.spring.context.annotation.EnableDubboConfig, we can see  the current comments  is:
 /**
     * It indicates whether binding to multiple Spring Beans.
     *
     * @return the default value is <code>false</code>
     * @revised 2.5.9
     */
we know after version 2.6.6, the default value of multiple is true, so the comments need to update, otherwise it will cause confusion, so change it to: 
/**
     * It indicates whether binding to multiple Spring Beans.
     *
     * @return the default value is <code>true</code>
     * @revised 2.6.6
     */


mercyblitz:
Duplicated

PR #6321, opened 2020-06-15
maybe we can remove null judge in this case

nineyang:

i have check object `a1` and `a2` carefully, their predecessor `o1`,`o2` had `NPE` check, and in this method context , no more where have changed them, so , object `a1` and `a2` always not be null.

image

image

## Brief changelog

dubbo-common/src/main/java/org/apache/dubbo/common/extension/support/ActivateComparator.java


tswstarplanet:
LGTM

PR #6376, opened 2020-06-28
fix NPE when check=false is set and provider is empty.

chickenlj:
1. Consumer process starts first with check='false' set
<dubbo:reference check="false" />


chickenlj:
https://github.com/apache/dubbo/issues/6228

PR #6414, opened 2020-07-03
Incorrect comments about multipleConfig

yuequan1997:

The comment is wrong, which is easy to cause misunderstanding。
org.apache.dubbo.config.spring.context.annotation.EnableDubboConfig#multiple
org.apache.dubbo.config.spring.context.annotation.EnableDubbo#multipleConfig

The default value is true but the comment is false


chickenlj:
only comment change, CI failed because of previous master failure.

PR #6422, opened 2020-07-06
double check lock

diguage:


chickenlj:
CI fails because of previous master failure.

PR #6432, opened 2020-07-08
provider url parameters should not override that of directory url

chickenlj:
fixes https://github.com/apache/dubbo/pull/6376


chickenlj:
fixed

PR #6439, opened 2020-07-10
[Dubbo-2.7.7] Fix problem: System property dubbo.service.delay invalid

yiocio:

Fixed issue #1728：System property dubbo.service.delay invalid 

## Brief changelog

ReferenceConfigBase.java
ServiceConfigBase.java

## Verifying this change

As described in issue #1728：we can't set dubbo.service.delay to dubbo through System Property，and System-Enviroment、Property-File mechanism also can't work. furthermore, other similar-style config(eg: dubbo.service.filter、dubbo.service.retries) are also invalid. 

The troublemaker which caused this problem is:

 **_public String getPrefix() { return DUBBO + ".service." + interfaceName; }_**

Due to these above codes，when getProperty through **CompositeConfiguration**, the packed key is （for examle）: dubbo.service._**org.yiocio.testService**_._**org.yiocio.testService**_.delay，this key's format is wrong according to dubbo's config strategy. In my view, dubbo's property-style config have a namespace concept: the first segment is **dubbo** at all times, the second segment is tag for config category（eg: application、registry、service），and the third segment is the **id** of  config object with specific category，of course，the third segment is optional, the last segment is config key. So the override of getPrefix for ServiceConfigBase will break the namespace rule of dubbo config.

**ReferenceConfig**'s situation is similar with **ServiceConfig**.


yiocio:
Travis CI build failed because of some unused imports, so close the PR and create a new one.

yiocio:
trigger travis-ci

PR #6472, opened 2020-07-15
[Dubbo-6470] Fix NPE and remove try/catch block in NetUtils

### LINKED ISSUE #6470: Fix NPE and remove try/catch block in NetUtils

Technoboy-:
Fix #6470

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
@@ -269,8 +269,8 @@ public class NetUtils {
         InetAddress localAddress = null;
 
         // @since 2.7.6, choose the {@link NetworkInterface} first
-        try {
-            NetworkInterface networkInterface = findNetworkInterface();
+        NetworkInterface networkInterface = findNetworkInterface();
+        if (networkInterface != null) {
             Enumeration<InetAddress> addresses = networkInterface.getInetAddresses();
             while (addresses.hasMoreElements()) {
                 Optional<InetAddress> addressOp = toValidAddress(addresses.nextElement());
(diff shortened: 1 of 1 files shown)


carryxyh:
Can you describe what causes NPE to happen? Use unit test to reproduce maybe a better way.

Technoboy-:
> Can you describe what causes NPE to happen? Use unit test to reproduce maybe a better way.

findNetworkInterface may return null.
UT is difficult to write for findNetworkInterface, it depends on environment.
You can find NetUtils has no UT for findNetworkInterface

carryxyh:
Hi, it seems to be difficult to reproduce this problem.

But I think we should make sure that this is indeed a problem before we try to fix it. Have you encountered this problem in a production environment? Can you post your stack?

If it is only for return value verification, then I tend not to merge it, which will cause our submission records to be messy...

Technoboy-:
> messy

So why don't you add UT for findNetworkInterface? 
From the respect of code reviewing, findNetworkInterface may return null.  If nothing happens, why you add try-catch, could please explain, which kind of exception you wanna catch ?  Do you also encounter some problem in a production and let you catch the exception ?

Technoboy-:
> Hi, it seems to be difficult to reproduce this problem.
> 
> But I think we should make sure that this is indeed a problem before we try to fix it. Have you encountered this problem in a production environment? Can you post your stack?
> 
> If it is only for return value verification, then I tend not to merge it, which will cause our submission records to be messy...

So why don't you add UT for findNetworkInterface?
From the respect of code reviewing, findNetworkInterface may return null. If nothing happens, why you add try-catch, could please explain, which kind of exception you wanna catch ? Do you also encounter some problem in a production and let you catch the exception ?

Technoboy-:
cc @CalvinKirs

PR #6480, opened 2020-07-16
fix expire cache time

wenbochang888:

- fix expire cache bug
- Optimize code

``` java
	Map<String, String> map = new ConcurrentHashMap<>();
	map.put("cache.seconds", "10");
	map.put("cache.interval", "10");
	URL url = new URL("", "", "", "", 8888, "", map);
	ExpiringCache cache = new ExpiringCache(url);

	long start = System.currentTimeMillis();
	boolean in = false;
	boolean out10 = false;
	boolean out20 = false;
	while (true) {
		long end = System.currentTimeMillis();
		// put k,v in 5s when the programmer start
		if (!in && (end - start) > 5 * 1000) {
			cache.put("aaa", "aaa");
			cache.put("bbb", "bbb");
			log.info("put val");
			in = true;
		}
		// get value after expire time(10s)
		// but I get can get the value where is no null. wrong
		if (!out10 && (end - start) > 16 * 1000) {
			log.info("val = {}", cache.get("aaa"));
			out10 = true;
		}

		// second thread clean。this is null, currect
		if (!out20 && (end - start) > 21 * 1000) {
			log.info("val = {}", cache.get("bbb"));
			out20 = true;
		}
	}
}
```
Let's look at this piece of code. Set the expiration time to 10s and clear the cache every 10s. At the 5s, we put "aaa" and "bbb" into memory. In theory, after 15s, he should be cleared. When we took the aaa-key in 16s, he was still in memory. And when we took the bbb-key in 21s, he was null. At this time we The bug can be located.

## Brief changelog

fix expire cache map

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-filter/dubbo-filter-cache/src/main/java/org/apache/dubbo/cache/support/expiring/ExpiringMap.java
+++ b/dubbo-filter/dubbo-filter-cache/src/main/java/org/apache/dubbo/cache/support/expiring/ExpiringMap.java
@@ -84,6 +84,11 @@ public class ExpiringMap<K, V> implements Map<K, V> {
     public V get(Object key) {
         ExpiryObject object = delegateMap.get(key);
         if (object != null) {
+            long timeIdle = System.currentTimeMillis() - object.getLastAccessTime();
+            if (timeIdle >= expireThread.getTimeToLive() * 1000) {
+                delegateMap.remove(object.getKey());
+                return null;
+            }
             object.setLastAccessTime(System.currentTimeMillis());
             return object.getValue();
         }
@@ -307,7 +312,7 @@ public class ExpiringMap<K, V> implements Map<K, V> {
          * start thread
          */
         public void startExpiryIfNotStarted() {
-            if (running) {
+            if (running && timeToLiveMillis <= 0) {
                 return;
             }
             startExpiring();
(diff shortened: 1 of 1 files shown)


dubbo-filter/dubbo-filter-cache/src/main/java/org/apache/dubbo/cache/support/expiring/ExpiringMap.java
```diff
@@ -84,6 +84,11 @@ public V put(K key, V value) {
     public V get(Object key) {
         ExpiryObject object = delegateMap.get(key);
         if (object != null) {
+            long timeIdle = System.currentTimeMillis() - object.getLastAccessTime();
+            if (timeIdle >= expireThread.getTimeToLive() * 1000) {
```
AlbumenJ, line 88: shall we check if `expireThread.getTimeToLive()` is zero or even negative

wenbochang888: oh yes. we should be

wenbochang888:
我用中文解释一下

dubbo的缓存 https://dubbo.apache.org/zh/docs/v2.7/user/examples/result-cache/ 其中有一个种为：ExpiringCache

它类似于Redis中的设置key的过期时间，比如我设置了10s后过期，那么我在10s就拿不到了

dubbo官方采用的是一个while循环，processExpires方法，每过一段时间检测map里面所有的key是否过期，如果过期 remove

这种是**定时删除**，但他有个问题如下描述：

我设置过期时间为10s，同时每10s定时删除。

processExpires会在第0s, 10s, 20s, 30s....执行定期删除

**我在第5s放入一个key，第10s检测没有过期，此时我在第17s get的时候，理论上是过期了，但此时我get不为null**

bug复现，应和Redis一样，不仅采用定时删除，也要采用惰性删除。

即每次get的时候，如果不为空，检测其是否过期。

AlbumenJ (approved):

AlbumenJ:
LGTM. 

I will merge it later.

@wenbochang888 would you please add some ut test cases for this pr in the future?

Issue #6491, opened 2020-07-17
[Issue] The URL's parameters are missing on the XML element "<dubbo:metadata-report >"


mercyblitz:
There are two issues about this :

- the useless URL parameters of the attributes of the element, like "sync-report"
-  without appending the URL parameters from the attribute "address"

PR #6493, opened 2020-07-17
[Bugfix] #6491

mercyblitz:
[Bugfix] #6491


dubbo-common/src/main/java/org/apache/dubbo/config/MetadataReportConfig.java
```diff
@@ -85,20 +86,25 @@ public MetadataReportConfig(String address) {
...
 
-    public URL toUrl() {
+    public URL toUrl() throws IllegalArgumentException {
         String address = this.getAddress();
-        if (StringUtils.isEmpty(address)) {
-            return null;
+        if (isEmpty(address)) {
+            throw new IllegalArgumentException("The address of metadata report is invalid.");
         }
         Map<String, String> map = new HashMap<String, String>();
```
lfzCarlosC, line 94: change the name to be `parametersByName`

PR #6506, opened 2020-07-23
[optimization] polish RegistryProtocol in slight and fix up some word spelling

yiocio:

polish RegistryProtocol in slight and fix up some word spelling

## Brief changelog

RegistryProtocol.java
CommonConstants.java
StringUtils.java
DubboBeanDefinitionParser.java
RegistryProtocol.java

## Verifying this change

The RegistryProtocol's implementation may be adjusted many times by different person, some logic is redundant。For example，in **export** method, **Registry** object is repeatedly acquired，Although It is guaranteed not to be repeatedly generated by **AbstractRegistryFactory** , but It is really a bit inelegant.

In addition to the above adjustments, Some abbreviation have been corrected（e.g. **Configed** to **Configured**）,Some variable name have been adjusted（e.g. **HIDE_KEY_PREFIX** to  **HIDDEN_KEY_PREFIX**）.


AlbumenJ:
@yiocio hi, thanks for your contribution.

Would you please reslove confilcts with the latest master branch.

yiocio:
@AlbumenJ OK，I will reslove confilcts  with the latest master branch.

yiocio:
@AlbumenJ I have reslove confilcts with the latest master branch, But the associated CI check can't be passed for several tiems, and the reason is different each time,I believe CI check will definitely be passed, as long as I try a few more times,Because I encountered this problem before. I can guarantee that the code to be merged will work properly. So Is this PR can be accpeted without CI check passed? Or can you tell me how to manually trigger the CI check.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java
```diff
@@ -214,7 +215,7 @@ private void registerStatedUrl(URL registryUrl, URL registeredProviderUrl, boole
         // decide if we need to delay publish
         boolean register = providerUrl.getParameter(REGISTER_KEY, true);
         if (register) {
-            register(registryUrl, registeredProviderUrl);
+            registry.register(registeredProviderUrl);
```
AlbumenJ, line 218: Could you please describe why you change this line

yiocio: I believe the extracted **register** method  is useless, for two reasons: 
1. the **register** method only been referred here(line 217  in RegistryProtocol.java file);
2. **register** method's logic have been implemented in **export** method,eg: get **registry** by **registryUrl**, and register ProviderUrl by **registry** object. But the **registry** object have been obtained in the context of **export** method(line 212).

AlbumenJ: got it, shall we remove `private void register(URL registryUrl, URL registeredProviderUrl)` is this pr? This method seems is no longer needed.

AlbumenJ (approved):
LGTM.

PR #6508, opened 2020-07-23
[Dubbo-6463] fix issue: start.sh can't work when dubbo.properties contain commented dubbo.protocol.port  property

yiocio:

fix issue #6463 : start.sh can't work when dubbo.properties contain commented dubbo.protocol.port  property

## Brief changelog

dubbo-container/dubbo-container-api/src/main/resources/META-INF/assembly/bin/start.sh

## Verifying this change

When configure the port of dubbo protocol in dubbo.properties file, start.sh will parse dubbo.properties using sed and get the dubbo protocol port number, then determine whether the application starts successfully by checking whether the port is opened.

But the mechanism of parsing dubbo protocol port isn't rigorous, when my dubbo.properties contains commented dubbo.protocol.port like this:

**#dubbo.protocol.port=20943
dubbo.protocol.port=20945**

start.sh will get wrong port number, and can't work normally.


AlbumenJ (approved):
LGTM.

PR #6590, opened 2020-08-13
DecodeableRpcResult解码参数丢失

XiaoweiM:
- Dubbo version: 2.7.8

```
public class DecodeableRpcResult{

public void setObjectAttachments(Map<String, Object> map) {
        this.attachments = map == null ? new HashMap<>() : map;
    }
}
```
DecodeableRpcResult解码的时候使用`setObjectAttachments`会丢失已经在IO线程解码的部分参数，例如input,output。

该方法应该替换为

```
  public void addObjectAttachments(Map<String, Object> map) {
        if (map == null) {
            return;
        }
            this.attachments = new HashMap<>();
        }
        this.attachments.putAll(map);
    }
```

### DIFF
--- a/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DecodeableRpcResult.java
+++ b/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DecodeableRpcResult.java
@@ -160,7 +160,7 @@ public class DecodeableRpcResult extends AppResponse implements Codec, Decodeabl
 
     private void handleAttachment(ObjectInput in) throws IOException {
         try {
-            setObjectAttachments(in.readAttachments());
+            addObjectAttachments(in.readAttachments());
         } catch (ClassNotFoundException e) {
             rethrow(e);
         }


chickenlj:
I think this is the first and only place trying to update Attachments, can u list the place that Attachments is possibly changed in the IO thread?

IndiraFinish:
```
public class DubboCountCodec{
public Object decode(Channel channel, ChannelBuffer buffer) throws IOException {
        int save = buffer.readerIndex();
        MultiMessage result = MultiMessage.create();
        do {
            Object obj = codec.decode(channel, buffer);
            if (Codec2.DecodeResult.NEED_MORE_INPUT == obj) {
                buffer.readerIndex(save);
                break;
            } else {
                result.addMessage(obj);
                logMessageLength(obj, buffer.readerIndex() - save);
                save = buffer.readerIndex();
            }
        } while (true);
        if (result.isEmpty()) {
            return Codec2.DecodeResult.NEED_MORE_INPUT;
        }
        if (result.size() == 1) {
            return result.get(0);
        }
        return result;
    }

    private void logMessageLength(Object result, int bytes) {
        if (bytes <= 0) {
            return;
        }
        if (result instanceof Request) {
            try {
                ((RpcInvocation) ((Request) result).getData()).setAttachment(INPUT_KEY, String.valueOf(bytes));
            } catch (Throwable e) {
                /* ignore */
            }
        } else if (result instanceof Response) {
            try {
                ((AppResponse) ((Response) result).getResult()).setAttachment(OUTPUT_KEY, String.valueOf(bytes));
            } catch (Throwable e) {
                /* ignore */
            }
        }
    }
}
```

IO线程解码会添加参数，业务线程解码会导致覆盖

IndiraFinish:
@chickenlj

AlbumenJ (approved):
LGTM.

PR #6612, opened 2020-08-17
Fix unit test error in jetcd

AlbumenJ:

- upgrade jectd version to adapt new version of netty
- upgrade grpc version to adatp new version of jetcd
- fix ASF header in dubbo-registry-dns pom file


chickenlj:
Ci failed on jdk 11 because of some environment-related issues.

PR #6617, opened 2020-08-18
k8s enhancement: only export MetadataService on Dubbo protocol

chickenlj:


AlbumenJ:
image

Pls modify it to `getAndListenInstanceMetadata.1.callback` which will enable parameter callback on consumer side.

PR #6618, opened 2020-08-18
Backport service discovery related changes on 3.0 to master

chickenlj:


dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/transport/AbstractEndpoint.java
```diff
@@ -51,7 +51,7 @@ public AbstractEndpoint(URL url, ChannelHandler handler) {
     }
 
     protected static Codec2 getChannelCodec(URL url) {
-        String codecName = url.getParameter(Constants.CODEC_KEY, "telnet");
+        String codecName = url.getProtocol(); // codec extension name must stay the same with protocol name
```
htynkn, line 54: what's the purpose for this change?
does this a break change?

htynkn: @chickenlj

ririliu: same question

PR #6623, opened 2020-08-18
Feat: add life cycle probe in dubbo-qos

AlbumenJ:

- add livenessProbe, readinessProbe, startupProbe SPIs in dubbo-qos
- add Bootstarp impl for readinessProbe, startupProbe
- add provider register check for readinessProbe


chickenlj:
I think we should only load those active extensions by using getActiveExtensions instead of load all extensions, so users can decide to enable/disable the default probe.

dubbo-plugin/dubbo-qos/src/main/resources/META-INF/dubbo/internal/org.apache.dubbo.qos.probe.ReadinessProbe
```diff
@@ -0,0 +1,2 @@
+bootstrap = org.apache.dubbo.qos.probe.impl.BootstrapReadinessProbe
```
chickenlj, line 1: I didn't find the definition of this class.

chickenlj (changes requested):

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/probe/impl/ProviderReadinessProbe.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.dubbo.rpc.model.ServiceRepository;
+
+import java.util.Collection;
+import java.util.List;
+
+public class ProviderReadinessProbe implements ReadinessProbe {
+    private static ServiceRepository serviceRepository = ApplicationModel.getServiceRepository();
+
+    @Override
+    public boolean check() {
```
chickenlj, line 31: I think checking whether ServiceInstance has been registered would be enough.

chickenlj (changes requested):

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/probe/impl/ProviderReadinessProbe.java
AlbumenJ: Readiness probe is used in total life cycle. If readiness probe failed, the endpoint will temporarily be removed from service. ServiceInstance is registered when startup.

AlbumenJ:
activate extensions support added

PR #6661, opened 2020-08-30
[issue #6648] fix check server start dependency

CalebZYC:
fix [#6648](urhttps://github.com/apache/dubbo/issues/6648l)


AlbumenJ:
would u pls add some ut cases for it

CalebZYC:
previous merged code seem to fix this bug, I rollback previous commited code and add unit test.

PR #6677, opened 2020-09-01
check default config status before add to ConfigManager

chickenlj:
To prevent default configs from taking place too early.


chickenlj:
fixed with

PR #6700, opened 2020-09-06
Fixed performance

daniel-liy:
there is have some method With too few parameters will cause memory waste And remove unnecessary paradigm parameters. eg:Arrays.asList() Can be replaced with Collections.singletonList() to save memory overhead. Thanks!


daniel-liy:
update

Issue #6758, opened 2020-09-17
Not change registryDirectory.forbidden when the dynamicConfig change  and registryDirectory's invoker is null


horizonzy:
### Environment

* Dubbo version: 2.7.x
* Operating System version: macOs
* Java version: jdk8

### Steps to reproduce this issue
1. start provider and consumer
2.stop all provider
3.change dynamicConfig which consumer used

### Expected Result
Not change the forbidden value if the invokers didn't change.

### Actual Result
The forbidden will be change to false.

horizonzy:
@AlbumenJ

horizonzy:
@manzhizhen pls have a look this issue.

PR #6759, opened 2020-09-17
[Dubbo-6758]  not refreshInvokers in some case mentioned in #6758

horizonzy:
fix-#6758


horizonzy:
@htynkn this pr also to fix bug. can you check it?

PR #6816, opened 2020-10-20
[Dubbo-6519] fix GenericService.$invokeAsync

kylixs:

fix GenericService.$invokeAsync() (https://github.com/apache/dubbo/issues/6519)

## Brief changelog

org.apache.dubbo.common.bytecode.Proxy#getProxy()

`GenericService.$invokeAsync()` wraps the result of $invoke() as CompletableFuture, does not need to be proxied

## Verifying this change


kylixs:
`org.apache.dubbo.rpc.service.GenericService.$invokeAsync()` is confusing, whether it is asynchronous call on the consumer side or asynchronous execution on the provider side ?

I think it is more reasonable to call asynchronously on the consumer side. In this case, the provider side should not be called the $invokeAsync method.

kylixs:
`dubbo-samples-generic-impl`  tests failed.
https://github.com/apache/dubbo/pull/6816/checks?check_run_id=2086452797
```
Caused by: org.apache.dubbo.remoting.RemotingException: Failed to send response: Response [id=2, version=2.0.2, status=20, event=false, error=null, result=AppResponse [value=java.util.concurrent.CompletableFuture@b00af6c[Completed normally], exception=null]], cause: java.lang.IllegalStateException: Serialized class java.util.concurrent.CompletableFuture must implement java.io.Serializable
java.lang.IllegalStateException: Serialized class java.util.concurrent.CompletableFuture must implement java.io.Serializable
	at com.alibaba.com.caucho.hessian.io.SerializerFactory.getDefaultSerializer(SerializerFactory.java:417)
	at com.alibaba.com.caucho.hessian.io.SerializerFactory.getSerializer(SerializerFactory.java:391)
	at com.alibaba.com.caucho.hessian.io.Hessian2Output.writeObject(Hessian2Output.java:411)
	at org.apache.dubbo.common.serialize.hessian2.Hessian2ObjectOutput.writeObject(Hessian2ObjectOutput.java:98)
	at org.apache.dubbo.rpc.protocol.dubbo.DubboCodec.encodeResponseData(DubboCodec.java:218)
	at org.apache.dubbo.remoting.exchange.codec.ExchangeCodec.encodeResponse(ExchangeCodec.java:303)
	at org.apache.dubbo.remoting.exchange.codec.ExchangeCodec.encode(ExchangeCodec.java:73)
	at org.apache.dubbo.rpc.protocol.dubbo.DubboCountCodec.encode(DubboCountCodec.java:40)

```

dubbo-common/src/main/java/org/apache/dubbo/rpc/service/GenericService.java
```diff
@@ -37,12 +37,24 @@
...
-            return (CompletableFuture<Object>) object;
-        }
-        return CompletableFuture.completedFuture(object);
+        return CompletableFuture.supplyAsync(() -> {
+            Object object = $invoke(method, parameterTypes, args);
+            // if (object instanceof CompletableFuture) {
+            //     return ((CompletableFuture) object).get();
+            // }
+            return object;
+        });
```
chickenlj, line 57: The initial motivation of introducing $invokeAsync is to support the direct calling of methods with the signature of CompletableFuture. 

This pr seems like a breaking API change for the CompletableFuture instance returned has been changed.

How about simply adding the following line to make $invoke asynchronously.
```java
RpcContext.getContext().setAttachment(ASYNC_KEY, "true");
```

chickenlj (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/rpc/service/GenericService.java
kylixs: > The initial motivation of introducing $invokeAsync is to support the direct calling of methods with the signature of CompletableFuture.

Could the $invoke method return a CompletableFuture object?
If `$invoke()` is a synchronous call, calling directly here will block the thread, so I want to change it to asynchronous execution, by calling `$invoke()` in `CompletableFuture.supplyAsync()`.

chickenlj: > Could the $invoke method return a CompletableFuture object?

Yes, only if the method signature is like:
```java
class Service {
     CompletableFuture<String> sayHello(); 
}
```

chickenlj: > If $invoke() is a synchronous call, calling directly here will block the thread,

Yes, you are right, the code before will block the thread. I think adding an ASYNC_KEY attachment would solve this issue while avoiding the previous CompletableFuture instance issue. 

`RpcContext.getContext().setAttachment(ASYNC_KEY, "true");`

kylixs: > Yes, only if the method signature is like:
> class Service {
>      CompletableFuture<String> sayHello(); 
> }

This code should be the implementation code of the Service. The returned CompletableFuture object will be processed during serialization and will not be received on the consumer side. Not sure how the injvm protocol handles this situation.

chickenlj: >  The returned CompletableFuture object will be processed during serialization and will not be received on the consumer side. 

Actually the CompletableFuture is not serialized and transmitted on the wire, it is disassembled at the consumer end and reassembled at the consumer side. Anyway, the completable future instance should not be changed.

chickenlj: Sorry, I think I missed something here.
```java
if (object instanceof CompletableFuture) {
      return ((CompletableFuture) object).get();
  }
```

chickenlj:
The thing that is certain now is that there indeed has an issue with the default implementation of ` GenericService.$invokeAsync() `, but the current patch is not able to fix it.

I am working on this issue right now, and hopefully,  to pull request very soon.

chickenlj:
BTW, this patch has nothing to do with the issue https://github.com/apache/dubbo/issues/6519 as explained by the issue https://github.com/apache/dubbo/pull/6520

PR #6953, opened 2020-11-29
[Dubbo-6954] Reset ExtensionLoader in ExtensionLoaderTest

lzx404243:
The tests are not idempotent and fails if run twice in the same JVM, because each of the tests pollutes state shared among tests:
* `dubbo-common,org.apache.dubbo.common.extension.ExtensionLoaderTest.test_replaceExtension_Adaptive`
* `dubbo-common,org.apache.dubbo.common.extension.ExtensionLoaderTest.test_AddExtension`
* `dubbo-common,org.apache.dubbo.common.extension.ExtensionLoaderTest.test_replaceExtension`

It may be good to clean this state pollution so that some other tests do not fail in the future due to the shared state polluted by these tests.

## Brief changelog

Call `ExtensionLoader.resetExtensionLoader()` to reset the ExtensionLoader for the certain extension types used in each tests in `test_replaceExtension_Adaptive`, `test_AddExtension`, and `test_replaceExtension`.

## Verifying this change
With the proposed fix, the test does not pollute the shared state (and passes when run twice in the same JVM).

Link to issue: https://github.com/apache/dubbo/issues/6954


AlbumenJ:
lgtm, thanks for your contribution to make unit test more strength.

lzx404243:
@AlbumenJ Thanks for the review! Is there anything I can do on my end for this to be merged?

PR #7039, opened 2020-12-19
[Dubbo-7038]Returns the same reference if no filtering has occurred, to save memory.

HeYuqiu:

[7038]https://github.com/apache/dubbo/issues/7038


kimmking:
I don't think keep original reference is a goog way.

dubbo-common/src/main/java/org/apache/dubbo/common/utils/UrlUtils.java
```diff
@@ -470,6 +470,16 @@ public static boolean isServiceKeyMatch(URL pattern, URL value) {
...
     public static List<URL> classifyUrls(List<URL> urls, Predicate<URL> predicate) {
+        boolean filter = false;
+        for (URL url : urls) {
+            if (!predicate.test(url)) {
+                filter = true;
+                break;
+            }
+        }
+        if (!filter) {
+            return urls;
```
kimmking, line 481: keep an outer variable is dangerous.

kimmking (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/utils/UrlUtils.java
chickenlj: Better wrapping `urls` with unmodifiable?

chickenlj: Let's keep this snippet as-is for this is not a critical path for performance improvement.

PR #7078, opened 2021-01-05
[Unit Test] Use matrix build workflow & Add Integration Test

AlbumenJ:

- Use matrix build workflow
- remove 3rd party step dependency due to apache limit
- add integration test base on dubbo-samples


kylixs (approved):
Good job!

Issue #6763, opened 2020-09-17
[Discuss] RegistryDirectory.refreshInvoker maybe exist concurrency problems


horizonzy:
### Environment

* Dubbo version: 2.7.x
* Operating System version: macOs
* Java version: jdk8

### Steps to reproduce this issue
1. provider shutdown and dynamicConfig change at the same time

What actually happens?
at the same time. 
provider shutdown will execute `destroyAllInvokers();`.
dynamicConfig change will execute  `destroyUnusedInvokers(oldUrlInvokerMap, newUrlInvokerMap); `.

in `destroyAllInvokers();`, it use ` Map<String, Invoker<T>> localUrlInvokerMap = this.urlInvokerMap; ` to destroy. 
in the process of dynamicConfig change, it will execute `Map<String, Invoker<T>> newUrlInvokerMap = toInvokers(invokerUrls);` and `this.urlInvokerMap = newUrlInvokerMap;`.   

maybe destroyAllInvokers(); use the urlInvokerMap before `this.urlInvokerMap = newUrlInvokerMap;`, it may destory the old invokers, then dynamicConfig change execute `destroyUnusedInvokers(oldUrlInvokerMap, newUrlInvokerMap);` the newUrlInvokerMap won't be destroy.

the detail logic in org.apache.dubbo.registry.integration.RegistryDirectory#refreshInvoker.

horizonzy:
@AlbumenJ

PR #7082, opened 2021-01-06
[ISSUE-#6763] Make refreshInvoker synchronized

horizonzy:
For-#6763
Make refreshInvoker synchronized, to avoid config-center change and registry-center change concurrency.


horizonzy:
@AlbumenJ This pr also fix concurency bug, PTAL.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
```diff
@@ -152,7 +152,9 @@ private String judgeCategory(URL url) {
         return "";
     }
 
-    private void refreshOverrideAndInvoker(List<URL> urls) {
+    // Issue: https://github.com/apache/dubbo/issues/6763.
```
AlbumenJ, line 155: remove this line, and we can refer these changes in git blame

AlbumenJ (changes requested):

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
horizonzy: done.

AlbumenJ (approved):
LGTM.

PR #7108, opened 2021-01-18
[Dubbo-7107] Reset TestProxyFactory.count after ServiceConfigTest.testProxy

lzx404243:
The test `org.apache.dubbo.config.ServiceConfigTest.testProxy` is not idempotent and fails if run twice in the same JVM, because each of the tests pollutes state shared among tests. It may be good to clean this state pollution so that some other tests do not fail in the future due to the shared state polluted by these tests.

## Brief changelog

Reset `TestProxyFactory.count` to 0 at the end of `ServiceConfigTest.testProxy()`.

## Verifying this change
With the proposed fix, the test does not pollute the shared state (and passes when run twice in the same JVM).

Link to issue: https://github.com/apache/dubbo/issues/7107


AlbumenJ (approved):
lgtm

lzx404243:
Thanks @AlbumenJ for the review. Is there anything I can do on my side before this can be merged?

Issue #7143, opened 2021-01-27
[Dubbo 3.0]Unappropriated decode operation in URLAddress.parse() will cause memory leak


kevinw66:
- [ ] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: xxx
* Operating System version: xxx
* Java version: xxx

### Steps to reproduce this issue

1. xxx
2. xxx
3. xxx

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result

Normal behavior.

### Actual Result

Memory leak.

The decode operation in URLAddress.parse() will store decoded string in rawAddress. But the stringAddress cache in CacheableFailbackRegistry still use encoded string as their key, make this cache value cannot be deleted, cause memory leak.

AlbumenJ:
close #7143 via #7144

PR #7144, opened 2021-01-27
[Dubbo-7143]Fix memory leak for stringAddress cache

kevinw66:

For-[#7143](https://github.com/apache/dubbo/issues/7143)


AlbumenJ (approved):
LGTM.

PR #7178, opened 2021-02-07
[Dubbo-7178] Fix 'java.io.FileNotFoundException' when 'dump.directory' not exists

zenuo:

Fix 'java.io.FileNotFoundException' when 'dump.directory' not exists

## Brief changelog

Try to create if not exists, if can't be created, use 'user.home' as dump directory. 

## Verifying this change

Link to issue: [#7178](https://github.com/apache/dubbo/issues/7178)

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/support/AbortPolicyWithReport.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/support/AbortPolicyWithReport.java
@@ -144,10 +144,14 @@ public class AbortPolicyWithReport extends ThreadPoolExecutor.AbortPolicy {
             return USER_HOME;
         }
         final File dumpDirectory = new File(dumpPath);
-        if (!dumpDirectory.exists() && !dumpDirectory.mkdirs()) {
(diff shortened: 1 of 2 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/support/AbortPolicyWithReport.java
```diff
@@ -134,4 +138,17 @@ private void dumpJStack() {
 
     }
 
+    private String getDumpPath() {
+        final String dumpPath = url.getParameter(DUMP_DIRECTORY);
+        if (StringUtils.isEmpty(dumpPath)) {
+            return USER_HOME;
+        }
+        final File dumpDirectory = new File(dumpPath);
+        if (!dumpDirectory.exists() && !dumpDirectory.mkdirs()) {
```
AlbumenJ, line 147: @zenuo maybe we need warn users if we create the folders ?

zenuo: Sounds like a good idea, how about make a `info` log? Because creating a folder may be a routine operation.

AlbumenJ (approved):
LGTM.

PR #7185, opened 2021-02-08
Split the start part from the DubboBootstrap/ReferenceConfig/ServiceConfig

ppzzyy11:

Make sure every layer only responsible for their own work.

## Brief changelog

Add the DubboServer to be the Class responsible to trigger the init/start process of the ReferenceConfig/ServiceConfig. 
Most methods of the DubboBootstrap call the methods with the same names in DubboServer.

## Verifying this change

I run Dubbo Demo in my PC. And demos based on DubboServer/DubooBootstrap/API works well.

##

The developing process is based on this link below and many unit testes can not pass before my modification.

https://github.com/apache/dubbo/issues/5666


AlbumenJ (changes requested):
@ppzzyy11 hi, could you please tell me the reason why you disable all of the test cases in MockProviderRpcExceptionTest and MockClusterInvokerTest ?

kylixs (commented):
@chickenlj @AlbumenJ 
Need to create a branch for API refactoring.

PR #7189, opened 2021-02-14
Remove parameters from registryUrl to prevent config pollution

### LINKED ISSUE #7188: Configuration in Spring env Inherit problem
When using xml configuration manager, `group` parameter in `dubbo:registry` section will be automaticly inherited to `dubbo:reference` section.

AlbumenJ:

Fix #7188

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/directory/AbstractDirectory.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/directory/AbstractDirectory.java
@@ -72,7 +72,10 @@ public abstract class AbstractDirectory<T> implements Directory<T> {
         this.consumedProtocol = this.queryMap.get(PROTOCOL_KEY) == null ? DUBBO : this.queryMap.get(PROTOCOL_KEY);
         this.url = url.removeParameter(REFER_KEY).removeParameter(MONITOR_KEY);
 
-        this.consumerUrl = this.url.setProtocol(consumedProtocol).setPath(path == null ? queryMap.get(INTERFACE_KEY) : path).addParameters(queryMap)
+        this.consumerUrl = this.url.setProtocol(consumedProtocol)
+                .setPath(path == null ? queryMap.get(INTERFACE_KEY) : path)
+                .clearParameters()
+                .addParameters(queryMap)
                 .removeParameter(MONITOR_KEY);
 
         setRouterChain(routerChain);
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/directory/AbstractDirectory.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/directory/AbstractDirectory.java
@@ -32,7 +32,6 @@ import java.util.List;
 import java.util.Map;
 
 import static org.apache.dubbo.common.constants.CommonConstants.DUBBO;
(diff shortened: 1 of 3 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/directory/AbstractDirectory.java
```diff
@@ -72,8 +75,13 @@ public AbstractDirectory(URL url, RouterChain<T> routerChain) {
         this.consumedProtocol = this.queryMap.get(PROTOCOL_KEY) == null ? DUBBO : this.queryMap.get(PROTOCOL_KEY);
         this.url = url.removeParameter(REFER_KEY).removeParameter(MONITOR_KEY);
 
-        this.consumerUrl = this.url.setProtocol(consumedProtocol).setPath(path == null ? queryMap.get(INTERFACE_KEY) : path).addParameters(queryMap)
-                .removeParameter(MONITOR_KEY);
+        URL consumerUrlFrom = this.url.setProtocol(consumedProtocol)
+                .setPath(path == null ? queryMap.get(INTERFACE_KEY) : path);
+        if (isUrlFromRegistry) {
+            // reserve parameters if url is already a consumer url
+            consumerUrlFrom = consumerUrlFrom.clearParameters();
```
kylixs, line 82: Will clearing all the parameters of the registry url cause the loss of some useful parameters of the consumer url?

kylixs (commented):
I think it's best to fix it based on `2.7.9-release` for  easy verification.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/directory/AbstractDirectory.java
AlbumenJ: > Will clearing all the parameters of the registry url cause the loss of some useful parameters of the consumer url?

The paramaters in url which is from `DynamicDirectory`  is the paramaters of registry. Real reference paramaters are bring encoded and set in `refer` paramater which is `queryMap` here.

chickenlj:
We will release it in the next version.

kylixs:
> We will release it in the next version.

This bug may affect users who set the `group` attribute of the `registry`. If there is no problem with the modified code, can it be merged into 2.7.9?

chickenlj:
I think there's no rush with this bugfix. I believe this is a regression of 2.7.8 or a few versions before. Considering that `registry group` feature is mostly used by some advanced users who most likely would stay in a stable version, there won't have many users facing this issue.

A new round of apache vote would take days to weeks, I think it's better to put it in the next release

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/directory/StaticDirectory.java
```diff
@@ -1,113 +1,113 @@
...
+            try {
+                finalInvokers = routerChain.route(getConsumerUrl(), invocation);
+            } catch (Throwable t) {
+                logger.error("Failed to execute router: " + getUrl() + ", cause: " + t.getMessage(), t);
+            }
+        }
+        return finalInvokers == null ? Collections.emptyList() : finalInvokers;
+    }
+
+}
```
chickenlj, line 113: Please check why the whole file is marked as changed.

chickenlj (changes requested):

PR #7201, opened 2021-02-19
[DUBBO-52] return true directly instead of box and unbox

XenoAmess:

return true directly instead of box and unbox


AlbumenJ (approved):
LGTM.

PR #7203, opened 2021-02-19
[DUBBO-54] bug: String.format have wrong number of parameters.

XenoAmess:

bug: String.format have wrong number of parameters.


AlbumenJ (approved):
LGTM.

PR #7209, opened 2021-02-19
[DUBBO-60] use Math.max and Math.min

XenoAmess:

use Math.max and Math.min


AlbumenJ (approved):
LGTM.

PR #7219, opened 2021-02-19
[DUBBO-70] use switch instead of if.

XenoAmess:

use switch instead of if.

also clean some codes in class ReflectUtils


AlbumenJ (approved):
LGTM.

PR #7224, opened 2021-02-19
[DUBBO-75] Remove useless boxing and unboxing

XenoAmess:

Remove useless boxing and unboxing


AlbumenJ (approved):
LGTM.

PR #7226, opened 2021-02-19
[DUBBO-77] use a way to disable disabled tests that normal people will choose to use.

XenoAmess:

use a way to disable disabled tests that normal people will choose to use.


AlbumenJ (approved):
LGTM.

PR #7248, opened 2021-02-23
3.0 api refactor

ppzzyy11:

Make sure every layer only responsible for their own work.

## Brief changelog

Add the DubboServer to be the Class responsible to trigger the init/start process of the ReferenceConfig/ServiceConfig. 
Most methods of the DubboBootstrap call the methods with the same names in DubboServer.

## Verifying this change

I run Dubbo Demo in my PC. And demos based on DubboServer/DubooBootstrap/API works well.

##

The developing process is based on this link below and many unit testes can not pass before my modification.

https://github.com/apache/dubbo/issues/5666

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/dubboserver/DubboServer.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/dubboserver/DubboServer.java
@@ -16,12 +16,7 @@
  */
 package org.apache.dubbo.config.dubboserver;
 
-import org.apache.dubbo.common.URL;
 import org.apache.dubbo.common.config.ConfigurationUtils;
-import org.apache.dubbo.common.config.Environment;
-import org.apache.dubbo.common.config.configcenter.DynamicConfiguration;
-import org.apache.dubbo.common.config.configcenter.DynamicConfigurationFactory;
-import org.apache.dubbo.common.config.configcenter.wrapper.CompositeDynamicConfiguration;
 import org.apache.dubbo.common.extension.ExtensionLoader;
 import org.apache.dubbo.common.lang.ShutdownHookCallback;
 import org.apache.dubbo.common.lang.ShutdownHookCallbacks;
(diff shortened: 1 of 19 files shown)


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/dubboserver/DubboServer.java
```diff
@@ -0,0 +1,1277 @@
...
+    /**
+     * A delayed exposure service timer
+     */
+    private static final ScheduledExecutorService DELAY_EXPORT_EXECUTOR = Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory("DubboServiceDelayExporter", true));
+
+    /**
+     * See {@link ApplicationModel} and {@link ExtensionLoader} for why DubboServer is designed to be singleton.
+     */
+    public static DubboServer getInstance(DubboBootstrap dubboBootstrap) {
+        if (instance == null) {
```
kylixs, line 219: Does it make sense to keep the dubbo server instance? The getInstance method with parameters looks strange, and the `DubboBootstrap` parameter is not used when the getInstance method is called for the second time.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/dubboserver/DubboServer.java
```diff
@@ -0,0 +1,1277 @@
...
+
+    public void init4ReferenceConfig(ReferenceConfig referenceConfig) {
+        if (referenceConfig.isInitialized()) {
+            return;
+        }
+
+        //use local variables to resonate with init() copied from ReferenceConfig
+        DubboBootstrap dubboBootstrap = referenceConfig.getDubboBootstrap();
+        if (dubboBootstrap == null) {
+            dubboBootstrap = DubboBootstrap.getInstance();
```
kylixs, line 484: The logic here is not clear, you should not initialize `DubboBootstrap` again, but use `this.dubboBootstrap`.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/dubboserver/DubboServer.java
```diff
@@ -0,0 +1,1277 @@
...
+        asyncReferringFutures.clear();
+        cache.destroyAll();
+    }
+
+    private void exportService(ServiceConfigBase sc) {
+        if (sc instanceof ServiceConfig) {
+            if (exportAsync) {
+                ExecutorService executor = executorRepository.getServiceExporterExecutor();
+                Future<?> future = executor.submit(() -> {
+                    sc.export();
```
kylixs, line 1053: Calling `export ServiceConfig(sc)` is more clear, `ServiceConfig Base.export()` is a compatible API

chickenlj:
Closed

Issue #7252, opened 2021-02-23
Some spi class already be removed but the internal files still exist it.


horizonzy:
https://github.com/apache/dubbo/blob/9fd271b6ecbe76b811856b18e9afc4b2b749572a/dubbo-config/dubbo-config-api/src/main/resources/META-INF/dubbo/internal/org.apache.dubbo.event.EventListener#L1

https://github.com/apache/dubbo/blob/9fd271b6ecbe76b811856b18e9afc4b2b749572a/dubbo-config/dubbo-config-api/src/main/resources/META-INF/dubbo/internal/org.apache.dubbo.metadata.MetadataServiceExporter#L3

PR #7253, opened 2021-02-23
[ISSUE-#7252] Remove some class in spi file which is removed.

horizonzy:

For-#7252


chickenlj (approved):
LGTM

Issue #7261, opened 2021-02-24
Remove logger which is already defined in superclass.


horizonzy:
Some class already define logger at superclass such as AbstractMetadataReport.

https://github.com/apache/dubbo/blob/15fc5b473bbeb73b0a6c9ac02f2c8d7e0f6fbe82/dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/report/support/AbstractMetadataReport.java#L87

The subclass needn't define logger again. But some class do as it, such as NacosMetadataReport.

https://github.com/apache/dubbo/blob/15fc5b473bbeb73b0a6c9ac02f2c8d7e0f6fbe82/dubbo-metadata/dubbo-metadata-report-nacos/src/main/java/org/apache/dubbo/metadata/store/nacos/NacosMetadataReport.java#L70

Remove the logger which is define at subclass.

AlbumenJ:
close #7261 via #7262

PR #7262, opened 2021-02-24
[ISSUE-#7261] Remove unnecessary logger which is defined  at subclass.

horizonzy:

For-#7261.


AlbumenJ (approved):
LGTM.

PR #7266, opened 2021-02-24
Code optimize about DubboBootstrap#ready

horizonzy:
Some logic optimize about DubboBootstrap#ready.


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java
```diff
@@ -172,7 +172,7 @@
 
     private AtomicBoolean started = new AtomicBoolean(false);
 
-    private AtomicBoolean ready = new AtomicBoolean(true);
```
AlbumenJ, line 175: We should adapt the way users use API to start ( such as directly refer with `ReferenceConfig` and this may cause dubbo-qos always return not ready state)

PR #7278, opened 2021-02-26
Fix jdk serialization, add ObjectInputStream wrapper

chickenlj:


AlbumenJ (approved):
LGTM.

chickenlj:
Is this patch still necessary, I noticed that the lates jdk 7/8/9 has provided a serialization data filtering mechanism JEP290 which the users can easily enable on their own.

Issue #7281, opened 2021-02-27
When using Telnet in QoS, the idle connect may not be closed in time, despite using the idle check.


Spongecaptain:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

## Bug Specification

When using Telnet in QoS, the idle connect may not be closed in time, despite using the idle check.

When using `IdleStateHandler` with `TelnetProcessHandler` in the `QosProcessHandler`, the timeout is set to  5*60 seconds, but **NO** handler deals with the `IdleStateEvent`. Maybe the `TelnetProcessHandler` should override the `userEventTriggered` method to handle this Event, and close the idle connect.

AlbumenJ:
close #7281 via #7282

PR #7282, opened 2021-02-27
fix the bug: the channel will not closed in time when timeout by using telnet in QOS

Spongecaptain:

Fix the bug: don't close the channel when timeout by using telnet in QOS.

see #7281

## Brief changelog

When using `IdleStateHandler` with `TelnetProcessHandler` in the `QosProcessHandler`, the timeout is set to  5*60 seconds, but **NO** handler deals with the `IdleStateEvent`. Maybe the `TelnetProcessHandler` should override the `userEventTriggered` method to handle this Event, and close the idle connect.

### DIFF
--- a/dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/server/handler/IdleEventHandler.java
+++ b/dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/server/handler/IdleEventHandler.java
@@ -1,3 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.apache.dubbo.qos.server.handler;
 
 import io.netty.channel.Channel;
--- a/dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/server/handler/IdleEventHandler.java
+++ b/dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/server/handler/IdleEventHandler.java
@@ -1,0 +1,27 @@
+package org.apache.dubbo.qos.server.handler;
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.timeout.IdleStateEvent;
+import org.apache.dubbo.common.logger.Logger;
+import org.apache.dubbo.common.logger.LoggerFactory;
+
+@ChannelHandler.Sharable
+public class IdleEventHandler extends ChannelDuplexHandler {
+    private static final Logger log = LoggerFactory.getLogger(IdleEventHandler.class);
+
+    @Override
+    public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+        // server will close channel when server don't receive any request from client util timeout.
+        if (evt instanceof IdleStateEvent) {
+            Channel channel = ctx.channel();
+            log.info("IdleStateEvent triggered, close channel " + channel);
+            channel.close();
+        } else {
+            super.userEventTriggered(ctx, evt);
+        }
+    }
+
+}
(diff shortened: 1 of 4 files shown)


horizonzy (commented):
Create a new handler to process this case will be better.

Spongecaptain:
@horizonzy 
Thnx.

In my opinion, for the reason of uniform code style, I prefer to add the `userEventTriggered` method in `TelnetProcessHandler`. Refer to class `NettyClientHandler` & `NettyServerHandler`. 

Could you tell me the reason why you think adding a Handler would be better in this case.

horizonzy:
Single responsibility, TelnetProcessHandler just handle the command from client. 
In this pr, it's supply a mechanism to close the channel which didn't exchange data for a long time. 
In the future, if need add some logic to handle the timeout channel, the logic shouldn't append in TelnetProcessHandler.
And if anyone didn't need this mechanism, just didn't add the new handler

Spongecaptain:
> Single responsibility, TelnetProcessHandler just handle the command from client.
> In this pr, it's supply a mechanism to close the channel which didn't exchange data for a long time.
> In the future, if need add some logic to handle the timeout channel, the logic shouldn't append in TelnetProcessHandler.
> And if anyone didn't need this mechanism, just didn't add the new handler

Yeah, I agree with you, but the fact is that `NettyClientHandler` & `NettyServerHandler` don't look like meeting the  rule of Single responsibility. I thought it as a code style.. QAQ

horizonzy:
> > Single responsibility, TelnetProcessHandler just handle the command from client.
> > In this pr, it's supply a mechanism to close the channel which didn't exchange data for a long time.
> > In the future, if need add some logic to handle the timeout channel, the logic shouldn't append in TelnetProcessHandler.
> > And if anyone didn't need this mechanism, just didn't add the new handler
> 
> Yeah, I agree with you, but the fact is that `NettyClientHandler` & `NettyServerHandler` don't look like meeting the rule of Single responsibility. I thought it as a code style.. QAQ

Hi, I'm not agree with you. `NettyClientHandler` and `NettyServerHandler` all didn't coupling with business. But TelnetProcessHandler is for business.

Spongecaptain:
> Hi, I'm not agree with you. `NettyClientHandler` and `NettyServerHandler` all didn't coupling with business. But TelnetProcessHandler is for business.

You are right and enlighten me a lot, now I can see the difference. It is better to use a New `Handler`.

Thnks a lot, have a good night !

horizonzy:
Hi, `apache license` should append at the top of the new class.

Spongecaptain:
> Hi, `apache license` should append at the top of the new class.

I forgot the commit specifications, plz forgive me. QAQ...

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/server/handler/IdleEventHandler.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.timeout.IdleStateEvent;
+import org.apache.dubbo.common.logger.Logger;
+import org.apache.dubbo.common.logger.LoggerFactory;
+
+@ChannelHandler.Sharable
```
horizonzy, line 27: It's not necessary.

Spongecaptain: if you means the `IdleEventHandler` is not necessary, then the `p.addLast(new IdleStateHandler(0, 0, 5 * 60));` in `QosProcessHandler` may be also not necessary.

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/server/handler/IdleEventHandler.java
```diff
@@ -0,0 +1,43 @@
...
+public class IdleEventHandler extends ChannelDuplexHandler {
```
horizonzy, line 28: The class name use `TelnetIdleEventHandler` will be better, what do you think about it.

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/server/handler/IdleEventHandler.java
Spongecaptain: it is kind of chinglish? QAQ hh

horizonzy: I mean the annotation `@Sharable`.

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/server/handler/IdleEventHandler.java
Spongecaptain: It is ok. Because when we use HTTP, All HTTP request is HTTP short connection, IdleEventHandler only used for telnet.

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/server/handler/IdleEventHandler.java
Spongecaptain: but why？

horizonzy: It's alway create new `TelnetIdleEventHandler` for channel.

horizonzy (approved):
LGTM

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/server/handler/IdleEventHandler.java
Spongecaptain: Ok, you are right.

AlbumenJ (approved):
LGTM.

PR #7294, opened 2021-03-01
Fix it in Dubbo 3.0

AlbumenJ:
- Remove parameters from registryUrl to prevent config pollution ( Related with #7189 )
- Change to `url.getRemoteApplication()` for `TagRouter`
- Distint `MigrationRuleHandler` with timestamp


chickenlj (approved):
LGTM

Issue #7299, opened 2021-03-01
Some problem about ZookeeperServiceDiscoveryChangeWatcher.


horizonzy:
ZookeeperServiceDiscoveryChangeWatcher received event and register self by `                zookeeperServiceDiscovery.registerServiceWatcher(serviceName, listener);
`
https://github.com/apache/dubbo/blob/51d5a88a067876abf9d9a8ba40a1426c4f805909/dubbo-registry/dubbo-registry-zookeeper/src/main/java/org/apache/dubbo/registry/zookeeper/ZookeeperServiceDiscoveryChangeWatcher.java#L55-L66

https://github.com/apache/dubbo/blob/51d5a88a067876abf9d9a8ba40a1426c4f805909/dubbo-registry/dubbo-registry-zookeeper/src/main/java/org/apache/dubbo/registry/zookeeper/ZookeeperServiceDiscovery.java#L181-L206

3 problems:

1.ZookeeperServiceDiscoveryChangeWatcher.line_61 and line_63, when it notify, query serviceName instance alonely, it should notify use same instanceList. Maybe exist this situation, line_61 query one instance, but line_63, query two instances.

2.the way of ZookeeperServiceDiscoveryChangeWatcher re register, it use  `zookeeperServiceDiscovery.registerServiceWatcher(serviceName, listener)`, it always create parent path, I think it's necessary.

3.th way of ZookeeperServiceDiscoveryChangeWatcher re register. it may cause ZookeeperServiceDiscoveryChangeWatcher can't be remove from watcherCaches when`remove wacter` and `zk event trigger` concurrently.

The remove logic:
https://github.com/apache/dubbo/blob/51d5a88a067876abf9d9a8ba40a1426c4f805909/dubbo-registry/dubbo-registry-zookeeper/src/main/java/org/apache/dubbo/registry/zookeeper/ZookeeperServiceDiscovery.java#L164-L169

Maybe exist this situation, when removeServiceInstancesChangedListener, the ZookeeperServiceDiscoveryChangeWatcher's 
keepWatching will be setted as `false`, but before it, zk event come in, it will invoke `ZookeeperServiceDiscoveryChangeWatcher#process`, before ` zookeeperServiceDiscovery.registerServiceWatcher(serviceName, listener)
`, watcherCaches removed ZookeeperServiceDiscoveryChangeWatcher,
then `zookeeperServiceDiscovery.registerServiceWatcher(serviceName, listener) 
[...]

PR #7300, opened 2021-03-01
[ISSUE-#7299] Logic optmization about zookeeperServiceDiscoveryChangeWatcher.

horizonzy:
For-#7299
1.use same instances to notify listener.onEvent and dispatchServiceInstancesChangedEvent.
2.modify the re watcher machine mechanism, re watcher by ZookeeperServiceDiscoveryChangeWatcher self.


wangchengming666:
pls resolve conflict

PR #7305, opened 2021-03-03
fix filter chain order

chickenlj:


AlbumenJ (approved):
LGTM.

AlbumenJ:
> I don't know that why to change the filter order? @AlbumenJ

For dubbo filter chain, smaller is higher

Issue #7347, opened 2021-03-09
经历2.7.8坑后，好不容易等来个2.7.9，结果升级整合报错，望指点指点


pangshuqiang:
环境：
springboot：2.3.7.RELEASE
springcloud：Hoxton.SR9
springcloudalibaba：2.2.5.RELEASE
nacos：1.4.1

正常情况下，springcloudalibaba-2.2.5.RELEASE里自带的事2.7.8版本的Dubbo，该版本下项目是能运行OK。

由于小马哥告知有新版，所以想单独替换Dubbo为2.7.9，调整后内容如下：
`
<dependency>
            <groupId>org.apache.dubbo</groupId>
            <artifactId>dubbo-spring-boot-starter</artifactId>
            <version>2.7.8</version>
            <exclusions>
                <exclusion>
                    <groupId>org.apache.dubbo</groupId>
                    <artifactId>dubbo</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
        <dependency>
            <groupId>org.apache.dubbo</groupId>
            <artifactId>dubbo</artifactId>
            <version>2.7.9</version>
        </dependency>
`

但是运行直接报错：
`2021-03-09 15:14:48.913  INFO 8364 --- [           main] c.a.s.u.BeanRegistrar                    : The Infrastructure bean definition [Root bean: class [org.apache.dubbo.config.spring.beans.factory.annotation.ReferenceAnnotationBeanPostProcessor]; scope=; abstract=false; lazyInit=null; autowireMode=0; dependencyCheck=0; autowireCandidate=true; primary=false; factoryBeanName=null; factoryMethodName=null; initMethodName=null; destroyMethodName=nullwith name [referenceAnnotationBeanPostProcessor] has been registered.
2021-03-09 15:14:48.914  INFO 8364 --- [           main] c.a.s.u.BeanRegistrar                    : The Infrastructure bean definition [Root bean: class [org.apache.dubbo.config.spring.beans.factory.annotation.DubboConfigAliasPostProcessor]; scope=; abstract=false; lazyInit=null; autowireMode=0; dependencyCheck=0; autowireCandidate=true; primary=false; factoryBeanName=null; factoryMethodName=null; initMethodName=null; destroyMethodName=nullwith name [dubboConfigAliasPostProcessor] has been registered.
[...]

PR #7357, opened 2021-03-10
Add empty constructor for DubboLifecycleComponentApplicationListener

AlbumenJ:

add empty constructor for DubboLifecycleComponentApplicationListener

related issue #7347


kylixs:
`DubboBootstrapApplicationListener` also an empty constructor for compatibility old version

AlbumenJ:
> `DubboBootstrapApplicationListener` also need an empty constructor for compatible with old version

done

dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/context/DubboBootstrapApplicationListener.java
```diff
@@ -43,6 +43,10 @@
 
     private final DubboBootstrap dubboBootstrap;
 
+    public DubboBootstrapApplicationListener() {
+        this.dubboBootstrap = DubboBootstrap.getInstance();
```
horizonzy, line 47: If init by this constructor, the applicationContext in super class `OnceApplicationContextEventListener` will null, when call getApplicationContext(), cause NPE, I'm not sure this case will be happen.

AlbumenJ: hi @horizonzy ,

I am not quite sure which class you mentioned which has `getApplicationContext()`.

I have found there are two places use `applicationContext`

- OnceApplicationContextEventListener

`ApplicationContextAware` has been implemented by `OnceApplicationContextEventListener`, so `applicationContext` will be automaticly injected by `public final void setApplicationContext(ApplicationContext applicationContext) throws BeansException`

- DubboBootstrapStartStopListenerSpringAdapter

The methods in `DubboBootstrapStartStopListenerSpringAdapter` will be called after `DubboBootstrap` started and before this `ApplicationContextEvent` will notify `DubboLifecycleComponentApplicationListener` with `applicationContext`.

horizonzy (approved):
LGTM.

PR #7378, opened 2021-03-16
Add arguments router to ConditionRouter

furaul:
Add arguments router logic to ConditionRouter.

Add arguments router to ConditionRouter.
Compatible to current ConditionRouter logic.

## Brief changelog

Add arguments router to ConditionRouter

## Verifying this change

Add arguments router to ConditionRouter

## Examples
In some cases, we need to choose the right provider while the arguments of the method is setted.
The condition rule is like

> arguments[1].str!=b  => host = 1.1.1.1

While the str property of the second argument is equal to b, the providers whose host is equal to 1.1.1.1 are to be choosed。
We could use this feature to route requests according to business params. 
The feature is compatible to current ConditionRouter logic.

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
@@ -23,6 +23,7 @@ import org.apache.dubbo.common.utils.CollectionUtils;
 import org.apache.dubbo.common.utils.NetUtils;
 import org.apache.dubbo.common.utils.StringUtils;
 import org.apache.dubbo.common.utils.UrlUtils;
+import org.apache.dubbo.common.utils.FieldUtils;
 import org.apache.dubbo.rpc.Invocation;
 import org.apache.dubbo.rpc.Invoker;
 import org.apache.dubbo.rpc.RpcException;
@@ -308,12 +309,12 @@ public class ConditionRouter extends AbstractRouter {
                 if (object == null) {
                     return false;
                 }
-                Field field = object.getClass().getDeclaredField(expressArray[i]);
+                Field field = FieldUtils.getDeclaredField(object.getClass(), expressArray[i]);
                 field.setAccessible(true);
                 object = field.get(object);
             }
 
-            if (matchPair.getValue().isMatch((String)object, null)) {
+            if (matchPair.getValue().isMatch(String.valueOf(object), null)) {
                 return true;
             }
         } catch (Exception e) {
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
@@ -23,14 +23,12 @@ import org.apache.dubbo.common.utils.CollectionUtils;
 import org.apache.dubbo.common.utils.NetUtils;
 import org.apache.dubbo.common.utils.StringUtils;
 import org.apache.dubbo.common.utils.UrlUtils;
-import org.apache.dubbo.common.utils.FieldUtils;
 import org.apache.dubbo.rpc.Invocation;
 import org.apache.dubbo.rpc.Invoker;
 import org.apache.dubbo.rpc.RpcException;
 import org.apache.dubbo.rpc.cluster.Constants;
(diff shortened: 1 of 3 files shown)


furaul:
https://github.com/apache/dubbo/pull/5649

furaul:
@AlbumenJ

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
```diff
@@ -262,6 +275,54 @@ private boolean matchCondition(Map<String, MatchPair> condition, URL url, URL pa
...
+                return false;
+            }
+
+            //extract the argument value
+            Object object = invocation.getArguments()[index];
+            for (int i = 1; i < expressArray.length; i++) {
+                if (object == null) {
+                    return false;
+                }
+                Field field = object.getClass().getDeclaredField(expressArray[i]);
```
AlbumenJ, line 311: pls use FieldUtils

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
```diff
@@ -262,6 +275,54 @@ private boolean matchCondition(Map<String, MatchPair> condition, URL url, URL pa
...
+                field.setAccessible(true);
+                object = field.get(object);
+            }
+
+            if (matchPair.getValue().isMatch((String)object, null)) {
```
AlbumenJ, line 316: Shall we adapt those types not match string? Or using toString?

AlbumenJ (changes requested):

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
furaul: Good advice.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
furaul: 👌

AlbumenJ (approved):

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionRouter.java
```diff
@@ -262,6 +276,54 @@ private boolean matchCondition(Map<String, MatchPair> condition, URL url, URL pa
         return result;
     }
 
+    /**
+     * analysis the arguments in the rule.
+     * Examples would be like this:
+     * "arguments[0]=1", whenCondition is that the first argument is equal to '1'.
+     * "arguments[1].param=a", whenCondition is that the 'param' attribute in the second argument is equal to 'a'.
+     * "arguments[2].inner.param=b", whenCondition is that the 'param' attribute in the 'inner' attribute of the third argument is equal to 'b'.
```
chickenlj, line 284: I think it's a good complement to have an argument matcher.

I think we can make the argument routing rules simpler while still covering more than 90% of the usage scenarios.

For example, we only support the matching of string type parameters and do not support t matching of complex objects like `arguments[1].param`. Or even our routing rule is like a switch, when the user tells us to turn it on, we will constantly match the first parameter of the method and require that this parameter be a string or primitive type.

furaul: > I think it's a good complement to have an argument matcher.
> 
> I think it's better if we can make the argument routing rules simpler while still covering more than 90% of the usage scenarios.
> 
> For example, we only support the matching of string type parameters and do not support t matching of complex objects like `arguments[1].param`. Or even our routing rule is like a switch, when the user tells us to turn it on, we will constantly match the first parameter of the method and require that this parameter be a string or primitive type.

I would like to simplify the arguments rule to make it easier to be understood and used.
For example,  we only support "one flour param", like  "arguments[0]=1", and only support java primary data type, like int, long, string and so on.
But I think it is not a good idea to make it like a switch, that would reduce flexibility. 
And in the future, I would like to config arguments router in the configuration pages, we could use drop-down box to choose the right arguments and its value. And in that time, we could expand the explementation of arguments router.
@AlbumenJ

AlbumenJ (approved):
LGTM.

PR #7420, opened 2021-03-21
About the application of spring management repeated monitoring processing dubbo shutdown

xiaoheng1:

About the application of spring management repeated monitoring processing dubbo shutdown.

see more detail from #7093 

## Brief changelog

XXXXX

## Verifying this change

XXXXX

### DIFF
--- a/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/context/DubboBootstrapApplicationListener.java
+++ b/dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/context/DubboBootstrapApplicationListener.java
@@ -19,6 +19,9 @@ package org.apache.dubbo.config.spring.context;
 import org.apache.dubbo.config.bootstrap.DubboBootstrap;
 
 import com.alibaba.spring.context.OnceApplicationContextEventListener;
+import org.apache.dubbo.config.event.DubboServiceDestroyedEvent;
+import org.apache.dubbo.event.EventDispatcher;
+import org.apache.dubbo.event.EventListener;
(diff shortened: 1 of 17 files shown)


kylixs:
There seems to be a problem with the stop logic. The Shutdown hook is called by the JVM before the JVM exits. **It is not good to wait for the dubbo shutdown hook to complete in the Spring close event.** Actually Spring also registered shutdown hook, please check org.springframework.context.support.AbstractApplicationContext#registerShutdownHook().

* In the Spring scenario, perhaps we should listen to Spring events to shut down dubbo instead of using shutdown hook. In this case, the life cycle of dubbo is consistent with spring. 
* In the Dubbo API scenario, the shutdown hook needs to be registered.

xiaoheng1:
Now it is precisely because of the existence of Spring hook and Dubbo Hook. As a result, Spring did not block when closing dubbo, and then continued to execute the destruction of Spring Bean. At this time, Dubbo has not completed the destruction, which will cause the database connection not to be obtained. It may be a way to cancel Dubbo Hook directly after hosting to the Spring container.

xiaoheng1:
like https://github.com/apache/dubbo/pull/7317.

 If the idea of #7317 is used, should the callback be removed from ShutdownHook? ShutDownHook only calls DubboBootstrap.stop to complete the resource wave work.

kylixs:
For clarity of responsibilities, the processing logic of dubbo destroy should extracted from the Dubbo shutdown hook, and  called it separately in the Spring scenario without relying on the Dubbo shutdown hook.

xiaoheng1:
So when shutting down the Dubbo service in Spring, do you tend to cancel the registration of DubboShutDownHook?

PR #7428, opened 2021-03-22
Use java box to prevent malicious code.

horizonzy:
If users want prevent maliciout code in router script. 
step one:  define a custom security file with content
```
grant {
	  permission java.security.AllPermission;
};
``` 
step two:  start dubbo-consumer with system property: -Djava.security.manager -Djava.security.policy=`customFile location`.

This solution maybe decrese the application performance.


chickenlj (approved):
LGTM

horizonzy:
The sciptRouter may exist security risk, some can inject malicious code to the zk nodes, becase most user didn't ssl the zk. So we remove scriptRouterFactory in dubbo internal spi file. If some still wants to use scriptRouter, add scriptRouterFactory in self spi file, and worry be attack by script, config as the first comment.

AlbumenJ (approved):
LGTM.

horizonzy:
The scriptRouter be removed at this pr, if user still wants use it, add scriptRouterFactory in selfs spi file.

PR #7435, opened 2021-03-23
Improved conditional test execution

eas5:
**Problem:**
Conditions within a test method may produce a false test outcome. Once a condition is not met, and the code block inside the condition is not executed, a false 'passed' outcome can be produced without the test ever being executed.

**Solution:**
Instead of using a conditional to control the test execution, we propose using JUnit Assumptions — a failed assumption results in a test being aborted. Assumptions are typically used whenever it does not make sense to continue executing a given test method.

**Result:**
_Before:_
```
if (address instanceof Inet6Address) {
    assertThat(NetUtils.isPreferIPV6Address(), equalTo(true));
}
```

_After:_
```
assumeTrue(address instanceof Inet6Address);
assertThat(isPreferIPV6Address, equalTo(true));
```


AlbumenJ (approved):
LGTM.

PR #7436, opened 2021-03-24
Add serialize class checker

AlbumenJ:

default block some class in gerenic filter deserializing


chickenlj (approved):
LGTM

owen200008:
image

we use generic to call 1W/s with dubbo2.7.8，dubbo update to 2.7.12 the speed is 7000/s and the memory use is too large。
what is the purpose of the CLASS_ALLOW_LFU_CACHE? one call add one node?

owen200008:
when get the CLASS_ALLOW_LFU_CACHE, why use moveToNextFreqQueue?

owen200008:
1. first i will add the switch
later， i want change this like，the function call frequently
1. name not need to lower
2. no use lock in this way(read more, write less)，use map the key is Class, the max size is same with class space

Issue #7434, opened 2021-03-23
Redis Registry 2.7.9 consumer 启动失败


hexw:
### Environment

* Dubbo version: 2.7.9
* Operating System version: xxx
* Java version: 1.8

hexw:
从2.7.8升级到2.7.9 使用redis Registry 客户端启动报错

hexw:
由 InterfaceCompatibleRegistryProtocol 引发这个问题

hexw:
直接使用dubbo-demo-annotation 
配置
dubbo.application.name=dubbo-demo-annotation-consumer
dubbo.registry.address=redis://redis:password@$127.0.0:6379?db.index=0
报错
Exception in thread "main" org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'demoServiceComponent': Injection of @DubboReference dependencies is failed; nested exception is java.lang.IllegalStateException: No such extension org.apache.dubbo.registry.client.ServiceDiscovery by name redis, no related exception was found, please check whether related SPI module is missing.
	at com.alibaba.spring.beans.factory.annotation.AbstractAnnotationBeanPostProcessor.postProcessPropertyValues(AbstractAnnotationBeanPostProcessor.java:149)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1425)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:593)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:516)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:324)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:226)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:322)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:202)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:897)
[...]

PR #7452, opened 2021-03-26
[Dubbo-7434]Change migration default step to force interface

kevinw66:

For-[#7434](https://github.com/apache/dubbo/issues/7434)
We should use interface invoker as default in 2.7.x

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/migration/MigrationRule.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/migration/MigrationRule.java
@@ -24,6 +24,8 @@ import org.yaml.snakeyaml.constructor.Constructor;
 
 import java.util.Optional;
 
+import static org.apache.dubbo.common.constants.RegistryConstants.INIT;
+
 public class MigrationRule {
     private static final String DUBBO_SERVICEDISCOVERY_MIGRATION_KEY = "dubbo.application.service-discovery.migration";
     public static final String DUBBO_SERVICEDISCOVERY_MIGRATION_GROUP = "MIGRATION";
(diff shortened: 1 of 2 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/migration/MigrationRule.java
```diff
@@ -82,9 +83,9 @@ public static MigrationRule queryRule() {
         return parse(rawRule);
     }
 
-    private  static MigrationRule getMigrationRule(String step) {
+    private static MigrationRule getMigrationRule(String step) {
         MigrationRule rule = new MigrationRule();
-        rule.setStep(Enum.valueOf(MigrationStep.class, StringUtils.isBlank(step) ? MigrationStep.APPLICATION_FIRST.name() : step));
+        rule.setStep(Enum.valueOf(MigrationStep.class, StringUtils.isBlank(step) ? MigrationStep.FORCE_INTERFACE.name() : step));
```
horizonzy, line 88: The APPLICATION_FIRST is ok, I think. The logic in org.apache.dubbo.registry.client.migration.MigrationInvoker#migrateToServiceDiscoveryInvoker will handle service discovery and interface discovery both.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/migration/MigrationInvoker.java
```diff
@@ -112,10 +112,10 @@ public synchronized void migrateToServiceDiscoveryInvoker(boolean forceMigrate)
             refreshServiceDiscoveryInvoker();
             refreshInterfaceInvoker();
             setListener(invoker, () -> {
-                this.compareAddresses(invoker, serviceDiscoveryInvoker);
+                this.compareAddresses(serviceDiscoveryInvoker, invoker);
```
horizonzy, line 115: nice.👍

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/migration/MigrationRule.java
kevinw66: The application-level service discovery is incomplete in 2.7.x, maybe we should use stable version(FORCE_INTERFACE) here. But keep APPLICATION_FIRST in 3.0, but also we should discuss this in more detail. And I'll extract other changes to a separated pull request and keep this here.

Issue #7459, opened 2021-03-28
Passing wrong parameters when calling MigrationInvoker.compareAddresses()


kevinw66:
I found the compareAddresses method declaration is like below
![image](https://user-images.githubusercontent.com/31196226/112747572-cf6ace00-8fe8-11eb-9e1f-119f9360af83.png)

But the caller pass wrong parameter order when calling this method.
![image](https://user-images.githubusercontent.com/31196226/112747552-b6fab380-8fe8-11eb-88d5-7db0a6e8177a.png)

PR #7460, opened 2021-03-28
[Dubbo-7459]Fix passing wrong parameter when calling MigrationInvoker.compareAddresses()

kevinw66:

For-[#7459](https://github.com/apache/dubbo/issues/7459)


AlbumenJ (approved):
LGMT.

PR #7461, opened 2021-03-28
enhance LRUCache

Augustvic:

Update LRU to LRU-2

## Brief changelog

Reduce cache pollution by filtering one access data with a history list. This strategy will not occupy too much extra memory because the history list only saves the keys.

When the data accessed for the first time, add it to the history list. If the size of the history list reaches max capacity, eliminate the earliest data. When the data already exists in the history list, and be accessed for the second time, then it will be put into the real cache.

## Verifying this change

Also need to update LRUCacheTest, CacheTest, and CacheFilterTest due to the two-time verification mechanism of LRU-2, and two calls need to be performed to put the cache into LRUCache, which shows as commit. 

The performance between LRU and LRU2 refer to https://blog.csdn.net/Victorgcx/article/details/104378378.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/LRUCache.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/LRUCache.java
@@ -20,6 +20,12 @@ import java.util.LinkedHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+/**
+ * LRU-2
+ * </p>
+ * When the data accessed for the first time, add it to history list. If the size of history list reaches max capacity, eliminate the earliest data (first in first out).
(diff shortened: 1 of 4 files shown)


AlbumenJ:
@Augustvic Please check why dubbo-samples-cache is failed in integration test. 
See https://github.com/apache/dubbo/pull/7461/checks?check_run_id=2212809120.

Augustvic:
> 
> 
> @Augustvic Please check why dubbo-samples-cache is failed in integration test.
> See https://github.com/apache/dubbo/pull/7461/checks?check_run_id=2212809120.

I think we can solve this problem through https://github.com/apache/dubbo-samples/pull/292, but I‘m not sure if it's appropriate to change dubbo-sample, please review. Thank you.

AlbumenJ:
@Augustvic I have merged the pull request in dubbo-samples, and re-run integration test. We can wait until the tests end.

Issue #7003, opened 2020-12-08
registry twice consumer service use @DubboReference annotation


JstCoding:
* Dubbo version: 2.7.8
* Operating System version: window
* Java version: 1.8
* Registry center: zookeeper 

example:

class A:
@DubboReference
ExampleService exampleSerivce

class B:
@DubboReference(check =false)
ExampleService exampleService;

where project start up, get a warm level log about zookeeper registry consumer fail, consumer exampleService registry twice to zookeeper
such as:

 org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /dubbo/net.xxxx.service,ExampleService/consumers/xxxxx

yuxiao97:
Maybe you use JavaConfig reference dubbo service, like this:
```
@Configuration
public class DubboServiceReferenceConfig {

    @DubboReference(version = "${dubbo.application.version}", check = false)
    private UserInfoService userInfoService;

    @Bean
    public UserInfoService userInfoService() {
        return userInfoService;
    }
    
}
```
Then use @Autowired annotation on your consumer service.

kylixs:
This bug has been fixed: https://github.com/apache/dubbo/pull/7383

kylixs:
> Maybe you can use JavaConfig reference dubbo service, like this:
> 
> ```
> @Configuration
> public class DubboServiceReferenceConfig {
> 
>     @DubboReference(version = "${dubbo.application.version}", check = false)
>     private UserInfoService userInfoService;
> 
>     @Bean
>     public UserInfoService userInfoService() {
>         return userInfoService;
>     }
>     
> }
> ```
> 
> Then use @Autowired annotation on your consumer service.

Only need to use @DubboReference to declare and register the reference bean, other places can be injected with `@Autowired`.
 
```java
@Configuration
public class DubboReferenceConfig {

    @DubboReference(version = "${dubbo.application.version}")
    private UserInfoService userInfoService;

    @DubboReference(version = "${dubbo.application.version}")
    private FooService fooService;
    ...
}
```

```java
@Service
public class MyService {
     @Autowired
     private UserInfoService userInfoService;
}
```

kylixs:
This can work in Dubbo 3: https://github.com/apache/dubbo/pull/7383 
This kind of reference annotation configuration is very clear and can be recommended to everyone.

Issue #6000, opened 2020-04-12
Autowired a ReferenceBean failed in some situation in Spring enviroment.


anLA7856:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.3 ~
* Operating System version: linux & windows
* Java version: 1.8

### Steps to reproduce this issue

In daily development, I am used to keeping `Dubbo reference bean` in `Spring` container.
Like using xml config, I will write all `reference bean` in a config class, and use `@Autowire` in other class. 
Sometime it runs well, sometime it will throw `NoSuchBeanDefinitionException`.
Problem Description:
My project tree(Consumer side):
```
│  Consumer.java
├─aaconfig
│      MyPriorityReferenceConfig.java
├─adubbo
│      HelloDubbo.java
└─dubbo
       MyReferenceConfig.java
```
I have two config file. `MyReferenceConfig.java` and `MyPriorityReferenceConfig.java` have the same code.
```
@Configuration
public class XXX {
    @Reference(version = "1.0.0", check = false)
    HelloService helloService;
}
```

`HelloDubbo.java` contains an annotation field(`@Autowired` or `@Resource`).
```
@Component
public class HelloDubbo {
    HelloService helloService;
    @Resource
    public void setHelloService(HelloService helloService) {
        this.helloService = helloService;
    }
}
```
When there is only `MyReferenceConfig.java` be register in Spring Context. It will throw an Exception:
```
org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'com.xxx.rpc.configcenter.provider.service.HelloService'
```
However, When only `MyPriorityReferenceConfig.java` be register in Spring Context. It will found that bean successfully.

### Expected Result

No Exception 

### Actual Result

throw a `NoSuchBeanDefinitionException`

nickwongwong:
Same problem~

kylixs:
This bug has been fixed:  https://github.com/apache/dubbo/pull/7383

PR #7497, opened 2021-04-01
3.0 improve reference bean register

kylixs:

Improve reference bean register, make xml parsing and annotation processing register bean definition in a similar way, solving the problem of reference injection.

## Brief changelog

**`@DubboReference` annotation and `<dubbo:reference />` xml elements:**
1. The main processing logic of `@DubboReference` annotation:
* First, scan all bean definitions, find the field and method annotated by `@DubboReference`, and register the reference bean definition;
* Then, inject the reference bean to other beans by matching type and name;

2. Support  using`@DubboReference` with `@Bean` in java-config class, and support `@Conditional` on ReferenceBean
By post processing `AnnotatedBeanDefinition` which created by java-config `@Configuration` class ,  parsing `@DubboReference` with `@Bean`  and set reference props to bean definition attribute.

3. Improve reference bean props populating

4. Through the reference key, if the attributes of `@DubboReference` annotation and the `<dubbo:reference />` xml bean  are the same, they share the same bean definition.

5. The default value of `@DubboReference/@DubboService/@Method/@Argument` annotated int type is changed to -1, and the default value of String type is an empty string.
   `@DubboReference` and `@DubboService` annotations add `scope` attribute, @Method annotation adds `parameters` attribute.

**Others:**
1. Upgrade Spring version to 5.x
2. Upgrade dubbo-config-spring  unit tests to junit 5
3. Add tests for issuse #6000 and #7003 , localcall and import xml
   and https://github.com/apache/dubbo-spring-boot-project/issues/779

**TODO:** 
1. fix dubbo-spring-boot unit tests

## Verifying this change

XXXXX


dubbo-common/src/main/java/com/alibaba/dubbo/config/annotation/Reference.java
```diff
@@ -63,9 +72,9 @@
 
     String cluster() default "";
 
-    int connections() default 0;
+    int connections() default -1;
```
chickenlj, line 75: I wonder if these default value changes will have side effects , have you checked how dubbo treats the default value of each item?

kylixs: **The default value of the annotation will be filtered as null**, so the default value is best set to a value that does not need to be used. 
`-1 `  may be used by some attributes? Change the default value of the int type to `Integer.MIN_VALUE`?  But `-1 ` is shorter.

chickenlj (approved):
LGTM.

PR #7556, opened 2021-04-15
[3.0] Seperate RpcContext to client and server

AlbumenJ:

Seperate `RpcContext` to `ServerContext`, `ClientAttachment`, `ServerAttachment` and `ServiceContext`.

ServiceContext: Using to pass environment parameters in the whole invocation. For example, `remotingApplicationName`, `remoteAddress`, etc. {@link RpcServiceContext}
ClientAttachment, ServerAttachment and ServiceContext are using to transfer attachments.

Imaging a situation like this, A is calling B, and B will call C, after that, B wants to return some attachments back to A.
ClientAttachment is using to pass attachments to next hop as a consumer. ( A --> B , in A side)
ServerAttachment is using to fetch attachments from previous hop as a provider. ( A --> B , in B side)
ServerContext is using to return some attachments back to client as a provider. ( A <-- B , in B side)

Add `PenetrateAttachmentSelector` SPI if user not want attachments pass always. `ServerAttachment` will pass to next hop in default. If one or more `PenetrateAttachmentSelector` SPI has been configured, Dubbo will only as those has been selected attachment.

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/support/ConsumerContextFilter.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/support/ConsumerContextFilter.java
@@ -17,6 +17,7 @@
 package org.apache.dubbo.rpc.cluster.filter.support;
 
 import org.apache.dubbo.common.extension.Activate;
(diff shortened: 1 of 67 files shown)


chickenlj (approved):
LGTM.

PR #7572, opened 2021-04-16
Optimize code to remove useless objects

srchen1987:
Optimize code to remove useless objects

update before
public void addAttachments(Map<String, String> attachments) {
if (attachments == null) {
return;
}
this.attachments = new HashMap<String, String>();
}
this.attachments.putAll(attachments);
}

update after
public void addAttachments(Map<String, String> attachments) {
if (attachments == null) {
return;
}
this.attachments = attachments;
}else{
this.attachments.putAll(attachments);
}
}

### DIFF
--- a/dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/RpcInvocation.java
+++ b/dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/RpcInvocation.java
@@ -160,7 +160,18 @@ public class RpcInvocation implements Invocation, Serializable {
             attachments.put(key, value);
         }
     }
-
+    //Optimize code to remove useless objects  srchen1987 add
+    public void addAttachments(Map<String, String> attachments) {
+        if (attachments == null) {
+            return;
+        }
+        if (this.attachments == null) {
+            this.attachments = attachments;
+        }else{
+            this.attachments.putAll(attachments);
+        }
+    }
+    /**
     public void addAttachments(Map<String, String> attachments) {
         if (attachments == null) {
             return;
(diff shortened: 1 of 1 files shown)


dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/RpcInvocation.java
```diff
@@ -160,15 +160,16 @@ public void setAttachmentIfAbsent(String key, String value) {
             attachments.put(key, value);
         }
     }
-
+    //Optimize code to remove useless objects
```
horizonzy, line 163: The comment is unnecessary

dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/RpcInvocation.java
```diff
@@ -160,15 +160,16 @@ public void setAttachmentIfAbsent(String key, String value) {
...
     public void addAttachments(Map<String, String> attachments) {
         if (attachments == null) {
             return;
         }
         if (this.attachments == null) {
-            this.attachments = new HashMap<String, String>();
+            this.attachments = attachments;
+        }else{
```
horizonzy, line 170: indent problem.

gonwan:
Hi there,

Happened to look through the release notes, but this merge introduced side effects. It assumes the argument `attachments` is not going to be modified outside the method, which is not always correct. Should revert this.

horizonzy:
> Hi there,
> 
> Happened to look through the release notes, but this merge introduced side effects. It assumes the argument `attachments` is not going to be modified outside the method, which is not always correct. Should revert this.

yep, it's indeed risk.

PR #7582, opened 2021-04-18
Optimize code to remove useless objects

srchen1987:
Optimize code to remove useless objects

update before
public void addAttachments(Map<String, String> attachments) {
if (attachments == null) {
return;
}
this.attachments = new HashMap<String, String>();
}
this.attachments.putAll(attachments);
}

update after
public void addAttachments(Map<String, String> attachments) {
if (attachments == null) {
return;
}
this.attachments = attachments;
}else{
this.attachments.putAll(attachments);
}
}


kevinw66 (commented):
duplicate with [#7572](https://github.com/apache/dubbo/pull/7572)

srchen1987:
> duplicate with [#7572](https://github.com/apache/dubbo/pull/7572)

Deadlock? That version because I wrote comments, this version removed the comments. The previous code is meaningless and will not be submitted any more

PR #7588, opened 2021-04-19
Reduce the multi-level nesting of if

BurningCN:

Reduce the multi-level nesting of `if`


AlbumenJ:
was these changes necessary?

PR #7600, opened 2021-04-22
De-duplicate the filter returned by the getActivateExtension method

### LINKED ISSUE #7587: dubbo supports duplicate filters which are defined by user，can be optimized?
In dubbo ,it supports  duplicate filters which are defined by user，this situation is normal or can be optimized?
and then i found there is no priority concept in some dubbo configurations 。
It's my opinion above,  can you help me out with any queries?
thanks!

xiaoheng1:

De-duplicate the filter returned by the getActivateExtension method

see more detail from fix #7587 

## Verifying this change

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
@@ -42,6 +42,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


AlbumenJ:
Shall we notify users that there contains two extension have the same name by loggeer?

xiaoheng1:
> Shall we notify users that there contains two extension have the same name by loggeer?

If just print the log, the user also needs to remove it manually, and may also ignore this log.
If the system supports filtering, it will be easier for users to use.

AlbumenJ:
> > Shall we notify users that there contains two extension have the same name by loggeer?
> 
> If just print the log, the user also needs to remove it manually, and may also ignore this log.
> If the system supports filtering, it will be easier for users to use.

Filter those duplicated and log out to notify that has been filtered.

xiaoheng1:
> Filter those duplicated and log out to notify that has been filtered.

Sorry, I misunderstand. Now let’s add a log.

PR #7611, opened 2021-04-23
replace nacos StringUtils with dubbo-common StringUtils.

wuwen5:
Nacos StringUtils has version compatibility issues.


xiaoheng1 (approved):
LGTM

PR #7637, opened 2021-04-26
Support sending generic requests in JSON format

goodjava:
Invoking a dubbo service via calling from api gateway or running dynamic scripts like groovy by simply passing string values instead of using maps or updating jar files.
zzz

像网关和groovy动态脚本这样的调用,可以直接通过传递字符串完成一次rpc调用(不必自己再组装Map之类,或者更新jar包).变的更简单.(golang 夸语言的泛化调用也会大幅度降低难度,毕竟字符串更好产出出来,相较于struct)

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
@@ -16,6 +16,10 @@
  */
 package org.apache.dubbo.rpc.filter;
 
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+import com.google.gson.reflect.TypeToken;
+
 import org.apache.dubbo.common.beanutil.JavaBeanAccessor;
 import org.apache.dubbo.common.beanutil.JavaBeanDescriptor;
 import org.apache.dubbo.common.beanutil.JavaBeanSerializeUtil;
@@ -45,10 +49,13 @@ import org.apache.dubbo.rpc.support.ProtocolUtils;
 
 import java.io.IOException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.stream.IntStream;
 
 import static org.apache.dubbo.common.constants.CommonConstants.$INVOKE;
 import static org.apache.dubbo.common.constants.CommonConstants.$INVOKE_ASYNC;
 import static org.apache.dubbo.common.constants.CommonConstants.GENERIC_SERIALIZATION_BEAN;
+import static org.apache.dubbo.common.constants.CommonConstants.GENERIC_SERIALIZATION_GSON;
 import static org.apache.dubbo.common.constants.CommonConstants.GENERIC_SERIALIZATION_NATIVE_JAVA;
 import static org.apache.dubbo.common.constants.CommonConstants.GENERIC_SERIALIZATION_PROTOBUF;
 import static org.apache.dubbo.rpc.Constants.GENERIC_KEY;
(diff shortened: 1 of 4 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
```diff
@@ -90,7 +94,9 @@ public Result invoke(Invoker<?> invoker, Invocation inv) throws RpcException {
                     generic = RpcContext.getContext().getAttachment(GENERIC_KEY);
                 }
 
-                if (StringUtils.isEmpty(generic)
+                if (ProtocolUtils.isGsonGenericSerialization(generic)) {
```
horizonzy, line 97: Code clean: This case should after origin first `if` judgement.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
```diff
@@ -174,6 +180,14 @@ public Result invoke(Invoker<?> invoker, Invocation inv) throws RpcException {
         return invoker.invoke(inv);
     }
 
+    private Object[] getArgs(final Object[] args, Type[] types) {
+        Gson gson = new Gson();
```
horizonzy, line 184: transfer to instance variable.

xiaoheng1 (commented):
I think it is better to catch exceptions when converting json to the corresponding type

horizonzy:
> I think it is better to catch exceptions when converting json to the corresponding type

agree, catch JsonSyntaxException expcetion and throw RpcException.

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
kimmking: agree with horizonzy

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
```diff
@@ -174,6 +180,14 @@ public Result invoke(Invoker<?> invoker, Invocation inv) throws RpcException {
...
+        return IntStream.range(0, args.length).mapToObj(i -> {
+            String s = args[i].toString();
```
kimmking, line 186: variable s is unnecessary

kimmking (changes requested):

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
goodjava: get

goodjava:
> I think it is better to catch exceptions when converting json to the corresponding type



> I think it is better to catch exceptions when converting json to the corresponding type

After resolving
org.apache.dubbo.rpc.RpcException: Generic serialization [gson] Json syntax exception thrown when parsing (message:abc type:int) error:java.lang.NumberFormatException: For input string: "abc"

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
goodjava: get

xiaoheng1 (approved):
LGTM

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java
goodjava: get

goodjava:
> > I think it is better to catch exceptions when converting json to the corresponding type
> 
> agree, catch JsonSyntaxException expcetion and throw RpcException.

done

AlbumenJ:
Please resolve confilcts

PR #7638, opened 2021-04-26
Remove EventListener implementation in DubboBootstrap

kevinw66:

DubboBootstrap implements EventListener but there is no handle event method, maybe we should remove the implementation.


AlbumenJ (approved):
LGTM.

PR #7639, opened 2021-04-26
Replace some orElse with orElseGet

kevinw66:

According to the [document](https://www.baeldung.com/java-optional-or-else-vs-or-else-get), the statement in orElse will be executed anyway, this may cause some extra consumption, we should replace it with orElseGet.


xiaoheng1 (approved):
LGTM

dubbo-common/src/main/java/org/apache/dubbo/config/ServiceConfigBase.java
```diff
@@ -194,7 +194,7 @@ public void checkDefault() throws IllegalStateException {
         if (provider == null) {
             provider = ApplicationModel.getConfigManager()
                     .getDefaultProvider()
-                    .orElse(new ProviderConfig());
+                    .orElseGet(ProviderConfig::new);
```
chickenlj, line 197: This will put the ProviderConfig instance into ConfigManager, which is not an action expected.

chickenlj (changes requested):

PR #7647, opened 2021-04-27
Refactoring to remove envious methods (extract methods)

binho1991:
Hi everyone,

I discovered an opportunity to improve Dubbo. I observed that the AccessLogData constructor is an envious method because it calls many times the RpcContext class and it gets worse the readability of the method. Then, I extracted two methods, separating a method only to set port and set host. I expect that it improves the source code. Thanks.

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/AccessLogData.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/AccessLogData.java
@@ -65,14 +65,22 @@ public final class AccessLogData {
     /**
      * Default constructor.
      */
-    private AccessLogData() {
-        RpcContext context = RpcContext.getContext();
-        data = new HashMap<>();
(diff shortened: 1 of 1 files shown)


AlbumenJ:
Is there any benefit to doing this?

binho1991:
Hi @AlbumenJ 

Thanks for your questions!
I think this pull request improves readability because we have a structural problem called Feature Envy. In this case, the AccessLogData method calls RpcContext many times and it is dealing with different concerns: port and host. Then, this pull request reduces the number of calls and divides concerns.

chickenlj:
> Hi @AlbumenJ
> 
> Thanks for your questions!
> I think this pull request improves readability because we have a structural problem called Feature Envy. In this case, the AccessLogData method calls RpcContext many times and it is dealing with different concerns: port and host. Then, this pull request reduces the number of calls and divides concerns.

Sorry, we tend to accept only bugfixes to the 2.7 branch, if you insist on this refactoring please file pr or against 3.0 branch.

PR #7671, opened 2021-05-03
[3.0] Use String Intern to reduce memory allocation

AlbumenJ:

Use String Intern to reduce memory allocation when create `ServiceKey` for each url.

Always, `ServiceKey` is associated with interfaces, and its counts is limited. While there are a large amount of providers, the objects related with url will be created often. These urls' `ServiceKey` will always the same content.


xiaoheng1 (approved):
LGTM

dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
```diff
@@ -1298,7 +1298,7 @@ public String getPathKey() {
     }
 
     public static String buildKey(String path, String group, String version) {
-        return BaseServiceMetadata.buildServiceKey(path, group, version);
+        return BaseServiceMetadata.buildServiceKey(path, group, version).intern();
```
horizonzy, line 1301: use intern() in BaseServiceMetadata.buildServiceKey?

AlbumenJ: done

horizonzy (approved):
LGTM.

dubbo-common/src/main/java/org/apache/dubbo/common/BaseServiceMetadata.java
```diff
@@ -42,7 +42,7 @@ public static String buildServiceKey(String path, String group, String version)
         if (version != null && version.length() > 0) {
             buf.append(":").append(version);
         }
-        return buf.toString();
+        return buf.toString().intern();
```
chickenlj, line 45: I am a bit confused. Does the memory already being allocated from `toString()`?
I think doing intern this way will only possibly save the memory if the String returned will be held by a constant reference, it won't save the temporary memory allocated.

chickenlj (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/BaseServiceMetadata.java
AlbumenJ: > I think doing intern this way will only possibly save the memory if the String returned will be held by a constant reference, it won't save the temporary memory allocated.

Yes, it is useless about temporarory memory allocation. But, it can save persistence memory allocation. That is the urls from directory can hold the same String object, and temporarory String object from `toString()` can be release by GC. URLs will persistence exist if the invoker is still alive, and it will keep the reference of `ServiceKey` as a cache.

chickenlj: I thought you were improving the efficiency of the RPC call, say, RT, QPS.

Issue #7674, opened 2021-05-05
Fix the concurrency problem when AbstractCacheFactory creates Cache


xiaoheng1:
### Environment

* Dubbo version: 2.7.x
* Operating System version: macos
* Java version: 1.8

Fix the concurrency problem when AbstractCacheFactory creates Cache.

Issue #7641, opened 2021-04-27
Revisit the design and usage of URL in Dubbo 3.0


chickenlj:

pinxiong:
The concept of URL-based design runs through Dubbo. There are two main threads, the provider and consumer.

+ #### Provider

  1. **registry-url**: It provides all information related the registry center, which is built by `RegistryConfig` and the protocol begins with **registry**, such as :

     ```java
     registry://127.0.0.1:2181/org.apache.dubbo.registry.RegistryService?application=dubbo-demo-api-provider&dubbo=2.0.2&pid=20855&registry=zookeeper&timestamp=1619492124796
     ```

     And then, **registry-url** will be enhanced if the proxy has been customized by users. You can see [ServicesConfig#doExportUrlsFor1Protocol](https://github.com/apache/dubbo/blob/3.0/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java)

     ```java
     // For providers, this is used to enable custom proxy to generate invoker
     String proxy = url.getParameter(PROXY_KEY);
     if (StringUtils.isNotEmpty(proxy)) {
     	registryURL = registryURL.addParameter(PROXY_KEY, proxy);
     }
     ```

     Finally, there are some changes in **registry-url** as follows:

     + **registry-url**'s protocol will be updated from **registry** to **zookeeper** (if you use zookeeper as the registry center)

     + **registry-url** will add the parameter `export` and its value is **provider-url**(see below)

       ```java
[...]

PR #7675, opened 2021-05-05
Fix the concurrency problem when AbstractCacheFactory creates Cache

xiaoheng1:

Fix the concurrency problem when AbstractCacheFactory creates Cache

see more detail from #7674 

## Verifying this change

### DIFF
--- a/dubbo-filter/dubbo-filter-cache/src/main/java/org/apache/dubbo/cache/support/AbstractCacheFactory.java
+++ b/dubbo-filter/dubbo-filter-cache/src/main/java/org/apache/dubbo/cache/support/AbstractCacheFactory.java
@@ -23,6 +23,7 @@ import org.apache.dubbo.rpc.Invocation;
 
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.locks.ReentrantLock;
 
 import static org.apache.dubbo.common.constants.CommonConstants.METHOD_KEY;
 
@@ -42,11 +43,14 @@ public abstract class AbstractCacheFactory implements CacheFactory {
     /**
      * This is used to store factory level-1 cached data.
      */
-    private final ConcurrentMap<String, Cache> caches = new ConcurrentHashMap<String, Cache>();
+    private final static ConcurrentMap<String, Cache> caches = new ConcurrentHashMap<String, Cache>();
+
+    private final static ReentrantLock LOCK = new ReentrantLock();
 
     /**
-     *  Takes URL and invocation instance and return cache instance for a given url.
-     * @param url url of the method
+     * Takes URL and invocation instance and return cache instance for a given url.
+     *
+     * @param url        url of the method
      * @param invocation invocation context.
      * @return Instance of cache store used as storage for caching return values.
      */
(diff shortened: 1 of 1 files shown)


dubbo-filter/dubbo-filter-cache/src/main/java/org/apache/dubbo/cache/support/AbstractCacheFactory.java
```diff
@@ -42,11 +43,14 @@
     /**
      * This is used to store factory level-1 cached data.
      */
-    private final ConcurrentMap<String, Cache> caches = new ConcurrentHashMap<String, Cache>();
+    private final static ConcurrentMap<String, Cache> caches = new ConcurrentHashMap<String, Cache>();
```
horizonzy, line 46: 没必要用static,不同种类的Cache用自己的caches

xiaoheng1: In my opinion, AbstractCacheFactory.caches should store jcache, lfu, lru cache, etc. Therefore, I changed caches to static, and let caches be shared by all other cache implementation classes. such as AbstractRegsitryFactory.

horizonzy: I use both lra and lfu, if it share cache, how did eliminate the data.

xiaoheng1: done

dubbo-filter/dubbo-filter-cache/src/main/java/org/apache/dubbo/cache/support/AbstractCacheFactory.java
```diff
@@ -23,6 +23,7 @@
 
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.locks.ReentrantLock;
```
horizonzy, line 26: unless import.

horizonzy (approved):
LGTM.

dubbo-filter/dubbo-filter-cache/src/main/java/org/apache/dubbo/cache/support/AbstractCacheFactory.java
```diff
@@ -55,15 +58,29 @@ public Cache getCache(URL url, Invocation invocation) {
         url = url.addParameter(METHOD_KEY, invocation.getMethodName());
```
pinxiong, line 58: You'd better not overwrite the value of `url`

xiaoheng1: I did not modify this piece of code. Can you describe your opinion in detail?

pinxiong: The `url` 's value is changed
```java
public Cache getCache(URL url, Invocation invocation) {
    url = url.addParameter(METHOD_KEY, invocation.getMethodName());
    String key = url.toFullString();
    Cache cache = caches.get(key);
}
```

The `url` 's value is unchanged
```java
public Cache getCache(URL url, Invocation invocation) {
    String key = url.addParameter(METHOD_KEY, invocation.getMethodName()).toFullString();
    Cache cache = caches.get(key);
}
```

I know you don't modify this piece of code, but I think it's unreasonable to change the value of an input parameter if the caller don't know that.

Also, I know that if you do it the way I said, you might need to change a lot of code (All the places where `METHOD_KEY` is used may need to be modified)

xiaoheng1: The URL is designed to be immutable. After adding parameters, if the url is not accepted, the new url cannot be obtained. If the METHOD_KEY parameter is added elsewhere, this parameter may not be needed in that place. Do you have any good suggestions?

pinxiong: This is indeed a complicated problem. You will change a lot of code if you want to solve this problem completely.

You can refer #7641

For this issue, your solution may be the simplest and least risky

xiaoheng1: Good suggestion. 

I think we should fix this problem first. The problem of url reconstruction will be dealt with later. What do you think so?

pinxiong: Yes, you're right!

AlbumenJ (approved):

PR #7702, opened 2021-05-08
Optimize turnRegistryUrlToConsumerUrl method

BurningCN:
In `turnRegistryUrlToConsumerUrl` method, the URL generation logic is basically the same as that of the superclass `AbstractDirectory#consumerUrl` attribute, so we can reuse the superclass's `consumerUrl`, and just modify the host and port


xiaoheng1 (commented):
URL is designed to be immutable, so here is a need to generate a new URL, if consumerUrl.setHost(host).setPort(0) is used directly, it will affect the place where consumerUrl is used.

PR #7705, opened 2021-05-09
Revert "Optimize code to remove useless objects (#7582)"

horizonzy:

This reverts commit d6252081

When RpcInvocation's attachemtns is null, it will share attachments.
It's indeed risk.
Thanks for your remind. @gonwan


xiaoheng1 (approved):
LGTM

Issue #7708, opened 2021-05-09
[3.0 service-discovery] Nacos instance healthy status is unhealthy when register to nacos.


horizonzy:
### Environment

* Dubbo version: 3.0

### Steps to reproduce this issue

1. modify registry address to nacos.
2. see nacos console, the instance healthly status is false.
![image](https://user-images.githubusercontent.com/22524871/117575073-d40bc180-b112-11eb-9b5b-19e99df3d8c3.png)

DefaultServiceInstance's property `healthy` and `enabled` default value is false.

https://github.com/apache/dubbo/blob/ca0b9828667dc981986b1235bc4e032f8cd46171/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/DefaultServiceInstance.java#L49-L51

https://github.com/apache/dubbo/blob/ca0b9828667dc981986b1235bc4e032f8cd46171/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/DefaultServiceInstance.java#L90-L92

https://github.com/apache/dubbo/blob/ca0b9828667dc981986b1235bc4e032f8cd46171/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java#L1225-L1230

PR #7709, opened 2021-05-09
[ISSUE-#7708] DefaultServiceInstance property enabled and healthy default value to true

horizonzy:
For-#7708

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/DefaultServiceInstance.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/DefaultServiceInstance.java
@@ -46,9 +46,9 @@ public class DefaultServiceInstance implements ServiceInstance {
 
     private int port;
 
-    private boolean enabled;
+    private boolean enabled = true;
 
-    private boolean healthy;
+    private boolean healthy = true;
 
     private Map<String, String> metadata = new UnifiedMap<>();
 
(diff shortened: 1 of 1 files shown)


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/DefaultServiceInstance.java
```diff
@@ -77,14 +77,12 @@ public DefaultServiceInstance(DefaultServiceInstance other) {
     }
 
     public DefaultServiceInstance(String serviceName, String host, Integer port) {
-        if (port != null && port.intValue() < 1) {
+        if (port != null && port < 1) {
             throw new IllegalArgumentException("The port must be greater than zero!");
         }
         this.serviceName = serviceName;
         this.host = host;
         this.port = port;
```
icankeep, line 85: (this.port = port) it will cause a npe when port == null, I'm not sure it's ok
why not add a null check
```
if (port == null) {
    throw new IllegalArgumentException("The port should not be null!");
}
if (port < 1) {
    throw new IllegalArgumentException("The port must be greater than zero!");
}
```

icankeep (commented):
(this.port = port) it will cause a npe when port == null, I'm not sure it's ok
why not add a null check
```
if (port == null) {
    throw new IllegalArgumentException("The port should not be null!");
}
if (port < 1) {
    throw new IllegalArgumentException("The port must be greater than zero!");
}
```

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/DefaultServiceInstance.java
horizonzy: yep, add null checker will be better.

PR #7711, opened 2021-05-10
Fix getAddress will cause npe problem.

horizonzy:

```
    @Override
    public String getAddress() {
        if (address == null) {
            address = getAddress(host, port);
        }
        return address;
    }

    private static String getAddress(String host, int port) {
        return port <= 0 ? host : host + ':' + port;
    }
```

when the port is null, the Integer unbox will cause npe.

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/DefaultServiceInstance.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/DefaultServiceInstance.java
@@ -115,8 +115,8 @@ public class DefaultServiceInstance implements ServiceInstance {
         return address;
     }
 
-    private static String getAddress(String host, int port) {
-        return port <= 0 ? host : host + ':' + port;
+    private static String getAddress(String host, Integer port) {
+        return port != null && port <= 0 ? host : host + ':' + port;
     }
 
     @Override


BurningCN:
And other places where caller might assign a value of type Integer to port

horizonzy:
> I looked at the two calls to the getAddress method, and the type it passed in was `int`, not `Integer`. If you are concerned about the NPE of port, I think it should be modified here(`this.port = Math.max(port, 0)` also cause NPE if port is Integer type and its values is null ):
> 
> ```
> public URL(String protocol,
>                String username,
>                String password,
>                String host,
>                int port, // here
>                String path,
>                Map<String, String> parameters,
>                Map<String, Map<String, String>> methodParameters) {
>         this.protocol = protocol;
>         this.username = username;
>         this.password = password;
>         this.host = host;
>         this.port = Math.max(port, 0); // here
>         this.address = getAddress(this.host, this.port);
> ........
> ```

Hi, this place  invoke `getAddress` isn't connected with DefaultServiceInstance. In DefaultServiceInstance, the property port type is wrap type `Integer`, at this class `URL`, the property port type is basic type.

BurningCN:
> > I looked at the two calls to the getAddress method, and the type it passed in was `int`, not `Integer`. If you are concerned about the NPE of port, I think it should be modified here(`this.port = Math.max(port, 0)` also cause NPE if port is Integer type and its values is null ):
> > ```
> > public URL(String protocol,
> >                String username,
> >                String password,
> >                String host,
> >                int port, // here
> >                String path,
> >                Map<String, String> parameters,
> >                Map<String, Map<String, String>> methodParameters) {
> >         this.protocol = protocol;
> >         this.username = username;
> >         this.password = password;
> >         this.host = host;
> >         this.port = Math.max(port, 0); // here
> >         this.address = getAddress(this.host, this.port);
> > ........
> > ```
> 
> Hi, this place invoke `getAddress` isn't connected with DefaultServiceInstance. In DefaultServiceInstance, the property port type is wrap type `Integer`, at this class `URL`, the property port type is basic type.

I'm sorry, I looked in the wrong place

horizonzy:
> I'm sorry, I looked in the wrong place

No biggie

Issue #7722, opened 2021-05-10
ServiceNameMapping didn't work when registry-type=service.


horizonzy:
### Environment

* Dubbo version: master

What actually happens?
https://github.com/apache/dubbo/blob/2ca0099c9029921507219e53ad907850dcc180c8/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java#L214-L227

line_219 judgement exist problem. It's communication protocol, not registry protocol.

PR #7723, opened 2021-05-10
[ISSUE-#7722] Fix serviceNameMapping didn't work problem.

horizonzy:
For-#7722

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
@@ -28,6 +28,7 @@ import org.apache.dubbo.common.utils.CollectionUtils;
 import org.apache.dubbo.common.utils.ConfigUtils;
 import org.apache.dubbo.common.utils.NamedThreadFactory;
 import org.apache.dubbo.common.utils.StringUtils;
+import org.apache.dubbo.common.utils.UrlUtils;
 import org.apache.dubbo.config.annotation.Service;
 import org.apache.dubbo.config.bootstrap.DubboBootstrap;
 import org.apache.dubbo.config.event.ServiceConfigExportedEvent;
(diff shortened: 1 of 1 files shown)


xiaoheng1 (commented):
if (SERVICE_REGISTRY_PROTOCOL.equals(registryURL.getProtocol())) {
                            url = url.addParameterIfAbsent(REGISTRY_TYPE_KEY, SERVICE_REGISTRY_TYPE);
                        }


Why do we need this paragraph of judgment? In the org.apache.dubbo.config.utils.ConfigValidationUtils#loadRegistries method, it has been converted to registryProtcol based on the registry-type parameter.

horizonzy:
The url is different. 
In method exported.
```
        List<URL> exportedURLs = this.getExportedUrls();
        exportedURLs.forEach(url -> {
            // dubbo2.7.x does not register serviceNameMapping information with the registry by default.
            // Only when the user manually turns on the service introspection, can he register with the registration center.
            boolean isServiceDiscovery = SERVICE_REGISTRY_PROTOCOL.equals(url.getProtocol());
            if (isServiceDiscovery) {
                Map<String, String> parameters = getApplication().getParameters();
                ServiceNameMapping.getExtension(parameters != null ? parameters.get(MAPPING_KEY) : null).map(url);
            }
        });
```
getExportedUrls return the url which didn't hold registry param. 

see:
https://github.com/apache/dubbo/blob/18c5b4cd78bd988fb24c7e570418e542cf45d6bd/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java#L348

xiaoheng1:
> The url is different.
> In method exported.
> 
> ```
>         List<URL> exportedURLs = this.getExportedUrls();
>         exportedURLs.forEach(url -> {
>             // dubbo2.7.x does not register serviceNameMapping information with the registry by default.
>             // Only when the user manually turns on the service introspection, can he register with the registration center.
>             boolean isServiceDiscovery = SERVICE_REGISTRY_PROTOCOL.equals(url.getProtocol());
>             if (isServiceDiscovery) {
>                 Map<String, String> parameters = getApplication().getParameters();
>                 ServiceNameMapping.getExtension(parameters != null ? parameters.get(MAPPING_KEY) : null).map(url);
>             }
>         });
> ```
> 
> getExportedUrls return the url which didn't hold registry param.
> 
> see:
> https://github.com/apache/dubbo/blob/18c5b4cd78bd988fb24c7e570418e542cf45d6bd/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java#L348

i get it. you are right.

xiaoheng1 (approved):
LGTM

AlbumenJ (approved):

PR #7730, opened 2021-05-11
Revert "Fix consumer side call DefaultExecutorRepository#getExecutor method cannot hit the cache when use the service discovery pattern "

BurningCN:
I'm sorry, There was a problem with #7717 PR before. There are some gaps in my understanding.

In consumer side,`DefaultExecutorRepository#getExecutor` in acquiring `side` parameter value is `null` because not RPC calls, but some of the other events, such as `channelActive` (use SHARED_EXECUTOR).

If an RPC call is made, `InvokerInvocationHandler#invoke` calls `RpcContext.setRpcContext(invoker.getUrl());` So `DefaultExecutorRepository#getExecutor` ->`InstanceAddressURL#getParameter `will get the `side` parameter and its value is `consumer`

So the cached Executor can only be used if the consumer initiates the RPC call, and SHARED_EXECUTOR is used for other network events

// ====================================================================
Here is the screenshot

In consumer side, `channelActive` event use SHARED_EXECUTOR:
image

In consumer side, RPC call use previous cached executor:
image


BurningCN:
@AlbumenJ I am sorry for the previous PR submission, Please help to merge this revert PR In your spare time

@horizonzy thank you, Because of your reminder, this made me recheck it

PR #7750, opened 2021-05-13
[3.0] Performance Optimization

AlbumenJ:
- Reduce case convert
- Reduce iterator creation
- Add attribute to MethodDescriptor


chickenlj:
Does LInkedHashMap has a significant effect in this scenario? That is, the cost saved from iteration is more worth than the pressure of storage allocation.

chickenlj:
LGTM.

PR #7752, opened 2021-05-13
fix_early_postprocessor_bug

owen200008:
fix bug 
early post processor must not Registry beanDefinition
BeanPostProcessorChecker check error, because it think the beanprocess size is less than create. so must have some beanprocess not deal with the bean
see spring error:
same 
https://github.com/apache/dubbo-spring-boot-project/issues/786

## Verifying this change

### DIFF
--- a/dubbo-config/dubbo-config-spring/src/test/java/org/apache/dubbo/config/spring/issues/Issue7752Test.java
+++ b/dubbo-config/dubbo-config-spring/src/test/java/org/apache/dubbo/config/spring/issues/Issue7752Test.java
@@ -16,21 +16,20 @@
  */
 package org.apache.dubbo.config.spring.issues;
 
-import org.apache.dubbo.config.spring.context.annotation.EnableDubboConfig;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+import org.mockito.MockedConstruction;
 import org.mockito.Mockito;
-import org.mockito.invocation.InvocationOnMock;
-import org.mockito.stubbing.Answer;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
-import org.springframework.beans.factory.support.AbstractBeanFactory;
 import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.context.annotation.*;
-import org.springframework.context.support.AbstractApplicationContext;
 import org.springframework.lang.Nullable;
 import org.springframework.test.annotation.DirtiesContext;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * The test-case for https://github.com/apache/dubbo/issues/7752
@@ -38,18 +37,19 @@ import org.springframework.test.annotation.DirtiesContext;
  * @since 2.7.8
  */
 @Configuration
-@EnableDubboConfig
-@PropertySource("classpath:/META-INF/issue-7752-test.properties")
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
 public class Issue7752Test {
 
+    /**
+     * code is same with org.springframework.context.support.PostProcessorRegistrationDelegate$BeanPostProcessorChecker
+     */
     private static final class Issue7752TestBeanPostProcessorChecker implements BeanPostProcessor{
 
         private ConfigurableListableBeanFactory beanFactory;
 
         private int beanPostProcessorTargetCount;
 
-        public void setParam(ConfigurableListableBeanFactory beanFactory, int beanPostProcessorTargetCount){
+        public Issue7752TestBeanPostProcessorChecker(ConfigurableListableBeanFactory beanFactory, int beanPostProcessorTargetCount){
             this.beanFactory = beanFactory;
             this.beanPostProcessorTargetCount = beanPostProcessorTargetCount;
         }
@@ -62,9 +62,9 @@ public class Issue7752Test {
         @Override
         public Object postProcessAfterInitialization(Object bean, String beanName) {
             if (!(bean instanceof BeanPostProcessor) && !isInfrastructureBean(beanName) &&
-                    //need + 1 self
-                    this.beanFactory.getBeanPostProcessorCount() < this.beanPostProcessorTargetCount + 1) {
-                Assertions.assertFalse(true);
+                    this.beanFactory.getBeanPostProcessorCount() < this.beanPostProcessorTargetCount) {
+                //means may be some BeanPostProcessor not deal with bean
+                Assertions.assertTrue(false);
             }
             return bean;
         }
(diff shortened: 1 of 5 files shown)


AlbumenJ:
LGTM. Would you please add some unit test cases for this?

zhongxiongzeng:
why not use BeanFactoryAware ？

owen200008:
DubboEarlyApplicationContext implements BeanFactoryAware or DubboConfigEarlyInitializationPostProcessor  implements BeanFactoryAware？

owen200008:
DubboConfigEarlyInitializationPostProcessor need early add in BeanPostProcessor.
i think in the step BeanDefinitionRegistry is good way.

owen200008:
it is like other early PostProcessor. see ImportAwareBeanPostProcessor

kylixs:
@owen200008 This test case is not a normal user scenario, and it seems difficult to understand.
From the description, it seems that AOP is unavailable, please add a test case for specific scenario.

kylixs:
I don't understand why the registration of `DubboConfigEarlyInitializationPostProcessor` should be adjusted in this pr.  For what?

owen200008:
it means maybe aop is not available。
because beanprocess not total process for the bean。
this is beanprocesscheck meaning of existence

owen200008:
is my fault. english is not so good. make puzzle.
BeanPostProcessorChecker check error, because it think the beanprocess size is less than create. so must have some beanprocess not deal with the bean. (so maybe aop beanprocess is not deal with bean, so it have the infomation ( is not eligible for getting processed by all BeanPostProcessors))

dubbo-config/dubbo-config-spring/src/test/java/org/apache/dubbo/config/spring/issues/Issue7752Test.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * The test-case for https://github.com/apache/dubbo/issues/7752
+ *
+ * @since 2.7.8
+ */
+@Configuration
+@EnableDubboConfig
+@PropertySource("classpath:/META-INF/issue-7752-test.properties")
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+public class Issue7752Test {
```
kylixs, line 44: Can you provide a test for a general application scenario?

owen200008: ok, i will try

kylixs: This will make it clearer what problem is solved, and it will be easier to understand.

kylixs:
If you just want to register DubboConfigEarlyInitializationPostProcessor earlier, you only need to modify the place where it is registered. It is not recommended to add a wrapper, which makes the registration logic difficult to understand. For example, register it in BeanDefinitionRegistryPostProcessor#postProcessBeanDefinitionRegistry.

kylixs:
`DubboConfigEarlyInitializationPostProcessor` is used to solve the problem that `AbstractConfig.addIntoConfigManager()` post construct does not take effect, because the config bean may be initialized before `AutowiredAnnotationBeanPostProcessor` is loaded, and `@PostConstruct` annotation cannot be processed at this time.
```
    @PostConstruct
    public void addIntoConfigManager() {
        ApplicationModel.getConfigManager().addConfig(this);
    }
```

The processing of DubboConfigEarlyInitializationPostProcessor is not elegant enough.
In Dubbo3, this premature initialization problem has been solved with another solution, so it has been deleted.

chickenlj:
Please follow @kylixs 's instructions to further enhance this patch.

owen200008:
@kylixs @chickenlj
Thank you for telling me the cause and effect.
later i will see dubbo3 how to solve.

DubboConfigEarlyInitializationPostProcessor must not be implements BeanDefinitionRegistryPostProcessor#postProcessBeanDefinitionRegistry.

owen200008:
spring code is not elegant enough.
BeanPostProcessorChecker is the process for check all beanprocessor is deal with beans.
it calc the BeanPostProcessor size code:
		String[] postProcessorNames = beanFactory.getBeanNamesForType(BeanPostProcessor.class, true, false);

		// Register BeanPostProcessorChecker that logs an info message when
		// a bean is created during BeanPostProcessor instantiation, i.e. when
		// a bean is not eligible for getting processed by all BeanPostProcessors.
		int beanProcessorTargetCount = beanFactory.getBeanPostProcessorCount() + 1 + postProcessorNames.length; 


the DubboConfigEarlyInitializationPostProcessor  in the getBeanPostProcessorCount and in the postProcessorNames.

so the size  > real BeanPostProcessor size

kylixs:
I saw this error elsewhere. 
```
 xxxx is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
```
It should be that new beans were created and registered during the execution of registerBeanPostProcessors(). These new beans cannot be processed by all BeanPostProcessors because there are still some BeanPostProcessors that are not registered. BeanPostProcessorChecker is used to detect this problem.

So, I think it can be broken down into several things:
1. Provide a general test case to reproduce this problem
2. Fix it, that is, find out the beanPostProcessor causing the problem, and avoid creating new beans when it is initialized.

owen200008:
@kylixs you are right.
so i think it is need to create DubboEarlyApplicationContext for add any early BeanPostProcessor.
is it ok i add some test for reproduct this problem and to check? 
if you agree, later i will add

kylixs:
@owen200008  I tested this pr locally. First, I removed the code related to `DubboEarlyApplicationContext` and registered `DubboConfigEarlyInitializationPostProcessor` in the original way, and then ran `Issue7752Test`. The result was successful but did not fail, which means that the `xxx is not eligible for getting processed by all BeanPostProcessors` problem was not triggered. 

1. To find the cause of the problem, we need a test case that can reproduce the problem.
2. The name of `DubboEarlyApplicationContext` may be ambiguous, it is a `PostProcessor`, not an `ApplicationContext`.

kylixs:
This problem was reproduced in `MultipleServicesWithMethodConfigsTest`. 
The reason for the `BeanPostProcessorChecker` alarm is that `DubboConfigEarlyInitializationPostProcessor` is calculated twice, but in fact the same bean registered twice will only keep one, so the `BeanPostProcessorCount` is one less than expected.

Two registered locations:
1. Because `DubboConfigEarlyInitializationPostProcessor` implements `BeanDefinitionRegistryPostProcessor`, register itself directly in `postProcessBeanDefinitionRegistry()` or `postProcessBeanFactory()`
```java
  private void initBeanFactory() {
      if (beanFactory != null) {
          // Register itself
          ...
          beanFactory.addBeanPostProcessor(this);
      }
  }
```
2. Because `DubboConfigEarlyInitializationPostProcessor` also implements the `BeanPostProcessor` interface and is registered to beanFactory, Spring will register it again as a normal `BeanPostProcessor`.

```java
	//org.springframework.context.support.PostProcessorRegistrationDelegate
	public static void registerBeanPostProcessors(
			ConfigurableListableBeanFactory beanFactory, AbstractApplicationContext applicationContext) {

		String[] postProcessorNames = beanFactory.getBeanNamesForType(BeanPostProcessor.class, true, false);

		// Register BeanPostProcessorChecker that logs an info message when
		// a bean is created during BeanPostProcessor instantiation, i.e. when
		// a bean is not eligible for getting processed by all BeanPostProcessors.
		int beanProcessorTargetCount = beanFactory.getBeanPostProcessorCount() + 1 + postProcessorNames.length;
		beanFactory.addBeanPostProcessor(new BeanPostProcessorChecker(beanFactory, beanProcessorTargetCount));
		....
	}
```

The same bean registered twice will only keep one: 
```java
        // org.springframework.beans.factory.support.AbstractBeanFactory#addBeanPostProcessor
	public void addBeanPostProcessor(BeanPostProcessor beanPostProcessor) {
		Assert.notNull(beanPostProcessor, "BeanPostProcessor must not be null");
		// Remove from old position, if any
		this.beanPostProcessors.remove(beanPostProcessor);
		// Track whether it is instantiation/destruction aware
		if (beanPostProcessor instanceof InstantiationAwareBeanPostProcessor) {
			this.hasInstantiationAwareBeanPostProcessors = true;
		}
		if (beanPostProcessor instanceof DestructionAwareBeanPostProcessor) {
			this.hasDestructionAwareBeanPostProcessors = true;
		}
		// Add to end of list
		this.beanPostProcessors.add(beanPostProcessor);
	}
```

**Solution:**
Ensure that `DubboConfigEarlyInitializationPostProcessor` can only be registered once. 
It is designed to be used before `CommonAnnotationBeanPostProcessor` is loaded, it needs to be registered to `beanFactory.beanPostProcessors` as early as possible. 

So the solution is keep the first registration, but remove the second registration, that is not to register it as a common bean to `beanFactory` to avoid repeated registration errors.

The improvement of this PR is generally right, but the name of `DubboEarlyApplicationContext` needs to be changed, such as `DubboEarlyRegistrationPostProcessor`.

Also add a use case to reproduce the problem, it is estimated that a common dubbo application example is enough.

owen200008:
#9397 fix it

PR #7759, opened 2021-05-13
[3.0] Support caching of dynamic property

chickenlj:
Support caching of dynamic properties to avoid unnecessary remote interaction cost.


chickenlj:
see https://github.com/apache/dubbo/pull/7760

PR #7764, opened 2021-05-14
Optimize the DubboBeanUtils#registerCommonBeans method to avoid multiple registrations

BurningCN:
issue #7765


icankeep (approved):

AlbumenJ:
Please fix unit test first

zhongxiongzeng:
I think it's not nessary to do this, registerInfrastructureBean() will check weather the bean has bean registered

BurningCN:
> I think it's not nessary to do this, registerInfrastructureBean() will check weather the bean has bean registered

This modification itself is problematic. Registered controls the class level, not the object level (so if two different classes call this method, such as provider and consumer, the latter cannot be registered correctly. From the unit test above This phenomenon can be seen), and I will consider whether to change it later.

In addition, perhaps as you said, it is not very necessary. I temporarily closed

PR #7774, opened 2021-05-17
Optimize the AbstractConfig#appendAnnotation method

BurningCN:

issue #7775


AlbumenJ (approved):
LGTM. This change will help filter those method implement from `Annotation` which is `Method`'s super class.

PR #7779, opened 2021-05-17
[3.0]Fix security problem

kevinw66:

Fix security problem


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/configurator/parser/model/ConfigItem.java
```diff
@@ -37,6 +38,50 @@
...
+        configItem.setType((String) map.get("type"));
+
+        Object enabled = map.get("enabled");
+        if (enabled != null) {
+            configItem.setEnabled(Boolean.parseBoolean(enabled.toString()));
+        }
+
+        Object addresses = map.get("addresses");
+        if (addresses != null) {
+            configItem.setAddresses(((List<Object>) addresses).stream()
```
AlbumenJ, line 53: it would be better to check if addresses is align from `List<Object>`

kevinw66: done

AlbumenJ (approved):

PR #7786, opened 2021-05-18
Make method and methodName first set.

horizonzy:
```
    private boolean needWrap() {
        if (CommonConstants.$INVOKE.equals(methodName) || CommonConstants.$INVOKE_ASYNC.equals(methodName)) {
            return true;
        } else if ($ECHO.equals(methodName)) {
            return true;
        } 
        ...
        ...
    }
```
method need methodName to judge, but methodName initial place after this judgement.


xiaoheng1 (approved):
LGTM

AlbumenJ (approved):

PR #7812, opened 2021-05-20
Replace the toString() method with a public parameter

jony-one:
Replace the toString() method with a public parameter to improve readability


BurningCN (approved):
LGTM

PR #7819, opened 2021-05-21
Add tag variables to avoid repeated calls to ExporterChangeableWrapper#unexport method

BurningCN:
When the dubbo program is destroyed, for example, start from spring's` context.stop()`. `DubboBootstrap#destroy` will be called, and the `ExporterChangeableWrapper#unexport` method will be called twice in the logic.
- The first is in `destroyProtocols()->protocol.destroy()->Registry#destroy()->exporter.unexport()->ExporterChangeableWrapper#unexport`
- the second place is in `unexportServices()->sc.unexport()->exporter.unexport()->DestroyableExporter#unexport->ExporterChangeableWrapper#unexport`

So, I add this `unexported` variable to avoid repeated `unexport`

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java
@@ -59,6 +59,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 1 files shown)


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryProtocol.java
```diff
@@ -795,6 +797,10 @@ public void setExporter(Exporter<T> exporter) {
 
         @Override
         public void unexport() {
+            if (unexported) {
```
AlbumenJ, line 800: It would be better to use `AtomicBoolean`, and use cas to set ture here to prevent concurrency problem

BurningCN: Thanks, according to your suggestion, it has been adjusted to AtomicBoolean type

AlbumenJ (approved):

PR #7824, opened 2021-05-22
[3.0] Service Name Mapping Refactor And Nacos model some support.

horizonzy:
Major changes:
1.Nacos model put in order Zookeeper model.
2.Service Name Mapping just support by metadata.
3.Some bug fix.


xiaoheng1 (commented):
getServiceAppMapping and getCasServiceAppMapping move to ServiceNameMapping may be better.

AlbumenJ (approved):
LGTM.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataServiceNameMapping.java
```diff
@@ -17,62 +17,109 @@
...
+        execute(() -> {
+            String serviceInterface = url.getServiceInterface();
+            if (IGNORED_SERVICE_INTERFACES.contains(serviceInterface)) {
+                return;
+            }
+            String registryCluster = getRegistryCluster(url);
+            MetadataReport metadataReport = MetadataReportInstance.getMetadataReport(registryCluster);
+            if (metadataReport.isSupportCas() && ServiceNameMappingHandler.isBothMapping()) {
+                doCasMap(metadataReport, url);
+                doNormalMap(metadataReport, url);
```
chickenlj, line 60: Why do we need two different types of mapping method?

chickenlj (changes requested):

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataServiceNameMapping.java
chickenlj: This is quite confusing on a concept or interface level. Will one single mapping method for all extensions to implement based on their own characteristics work?

chickenlj (approved):
LGTM.

PR #7841, opened 2021-05-23
Fix the problem that the AbstractInterfaceConfig#setRegistry method does not fill the RegistryConfig to the ConfigManager

BurningCN:
When I use the` dubbo-samples-api` test program of the `dubbo-samples` project, the provider manually calls `ServiceConfig#setRegistry`. This method does not fill the `RegistryConfig `into the `ConfigManager`

This will cause some problems, such as the inability to rely on `RegistryConfig` to build the corresponding `ConfigCenterConfig `when calling the `DubboBootstap#startConfigCenter` method


AlbumenJ:
Please check why tests are failure

PR #7852, opened 2021-05-24
Move URL usage to ServiceConfigURL & Make Monitor Work

AlbumenJ:
- Move URL usage to ServiceConfigURL
- Make MonitorFilter work on Consumer Side
- Fix attributes pass in URL


chickenlj (approved):
LGTM.

PR #7870, opened 2021-05-26
Remove the forced conversion in the ServiceInstanceMetadataCustomizer#customize method

BurningCN:

Remove the forced conversion in the ServiceInstanceMetadataCustomizer#customize method


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/ServiceInstanceMetadataCustomizer.java
```diff
@@ -40,13 +39,11 @@
 
     @Override
     public void customize(ServiceInstance serviceInstance) {
-        Map<String, String> params = new HashMap<>();
```
BurningCN, line 43: This variable is not used and can be deleted

PR #7903, opened 2021-05-28
Fix some bugs in publishMetadata command

BurningCN:
1.Fix the bug that publishMetadata cmd does not work
2.Delay scheduling should be used instead of scheduled scheduling
3.The delay unit should be seconds, not milliseconds

### DIFF
--- a/dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/command/DefaultCommandExecutor.java
+++ b/dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/command/DefaultCommandExecutor.java
@@ -17,13 +17,17 @@
 package org.apache.dubbo.qos.command;
 
 import org.apache.dubbo.common.extension.ExtensionLoader;
+import org.apache.dubbo.common.utils.StringUtils;
+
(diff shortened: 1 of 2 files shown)


dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/command/DefaultCommandExecutor.java
```diff
@@ -17,13 +17,17 @@
...
+
+import static org.apache.dubbo.common.constants.CommonConstants.PROPERTIES_CHAR_SEPARATOR;
 
 public class DefaultCommandExecutor implements CommandExecutor {
     @Override
     public String execute(CommandContext commandContext) throws NoSuchCommandException {
         BaseCommand command = null;
         try {
-            command = ExtensionLoader.getExtensionLoader(BaseCommand.class).getExtension(commandContext.getCommandName());
+            String extName = StringUtils.camelToSplitName(commandContext.getCommandName(), PROPERTIES_CHAR_SEPARATOR);
```
BurningCN, line 29: The split here is based on the camel case because the extension of PublishMetadata in the spi file is like this
publish-metadata=org.apache.dubbo.qos.command.impl.PublishMetadata

PR #7919, opened 2021-05-30
Cache the parsed result of the HelpCommand and HelpTelnetHandler

BurningCN:

Cache the parsed result of the HelpCommand and HelpTelnetHandler to avoid repeated parsing and processing


AlbumenJ:
Would it be better to change to WeakReference?

BurningCN:
> Would it be better to change to WeakReference?

Thank you for your suggestion. The help command is less important than others. I have modified the map type to WeakHashMap

AlbumenJ (approved):

Issue #3531, opened 2019-02-21
use autobox in ReferenceConfig and ServiceConfig


beiwei30:
for example, if we use Boolean, then it has three values: null, true and false. We should avoid of it by using boolean.

RainYuY:
I think this depends on the field's practical utility, and it works fine for now.

PR #7931, opened 2021-05-31
[3.0] Add support for async export/refer

kevinw66:

Add support for async export/refer


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java
```diff
@@ -1092,15 +1100,17 @@ private void unexportMetadataService() {
     private void exportServices() {
         configManager.getServices().forEach(sc -> {
             // TODO, compatible with ServiceConfig.export()
-            ServiceConfig serviceConfig = (ServiceConfig) sc;
+            ServiceConfig<?> serviceConfig = (ServiceConfig<?>) sc;
             serviceConfig.setBootstrap(this);
 
-            if (exportAsync) {
+            ApplicationConfig config = getApplication();
+            if (config.getExportAsync() != null && config.getExportAsync()) {
```
pinxiong, line 1107: Why not design the `exportServices` to be asynchronous instead of relying on `export-async` parameter?

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java
```diff
@@ -1131,15 +1141,14 @@ private void referServices() {
...
             referenceConfig.setBootstrap(this);
 
             if (rc.shouldInit()) {
-                if (referAsync) {
-                    CompletableFuture<Object> future = ScheduledCompletableFuture.submit(
-                            executorRepository.getServiceExporterExecutor(),
-                            () -> cache.get(rc)
-                    );
+                ApplicationConfig config = getApplication();
+                if (config.getReferAsync() != null && config.getReferAsync()) {
```
pinxiong, line 1149: Why not design the `referServices` to be asynchronous instead of relying on `refer-async` parameter?

pinxiong (commented):
There are some suggestions as follows:
+ We don't need to define new parameters `export-async` and `refer-async`, we can make `exportServices` and `referServices` asynchronous
+ We don't to define different ThreadPools. In most cases, only one of serviceExporterExecutor and serviceRefererExecutor will work, because it is less likely that a service is both a service and a reference. We'd better design an shared ThreadPool to handle all temporary tasks

dubbo-common/src/main/java/org/apache/dubbo/config/ApplicationConfig.java
```diff
@@ -547,6 +551,24 @@ public void setStartupProbe(String startupProbe) {
         this.startupProbe = startupProbe;
     }
 
+    @Parameter(key = "export-async")
+    public Boolean getExportAsync() {
```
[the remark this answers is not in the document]
kevinw66: Also there is an issue to suggest we use primitive types. But still there are a lot non-primitive types is using in project. And here I was just trying to keep the same style as before. Maybe we can find a time to change it all together?

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java
kevinw66: I think it's better to let user to choose.

dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/manager/DefaultExecutorRepository.java
```diff
@@ -81,8 +83,9 @@ public DefaultExecutorRepository() {
 //        reconnectScheduledExecutor = Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory("Dubbo-reconnect-scheduler"));
         poolRouterExecutor = new ThreadPoolExecutor(1, 10, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(1024),
             new NamedInternalThreadFactory("Dubbo-state-router-pool-router", true), new ThreadPoolExecutor.AbortPolicy());
-        serviceExporterExecutor = Executors.newScheduledThreadPool(1, new NamedThreadFactory("Dubbo-exporter-scheduler"));
-        serviceDiscveryAddressNotificationExecutor = Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory("Dubbo-SD-address-refresh"));
+        serviceExporterExecutor = Executors.newScheduledThreadPool(10, new NamedThreadFactory("Dubbo-exporter-scheduler", true));
```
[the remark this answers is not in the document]
kevinw66: good idea

dubbo-common/src/main/java/org/apache/dubbo/config/ApplicationConfig.java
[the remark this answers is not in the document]
pinxiong: Yes, we can open an new PR to fix the issue #3531

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java
pinxiong: In this case, we know that `async` is better than `sync`, why don't we recomend `async` or set it as default?

kevinw66: fair point. I'll change it.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java
```diff
@@ -1141,6 +1163,16 @@ private void unreferServices() {
         cache.destroyAll();
     }
 
+    private ExecutorService getExportReferExecutor() {
```
chickenlj, line 1166: Please consider moving this executor to ExecutorRepository

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java
```diff
@@ -1067,19 +1082,21 @@ private void unexportMetadataService() {
     private void exportServices() {
         configManager.getServices().forEach(sc -> {
             // TODO, compatible with ServiceConfig.export()
-            ServiceConfig serviceConfig = (ServiceConfig) sc;
+            ServiceConfig<?> serviceConfig = (ServiceConfig<?>) sc;
             serviceConfig.setBootstrap(this);
 
-            if (exportAsync) {
```
chickenlj, line 1073: Has the definition of exportAsync cleaned up in Bootstrap?

chickenlj (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/config/ApplicationConfig.java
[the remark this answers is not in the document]
chickenlj: I think Boolean is ok for it's better for us to recognize if default value is set.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java
kevinw66: Already annotated with @Deprecated, But It's better to remove the definition, I'll change it

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java
kevinw66: done

chickenlj:
>  There are some suggestions as follows:
> 
> We don't need to define new parameters export-async and refer-async, we can make exportServices and referServices asynchronous
> We don't to define different ThreadPools. In most cases, only one of serviceExporterExecutor and serviceRefererExecutor will work, because it is less likely that a service is both a service and a reference. We'd better design an shared ThreadPool to handle all temporary tasks
> 
> 

I think these are very good suggestions we should take into consideration.

kevinw66:
> > There are some suggestions as follows:
> > We don't need to define new parameters export-async and refer-async, we can make exportServices and referServices asynchronous
> > We don't to define different ThreadPools. In most cases, only one of serviceExporterExecutor and serviceRefererExecutor will work, because it is less likely that a service is both a service and a reference. We'd better design an shared ThreadPool to handle all temporary tasks
> 
> I think these are very good suggestions we should take into consideration.

1、We don't need to define new parameters export-async and refer-async, we can make exportServices and referServices asynchronous

We can consider to set async as default mode, but I think we still need these parameters.

2、We don't to define different ThreadPools. In most cases, only one of serviceExporterExecutor and serviceRefererExecutor will work, because it is less likely that a service is both a service and a reference. We'd better design an shared ThreadPool to handle all temporary tasks

I have done that.

chickenlj:
> 1、We don't need to define new parameters export-async and refer-async, we can make exportServices and referServices asynchronous

> We can consider to set async as default mode, but I think we still need these parameters.

I think it's better to move `export-async and refer-async` to Service&Provider or Reference&Consumer.

kevinw66:
> > 1、We don't need to define new parameters export-async and refer-async, we can make exportServices and referServices asynchronous
> 
> > We can consider to set async as default mode, but I think we still need these parameters.
> 
> I think it's better to move `export-async and refer-async` to Service&Provider or Reference&Consumer.

Agree, I'll change it

chickenlj:
> > > 1、We don't need to define new parameters export-async and refer-async, we can make exportServices and referServices asynchronous
> > 
> > 
> > > We can consider to set async as default mode, but I think we still need these parameters.
> > 
> > 
> > I think it's better to move `export-async and refer-async` to Service&Provider or Reference&Consumer.
> 
> Agree, I'll change it

Thanks, Kevin. It's ready to merge once it's done

kevinw66:
> > > > 1、We don't need to define new parameters export-async and refer-async, we can make exportServices and referServices asynchronous
> > > 
> > > 
> > > > We can consider to set async as default mode, but I think we still need these parameters.
> > > 
> > > 
> > > I think it's better to move `export-async and refer-async` to Service&Provider or Reference&Consumer.
> > 
> > 
> > Agree, I'll change it
> 
> Thanks, Kevin. It's ready to merge once it's done

Done. By the way, have any suggestions for the async-pool-core-size config? I don't think it's a good way to move it to reference/service config, neither stay in application config.

dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/manager/DefaultExecutorRepository.java
```diff
@@ -177,8 +186,41 @@ public ExecutorService nextExecutorExecutor() {
...
+                        new NamedThreadFactory("Dubbo-export-refer", true));
+                }
+            }
+        }
+
+        return exportReferExecutor;
+    }
+
+    private Integer getExportReferThreadNum() {
+        Stream<Integer> provider = getConfigManager().getProviders()
```
AlbumenJ, line 204: This might cause NPE in some situations. It is better to return default value.

AlbumenJ:
- Please add these new configuration into `@DubboReference` and `@DubboService` and spring xds schema.
- I think is better to shutdown the`ScheduledExecutorService` after the last task finished. After that, if user wants to refer or export a new one, such as by API, we can create a new`ScheduledExecutorService`. What's more, we should make sure that there is always one`ScheduledExecutorService` at most all the time.

dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/manager/DefaultExecutorRepository.java
kevinw66: I don't think this is necessary, there always exists default provider/consumer.

kevinw66:
> * Please add these new configuration into `@DubboReference` and `@DubboService` and spring xds schema.
> * I think is better to shutdown the`ScheduledExecutorService` after the last task finished. After that, if user wants to refer or export a new one, such as by API, we can create a new`ScheduledExecutorService`. What's more, we should make sure that there is always one`ScheduledExecutorService` at most all the time.

done.

AlbumenJ (approved):
LGTM.

PR #7959, opened 2021-06-02
Opt Migration Invoker

AlbumenJ:

- migration rule support changed on work
- re-introduce FORCE_INTERFACE, APPLICATION_FIRST, FORCE_APPLICATION mode
- change default dual register
- change default APPLICATION_FIRST, default use service discovery invoker


chickenlj (approved):
LGTM.

PR #7961, opened 2021-06-02
Add unit test for InternalThreadLocalMap

BurningCN:
Add unit test for InternalThreadLocalMap


BurningCN:
Temporarily close and reopen after the check is passed

PR #7962, opened 2021-06-02
Use constants instead of literals

BurningCN:

Use constants instead of literals


chickenlj (approved):
LGTM.

AlbumenJ (approved):

PR #7964, opened 2021-06-03
Add generics to avoid forced conversion

BurningCN:

Add generics to avoid forced conversion


chickenlj (approved):
LGTM

PR #7968, opened 2021-06-03
Remove unused variable and the double calculation of rootPath

BurningCN:
Remove unused variable and the double calculation of rootPath


chickenlj (commented):
Let's avoid changing this part for the time being for someone from the community is working on that too.

chickenlj:
Thanks. I am closing this pr.

PR #7978, opened 2021-06-04
Fix unit test for ConfigUtils#loadProperties method

BurningCN:

- Fix the bug that the ConfigUtils#loadProperties method does not handle exceptions when allowMultiFile=false and multiple files are satisfied

- Fix unit test for ConfigUtils#loadProperties method

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/ConfigUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/ConfigUtils.java
@@ -220,7 +220,7 @@ public class ConfigUtils {
      * </ul>
      * @throws IllegalStateException not allow multi-file, but multi-file exist on class path.
      */
-    public static Properties loadProperties(String fileName, boolean allowMultiFile, boolean optional) {
+    public static Properties loadProperties(String fileName, boolean allowMultiFile, boolean optional) throws IllegalStateException {
         Properties properties = new Properties();
         // add scene judgement in windows environment Fix 2557
         if (checkFileNameExist(fileName)) {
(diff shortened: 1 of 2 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/utils/ConfigUtils.java
```diff
@@ -259,7 +259,7 @@ public static Properties loadProperties(String fileName, boolean allowMultiFile,
             if (list.size() > 1) {
                 String errMsg = String.format("only 1 %s file is expected, but %d dubbo.properties files found on class path: %s",
                         fileName, list.size(), list.toString());
-                logger.warn(errMsg);
+                throw new IllegalStateException(errMsg);
```
chickenlj, line 262: To avoid the following sentence from being executed?

`properties.load(ClassUtils.getClassLoader().getResourceAsStream(fileName));`

https://github.com/apache/dubbo/pull/7978/files#diff-04e43b4a240d7dc90d1b515a05af5baca07e3e96e3d2fae4c7b14722d0b4e145R267

BurningCN: yes, It looks like what you said.

I see that the code record from a long time ago threw an exception, and then it was deleted, but I don’t know why, I can’t open the attached connection here.

http://code.alibabatech.com/jira/browse/DUBBO-133

chickenlj: I mean if `properties.load(ClassUtils.getClassLoader().getResourceAsStream(fileName));` can work well with multiple properties file, we'd better not not add this restriction.

BurningCN: OK

chickenlj:
Could you let me know your Dingding id?

BurningCN:
> Could you let me know your Dingding id?

19w-qfwe0xmhd2

chickenlj:
Thanks, got it.

PR #7988, opened 2021-06-05
Remove unused method calls

BurningCN:
Remove unused method calls

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/extension/support/WrapperComparator.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/extension/support/WrapperComparator.java
@@ -55,25 +55,6 @@ public class WrapperComparator implements Comparator<Object> {
         return n1 > n2 ? 1 : -1;
     }
 
-    private Class<?> findSpi(Class clazz) {
-        if (clazz.getInterfaces().length == 0) {
-            return null;
-        }
-
-        for (Class<?> intf : clazz.getInterfaces()) {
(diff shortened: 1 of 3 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/extension/support/WrapperComparator.java
```diff
@@ -43,10 +43,8 @@ public int compare(Object o1, Object o2) {
             return 0;
         }
 
-        Class clazz1 = (Class) o1;
-        Class clazz2 = (Class) o2;
-
-        Class<?> inf = findSpi(clazz1);
```
BurningCN, line 49: The result of the method call here is not used, so we can directly delete this line of code and the findSpi method below

dubbo-common/src/main/java/org/apache/dubbo/common/utils/TypeUtils.java
```diff
@@ -92,7 +92,7 @@ static boolean isClass(Type type) {
                         }
                     }
                     Class<?> superClass = rawClass.getSuperclass();
-                    if (superClass != null) {
+                    if (superClass != null && NON_OBJECT_TYPE_FILTER.test(superClass)) {
```
BurningCN, line 95: No need to deal with Object type classes

chickenlj:
Please submit to 3.0 branch for this kind of refactoring, we tend to accept bugfixes and necessary enhancements on 2.7 branch.

Issue #7985, opened 2021-06-04
Dubbo 2.6.x may not support multi group consumer


qixiaobo:
- [ ] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.6.4
* Operating System version: *unix
* Java version: 1.8

### Steps to reproduce this issue

1. set dubbo producer group   test
2. not set dubbo service group
3. set dubbo reference group multi_group(test,xxx)

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result

dubbo can invoke specify group service 

### Actual Result

dubbo throws exception

If there is an exception, please attach the exception trace:

```
com.alibaba.dubbo.remoting.RemotingException: Not found exported service: devp495,f6-test/xxx.UserOrgApi:20880 in [f6-test/xxx], may be version or group mismatch , channel: consumer: /10.203.67.52:33238 --> provider: /10.203.64.14:20880, message:RpcInvocation [methodName=queryPermissionOrgIdsByUserId, parameterTypes=[class java.math.BigInteger], arguments=[xxx], attachments={path=xxx.UserOrgApi, input=782, dubboApplication=f6-maintain, _clientInfo={"idOwnOrg":"10907434497393153693","remoteIp":null,"idOwnGroup":"10907434497393152194","udf":null,"from":"maintain","user":"14870719207238203027"}, dubbo=2.0.2, interface=xxx.UserOrgApi version=0.0.0, timeout=3000, group=devp495,f6-test}]
        at com.alibaba.dubbo.rpc.protocol.dubbo.DubboProtocol.getInvoker(DubboProtocol.java:213)
        at com.alibaba.dubbo.rpc.protocol.dubbo.DubboProtocol$1.reply(DubboProtocol.java:79)
        at com.alibaba.dubbo.remoting.exchange.support.header.HeaderExchangeHandler.handleRequest(HeaderExchangeHandler.java:96)
        at com.alibaba.dubbo.remoting.exchange.support.header.HeaderExchangeHandler.received(HeaderExchangeHandler.java:173)
        at com.alibaba.dubbo.remoting.transport.DecodeHandler.received(DecodeHandler.java:51)
[...]

PR #7997, opened 2021-06-07
Dubbo 2.6.x may not support multi group consumer #7985

qixiaobo:
ref to #7985 

Just add some default keys to avoid such special cases


chickenlj (approved):
LGTM.

PR #8007, opened 2021-06-08
Optimize the AbstractConfig#refresh method to avoid multiple refreshes

BurningCN:

For example, the DubboBootstrap#getApplication method will be called multiple times, which will cause AbstractConfig#refresh to be called multiple times. Actually, it only needs to be called once.


BurningCN:
- [ ]

PR #8018, opened 2021-06-09
Replace ProviderMethodModel and ConsumerMethodModel with MethodDescriptor and add some unit test

BurningCN:

Replace ProviderMethodModel and ConsumerMethodModel with MethodDescriptor and add some unit test

### DIFF
--- a/dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/report/MetadataReportInstance.java
+++ b/dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/report/MetadataReportInstance.java
@@ -53,7 +53,7 @@ public class MetadataReportInstance {
                     .removeParameter(METADATA_REPORT_KEY)
                     .build();
         }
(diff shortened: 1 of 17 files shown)


dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/report/MetadataReportInstance.java
```diff
@@ -53,7 +53,7 @@ public static void init(MetadataReportConfig config) {
                     .removeParameter(METADATA_REPORT_KEY)
                     .build();
         }
-        url = url.addParameterIfAbsent(APPLICATION_KEY, ApplicationModel.getApplicationConfig().getName());
```
BurningCN, line 56: `ApplicationModel.getApplicationConfig().getName()` is actually the called `ApplicationModel.getName()`

chickenlj:
Thanks, Burning, that's a great job. 

But I am afraid we cannot accept this patch for the time being. For some historical reasons, some user's inner frameworks rely heavily on these models, so we need to wait for them to eliminate dependencies first before we can refactor them.

Issue #7937, opened 2021-05-31
Consumer can not get server attachment info with using explicit-target


mr3:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.x
* Operating System version: Windows 10
* Java version: [8,)

### Steps to reproduce this issue

1. Refer to this article: [Explicit Target-Specify explicit target in dubbo](https://dubbo.apache.org/en/docs/v2.7/user/examples/explicit-target/), we use consumer directly connected to provider, but consumer can not get server attachment info  through method RpcContext.getServerContext().getAttachment(). The reason is that ConsumerContextClusterInterceptor.onMeesage will not be executed when calling directly.
```java
public void onMessage(Result appResponse, AbstractClusterInvoker<?> invoker, Invocation invocation) {
    RpcContext.getServerContext().setObjectAttachments(appResponse.getObjectAttachments());
}
```
  I checked the history of the class ConsumerContextListener and found that the method onResponse was removed in this [commit](https://github.com/apache/dubbo/commit/2dd3d1a0c25b6414316c5577928b4ba207c0b310#diff-485e8c8c5707a33d551821a70cd294230f0bfdb4d59527a0ca8fd81190b05760)
```java
static class ConsumerContextListener implements Listener {
    @Override
    public void onResponse(Result appResponse, Invoker<?> invoker, Invocation invocation) {
        RpcContext.getServerContext().setAttachments(appResponse.getAttachments());
    }
}
```

  I think we still need these codes, pls give some ideas ASAP.

  If yes, I would submit the fix commit for your guys review.

2. I have wrote a demo to reproduce the problem, [demo address](https://github.com/mr3/dubbo-samples-direct). 
- org.apache.dubbo.samples.direct.DirectServiceIT#testApiGetAttachment
- org.apache.dubbo.samples.direct.DirectServiceIT#testGenericGetAttachment

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result

Consumer can get server attachment info with using explicit-target.

### Actual Result

Consumer can not get server attachment info with using explicit-target.

If there is an exception, please attach the exception trace:

```
Just put your stack trace here!
```

AlbumenJ:
[...]

PR #8024, opened 2021-06-10
Fix consumer can not get server attachment info with using explicit-target(#7937)

mr3:
Fix issue[#7937]
Revert ConsumerContextFilter
Remove ConsumerContextClusterInterceptor
see https://github.com/apache/dubbo/pull/7986

## Checklist


chickenlj:
I don't think we cannot simply replacing Consumer ClusterInterceptor with Filter. ConsumerContext was a Filter and was changed to a ClusterInterceptor after some careful thinking.

While target server is only a test feature that is not recommended to use in production, I would recommend trying to find other ways to enhance it.

chickenlj:
I will let #7937 open and close this or for it's clearly not suitable.

mr3:
> I don't think we cannot simply replacing Consumer ClusterInterceptor with Filter. ConsumerContext was a Filter and was changed to a ClusterInterceptor after some careful thinking.
> 
> While target server is only a test feature that is not recommended to use in production, I would recommend trying to find other ways to enhance it.

Agree with you. We use explicit-target when replay product traffic in test environment, 

Now we have added it a filter(be activated with parameter) to fix this in the internal verison.

_"after some careful thinking."_  more details?

AlbumenJ:
> > I don't think we cannot simply replacing Consumer ClusterInterceptor with Filter. ConsumerContext was a Filter and was changed to a ClusterInterceptor after some careful thinking.
> > While target server is only a test feature that is not recommended to use in production, I would recommend trying to find other ways to enhance it.
> 
> Agree with you. We use explicit-target when replay product traffic in test environment,
> 
> Now we have added it a filter(be activated with parameter) to fix this in the internal verison.
> _"after some careful thinking."_ more details?

Imaging that there is an invocation invoke failed at the first time and it success in the second time. If we use `ConsumerContext`, the rpc context will be cleared in the first and the second time cannot fetch it due to `RpcContext.removeContext(true);`

mr3:
@AlbumenJ Got it, thanks.

PR #7477, opened 2021-03-30
issue: Component scan two times for @Service

zhangjiezhang:
Add cache to solve this issue

PR #8049, opened 2021-06-15
add cache for scan result. (#7477)

zhangjiezhang:
What is the purpose of the change:
 #7477 
The method(ServiceClassPostProcessor's registerServiceBeans) is executed to scan the package path twice, so the solution is to add a cache to the scan class(DubboClassPathBeanDefinitionScanner).


AlbumenJ:
@kylixs PTAL

kylixs (commented):
Please remove unused method: 
```
public Set<BeanDefinitionHolder> doScan(String... basePackages) {
...
}
```

kylixs:
Please fix it:
```
Error: ....DubboClassPathBeanDefinitionScanner.java:19:8: Unused import - org.apache.dubbo.common.utils.CollectionUtils. [UnusedImports]
```

dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/context/annotation/DubboClassPathBeanDefinitionScanner.java
```diff
@@ -57,14 +64,20 @@ public DubboClassPathBeanDefinitionScanner(BeanDefinitionRegistry registry, Envi
...
     @Override
     public boolean checkCandidate(String beanName, BeanDefinition beanDefinition) throws IllegalStateException {
         return super.checkCandidate(beanName, beanDefinition);
     }
 
+    @Override
+    public Set<BeanDefinition> findCandidateComponents(String basePackage) {
+        Set<BeanDefinition> beanDefinitions = beanDefinitionMap.get(basePackage);
+        // if beanDefinitions size ge 0 => don't scan
+        // if beanDefinitions size is null => scan
```
kylixs, line 76: > // if beanDefinitions size ge 0 => don't scan

The comment is inaccurate, it is possible that the scan result is empty

kylixs:
Please remove unused method in Dubbo 3.0: 
```java
    @Override
    public boolean checkCandidate(String beanName, BeanDefinition beanDefinition) throws IllegalStateException {
        return super.checkCandidate(beanName, beanDefinition);
    }
```

dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/context/annotation/DubboClassPathBeanDefinitionScanner.java
zhangjiezhang: > > // if beanDefinitions size ge 0 => don't scan
> 
> The comment is inaccurate, it is possible that the scan result is empty.
> Scan only when the cache is not found.

beanDefinitions size >= 0, means that the scan has been executed.
Because the result of the scan may be equal to 0, only beanDefinitions is null to scan.

kylixs: Please remove redundant comments.

kylixs:
> No description provided.

Please add some description of this pr.

kylixs:
@AlbumenJ Please approve workflow runs
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

kylixs (approved):

PR #8059, opened 2021-06-16
forward #8039  Registry cannot work under k8s API server using SSL

m1ngyuan:
forward #8039  Registry cannot work under k8s API server using SSL


m1ngyuan:
reopen

PR #8065, opened 2021-06-16
Fix the issue that the ReferenceConfigCache#destroy method does not call proxy.$destroy()

BurningCN:

The destroy method should trigger the call of proxy.$destroy() like the destroyAll method


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/utils/ReferenceConfigCache.java
```diff
@@ -175,7 +175,8 @@ public void destroy(String key, Class<?> type) {
 
         Map<String, Object> proxiesOftype = proxies.get(type);
         if (CollectionUtils.isNotEmptyMap(proxiesOftype)) {
-            proxiesOftype.remove(key);
+            Destroyable proxy = (Destroyable) proxiesOftype.remove(key);
```
AlbumenJ, line 178: it would be better to check if the object is an instance of Destroyable first

BurningCN: The generated proxy instances all automatically implement the Destroyable interface, right?

image
image

PR #8136, opened 2021-06-24
fix #8135 add a default value to accessKeyId

### LINKED ISSUE #8135: Auth properties accessKeyId is not necessary
[x] I have searched the issues of this repository and believe that this is not a duplicate.
[x] I have checked the FAQ of this repository and believe that this is not a duplicate.
Environment
Dubbo version: 2.7.9
Operating System version: macos 10.14.6
Java version: oracle 1.8.0_211
Steps to reproduce this issue

Expected Result
As an atuh configuration properties,accessKeyId makes no difference to signature,however, I have to define it in my auth api。In most case，I would like handle  less config  rather than handle everything。By the way，accessKeyId that has a default value maybe a better choice？

24kpure:
#8135 

Reducing auth configuration

## Brief changelog
Adding a default value to accessKeyId

## Checklist

### DIFF
--- a/dubbo-plugin/dubbo-auth/src/main/java/org/apache/dubbo/auth/Constants.java
+++ b/dubbo-plugin/dubbo-auth/src/main/java/org/apache/dubbo/auth/Constants.java
@@ -42,4 +42,6 @@ public interface Constants {
     String SIGNATURE_STRING_FORMAT = "%s#%s#%s#%s";
 
     String PARAMETER_SIGNATURE_ENABLE_KEY = "param.sign";
+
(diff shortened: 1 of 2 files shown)


AlbumenJ:
Was is necessary to add a default value? This might cause some compatible problem for other users.

24kpure:
Fist of all,thanks for your advice.
It's not necessary to add a default value,but  it  does reduce some auth config which make no difference to  result in most case。 As for compatible problem，as far as I am concerned, it works well when new users add new auth service config ,which thows AccessKeyNotFoundException in current version。Similarly，old users  will not  be in trouble for they have added the config for some time。

PR #8142, opened 2021-06-25
Print warn log instead of throw exception.

chickenlj:


kylixs:
Some types of configuration are unique, and only one instance takes effect. If there are multiple configurations, ignoring some of them will cause confusion. It is difficult for users to understand which configuration is effective, which may lead to unexpected results. Alarm logs are often ignored, it is difficult for every user to deal with it, and throwing errors is more clear.

PR #8184, opened 2021-06-30
Use service name mapping key to avoid logic conflict.

horizonzy:

Now use REGISTRY_TYPE_KEY to judge is publish service name mapping or not, but this key is meaningful for RegistryProtol, maybe it will influence the origin logic. 

So add a particular key to judge is publish service name mapping info or not.


xiaoheng1 (approved):
LGTM

PR #8186, opened 2021-06-30
[3.0] Improve async export / refer

kevinw66:

Improve async export / refer

## Brief changelog

Currently we use 6 parameters to support async and background export / refer
Usage: 
`<dubbo:provider export-async="true" export-thread-num="10" export-background="true"/>` 
`<dubbo:consumer refer-async="true" refer-thread-num="10" refer-background="true"/>` 

export-async(refer-async): we create a thread pool to export(refer) services if this set to `true`, or we do export(refer) in main thread if this set to `false`, default `false`.
export-thread-num(refer-thread-num): means the core size of async thread pool, default `10`.
export-background(refer-background): async export(refer) will run in background if this set to `true`, or bootstrap will wait it to finish if this set to `false`, default `false`. this can only take effects when `export-async(refer-async)` is set to `true`


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/DubboBootstrap.java
```diff
@@ -1103,26 +1108,25 @@ public DubboBootstrap start() {
             }
 
             referServices();
-            if (asyncExportingFutures.size() > 0 || asyncReferringFutures.size() > 0) {
-                new Thread(() -> {
+
+            // wait async export / refer finish if needed
+            awaitFinish();
+
+            new Thread(() -> {
```
AlbumenJ, line 1115: Is it necessary to create a thread even if users not specify async export?

kevinw66: Changed

Issue #8195, opened 2021-07-01
One NPE in dubbo/container/Main.java


containerAnalyzer:
Hello,
Our static analyzer found a following potential NPE. We have checked the feasibility of this execution trace. It is necessary to defend this vulnerability to improve the code quality.

1. Return **null** to caller (Trace staring point)
https://github.com/apache/dubbo/blob/f26ba91b67f642148a10d3b197502e29928b77bf/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java#L466

2. Function **getDefaultExtension** executes, stores the return value to **defaultextension** (**defaultextension** can be **null**) and return **defaultextension** to caller, which can be null 
https://github.com/apache/dubbo/blob/f26ba91b67f642148a10d3b197502e29928b77bf/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java#L434

3. Return the return value of function **getExtension** to caller
https://github.com/apache/dubbo/blob/f26ba91b67f642148a10d3b197502e29928b77bf/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java#L426

4. Function **getExtension** executes and stores the return value to **extension** (extension can be null)
https://github.com/apache/dubbo/blob/f26ba91b67f642148a10d3b197502e29928b77bf/dubbo-container/dubbo-container-api/src/main/java/org/apache/dubbo/container/Main.java#L63

5. Function **add** executes and **containers** contains null as its elements.
https://github.com/apache/dubbo/blob/f26ba91b67f642148a10d3b197502e29928b77bf/dubbo-container/dubbo-container-api/src/main/java/org/apache/dubbo/container/Main.java#L63

6. Function **next** executes and return **null** value
https://github.com/apache/dubbo/blob/f26ba91b67f642148a10d3b197502e29928b77bf/dubbo-container/dubbo-container-api/src/main/java/org/apache/dubbo/container/Main.java#L89

7. The return value of function **next** is passed as the this pointer to function **start** (the return value of function **next** can be null), which will leak to null pointer dereference
https://github.com/apache/dubbo/blob/f26ba91b67f642148a10d3b197502e29928b77bf/dubbo-container/dubbo-container-api/src/main/java/org/apache/dubbo/container/Main.java#L90

Commit: f26ba91b67f642148a10d3b197502e29928b77bf

ContainerAnalyzer

xh1202:
I will fix it.

[...]

PR #8207, opened 2021-07-03
Fix the NPE (#8195)

xh1202:


dubbo-container/dubbo-container-api/src/main/java/org/apache/dubbo/container/Main.java
```diff
@@ -60,7 +60,10 @@ public static void main(String[] args) {
 
             final List<Container> containers = new ArrayList<Container>();
             for (int i = 0; i < args.length; i++) {
-                containers.add(LOADER.getExtension(args[i]));
+                Container container = LOADER.getExtension(args[i]);
```
horizonzy, line 63: here maybe throw exception will be better, tell user the args maybe wrong.

dubbo-container/dubbo-container-api/src/main/java/org/apache/dubbo/container/Main.java
```diff
@@ -61,9 +61,10 @@ public static void main(String[] args) {
             final List<Container> containers = new ArrayList<Container>();
             for (int i = 0; i < args.length; i++) {
                 Container container = LOADER.getExtension(args[i]);
-                if (null != container) {
-                    containers.add(LOADER.getExtension(args[i]));
+                if (null == container) {
+                    throw new IllegalArgumentException("Default extension is not configured, but the extension name is 'true'");
                 }
+                containers.add(LOADER.getExtension(args[i]));
```
horizonzy, line 67: here can reuse container

dubbo-container/dubbo-container-api/src/main/java/org/apache/dubbo/container/Main.java
```diff
@@ -61,9 +61,10 @@ public static void main(String[] args) {
             final List<Container> containers = new ArrayList<Container>();
             for (int i = 0; i < args.length; i++) {
                 Container container = LOADER.getExtension(args[i]);
-                if (null != container) {
-                    containers.add(LOADER.getExtension(args[i]));
+                if (null == container) {
+                    throw new IllegalArgumentException("Default extension is not configured, but the extension name is 'true'");
```
horizonzy, line 65: msg can be String.format("The extension is not find, pls check your param: %s is match spi", args[i]);

horizonzy:
This issue maybe be fixed by #8215.

Issue #8218, opened 2021-07-05
Abuse in variables and getters or setters


pinxiong:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: `3.0.2-SNAPSHOT`
* Operating System version: `Mac OS`
* Java version: `1.8`

### Steps to reproduce this issue

+ Don't need to expose private variables

```java
public abstract class AbstractRouter implements Router {
    // The getter and setter already exist, so don't need to expose the variable
    protected int priority = DEFAULT_PRIORITY;
    // The getter and setter already exist, so don't need to expose the variable
    protected boolean force = false;
    // The getter already exists, so don't need to expose the variable
    protected URL url;
    // Don't need to expose the variable
    protected GovernanceRuleRepository ruleRepository;

    public AbstractRouter(URL url) {
        this.ruleRepository = ExtensionLoader.getExtensionLoader(GovernanceRuleRepository.class).getDefaultExtension();
        this.url = url;
    }

    @Override
    public URL getUrl() {
        return url;
    }

    public void setUrl(URL url) {
        this.url = url;
    }

    @Override
    public boolean isForce() {
        return force;
    }

    public void setForce(boolean force) {
        this.force = force;
    }

    @Override
    public int getPriority() {
        return priority;
    }

    public void setPriority(int priority) {
        this.priority = priority;
    }
}
```

+ Why not use setter or getter
```java
public class ConditionRouter extends AbstractRouter {
    ......
    public ConditionRouter(String rule, boolean force, boolean enabled) {
        // use setter
        this.force = force;
        this.enabled = enabled;
        if (enabled) {
            this.init(rule);
        }
    }

    public ConditionRouter(URL url) {
        // use setter
        this.url = url;
        // use setter
        this.priority = url.getParameter(PRIORITY_KEY, 0);
        // use setter
        this.force = url.getParameter(FORCE_KEY, false);
        this.enabled = url.getParameter(ENABLED_KEY, true);
        if (enabled) {
            init(url.getParameterAndDecoded(RULE_KEY));
        }
    }
}
```

PR #8219, opened 2021-07-05
perf: Optimize access permissions of variables

pinxiong:

see #8218 

## Checklist


horizonzy (commented):
Hi, I think the modification is not necessary, the origin is a way use `extend`. 
protected modifier use at subclass directly is normal.

pinxiong:
For Java syntax, it does work well. However, there are two reasons for those changes

+ All of variables with `protected` modifier have defined setters and getters in `AbstractRouter`.

+ It is best to provide only one way to modify the same variables, however, there are two ways to visite those variables:

    + use `protected` modifier to access directly
    + use setter and getter to access

I think the best way is to use setter and getter to access variables, because the interface of `Router` already defines some abstract getter methods, like `Router#getUrl()`, `Router#isForce()`, `Router#getPriority()` etc.

```java
public interface Router extends Comparable<Router> {
    ......
    URL getUrl();

    boolean isForce();

    int getPriority();
    ......
    @Override
    default int compareTo(Router o) {
        if (o == null) {
            throw new IllegalArgumentException();
        }
        return Integer.compare(this.getPriority(), o.getPriority());
    }
}
```

PR #8230, opened 2021-07-05
Don't trust the response format from nacos server.

horizonzy:

Don't trust the response format from nacos server.


xiaoheng1 (approved):
LGTM

PR #8234, opened 2021-07-06
support native image

goodjava:
support native image


dubbo-native/src/main/java/org/apache/dubbo/common/serialize/Serialization$Adaptive.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.dubbo.common.serialize;
+import org.apache.dubbo.common.extension.ExtensionLoader;
+public class Serialization$Adaptive implements org.apache.dubbo.common.serialize.Serialization {
```
kylixs, line 19: It would be more useful if `XXX$Adaptive.java` could be generated automatically through a plugin or command line.

goodjava: Plug-in is being developed

Issue #8303, opened 2021-07-16
Dubbo3.0使用nacos2.0注册时parameters: namespace无效


xiaohuige99:
- [ ] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 3.0.1
* Operating System version: mac os
* Java version: 1.8

### Steps to reproduce this issue

配置信息
  registry:
    address: nacos://192.168.88.90:8848?namespace=dubbo-test # 这样是可以的
    parameters:
      namespace: dubbo-test #这种不行
    group: TEST_GROUP

xiaoheng1:
let me have a look

CrazyHZM:
Resolved in #8308

PR #8308, opened 2021-07-17
support  dubbo.registry.parameters.item3=value3 configuration properties

xiaoheng1:

Support the following format configuration properties. such as:

dubbo.registry.parameters.item3=value3

see #8303 

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractConfig.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractConfig.java
@@ -561,6 +561,9 @@ public abstract class AbstractConfig implements Serializable {
     }
 
     private void invokeSetParameters(Map<String, String> values) {
+        if (CollectionUtils.isEmptyMap(values)) {
+            return;
+        }
         Map<String, String> map = invokeGetParameters(getClass(), this);
         map = map == null ? new HashMap<>() : map;
         map.putAll(values);
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractConfig.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractConfig.java
@@ -540,12 +540,14 @@ public abstract class AbstractConfig implements Serializable {
                 } else if (isParametersSetter(method)) {
                     String propertyName = extractPropertyName(method.getName());
                     String value = StringUtils.trim(subPropsConfiguration.getString(propertyName));
+                    Map<String, String> parameterMap = null;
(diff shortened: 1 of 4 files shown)


dubbo-common/src/main/java/org/apache/dubbo/config/AbstractConfig.java
```diff
@@ -559,6 +560,13 @@ public void refresh() {
         postProcessRefresh();
     }
 
+    private void invokeSetParameters(Map<String, String> values) {
+        Map<String, String> map = invokeGetParameters(getClass(), this);
```
kylixs, line 564: It is better to check argument `Map<String, String> values`, If it is null or empty, just return

dubbo-common/src/main/java/org/apache/dubbo/config/AbstractConfig.java
```diff
@@ -540,10 +541,10 @@ public void refresh() {
...
                     String value = StringUtils.trim(subPropsConfiguration.getString(propertyName));
                     if (StringUtils.hasText(value)) {
-                        Map<String, String> map = invokeGetParameters(getClass(), this);
-                        map = map == null ? new HashMap<>() : map;
-                        map.putAll(convert(StringUtils.parseParameters(value), ""));
-                        invokeSetParameters(getClass(), this, map);
+                        invokeSetParameters(convert(StringUtils.parseParameters(value), ""));
+                    } else {
+                        // in this case, maybe parameters.item3=value3.
+                        invokeSetParameters(ConfigurationUtils.getSubProperties(subProperties, PARAMETERS));
```
kylixs, line 547: Should we do `convert()` parameters or not? 
For compatibility, convert is needed, but additional parameters will be added, which leads to more url parameters, which is not good. This compatibility does not know what the specific problem is, it is best to deal with it when the parameters are read.

kylixs (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/config/AbstractConfig.java
xiaoheng1: ok, i will modify it.

kylixs (approved):

dubbo-common/src/main/java/org/apache/dubbo/config/AbstractConfig.java
kylixs: Some use a.b, some a-b, we will normalize parameter keys later.

PR #8430, opened 2021-08-05
[3.0] compatible with spring 3.2.x and add tests

kylixs:

compatible with spring 3.2.x and add tests


dubbo-test/dubbo-test-spring4.2/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+        <groupId>org.apache.dubbo</groupId>
+        <version>${revision}</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>dubbo-test-spring4.2</artifactId>
+
+    <properties>
+        <spring_version>4.2.4.RELEASE</spring_version>
```
AlbumenJ, line 31: please add maven skip deploy property

PR #8471, opened 2021-08-11
Make ServiceConfigTest#testDelayExport more reasonable

horizonzy:
https://github.com/apache/dubbo/blob/b64ea1d8c6c5b64828ef3d08f8cc432de59bb559/dubbo-config/dubbo-config-api/src/test/java/org/apache/dubbo/config/ServiceConfigTest.java#L205-L212

Here use sleep to test delay, it's not suitable. It can't guarantee that delayService already exported.

So here use listener to check it.


xiaoheng1 (approved):
LGTM

BurningCN (approved):
LGTM

PR #8519, opened 2021-08-16
fix ShortestResponseLoadBalanceTest failure

24kpure:
ShortestResponseLoadBalanceTest#testSelectByResponse sometimes runs failed，for weightInvoker1's weight is standard unit and it maybe especially big or small. In my opinion, the asserts of weightInvoker2 and weightInvoker5  are enough，while maxDeviation should be bigger.
 
[Test failure](https://github.com/apache/dubbo/pull/8510/checks?check_run_id=3336766043)
[Relate pr](https://github.com/apache/dubbo/pull/8441#issuecomment-899264224)

### DIFF
--- a/dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/loadbalance/ShortestResponseLoadBalanceTest.java
+++ b/dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/loadbalance/ShortestResponseLoadBalanceTest.java
@@ -97,14 +97,13 @@ public class ShortestResponseLoadBalanceTest extends LoadBalanceBaseTest {
             }
         }
         Map<Invoker<LoadBalanceBaseTest>, Integer> weightMap = weightInvokersSR.stream()
-                .collect(Collectors.toMap(Function.identity(), e -> Integer.valueOf(e.getUrl().getParameter("weight"))));
+            .collect(Collectors.toMap(Function.identity(), e -> Integer.valueOf(e.getUrl().getParameter("weight"))));
         Integer totalWeight = weightMap.values().stream().reduce(0, Integer::sum);
-        // max deviation 10%
+        // max deviation = expectWeightValue * 2
         int expectWeightValue = loop / totalWeight;
-        int maxDeviation = expectWeightValue / 10;
+        int maxDeviation = expectWeightValue * 2;
 
         Assertions.assertEquals(sumInvoker1 + sumInvoker2 + sumInvoker5, loop, "select failed!");
-        Assertions.assertTrue(Math.abs(sumInvoker1 / weightMap.get(weightInvoker1) - expectWeightValue) < maxDeviation, "select failed!");
         Assertions.assertTrue(Math.abs(sumInvoker2 / weightMap.get(weightInvoker2) - expectWeightValue) < maxDeviation, "select failed!");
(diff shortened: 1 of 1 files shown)


dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/loadbalance/ShortestResponseLoadBalanceTest.java
```diff
@@ -97,14 +97,13 @@ public void testSelectByResponse() throws NoSuchFieldException, IllegalAccessExc
...
         }
         Map<Invoker<LoadBalanceBaseTest>, Integer> weightMap = weightInvokersSR.stream()
-                .collect(Collectors.toMap(Function.identity(), e -> Integer.valueOf(e.getUrl().getParameter("weight"))));
+            .collect(Collectors.toMap(Function.identity(), e -> Integer.valueOf(e.getUrl().getParameter("weight"))));
         Integer totalWeight = weightMap.values().stream().reduce(0, Integer::sum);
-        // max deviation 10%
+        // max deviation 12.5%
         int expectWeightValue = loop / totalWeight;
-        int maxDeviation = expectWeightValue / 10;
+        int maxDeviation = expectWeightValue / 8;
```
AlbumenJ, line 104: As for test stability, it would be better to set maxDeviation to `expectWeightValue*2` to tolerant random seed problem.

AlbumenJ (changes requested):

dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/loadbalance/ShortestResponseLoadBalanceTest.java
24kpure: Good idea!

pinxiong:
@24kpure please fix the failed testcase

24kpure:
> @24kpure please fix the failed testcase
Thanks for your notice. I have reviewed my changed code which make no relate with org.apache.dubbo.config.spring.schema.GenericServiceTest#testGeneric. As a matter of fact,I don't commit anything, however,ut offen run failed.

pinxiong:
> @pinxiong Thanks for your notice. I have reviewed my changed code which make no relate with org.apache.dubbo.config.spring.schema.GenericServiceTest#testGeneric. As a matter of fact,I don't commit anything, however,ut offen run failed.

You also can have a try😎

24kpure:
Luckily，it worked，but it did cost a lot of time. @pinxiong

AlbumenJ (approved):
LGTM

AlbumenJ:
@24kpure Thanks for your contribution. I think if you can directly using merge instead of force-pushed would be better for others to track the commits.

24kpure:
> @24kpure Thanks for your contribution. I think if you can directly using merge instead of force-pushed would be better for others to track the commits.

I agree with you. It can be more clear to track the commits ,however,I have some doubts.Would it be more elegant to deal a issue with one commit. 
You are always helping me and giving me advice and I have great respect for you.

AlbumenJ:
> I agree with you. It can be more clear to track the commits ,however,I have some doubts.Would it be more elegant to deal a issue with one commit.

I agree with you. It can be more clear to track the commits ,however,I have some doubts.Would it be more elegant to deal a issue with one commit.

For Apache Dubbo project, we use squash and merge to merge a PR. That means all of the changes will be compressed into a single commit. Therefore, even if there are a lot of commits in a single PR, only one commit will be on main branch.

24kpure:
It's very nice of you ~

PR #8522, opened 2021-08-16
Dubbo ProviderConsumerRegTable class cause OutOfMemoryException

qixiaobo:
fix https://github.com/apache/dubbo/issues/8515

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/ReferenceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/ReferenceConfig.java
@@ -29,6 +29,8 @@ import com.alibaba.dubbo.config.annotation.Reference;
 import com.alibaba.dubbo.config.model.ApplicationModel;
 import com.alibaba.dubbo.config.model.ConsumerModel;
 import com.alibaba.dubbo.config.support.Parameter;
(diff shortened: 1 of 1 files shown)


dubbo-config/dubbo-config-api/src/main/java/com/alibaba/dubbo/config/ReferenceConfig.java
```diff
@@ -423,7 +427,14 @@ private T createProxy(Map<String, String> map) {
         if (c && !invoker.isAvailable()) {
             // make it possible for consumer to retry later if provider is temporarily unavailable
             initialized = false;
-            throw new IllegalStateException("Failed to check the status of the service " + interfaceName + ". No provider available for the service " + (group == null ? "" : group + "/") + interfaceName + (version == null ? "" : ":" + version) + " from the url " + invoker.getUrl() + " to the consumer " + NetUtils.getLocalHost() + " use dubbo version " + Version.getVersion());
+            final String serviceKey = (group == null ? "" : group + "/") + interfaceName + (version == null ? "" :
```
BurningCN, line 430: Here you can use `BaseServiceMetadata.buildServiceKey(path, group, version);`

qixiaobo: @BurningCN  Thanks. But you may see this is 2.6.x version

PR #8532, opened 2021-08-18
Fix bug of BroadcastClusterInvoker and add test cases

BurningCN:

Fix bug of BroadcastClusterInvoker and add test cases

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/BroadcastClusterInvoker.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/BroadcastClusterInvoker.java
@@ -72,20 +72,18 @@ public class BroadcastClusterInvoker<T> extends AbstractClusterInvoker<T> {
                     if (null != resultException) {
                         exception = getRpcException(result.getException());
                         logger.warn(exception.getMessage(), exception);
+                        failIndex++;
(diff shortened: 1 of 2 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/BroadcastClusterInvoker.java
```diff
@@ -72,20 +72,18 @@ public Result doInvoke(final Invocation invocation, List<Invoker<T>> invokers, L
                     if (null != resultException) {
                         exception = getRpcException(result.getException());
                         logger.warn(exception.getMessage(), exception);
+                        failIndex++;
```
BurningCN, line 75: The accumulation of failIndex should be placed above the judgment of failIndex. Suppose 4 invokers (their invoke methods will throw exceptions), and the failure percentage is configured to 50/100. It is reasonable to throw exceptions after 2 invokers fail. If it is the original code method, it will throw exception after 3 invokers fail.

failIndex的累加应该放到对failIndex判断的上面。假设4个invoker（他们的invoke方法都会抛异常），配置了失败百分比50/100，按道理是2个invoker失败后上抛异常，如果是原有代码方式的话，需要3个invoker调用失败才会抛异常

AlbumenJ (approved):

PR #8558, opened 2021-08-20
fix Issue 8516 on branch 3.0: fix mock:true configuration.

changfubai:
same as #8520


BurningCN:
The name of pr should be more specific, such as what problem is fixed

changfubai:
> The name of pr should be more specific, such as what problem is fixed

ok.thanks

AlbumenJ (approved):

PR #8560, opened 2021-08-20
Add test cases to the Configurator after version 2.7

BurningCN:

Add test cases to the Configurator after version 2.7


chickenlj (approved):
LGTM

PR #8566, opened 2021-08-22
[3.0-tri-compiler] dubbo compiler support stream

EarthChen:

- change dubbo-compiler version to `0.0.3-SNAPSHOT`
- dubbo compiler support stream on dubbo3 gen
- drop compiler dependency `dubbo-serialization-protobuf`
- change TriUtils pbserialization to SingleProtobufUtils

## Checklist

### DIFF
--- a/compiler/src/main/resources/Dubbo3InterfaceStub.mustache
+++ b/compiler/src/main/resources/Dubbo3InterfaceStub.mustache
@@ -1,5 +1,5 @@
 {{#packageName}}
-    package {{packageName}};
+package {{packageName}};
 {{/packageName}}
 
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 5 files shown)


compiler/src/main/resources/Dubbo3InterfaceStub.mustache
```diff
@@ -18,11 +18,37 @@ static final String SERVICE_NAME = "{{commonPackageName}}.{{serviceName}}";
...
+
+{{#biStreamingMethods}}
+    {{#javaDoc}}
+        {{{javaDoc}}}
+    {{/javaDoc}}
+    {{#deprecated}}
+        @java.lang.Deprecated
+    {{/deprecated}}
+    org.apache.dubbo.common.stream.StreamObserver<{{inputType}}> {{methodName}}(org.apache.dubbo.common.stream.StreamObserver<{{outputType}}> responseObserver);
+{{/biStreamingMethods}}
```
chickenlj, line 52: I think we should consider how the developers will use this before determining support of streaming interface:
* For the consumer side, developers can call any of the `sync`, `async`, `streaming` methods from the proxy they get.
* For the provider side, developers don't have to provider implementation for all methods, one bi-directional streaming implementation would be enough.

chickenlj (changes requested):

compiler/src/main/resources/Dubbo3InterfaceStub.mustache
EarthChen: i think we should support like grpc. That's enough.This is also how tri Stream is implemented

chickenlj: Yes, but simply add those streaming method declarations will be a burden for the Provider, we should think of letting the Provider only care about the bi-directional streaming implementation.

chickenlj: not have to give the implementation of all methods.

EarthChen: server stream and bi stream is diff, Developers should decide which to use

chickenlj: They are different and that's the key point we need to determine. I think for the most generic case, Provider can provider one streaming implementation that can serve all different kinds of calls from the Consumer side.

chickenlj: Sure we can let the Provider decide to only support unary or streaming or bi streaming, but is that necessary, or at least whey they want to choose they don't need to give the implementation of all kinds.

EarthChen: You mean to provide a default implementation that wraps the service stream with a bi-directional stream

EarthChen:
drop `@Generated` Because a higher version of JDK not exists

EarthChen:
Stub will be supported later similar to GRPC

guohao (approved):
LGTM

guohao:
Basic interface containing simplest method definitions has been generated by compiler,  then `Stub` should be considerdto support multiple programming API :future,unary,stream...etc

Issue #7463, opened 2021-03-28
dubbo-demo-xml  can't startup in jdk15


wenchaomartin:
- [ ] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.10-SNAPSHOT
* Operating System version: win10 
* Java version: 15

### Steps to reproduce this issue

1. change all zookeeper listenning port 2181 to 2182(customed by myself)
2. set enviroment -Djava.net.preferIPv4Stack=true
3. run dubbo-demo-xml  provider

Pls. provide [GitHub address] to reproduce this issue.

### Expected Result

start up rightly 

when i switch jdk15 to jdk8 ,start up rightly

### Actual Result

What actually happens?

If there is an exception, please attach the exception trace:

```
[28/03/21 23:15:07:296 CST] main  INFO zookeeper.ZooKeeper: Initiating client connection, connectString=127.0.0.1:2182 sessionTimeout=60000 watcher=org.apache.curator.ConnectionState@ebb6851
[28/03/21 23:15:07:373 CST] main-SendThread(127.0.0.1:2182)  INFO zookeeper.ClientCnxn: Opening socket connection to server 127.0.0.1/<unresolved>:2182. Will not attempt to authenticate using SASL (unknown error)
[28/03/21 23:15:07:375 CST] main  INFO imps.CuratorFrameworkImpl: Default schema
[28/03/21 23:15:07:377 CST] main-SendThread(127.0.0.1:2182)  WARN zookeeper.ClientCnxn: Session 0x0 for server 127.0.0.1/<unresolved>:2182, unexpected error, closing socket connection and attempting reconnect
java.nio.channels.UnresolvedAddressException
	at java.base/sun.nio.ch.Net.checkAddress(Net.java:149)
	at java.base/sun.nio.ch.Net.checkAddress(Net.java:157)
	at java.base/sun.nio.ch.SocketChannelImpl.checkRemote(SocketChannelImpl.java:753)
	at java.base/sun.nio.ch.SocketChannelImpl.connect(SocketChannelImpl.java:774)
	at org.apache.zookeeper.ClientCnxnSocketNIO.registerAndConnect(ClientCnxnSocketNIO.java:277)
[...]

PR #8568, opened 2021-08-22
Fix issue #7463, upgrade zk version on branch 3.0

BazookaW:
issue no #7463


chickenlj:
Does this mean users have to upgrade zookeeper server to 3.5 or above to use Dubbo 3.0?

PR #8636, opened 2021-08-30
Resource clean before destroy or stop

BurningCN:
Resource clean before destroy or stop


wangchengming666 (approved):
LGTM

PR #8648, opened 2021-09-01
2.7.13 fix add serialize check switch

owen200008:
2.7.13 fix add serialize check switch，use config to read SerializeClassChecker config


BurningCN (approved):
LGTM

PR #8671, opened 2021-09-03
[3.0] Enhance: Remove similar code

wangchengming666:
在3.0分支的NacosNamingServiceUtils中有两个方法：

```
private static void putPropertyIfAbsent(URL url, Properties properties, String propertyName) {
        String propertyValue = url.getParameter(propertyName);
        if (StringUtils.isNotEmpty(propertyValue)) {
            properties.setProperty(propertyName, propertyValue);
        }
    }
```

```
private static void putPropertyIfAbsent(URL url, Properties properties, String propertyName, String defaultValue) {
        String propertyValue = url.getParameter(propertyName);
        if (StringUtils.isNotEmpty(propertyValue)) {
            properties.setProperty(propertyName, propertyValue);
        } else {
            properties.setProperty(propertyName, defaultValue);
        }
    }
```

这两个方法做的事情基本都是一样的，只不过一个方法的形参支持默认值，所以考虑合并成一个方法，然后将`NACOS_NAMING_LOG_NAME`给一个默认值`null`。

`NACOS_NAMING_LOG_NAME`是指定nacos的log地址，考虑给了`NACOS_NAMING_LOG_NAME`默认值`null`，所以去看了一下nacos的源码，`NACOS_NAMING_LOG_NAME`有三个地方引用。
image
代码里都做了判空，故不会影响现在的逻辑
image


dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/util/NacosNamingServiceUtils.java
```diff
@@ -133,7 +133,7 @@ private static void setServerAddr(URL url, Properties properties) {
     }
 
     private static void setProperties(URL url, Properties properties) {
-        putPropertyIfAbsent(url, properties, NACOS_NAMING_LOG_NAME);
+        putPropertyIfAbsent(url, properties, NACOS_NAMING_LOG_NAME, null);
```
AlbumenJ, line 136: `null` should not be default value here. We should not call `properties.setProperty` with null.

AlbumenJ: Also refer to Integration Tests

AlbumenJ (changes requested):

xiaoheng1 (commented):
It may be a good idea to extract a method that supports default values at the upper level, and then use a public method in this method.

wangchengming666:
@AlbumenJ could you please help me review the code again. Thank you.

PR #8697, opened 2021-09-06
Caculate capacity in the default constructor of InternalThreadLocalMap

kingkh1995:
avoid foreseeable expanding in InternalThreadLocalMap

## Brief changelog
use the same code which in method ***expandIndexedVariableTableAndSet*** to caculate capacity when create a InternalThreadLocalMap instance


kingkh1995:
Caculate capacity in the default constructor of InternalThreadLocalMap

PR #8721, opened 2021-09-08
[3.0] Judge whether success or not by user's perspective in interface-appName mapping.

### LINKED ISSUE #8722: [3.0] the same service export multi protocol error
- [ ] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 3.0.3-SNAPSHOT
* Operating System version: macos
* Java version: 1.8

1.  provider export multi protocol 
```java
class ApiMigrationBothProvider {

        ServiceConfig<IWrapperGreeter> service = new ServiceConfig<>();
        service.setInterface(IWrapperGreeter.class);
        service.setRef(new IGreeter2Impl());

        DubboBootstrap bootstrap = DubboBootstrap.getInstance();
        bootstrap.application(new ApplicationConfig("demo-migration-both-provider"))
                .registry(new RegistryConfig(TriSampleConstants.ZK_ADDRESS))
                .protocol(new ProtocolConfig(CommonConstants.TRIPLE, TriSampleConstants.SERVER_PORT))
                .protocol(new ProtocolConfig(CommonConstants.DUBBO, TriSampleConstants.DEFAULT_DUBBO_PORT))
                .service(service)
                .start()
                .await();

    }
}
```

2. run provider

## reason

image
 
when the second protocol export, contains always true 

Pls. provide [GitHub address] to reproduce this issue.

https://github.com/apache/dubbo-samples/pull/356

### Expected Result

everything is ok

### Actual Result

throw exception

If there is an exception, please attach the exception trace:

```
java.lang.RuntimeException
	at org.apache.dubbo.registry.client.metadata.MetadataServiceNameMapping.map(MetadataServiceNameMapping.java:88)
	at org.apache.dubbo.config.ServiceConfig.lambda$exported$1(ServiceConfig.java:270)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
	at org.apache.dubbo.config.ServiceConfig.exported(ServiceConfig.java:266)
	at org.apache.dubbo.config.ServiceConfig.doExport(ServiceConfig.java:371)
	at org.apache.dubbo.config.ServiceConfig.export(ServiceConfig.java:244)
	at org.apache.dubbo.config.bootstrap.DubboBootstrap.exportServices(DubboBootstrap.java:1476)
	at org.apache.dubbo.config.bootstrap.DubboBootstrap.doStart(DubboBootstrap.java:1190)
	at org.apache.dubbo.config.bootstrap.DubboBootstrap.start(DubboBootstrap.java:1166)
	at org.apache.dubbo.sample.tri.migration.ApiMigrationBothProvider.main(ApiMigrationBothProvider.java:44)
```

plusmancn:
* Judge whether success or not by user's perspective in interface-appName mapping.
* By the way, format the function `MetadataServiceNameMapping#map`: change the indent spaces from 8 to 4.

Fixes #8722 

## Brief changelog
Every time the provider restarted after the first start,  it would raise an error "Failed register interface application mapping......".
But from the user's perspective, it means successful when the oldConfigContent has contained the current appName. So we should not throw an Exception to the user, it will confuse the user.

Just change one line code：
```java
// => org.apache.dubbo.registry.client.metadata.MetadataServiceNameMapping#map
public boolean map(URL url) {
    // ......
    if (StringUtils.isNotEmpty(oldConfigContent)) {
        boolean contains = StringUtils.isContains(oldConfigContent, appName);
        if (contains) {
            // From user's perspective, it means successful when the oldConfigContent has contained the current appName. So we should not throw an Exception to user, it will confuse user.
            succeeded = true;   // <= The only changes for this function.
            break;
        }
        newConfigContent = oldConfigContent + COMMA_SEPARATOR + appName;
    }
    // ......
}
```

What's more, the `map` function is only called by `ServiceConfig#export`，the changes are safe.
```java
// => org.apache.dubbo.config.ServiceConfig#export
protected void exported() {
    exported = true;
    List<URL> exportedURLs = this.getExportedUrls();
    exportedURLs.forEach(url -> {
        if (url.getParameters().containsKey(SERVICE_NAME_MAPPING_KEY)) {
            ServiceNameMapping serviceNameMapping = ServiceNameMapping.getDefaultExtension(getScopeModel());
            try {
                // try to register
                boolean succeeded = serviceNameMapping.map(url);
                if (succeeded) {
                    logger.info("Successfully registered interface application mapping for service " + url.getServiceKey());
                }
            } catch (Exception e) {
                logger.error("Failed register interface application mapping for service " + url.getServiceKey(), e);
            }
        }
    });
    onExported();
}
```

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataServiceNameMapping.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataServiceNameMapping.java
@@ -78,7 +78,7 @@ public class MetadataServiceNameMapping extends AbstractServiceNameMapping imple
             if (StringUtils.isNotEmpty(oldConfigContent)) {
                 boolean contains = StringUtils.isContains(oldConfigContent, appName);
                 if (contains) {
-                    // From user's perspective, it means successful when the oldConfigContent has contained the current appName. So we should not throw an Exception to user, it will confused user.
+                    // From the user's perspective, it means successful when the oldConfigContent has contained the current appName. So we should not throw an Exception to user, it will confuse the user.
                     succeeded = true;
                     break;
                 }
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataServiceNameMapping.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataServiceNameMapping.java
@@ -53,38 +53,40 @@ public class MetadataServiceNameMapping extends AbstractServiceNameMapping imple
 
     @Override
     public boolean map(URL url) {
-            if (CollectionUtils.isEmpty(applicationModel.getApplicationConfigManager().getMetadataConfigs())) {
-                return false;
-            }
-            String serviceInterface = url.getServiceInterface();
-            if (IGNORED_SERVICE_INTERFACES.contains(serviceInterface)) {
-                return false;
-            }
-            String registryCluster = getRegistryCluster(url);
-            MetadataReport metadataReport = metadataReportInstance.getMetadataReport(registryCluster);
+        if (CollectionUtils.isEmpty(applicationModel.getApplicationConfigManager().getMetadataConfigs())) {
+            return false;
+        }
+        String serviceInterface = url.getServiceInterface();
+        if (IGNORED_SERVICE_INTERFACES.contains(serviceInterface)) {
+            return false;
+        }
+        String registryCluster = getRegistryCluster(url);
+        MetadataReport metadataReport = metadataReportInstance.getMetadataReport(registryCluster);
 
-            String appName = applicationModel.getApplicationName();
-            if (metadataReport.registerServiceAppMapping(serviceInterface, appName, url)) {
-                // MetadataReport support directly register service-app mapping
-                return true;
-            }
+        String appName = applicationModel.getApplicationName();
+        if (metadataReport.registerServiceAppMapping(serviceInterface, appName, url)) {
+            // MetadataReport support directly register service-app mapping
+            return true;
+        }
 
-            int currentRetryTimes = 1;
-            boolean succeeded = false;
-            String newConfigContent = appName;
-            do {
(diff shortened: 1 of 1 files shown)


guohao (changes requested):
PLZ do not format all code ... Just modify the small part

EarthChen:
#8722 #8721

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataServiceNameMapping.java
```diff
@@ -53,38 +53,40 @@ public MetadataServiceNameMapping() {
...
+        boolean succeeded = false;
+        String newConfigContent = appName;
+        do {
+            ConfigItem configItem = metadataReport.getConfigItem(serviceInterface, DEFAULT_MAPPING_GROUP);
+            String oldConfigContent = configItem.getContent();
+            if (StringUtils.isNotEmpty(oldConfigContent)) {
+                boolean contains = StringUtils.isContains(oldConfigContent, appName);
+                if (contains) {
+                    // From the user's perspective, it means successful when the oldConfigContent has contained the current appName. So we should not throw an Exception to user, it will confuse the user.
+                    succeeded = true;
```
EarthChen, line 82: Is it better to just return true?

EarthChen (changes requested):

plusmancn:
> PLZ do not format all code ... Just modify the small part

@guohao  The original style is unreadable, see follow:

image

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataServiceNameMapping.java
plusmancn: According to the original function,  setting the `succeeded` variable may be better.
The latter code may do something base on `succeeded` status.

EarthChen: Based on existing code, a direct return is better

guohao (approved):

Issue #8695, opened 2021-09-06
org.apache.dubbo.remoting.exchange.support.DefaultFuture doWithCloseChannel will logRequestData


vio-lin:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version:  2.7.14-SNAPSHOT
* Operating System version: Win10
* Java version: JDK8

### Actual Result
打印出来的日志带有请求toString的内容
```
 public static void closeChannel(Channel channel) {
        for (Map.Entry<Long, Channel> entry : CHANNELS.entrySet()) {
            if (channel.equals(entry.getValue())) {
                DefaultFuture future = getFuture(entry.getKey());
                if (future != null && !future.isDone()) {
                    Response disconnectResponse = new Response(future.getId());
                    disconnectResponse.setStatus(Response.CHANNEL_INACTIVE);
                    disconnectResponse.setErrorMessage("Channel " +
                            channel +
                            " is inactive. Directly return the unFinished request : " +
                            // 这个方法会打印整个 request 对象toString的值
                            future.getRequest());
                    DefaultFuture.received(channel, disconnectResponse);
                }
            }
        }
    }
```
timeout打印的请求对象是不带有Request的
```
    private String getTimeoutMessage(boolean scan) {
        long nowTimestamp = System.currentTimeMillis();
        return (sent > 0 ? "Waiting server-side response timeout" : "Sending request timeout in client-side")
                + (scan ? " by scan timer" : "") + ". start time: "
                + (new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS").format(new Date(start))) + ", end time: "
                + (new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS").format(new Date(nowTimestamp))) + ","
                + (sent > 0 ? " client elapsed: " + (sent - start)
                + " ms, server elapsed: " + (nowTimestamp - sent)
                : " elapsed: " + (nowTimestamp - start)) + " ms, timeout: "
                + timeout + " ms, request: " + (logger.isDebugEnabled() ? request : getRequestWithoutData()) + ", channel: " + channel.getLocalAddress()
                + " -> " + channel.getRemoteAddress();
    }
```

这个是漏加了吧

wangchengming666:
[...]

PR #8739, opened 2021-09-09
Fix DefaultFuture turn off exception logging optimization for Channel…

candyaaa:

fix #8695 on 3.0

## Brief changelog

DefaultFuture turn off logging optimization for Channel.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/config/Environment.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/config/Environment.java
@@ -95,7 +95,7 @@ public class Environment extends LifecycleAdapter implements FrameworkExt {
                 path = CommonConstants.DEFAULT_DUBBO_MIGRATION_FILE;
             }
         }
-        this.localMigrationRule = ConfigUtils.loadMigrationRule(path);
+        this.localMigrationRule = ConfigUtils.loadMigrationRule(applicationModel.getClassLoaders(), path);
     }
 
     @Deprecated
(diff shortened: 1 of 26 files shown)


dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
```diff
@@ -251,11 +251,11 @@ private String getTimeoutMessage(boolean scan) {
                 + (sent > 0 ? " client elapsed: " + (sent - start)
                 + " ms, server elapsed: " + (nowTimestamp - sent)
                 : " elapsed: " + (nowTimestamp - start)) + " ms, timeout: "
-                + timeout + " ms, request: " + (logger.isDebugEnabled() ? request : getRequestWithoutData()) + ", channel: " + channel.getLocalAddress()
+                + timeout + " ms, request: " + (logger.isDebugEnabled() ? request : rinseRequestData(request)) + ", channel: " + channel.getLocalAddress()
                 + " -> " + channel.getRemoteAddress();
     }
 
-    private Request getRequestWithoutData() {
+    private static Request rinseRequestData(Request request) {
```
horizonzy, line 258: Here maintain original method name, `rinseRequestData` sounds like do some operation for request.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
```diff
@@ -251,11 +251,11 @@ private String getTimeoutMessage(boolean scan) {
...
         Request newRequest = request.copy();
```
horizonzy, line 259: And here we can optimize the logic. Add new method copy without data method in `Request`, avoid to copy data.

candyaaa:
@horizonzy I recommitted and found that there were redundant commit records, so "reset head" caused it to close，I have now resubmitted a PR，link：https://github.com/apache/dubbo/pull/8775

PR #8744, opened 2021-09-09
[3.0] fix issue #8695, exception log optimization

wangchengming666:
issue no #8695

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
@@ -147,7 +147,7 @@ public class DefaultFuture extends CompletableFuture<Object> {
                     disconnectResponse.setErrorMessage("Channel " +
                             channel +
                             " is inactive. Directly return the unFinished request : " +
-                            future.getRequest());
+                            getRequestWithoutData(future.getRequest()));
                     DefaultFuture.received(channel, disconnectResponse);
                 }
             }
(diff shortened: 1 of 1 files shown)


dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
```diff
@@ -251,11 +251,11 @@ private String getTimeoutMessage(boolean scan) {
                 + (sent > 0 ? " client elapsed: " + (sent - start)
                 + " ms, server elapsed: " + (nowTimestamp - sent)
                 : " elapsed: " + (nowTimestamp - start)) + " ms, timeout: "
-                + timeout + " ms, request: " + (logger.isDebugEnabled() ? request : getRequestWithoutData()) + ", channel: " + channel.getLocalAddress()
+                + timeout + " ms, request: " + (logger.isDebugEnabled() ? request : getRequestWithoutData(request)) + ", channel: " + channel.getLocalAddress()
                 + " -> " + channel.getRemoteAddress();
     }
 
-    private Request getRequestWithoutData() {
+    private static Request getRequestWithoutData(Request request) {
```
plusmancn, line 258: Don't use static.

Static results in many problems during the multi-instance refactor.

changfubai: would you mind explaining it? I think this useage is not perfect, but I don't understand the meaning of `Static results in many problems during the multi-instance refactor.`
please help me, thanks.

wangchengming666: I think private static method it's ok.

AlbumenJ:
fixed in #8775

PR #8758, opened 2021-09-09
add peakEwmaLoadBalance

24kpure:
I am disappointment with shortestResponseLoadBalance，though  it  is better than others in latency spike. However, it’s slow for it  depends on average.I am looking forwad to a sensitive loadbance.

PeakEwmaLoadBalance is designed to converge quickly when encountering slow endpoints.
It is quick to react to latency spikes recovering only cautiously.Peak EWMA takes history into account,so that slow behavior is penalized relative to the supplied `decayTime`.

Inspiration drawn from:[finagle](https://github.com/twitter/finagle/blob/1bc837c4feafc0096e43c0e98516a8e1c50c4421/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/PeakEwma.scala)

## note
We will have a try first and  feed back to the community in time.

## Brief changelog
-  calculate cost by exponential moving averages
-  add decayTime
-  ignore invoker‘s weight

## Reference
[finagle](https://github.com/twitter/finagle/blob/1bc837c4feafc0096e43c0e98516a8e1c50c4421/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/PeakEwma.scala)
[apisix](https://github.com/apache/apisix/blob/131998df4f44720b7ddf0424f4d381efb689d4ce/apisix/balancer/ewma.lua
)
[paper](http://www.eckner.com/papers/Algorithms%20for%20Unevenly%20Spaced%20Time%20Series.pdf)

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/Constants.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/Constants.java
@@ -139,4 +139,9 @@ public interface Constants {
      * The key of shortestResponseSlidePeriod
      */
     String SHORTEST_RESPONSE_SLIDE_PERIOD = "shortestResponseSlidePeriod";
+
+    /**
+     * The key of peakEwmaDecayTime
+     */
+    String PEAK_EWMA_DECAY_TIME = "peakEwmaDecayTime";
 }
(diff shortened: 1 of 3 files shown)


AlbumenJ:
Please move this feature to dubbo-spi-extensions

24kpure:
That's a good idea. Dubbo is gradually weakening the concepts of load balancing and routing in 3.0,but I still want to fight for it. If we don't use service mesh for traffic distribution, a new smart loadBalance will be good for Dubbo. The existing loadBalance can not be adjusted adaptively and peakEwma  is a good choice.
I'd like to ask for your confirmation.If you stick to your ideas, I will move this feature to dubbo-spi-extensions. Thanks for your advice.

24kpure:
I moved this feature to [dubbo-spi-extensions](https://github.com/apache/dubbo-spi-extensions/pull/68)

PR #8772, opened 2021-09-11
Use MapUtils instead of AttachmentsAdapter

manzhizhen:
In order to realize the conversion from Map<String, Object> to Map<String, String>, the implementation of AttachmentsAdapter is too complicated, and the inner class of the class is also used. For the sake of simplicity, the tool class MapUtils with less code is used to achieve the same function.

为了实现Map<String, Object>到Map<String, String>的转换，AttachmentsAdapter实现的太复杂，还使用类内部类。为了简单起见，使用更少代码的工具类MapUtils来实现一样的功能。

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AttachmentsAdapter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AttachmentsAdapter.java
@@ -57,10 +57,11 @@ public class AttachmentsAdapter {
         }
 
         private String convert(Object obj) {
-            if (obj instanceof String) {
-                return (String) obj;
+            if (obj == null) {
+                return null;
+            } else {
+                return obj.toString();
             }
-            return null; // or JSON.toString(obj);
         }
 
         @Override
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AttachmentsAdapter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AttachmentsAdapter.java
@@ -22,12 +22,18 @@ import java.util.Map;
 /**
  * This class provides map adapters to support attachments in RpcContext, Invocation and Result switch from
  * <String, String> to <String, Object>
+ *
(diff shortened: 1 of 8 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AttachmentsAdapter.java
```diff
@@ -22,12 +22,18 @@
...
+ *
  */
+@Deprecated
 public class AttachmentsAdapter {
 
+    @Deprecated
     public static class ObjectToStringMap extends HashMap<String, String> {
         private Map<String, Object> attachments;
 
+        @Deprecated
```
horizonzy, line 36: this file method `convert` has problem, if object type is not string, it will return null, cound you fix it conveniently

manzhizhen: Hmm, the new method takes this into consideration, calling Object#toString() directly。

horizonzy (approved):
lgtm.

horizonzy:
Now branch 3.0 is the default, if you want 3.0 need this change also, pick it to 3.0

wuwen5:
This modification breaks compatibility ☹️  https://github.com/apache/skywalking/issues/8415

wu-sheng:
We are surprised to receive a compatibility bug due to a minor version change, and include an implementation level change.

I am following changes at https://github.com/apache/skywalking-java/pull/88, but it is better Dubbo community could provide some suggestions. And hope this kind of change would not be in patch version in the future :(

horizonzy:
> `openzipkin/brave` is also affected
> 
> ```java
> Map<String, String> attachments = RpcContext.getContext().getAttachments();
> DubboClientRequest clientRequest = new DubboClientRequest(invoker, invocation, attachments);
> ```
> 
> If the user's business code is used, it may be fatal.
> 
> @chickenlj @horizonzy @CrazyHZM @manzhizhen

yep, the code has breaking change, It will influence the user who modify the attachments by operate `RpcContext.getContext().getAttachments()`. 
It's a problem indeed, it shouldn't change original logic.
sorry for my careless review.

PR #8784, opened 2021-09-13
Inefficient Usages of Java Collections

DittoTool:
Hi,

We find that there are several inefficient usages of Java Collections:

1. The contains method is invoked upon a list object in a loop. We recommend replacing it with a HashSet.
2. Random access can occur at several LinkedList objects, which run in linear time complexity. We recommend replacing them with Arraylist objects.
3. There is no iteration occurring upon a LinkedHashMap and LinkedHashSet, thus the insertion order does not matter. We recommend replacing it with a HashMap and HashSet.
4. ArrayList is inserted before an iteration, while multiple memory reallocation might occur when the size of the list exceeds its capacity. We recommend replacing it with a LinkedList.

We discovered the above inefficient usage of containers by our tool Ditto. The patch is submitted. Could you please check and accept it? We have tested the patch on our PC. The patched program works well.

Bests

Ditto


mercyblitz:
Please fix the checkstyle issue first, thanks.

PR #8824, opened 2021-09-16
[3.0] Refactor application/module deploy processing and events

kylixs:
 * Refactor application/module deploy processing and events,  more clearer state.
 * Start and stop the bound dubbo module according to the spring context event, instead of controlling whole dubbo application.
 * Remove `DubboBootstrap` from `dubbo-config-spring` ,  add `DubboSpringInitializationCustomizerHolder` for register customizers by programing.
 * Split `DubboShutdownHook` and `ShutdownHookCallback`, adjust call chains.
 * Improve `SpringExtensionInjector` initialization


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/DubboShutdownHook.java
```diff
@@ -81,7 +85,11 @@ public void register() {
      */
     public void unregister() {
         if (registered.compareAndSet(true, false)) {
-            Runtime.getRuntime().removeShutdownHook(this);
+            try {
+                Runtime.getRuntime().removeShutdownHook(this);
```
zrlw, line 89: ShutdownHookCallbacks.callback() will be called at Runtime.getRuntime().removeShutdownHook(this):
```
##[error]Exception in thread "DubboShutdownHook" java.lang.RuntimeException: java.lang.IllegalStateException: Shutdown in progress
 	at org.apache.dubbo.common.function.ThrowableAction.execute(ThrowableAction.java:48)
 	at org.apache.dubbo.common.lang.ShutdownHookCallbacks.lambda$callback$0(ShutdownHookCallbacks.java:72)
 	at java.lang.Iterable.forEach(Iterable.java:75)
 	at org.apache.dubbo.common.lang.ShutdownHookCallbacks.callback(ShutdownHookCallbacks.java:72)
 	at org.apache.dubbo.config.DubboShutdownHook.callback(DubboShutdownHook.java:89)
```
we might add try-catch in ShutdownHookCallbacks callback() method like this, otherwise the callbacks in getCallbacks() after the callback which throw the exception will lost running opptunity:
```
    private static final Logger logger = LoggerFactory.getLogger(ShutdownHookCallbacks.class);
    public void callback() {
        getCallbacks().forEach(callback -> {
            try {
                execute(callback::callback);
            } catch (Exception e) {
                logger.warn("callback exception.", e);
            }
        });
    }
```

PR #8840, opened 2021-09-17
Add test case for ZookeeperMetadataReportTest

BurningCN:

1.test case
2.fix typo
3.remove unused param


dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/report/MetadataReport.java
```diff
@@ -45,7 +45,7 @@
     default void publishAppMetadata(SubscriberMetadataIdentifier identifier, MetadataInfo metadataInfo) {
     }
 
-    default MetadataInfo getAppMetadata(SubscriberMetadataIdentifier identifier, Map<String, String> instanceMetadata) {
+    default MetadataInfo getAppMetadata(SubscriberMetadataIdentifier identifier) {
```
AlbumenJ, line 48: Please do not change this method signature. `instanceMetadata` contains some instance releated properties which will take effect when getting metadata

AlbumenJ (changes requested):

AlbumenJ (approved):

PR #8876, opened 2021-09-22
[3.0] socks proxy filtering local address

kylixs:

socks proxy filtering local address

### DIFF
--- a/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyClient.java
+++ b/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyClient.java
@@ -32,6 +32,7 @@ import org.apache.dubbo.common.config.ConfigurationUtils;
 import org.apache.dubbo.common.logger.Logger;
 import org.apache.dubbo.common.logger.LoggerFactory;
 import org.apache.dubbo.common.utils.NetUtils;
+import org.apache.dubbo.common.utils.StringUtils;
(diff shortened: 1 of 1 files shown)


dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyClient.java
```diff
@@ -127,6 +128,16 @@ protected void initChannel(SocketChannel ch) throws Exception {
         });
     }
 
+    private boolean isFilteredAddress(String host) {
+        // filter local address
+        if (StringUtils.isEquals(NetUtils.getLocalHost(), host) ||
+            "localhost".equals(host) ||
```
changfubai, line 134: 这个有常量可以用的吧~  
另外，用StringUtils.equalsAny()就可以了，不用equals两次

kylixs: I found this method to check whether is localhost: `org.apache.dubbo.common.utils.NetUtils.isLocalHost()`

`StringUtils.equalsAny()`  is essentially the same, checking each string once for equality.
```java
    public static boolean equalsAny(final CharSequence string, final CharSequence... searchStrings) {
        if (ArrayUtils.isNotEmpty(searchStrings)) {
            for (final CharSequence next : searchStrings) {
                if (equals(string, next)) {
                    return true;
                }
            }
        }
        return false;
    }
```

PR #8879, opened 2021-09-22
[3.0] Fix when subscribing, the check configuration section is unavailable(#8841)

### LINKED ISSUE #8841: When subscribing, the check configuration section is unavailable
- [ ] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 3.0

When the subscription is initiated, check is forcibly set to false, which should be determined by the upper-level configuration
For example：
Mandatory set the check value to false in ZookeeperRegistry

Expectation: If check is an interface-level service discovery, it can be configured for different interfaces. If it is an application-level service discovery, it can be configured for an application.

candyaaa:

Fix #8841 on 3.0


BurningCN (approved):
LGTM

PR #8881, opened 2021-09-23
[3.0] set client to null at LazyConnectExchangeClient close methods

zrlw:
#8880
* set the client field of LazyConnectExchangeClient  to null at the end of close() and close(int timeout) methods
* move close status setting from HeaderExchangeChannel close(timeout) to close()


wangchengming666 (approved):
LGTM

PR #8913, opened 2021-09-24
Add unit test for StandardMetadataServiceURLBuilder

BurningCN:
Add unit test for StandardMetadataServiceURLBuilder

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/test/java/org/apache/dubbo/registry/client/metadata/StandardMetadataServiceURLBuilderTest.java
+++ b/dubbo-registry/dubbo-registry-api/src/test/java/org/apache/dubbo/registry/client/metadata/StandardMetadataServiceURLBuilderTest.java
@@ -21,6 +21,8 @@ import org.apache.dubbo.config.ApplicationConfig;
 import org.apache.dubbo.metadata.MetadataService;
 import org.apache.dubbo.registry.client.DefaultServiceInstance;
 import org.apache.dubbo.rpc.model.ApplicationModel;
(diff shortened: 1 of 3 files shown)


dubbo-registry/dubbo-registry-api/src/test/java/org/apache/dubbo/registry/client/metadata/StandardMetadataServiceURLBuilderTest.java
```diff
@@ -0,0 +1,85 @@
...
+    @AfterAll
+    public static void clearUp() {
+        ApplicationModel.reset();
+    }
+
+    @Test
+    public void testBuild() {
+        // test generateUrlWithoutMetadata
+        List<URL> urls = builder.build(new DefaultServiceInstance("test", "127.0.0.1", 8080, ApplicationModel.defaultModel()));
+        assertEquals(1, urls.size());
```
DiegoKrupitza, line 57: You could use Hamcrest here to test the size in a more `fluent` way. 
For example here you can write: `assertThat(urls, hasSize(1));`

In general this is not really necessary but since Hamcrest is in the project dependencies it would be nice to use it to make Testing more pleasing for the eye (easier reading and writing)

DiegoKrupitza (commented):
All in all it looks good to me. I would suggest to write your test with Hamcrest. This will make the tests easier to read. 
I suggest this since Hamcrest is already used in few tests in this project and already in the projects dependencies.

changfubai (approved):
LGTM.

PR #8924, opened 2021-09-26
An optimization of generic invocation in gson format

goodjava:
+ An optimization of generic invocation in gson format


wangchengming666 (approved):
LGTM

PR #8930, opened 2021-09-27
[3.0] catch handler.caught() throwable of MultiMessageHandler

zrlw:
#8929


changfubai (approved):
LGTM.

PR #8931, opened 2021-09-27
[master] catch handler.caught() throwable of MultiMessageHandler

zrlw:
#8929
* catch handler.caught() throwable of MultiMessageHandler
* reset ChannelHandlers after each testing of HeartbeatHandlerTest


changfubai (approved):
LGTM.

PR #9003, opened 2021-10-09
[3.0] close dynamic configuration to fix configCenter connection leak

zrlw:
#8993
* close config center connection at destroyDynamicConfigurations()


zrlw:
> The pr #9033 is also fixing zk connections leaks, please check it. This PR seems unnecessary.
let zkClient keep alive will cause more integration testing failure because curator event processing will be blocked by huge reconnection trying failure events which triggered by the send thread of these unclosed clients.

PR #9033, opened 2021-10-13
[3.0] Manage global resources and executor services, fix zk client connections

kylixs:
* Improve and destroy executor services
* Support manage global static resources, including `EventLoopGroup`, `HashedWheelTimer`, etc.
* Improve application/module deploying
* Fix zk client connection leaks and early closing problem, change `ZookeeperTransporter` to application scope, see https://github.com/apache/dubbo/pull/9033#discussion_r728599330
* Destroy nacos client and `MetadataReportFactory`
* Support check unclosed threads in unit tests by `mvn test -DcheckThreads=true`, see `org.apache.dubbo.test.check.DubboTestChecker`
* Sort methods of adaptive classes in `dubbo-native`

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,11 +1,14 @@
 ---
-name: Dubbo issue report template
-about: If you would like to report a issue to Dubbo, please use this template.
+name: Bug Report
+about: Report a bug in Dubbo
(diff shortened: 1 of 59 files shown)


kylixs:
> working

Most of the problems have been solved, because the processing flow has changed greatly, some of the previous repairs are unnecessary, and if there are other problems, they need to be dealt with again.

PR #9111, opened 2021-10-24
[3.0] Optimization code of ClassUtils

stone-98:

Optimization code of ClassUtils


stone-98:
@AlbumenJ I think we can provide something like flip in the Converter interface.Can use it in a lot of places.

PR #9112, opened 2021-10-24
[3.0] Add connectivity validation, router snapshot

AlbumenJ:

1. Refactor BitList implementation. Remove RoaringBitmap dependency. Extend AbstractList, implement most of the method implemented from List. `BitList.java` `BitListTest.java`
2. Refactor the collection using in Directory of invokers to BitList to improve performance. `AbstractDirectory.java` `RouterChain.java`
3. Add valid, invalid, disabled invokers support. `AbstractDirectory.java`
4. Add connectivity validation to AbstractClusterInvoker, automatically add invaild invokers to Directory. `AbstractClusterInvoker.java`
5. Add reconnect check support for Directory. `AbstractDirectory.java`
6. Add router snapshot print when route result is empty support. `RouterChain.java` `RouterSnapshotNode.java`


chickenlj (approved):
From a functional point of view, I think this patch is fine.

But the `addDisabledInvokers` `removeDisabledInvokers`  operations added on the Directory seems a little strange from the design point. I personally think it will be better if Directory itself can take care of the status of all invokers inside while the ClusterInvoker can only interact with the `list` method and maintain its reselected list or choose to do retry or not in case of RPC failures.

AlbumenJ:
> But the addDisabledInvokers removeDisabledInvokers operations added on the Directory seems a little strange from the design point. I personally think it will be better if Directory itself can take care of the status of all invokers inside while the ClusterInvoker can only interact with the list method and maintain its reselected list or choose to do retry or not in case of RPC failures.

`addDisabledInvokers` and `removeDisabledInvokers` are designed for offline notification. For example, server can notify all of the clients that server will be shutdown in moments, and clients should not continue invoke this server any more.

PR #9145, opened 2021-10-27
change way to determine the port that metadata service will listen on

chickenlj:
The protocol will always be Dubbo while the port is determined by following below steps:

1. Check explicitly specified port
2. Try to find currently working Dubbo protocol port by terating protocols 
3. use random port

### DIFF
--- a/dubbo-config/dubbo-config-api/pom.xml
+++ b/dubbo-config/dubbo-config-api/pom.xml
@@ -68,6 +68,11 @@
             <version>${project.parent.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.dubbo</groupId>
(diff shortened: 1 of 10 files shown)


dubbo-config/dubbo-config-api/pom.xml
```diff
@@ -68,6 +68,11 @@
             <version>${project.parent.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.dubbo</groupId>
+            <artifactId>dubbo-rpc-triple</artifactId>
+            <version>${project.parent.version}</version>
+        </dependency>
```
AlbumenJ, line 75: What is the purpose of adding this dependency? Can it be changed to the test scope?

AlbumenJ (changes requested):

PR #9223, opened 2021-11-06
Fix Compressor's getCompressor method

BurningCN:

IdentityCompressor and Compressor#NONE obtained through SPI are not the same object, causing OutboundTransportObserver#calcCompressFlag to perform IdentityCompressor.NONE.equals(compressor) to return false

image


wangchengming666 (approved):
LGTM

EarthChen (approved):
LGTM.

Not an object may seem unimportant, but it is still better after modification

guohao (approved):
LGTM

PR #9228, opened 2021-11-07
[3.0] Refactor State Router

AlbumenJ:

- Remove RouterChain State Router Cache.
- Refactor State Router Interface. Remove cache related methods. Except that `List` is changed to `BitList`, all other methods are the same as `Router`.
- Remove TagStaticStateRouter, TagDynamicStateRouter
- Move `tag`, `condition`, `service`, `app`, `file`, `mock` and `script` router to state router
- Add router message
- BREAKING CHANGE: Remove STATE_ROUTER_KEY. Use ROUTER_KEY instead


guohao (approved):
LGTM

PR #9260, opened 2021-11-12
Fix the problem of InvokeTelnet that user use [invoke sayHello(xxx)] without executing the cd command first

BurningCN:
If the user directly uses the invocation method without the service name prefix, that is, `invoke sayHello (xxxx)`, and has not executed the `cd xxx `command beforehand, the NPE problem will occur


wangchengming666 (approved):
LGTM

PR #9265, opened 2021-11-13
Including setup method to clean state between tests

odrepa:
In InternalThreadLocalTest, testSize fails when run after testSetAndGet, testRemove, or testRemoveAll in the same test class. I found that running InternalThreadLocalMap.remove() before testSize ensures the starting state is correct such that testSize can pass even when run after those three tests. I introduced the setup method to ensure this resetting logic is run before each test. Please let me know if there is a more preferred way to address this issue.

## Brief changelog
Including a setup method that sets the state through InternalThreadLocalMap.remove().

## Checklist

### DIFF
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/threadlocal/InternalThreadLocalTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/threadlocal/InternalThreadLocalTest.java
@@ -18,6 +18,7 @@
 package org.apache.dubbo.common.threadlocal;
 
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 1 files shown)


BurningCN:
hi, why my local test did not reproduce
image

odrepa:
Hi,

Sorry for the late response. On my side, what I could to reproduce the failure in the specific order is to add the annotation "@org.junit.jupiter.api.TestMethodOrder(org.junit.jupiter.api.MethodOrderer.Alphanumeric.class)" to the test class (as to force alphabetical test method order, which would put testSize at the end), leading to testSize to fail. The failure message I would see is:

```
org.opentest4j.AssertionFailedError: size method is wrong! ==> expected: <1> but was: <4>
        at org.apache.dubbo.common.threadlocal.InternalThreadLocalTest.testSize(InternalThreadLocalTest.java:78)
```

BurningCN (approved):
LGTM

BurningCN:
Without the `TestMethodOrder` annotation, my side has always been normal.
But as far as this `pr` is concerned, I think it’s good to do a pre-cleaning operation before each test.

odrepa:
As per the suggestion, I've switched to using `@AfterEach` to clean up instead of setting up.

PR #9273, opened 2021-11-14
[3.0] fix #9272, add thread pool info to DataStore

### LINKED ISSUE #9272: [3.0] provider开启MetricsFilter，但是在dubbo-admin获取不到线程池的统计数据
provider开启MetricsFilter，但是在dubbo-admin获取不到线程池的统计数据

原因如下
`    
private List<MetricObject> getThreadPoolMessage() {
        DataStore dataStore = extensionAccessor.getExtensionLoader(DataStore.class).getDefaultExtension();
        Map<String, Object> executors = dataStore.get(EXECUTOR_SERVICE_COMPONENT_KEY);
`

解决方案：在创建的时候将线程池信息放到`DataStore`中

wangchengming666:
add thread pool info to DataStore


dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/manager/DefaultExecutorRepository.java
```diff
@@ -352,7 +354,11 @@ public ScheduledExecutorService getSharedScheduledExecutor() {
     }
 
     private ExecutorService createExecutor(URL url) {
-        return (ExecutorService) extensionAccessor.getExtensionLoader(ThreadPool.class).getAdaptiveExtension().getExecutor(url);
+        ExecutorService executorService = (ExecutorService) ExtensionLoader.getExtensionLoader(ThreadPool.class).getAdaptiveExtension().getExecutor(url);
```
BurningCN, line 357: The original code should be used here
`ExecutorService executorService = (ExecutorService) extensionAccessor.getExtensionLoader(ThreadPool.class).getAdaptiveExtension().getExecutor(url);`

dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/manager/DefaultExecutorRepository.java
```diff
@@ -352,7 +354,11 @@ public ScheduledExecutorService getSharedScheduledExecutor() {
...
+
+        DataStore dataStore = ExtensionLoader.getExtensionLoader(DataStore.class).getDefaultExtension();
+        dataStore.put(EXECUTOR_SERVICE_COMPONENT_KEY, Integer.toString(url.getPort()), executorService);
```
BurningCN, line 360: In the `destroyAll` method, the DataSource should be cleaned up

dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/manager/DefaultExecutorRepository.java
wangchengming666: doen, thanks

zrlw: all ```ExtensionLoader.getExtensionLoader``` might be changed to  ```extensionAccessor.getExtensionLoader``` , otherwise it will throws exception at destroyAll.

dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/manager/DefaultExecutorRepository.java
```diff
@@ -354,10 +346,11 @@ public ScheduledExecutorService getSharedScheduledExecutor() {
     }
 
     private ExecutorService createExecutor(URL url) {
-        ExecutorService executorService = (ExecutorService) ExtensionLoader.getExtensionLoader(ThreadPool.class).getAdaptiveExtension().getExecutor(url);
+        ExecutorService executorService = (ExecutorService) extensionAccessor.getExtensionLoader(ThreadPool.class).getAdaptiveExtension().getExecutor(url);
 
         DataStore dataStore = ExtensionLoader.getExtensionLoader(DataStore.class).getDefaultExtension();
```
zrlw, line 351: ```extensionAccessor```'s scope might not be default application, ```executorService``` created by ```extensionAccessor``` will be destroyed with ```extensionAccessor```'s application by DefaultApplicationDeployer.destroyExecutorRepository().
```dataStore```'s scope might be same as  ```executorService``` to avoid multiple applications problem.

**Comment by zrlw** [comment]
app A and app B will create different executors even they use the same URLs, the key of dataStore should include application name.

**Comment by zrlw** [comment]
if dataStore includes consumer apps, the key of dataStore should include full url as the port of different server might be same.

**Comment by zrlw** [comment]
there are one comment at createExecutorIfAbsent():
```
// Consumer's executor is sharing globally, key=Integer.MAX_VALUE. Provider's executor is sharing by protocol.
Integer portKey = CONSUMER_SIDE.equalsIgnoreCase(url.getParameter(SIDE_KEY)) ? Integer.MAX_VALUE : url.getPort();
```
two providers might share one executor by url.getPort()?
how to remove it from dataStore at destroyAll? add reference count and remove it only when reference count == 0?

zrlw:
another problem:
getExecutor() will remove executor which is shutdown or terminated and use sharedExecutor instead of it. does dataStore need to remove it or not?

chickenlj:
> does dubbo-admin need to put it in dataStore?

I have the same question here. Are we doing this only for Admin or Metrics being able to get the executor information from DataStore?

chickenlj:
Reopen if necessary

PR #9317, opened 2021-11-24
Fix the problem that Curator[5]ZookeeperClient does not handle RegistryConfig#session value

BurningCN:
Fix the problem that Curator[5]ZookeeperClient does not handle RegistryConfig#session value


wangchengming666 (approved):
lgtm

PR #9329, opened 2021-11-29
Code adjustment and optimization of config-api/common module

BurningCN:

Code adjustment and optimization of config-api/common module

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ReferenceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ReferenceConfig.java
@@ -23,6 +23,7 @@ import org.apache.dubbo.common.extension.ExtensionLoader;
 import org.apache.dubbo.common.logger.Logger;
 import org.apache.dubbo.common.logger.LoggerFactory;
 import org.apache.dubbo.common.url.component.ServiceConfigURL;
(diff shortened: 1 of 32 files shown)


wangchengming666 (approved):
lgtm

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ReferenceConfig.java
```diff
@@ -241,7 +242,6 @@ public synchronized void destroy() {
             ModuleServiceRepository repository = getScopeModel().getServiceRepository();
             repository.unregisterConsumer(consumerModel);
         }
-        getScopeModel().getConfigManager().removeConfig(this);
```
AlbumenJ, line 244: pls desc why remove this

AlbumenJ (changes requested):

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ReferenceConfig.java
BurningCN: The first step of the destroy method will call super.destroy, which is `ReferenceConfigBase#destroy`, which has already done removeConfig

AlbumenJ (approved):

PR #9333, opened 2021-11-30
Use existing methods to generate serviceKey for SimpleReferenceCache

BurningCN:

Use existing methods to generate serviceKey for SimpleReferenceCache 

1.Use existing methods to generate serviceKey for SimpleReferenceCache
2.add test case for ServiceInstanceHostPortCustomizer
3.Remove the toString method of AbstractConfig subclass
4.Add unit test for ConfigScopeModelInitializer


wangchengming666 (approved):
lgtm

PR #9372, opened 2021-12-09
Fix the logic of removing REFER and EXPORT attributes from URL

BurningCN:
Fix the logic of removing REFER and EXPORT attributes from URL


dubbo-configcenter/dubbo-configcenter-zookeeper/src/main/java/org/apache/dubbo/configcenter/support/zookeeper/ZookeeperDynamicConfiguration.java
```diff
@@ -40,27 +40,22 @@
...
     private CacheListener cacheListener;
-    private URL url;
     private static final int DEFAULT_ZK_EXECUTOR_THREADS_NUM = 1;
     private static final int DEFAULT_QUEUE = 10000;
     private static final Long THREAD_KEEP_ALIVE_TIME = 0L;
 
     ZookeeperDynamicConfiguration(URL url, ZookeeperTransporter zookeeperTransporter) {
         super(url);
-        this.url = url;
-        rootPath = getRootPath(url);
```
BurningCN, line 56: The parent class already has a rootPath, so we don’t need to get it again
The url is not used so it is deleted

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/ServiceDiscoveryRegistry.java
```diff
@@ -79,13 +79,10 @@
     /* apps - listener */
     private final Map<String, ServiceInstancesChangedListener> serviceListeners = new ConcurrentHashMap<>();
 
-    private RegistryManager registryManager;
-
     public ServiceDiscoveryRegistry(URL registryURL) {
         super(registryURL);
         this.serviceDiscovery = createServiceDiscovery(registryURL);
         this.writableMetadataService = WritableMetadataService.getDefaultExtension(registryURL.getScopeModel());
-        this.registryManager = registryURL.getOrDefaultApplicationModel().getBeanFactory().getBean(RegistryManager.class);
```
BurningCN, line 88: The parent class already has this attribute, so we don’t need to get it again

wangchengming666 (approved):
LGTM

PR #9381, opened 2021-12-11
Adjust the logic of Deployer to handle concurrent initialization

BurningCN:
- Replace some AtomicBoolean types with Boolean
- Fix the error of setting the initialized attribute under abnormal conditions of ConfigCenterConfig
- Extract constants


wangchengming666 (approved):
lgtm

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/deploy/DefaultApplicationDeployer.java
```diff
@@ -184,14 +183,8 @@ public ReferenceCache getReferenceCache() {
...
-        if (initialized.get()) {
-            return;
-        }
         // Ensure that the initialization is completed when concurrent calls
-        synchronized (startLock) {
-            if (initialized.get()) {
-                return;
-            }
+        if (initialized.compareAndSet(false, true)) {
             // register shutdown hook
```
kylixs, line 188: The purpose of the `startLock` is:
> Ensure that the initialization is completed when concurrent calls

If change to `compareAndSet`, it will not block until the `DefaultApplicationDeployer` was initialized, it is unsafe.

Possible case:
1. The first call, start to initialize, but not complete
2. In the second call, check that it has been initialized, skip it, and proceed to the following startup. At this time, the initialization may not be completed, resulting in an error

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/deploy/DefaultApplicationDeployer.java
```diff
@@ -641,11 +632,7 @@ public void prepareApplicationInstance() {
     }
 
     private void prepareInternalModule() {
-        synchronized (internalModuleLock) {
-            if (!hasPreparedInternalModule.compareAndSet(false, true)) {
-                return;
-            }
-
```
kylixs, line 648: The `internalModuleLock` is used to ensure prepare internal module is finished

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/deploy/DefaultModuleDeployer.java
```diff
@@ -92,14 +92,8 @@ public DefaultModuleDeployer(ModuleModel moduleModel) {
...
-        if (initialized.get()) {
-            return;
-        }
         // Ensure that the initialization is completed when concurrent calls
-        synchronized (this) {
-            if (initialized.get()) {
-                return;
-            }
+        if (initialized.compareAndSet(false, true)) {
             loadConfigs();
```
kylixs, line 97: see `DefaultApplicationDeployer.initialize()`

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/deploy/DefaultApplicationDeployer.java
```diff
@@ -691,7 +694,7 @@ private DynamicConfiguration prepareEnvironment(ConfigCenterConfig configCenter)
             } catch (Exception e) {
                 if (!configCenter.isCheck()) {
                     logger.warn("The configuration center failed to initialize", e);
-                    configCenter.checkOrUpdateInitialized(false);
+                    configCenter.setInitialized(false);
```
BurningCN, line 697: As shown below, sequence number 1 will change `initialized` from false to true, but sequence number 2 cannot set `initialized` to false, so adjust to `configCenter.setInitialized(false);`
image

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/deploy/DefaultApplicationDeployer.java
BurningCN: thanks, i got it.
Adjust the type of initialized to volatile boolean

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/deploy/DefaultModuleDeployer.java
BurningCN: Adjust the type of initialized to volatile boolean

dubbo-common/src/main/java/org/apache/dubbo/config/ConfigCenterConfig.java
```diff
@@ -149,7 +146,7 @@ public URL toUrl() {
         if (StringUtils.isEmpty(address)) {
             address = ANYHOST_VALUE;
         }
-        map.put(PATH_KEY, ConfigCenterConfig.class.getSimpleName());
+        map.put(PATH_KEY, ConfigCenterConfig.class.getName());
```
kylixs, line 149: why change PATH_KEY value to full class name ?

BurningCN: I see that the path of `MetadataReport`'s url is full class name, and `RegistryService` is also like this
image
image

kylixs: @AlbumenJ  PTAL the `PATH_KEY`

AlbumenJ (approved):

kylixs (approved):

andrewchoi5 (approved):

PR #9391, opened 2021-12-12
change getting order of wrapper class from @Activate to @Wrapper

HetaoWang:
1. add order() method in @Wrapper annotation 
2. change getting order of wrapper class from @Activate to @Wrapper
3. backward compatibility


dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
```diff
@@ -771,13 +772,15 @@ private T createExtension(String name, boolean wrap) {
...
-                        }
+                    wrapperClassesList = wrapperClassesList.stream().filter(wrapperClz -> {
+                        Wrapper wrapper = wrapperClz.getAnnotation(Wrapper.class);
+                        return (wrapper == null) ||
+                            ((ArrayUtils.isEmpty(wrapper.matches()) || ArrayUtils.contains(wrapper.matches(), name)) &&
+                                !ArrayUtils.contains(wrapper.mismatches(), name));
+                    }).collect(Collectors.toList());
+                    for (Class<?> wrapperClz : wrapperClassesList) {
+                        instance = injectExtension((T) wrapperClz.getConstructor(type).newInstance(instance));
+                        instance = postProcessAfterInitialization(instance, name);
```
chickenlj, line 783: I think streaming is not a good option used here. Normal for iterating can be more memory efficient while clear enough.

HetaoWang: got it

chickenlj (approved):
LGTM.

Issue #1508, opened 2018-03-26
feature required: dubbo should consider empty protection.


HouZhiHouJue:
when register center lost all provider instance , it's doesn't mean that the consumers are all disconnected from the providers,so the method  refreshInvoker(List<URL> invokerUrls) ; in registrydirectory is not so fault tolerance.the method code is :

` private void refreshInvoker(List<URL> invokerUrls) {

        if (invokerUrls != null && invokerUrls.size() == 1 && invokerUrls.get(0) != null

                && Constants.EMPTY_PROTOCOL.equals(invokerUrls.get(0).getProtocol())) {

            this.forbidden = true; // Forbid to access

            this.methodInvokerMap = null; // Set the method invoker map to null

            destroyAllInvokers(); // Close all invokers

        } else {
         ......
`

here when invokerUrls  is empty ,all the invokers are destryed, but in fact,ignore it  will be more fault tolerance.

beiwei30:
dubbo already considered this scenario:

```java
            if (invokerUrls.isEmpty() && this.cachedInvokerUrls != null) {
                invokerUrls.addAll(this.cachedInvokerUrls);
            } else {
                this.cachedInvokerUrls = new HashSet<URL>();
                this.cachedInvokerUrls.addAll(invokerUrls);//Cached invoker urls, convenient for comparison
            }
            if (invokerUrls.isEmpty()) {
                return;
            }
```

if the addresses from the register center is empty, then the cached address list `cachedInvokerUrls` will be used. 

feel free to reopen this issue if you do not agree.

HouZhiHouJue:
but when register center lost all provider instance , invokerUrls will not be empty,it has one element and the protocol is Constants.EMPTY_PROTOCOL,so the code may not be right.

beiwei30:
with which registry center did you experience the problem? we may need to discuss this further.

HouZhiHouJue:
` ChildListener zkListener = listeners.get(listener);

                    if (zkListener == null) {

                        listeners.putIfAbsent(listener, new ChildListener() {

                            public void childChanged(String parentPath, List<String> currentChilds) {

                                ZookeeperRegistry.this.notify(url, listener, toUrlsWithEmpty(url, parentPath, currentChilds));

                            }

                        });

                        zkListener = listeners.get(listener);

                    }

                    zkClient.create(path, false);

                    List<String> children = zkClient.addChildListener(path, zkListener);

                    if (children != null) {

[...]

PR #9408, opened 2021-12-15
support empty address protection

chickenlj:
Stop empty address notification from taking effect on consumer side.

application.yml
```yaml
dubbo: 
   registry:
     enable-empty-protection: true/false
```

dubbo.properties
```properties
dubbo.registry.enable-empty-protection: true/false
```


chickenlj:
Related https://github.com/apache/dubbo/pull/9112

chickenlj:
#1508

PR #9423, opened 2021-12-15
[master] Optimize of MetadataInfo

wangchengming666:
Optimize of MetadataInfo


dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/MetadataInfo.java
```diff
@@ -315,27 +315,27 @@ public void setParams(Map<String, String> params) {
...
+                if (StringUtils.isNotEmpty(value)) {
                     return value;
                 }
             }
             return params.get(key);
         }
 
         public String getMethodParameter(String method, String key, String defaultValue) {
-            if (methodParams == null) {
+            if (CollectionUtils.isNotEmptyMap(methodParams)) {
```
chickenlj, line 328: Be careful! The new code checks exactly the opposite condition.

chickenlj (changes requested):

PR #9433, opened 2021-12-17
upgrade log4j2 to 2.16.0

wuwen5:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-45046

> It was found that the fix to address CVE-2021-44228 in Apache Log4j 2.15.0 was incomplete in certain non-default configurations. This could allows attackers with control over Thread Context Map (MDC) input data when the logging configuration uses a non-default Pattern Layout with either a Context Lookup (for example, $${ctx:loginId}) or a Thread Context Map pattern (%X, %mdc, or %MDC) to craft malicious input data using a JNDI Lookup pattern resulting in a denial of service (DOS) attack. Log4j 2.15.0 makes a best-effort attempt to restrict JNDI LDAP lookups to localhost by default. Log4j 2.16.0 fixes this issue by removing support for message lookup patterns and disabling JNDI functionality by default.

### DIFF
--- a/compiler/pom.xml
+++ b/compiler/pom.xml
@@ -103,7 +103,7 @@
         <profile>
             <id>release</id>
             <properties>
-                <log4j2_version>2.11.1</log4j2_version>
+                <log4j2_version>2.16.0</log4j2_version>
             </properties>
             <build>
                 <plugins>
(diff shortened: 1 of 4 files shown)


CrazyHZM (approved):
LGTM

guohao (approved):
LGTM

justinmclean:
2.17.0 is out you would probably want to use that instead.

AlbumenJ:
> 2.17.0 is out you would probably want to use that instead.

Upgrade to 2.17.0 in https://github.com/apache/dubbo/pull/9444.

PR #9448, opened 2021-12-19
Update log4j to 2.17 to address CVE-2021-45105.

theArtechnology:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-45105

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -379,7 +379,7 @@
         <profile>
             <id>release</id>
             <properties>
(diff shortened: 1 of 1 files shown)


CrazyHZM:
Thank you for your fix, but this problem will be fixed in version 3.0.5, the relevant pr is https://github.com/apache/dubbo/pull/9443

PR #9451, opened 2021-12-20
cluserter merger refine

HetaoWang:


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/MergerFactory.java
```diff
@@ -49,34 +56,51 @@ public void setScopeModel(ScopeModel scopeModel) {
             throw new IllegalArgumentException("returnType is null");
         }
 
-        Merger result;
-        if (returnType.isArray()) {
-            Class type = returnType.getComponentType();
```
AlbumenJ, line 54: What is the purpose of removing `returnType.getComponentType()`?

AlbumenJ (changes requested):

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/merger/MergerFactory.java
HetaoWang: one question is why the parameter type of Merger.merge() must be array? If use returnType.getComponentType() in MergerFactory, other non-array types can not exist in MergerFactory. 
for example int[] and int type can not both exist in MergerFactory.

HetaoWang: > Another problem is MergerFactory caches merger by return type. But there may exist more than one merger with same return type, for example IntFindAnyMerger and IntFindFirstMerger in this pr.

chickenlj:
As mentioned above, please provide the purposes of this PR. What problems have you been facing and how does this patch can solve.

PR #9463, opened 2021-12-22
Fix callback timeout

wuwen5:
callback use correct timeout

### DIFF
--- a/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/ChannelWrappedInvoker.java
+++ b/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/ChannelWrappedInvoker.java
@@ -28,6 +28,7 @@ import org.apache.dubbo.rpc.AppResponse;
 import org.apache.dubbo.rpc.AsyncRpcResult;
 import org.apache.dubbo.rpc.Invocation;
 import org.apache.dubbo.rpc.Result;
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/ChannelWrappedInvoker.java
```diff
@@ -71,7 +71,9 @@ protected Result doInvoke(Invocation invocation) throws Throwable {
                 currentClient.send(inv, getUrl().getMethodParameter(invocation.getMethodName(), SENT_KEY, false));
                 return AsyncRpcResult.newDefaultAsyncResult(invocation);
             } else {
-                CompletableFuture<AppResponse> appResponseFuture = currentClient.request(inv).thenApply(obj -> (AppResponse) obj);
+                final String methodName = RpcUtils.getMethodName(invocation);
+                final int timeout = calculateTimeout(invocation, methodName);
+                CompletableFuture<AppResponse> appResponseFuture = currentClient.request(inv, timeout, null).thenApply(obj -> (AppResponse) obj);
```
chickenlj, line 76: I have two thoughts over this issue:
1. Should we implicitly set all server push invokers to one-way mode inside the framework? 
2. If 1 is not right, I think adding a timeout to server push RPC call is necessary, but simply using `url.getParameter("timeout")` would be enough, times spent during the should not be considered into the calculation.

chickenlj (changes requested):

PR #9464, opened 2021-12-22
fix async call issues, rpc context and response future callback race conditions

chickenlj:
related issues https://github.com/apache/dubbo/issues/9461 https://github.com/apache/dubbo/issues/8602

* RpcContext deep/shallow copy before async context switch, see AsyncRpcResult constructor.
* Update Future instance in RpcContext to guarantee user registered callbacks will be triggered after framework filter callbacks. See AyncRpcResult.whenCompleteWithContext()
* Refactor recursive filter callback stack to flat iteration to reduce context switch, see FilterChainBuilder.java

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/FilterChainBuilder.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/FilterChainBuilder.java
@@ -18,6 +18,8 @@ package org.apache.dubbo.rpc.cluster.filter;
 
 import org.apache.dubbo.common.URL;
 import org.apache.dubbo.common.extension.SPI;
+import org.apache.dubbo.common.logger.Logger;
+import org.apache.dubbo.common.logger.LoggerFactory;
 import org.apache.dubbo.rpc.BaseFilter;
 import org.apache.dubbo.rpc.Filter;
 import org.apache.dubbo.rpc.Invocation;
@@ -27,6 +29,9 @@ import org.apache.dubbo.rpc.Result;
 import org.apache.dubbo.rpc.RpcException;
 import org.apache.dubbo.rpc.cluster.ClusterInvoker;
 import org.apache.dubbo.rpc.cluster.Directory;
+
+import java.util.ArrayList;
+import java.util.List;
 
 import static org.apache.dubbo.common.extension.ExtensionScope.APPLICATION;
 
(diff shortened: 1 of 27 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/FilterChainBuilder.java
```diff
@@ -144,12 +126,116 @@ public String toString() {
...
+                            }
+                        } else if (filter instanceof FILTER.Listener) {
+                            FILTER.Listener listener = (FILTER.Listener) filter;
+                            if (t == null) {
+                                listener.onResponse(r, filterInvoker, invocation);
+                            } else {
+                                listener.onError(t, filterInvoker, invocation);
+                            }
+                        }
+                    } catch (Throwable filterThrowable) {
```
AlbumenJ, line 190: Calling FILTER.Listener after all filters invoked may cause influence semantics of invokers, such as this may effect some environment clean up operation in finally block.

Example:
```java
public class DemoFilter implements Filter, Filter.Listener {

    @Override
    public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
        try {
            return invoker.invoke(invocation);
        } finally {
            // action 1
        }
    }

    @Override
    public void onResponse(Result appResponse, Invoker<?> invoker, Invocation invocation) {
            // action 2
    }

    @Override
    public void onError(Throwable t, Invoker<?> invoker, Invocation invocation) {
            // action 2
    }
}
```

After this PR change, action 1 will happends before action 2.

AlbumenJ (changes requested):

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/FilterChainBuilder.java
chickenlj: > After this PR change, action 1 will happends before action 2.

After thinking about this issue, I came up with two possible problematic scenarios:

```java
        @Override
        public Result invoke(Invocation invocation) throws RpcException {
            Result asyncResult = filterInvoker.invoke(invocation);
            asyncResult.whenCompleteWithContext((r, t) -> {
                for (int i = filters.size() - 1; i >= 0; i--) {
                    FILTER filter = filters.get(i);
```

## 1 The execution order of action1 and action2 in the same Fitler instance

`Result asyncResult = filterInvoker.invoke(invocation);` happens before `asyncResult.whenCompleteWithContext`(the onResponse callback registration). This sequence can make sure  that`action1` will always happen before `action2`, which I think is right and consistent with the previous behavior. 

## 2 The sequence of action1 and action2 among different filters

For filter chain like `Pxoxy -> Filter1 -> Filter2 -> Invoker`

Filter1's finally block(action1) can be executed before Filter2's onRespose callback(action2), which can be different from the previous behavior when making synchronous calls.  Typical example is `ClassLoaderFilter` which tries to set context that will be used by all the following filters and stacks, that is, all filters that set context across filters may be affected.

But in asynchronous scenarios, this behavior is also not guaranteed before.

chickenlj:
I added support of classloader context recovery for `ClassLoaderFilter`, but I personally don't think `ClassLoaderFilter` is very useful now for the following reasons:
* ClassLoaderFilter classloader switching only works for execution of Filter and service impl but not serialization
* There's classloader switching implemented by ServiceModel.classloader.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/FilterChainBuilder.java
chickenlj: To conclude, there may still have some problems in certain scenarios for this `onResponse` model we are having now, but I don't think it's brought by changes in this patch.

PR #9467, opened 2021-12-22
[3.0] Save url attributes on initialization

chickenlj:
Followup of #9453


AlbumenJ (approved):

chickenlj:
Notice that currently both `lazyUrl` and `atrributes` can be possibly used and they can not be equavelant with each other because one of them may get modified, more likely lazyUrl can be modified.

Issue #9489, opened 2021-12-24
Optimize argument decode exception handling.


tianshuang:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.5
* Operating System version: Mac OS
* Java version: 8

### Steps to reproduce this issue

[dubbo-samples/dubbo-samples-serialization/dubbo-samples-serialization-hessian2/src/main/java/org/apache/dubbo/samples/serialization at master · tianshuang/dubbo-samples · GitHub](https://github.com/tianshuang/dubbo-samples/tree/master/dubbo-samples-serialization/dubbo-samples-serialization-hessian2/src/main/java/org/apache/dubbo/samples/serialization)
Run org.apache.dubbo.samples.serialization.DubboProvider, and run org.apache.dubbo.samples.serialization.DubboConsumer.

### Expected Behavior
When the [User](https://github.com/tianshuang/dubbo-samples/blob/master/dubbo-samples-serialization/dubbo-samples-serialization-hessian2/src/main/java/org/apache/dubbo/samples/serialization/dto/User.java) constructor contains non-primitive objects that cause an exception in the constructor, we hope that the consumer can obtain the original exception information. We want the consumer to get IllegalArgumentException instead of NullPointerException, that is, it should get the original exception.

### Actual Behavior

What we get is the NullPointerException thrown in the provider method body, not the original IllegalArgumentException in the constructor. The reason is that the decode exception is caught and the exception is not thrown to the consumer.

```
Exception in thread "main" java.lang.NullPointerException
	at org.apache.dubbo.samples.serialization.impl.UserServiceImpl.find(UserServiceImpl.java:31)
	at org.apache.dubbo.common.bytecode.Wrapper1.invokeMethod(Wrapper1.java)
	at org.apache.dubbo.rpc.proxy.javassist.JavassistProxyFactory$1.doInvoke(JavassistProxyFactory.java:47)
	at org.apache.dubbo.rpc.proxy.AbstractProxyInvoker.invoke(AbstractProxyInvoker.java:84)
	at org.apache.dubbo.config.invoker.DelegateProviderMetaDataInvoker.invoke(DelegateProviderMetaDataInvoker.java:56)
	at org.apache.dubbo.rpc.protocol.InvokerWrapper.invoke(InvokerWrapper.java:56)
[...]

PR #9490, opened 2021-12-24
Optimize argument decode exception handling.

tianshuang:
In the current implementation, org.apache.dubbo.rpc.protocol.dubbo.DecodeableRpcInvocation#decode(org.apache.dubbo.remoting.Channel, java.io.InputStream) will catch exceptions in readObject, and then only print WARN-level log, If the log level of the provider is ERROR, the error cannot be seen. A common problem is that a NullPointerException be throw when the provider executes business logic. Finally, the consumer gets a NullPointerException, which makes the problem difficult to troubleshoot.

For example, when the argument class has only one constructor with parameters, and the constructor parameter is a non-primitive type, Hession will initialize the parameter to be null. If the parameter is used in the constructor to execute business logic, it will trigger related exceptions, such as IllegalArgumentException, but in the current implementation, parameter decoding exceptions will be caught. The provider actually gets an null parameter when execute business logic. In most scenarios, a NullPointerException will be thrown to the consumer.

Allow the consumer to get the original argument decode exception to facilitate troubleshooting.

## Brief changelog

Remove the exception catch and log printing logic when calling readObject, and throw it directly to the outer decode method so that the exception is set into the request instance, and finally sent to the consumer.

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DecodeableRpcInvocation.java
+++ b/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DecodeableRpcInvocation.java
@@ -210,13 +210,7 @@ public class DecodeableRpcInvocation extends RpcInvocation implements Codec, Dec
 
                 args = new Object[pts.length];
                 for (int i = 0; i < args.length; i++) {
-                    try {
-                        args[i] = in.readObject(pts[i]);
-                    } catch (Exception e) {
-                        if (log.isWarnEnabled()) {
-                            log.warn("Decode argument failed: " + e.getMessage(), e);
-                        }
-                    }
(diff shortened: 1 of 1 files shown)


tianshuang:
#9489

chickenlj:
How about changing the log level to error?

I am a little bit worried that this change will bring unexpected behavior to some of the previous users.

tianshuang:
I checked the Git commit log and this code has been committed since 2012.07.01. The first version used `e.printStackTrace()`, and then 2012.07.06 was optimized to use WARN log printing. This code has not changed since then. It is not a good design to initialize to null when deserialization fails, which delays the time when the error occurs.

Consumers should be aware of deserialization exceptions, not just print logs on the provider.

For previous users, if the deserialization fails, they should deal with the deserialization problem instead of using null parameters, because it seems that the parameters are lost after the Dubbo call.

Reference: [Java Techies Zone: Item 65: Don’t ignore exceptions](http://jtechies.blogspot.com/2012/07/item-65-dont-ignore-exceptions.html)

PR #9578, opened 2022-01-15
refactor: Optimize performance

juzi214032:
If there is already an element in the queue, it is no longer requested

## Checklist

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/ForkingClusterInvoker.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/ForkingClusterInvoker.java
@@ -81,7 +81,7 @@ public class ForkingClusterInvoker<T> extends AbstractClusterInvoker<T> {
             }
             RpcContext.getServiceContext().setInvokers((List) selected);
             final AtomicInteger count = new AtomicInteger();
-            final BlockingQueue<Object> ref = new LinkedBlockingQueue<>();
+            final BlockingQueue<Object> ref = new LinkedBlockingQueue<>(1);
             for (final Invoker<T> invoker : selected) {
                 URL consumerUrl = RpcContext.getServiceContext().getConsumerUrl();
                 executor.execute(() -> {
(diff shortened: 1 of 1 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/ForkingClusterInvoker.java
```diff
@@ -86,6 +86,9 @@ public Result doInvoke(final Invocation invocation, List<Invoker<T>> invokers, L
                 URL consumerUrl = RpcContext.getServiceContext().getConsumerUrl();
                 executor.execute(() -> {
                     try {
+                        if (ref.size() > 0) {
```
BurningCN, line 89: Very good, but the concurrency problem must be taken into account. Suppose that 10 threads execute this step and judge size=0. At this time, multiple elements will still be stored in the queue. The size judgment can use the atomic class, like the previous `final AtomicInteger count = new AtomicInteger();`

BurningCN (changes requested):

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/ForkingClusterInvoker.java
juzi214032: I think that in the case of 10 threads judging at the same time, there should not be a limit, because the results are not obtained at this time and as many threads as possible need to initiate the access, so as to speed up the acquisition of results

juzi214032: However, I have added a capacity limit to the queue, which prevents the memory usage caused by multiple threads getting the results at the same time

BurningCN: Setting size to 1 can only ensure that the result of each call by multiple threads will not be stored when queue.offer is executed, but the actual call still occurs.
Of course, this is a probabilistic event under multithreading.

juzi214032: > I think that in the case of 10 threads judging at the same time, there should not be a limit, because the results are not obtained at this time and as many threads as possible need to initiate the access, so as to speed up the acquisition of results

Yes, as stated above, concurrent calls cannot be limited because of the need to get the fastest results

PR #9642, opened 2022-01-29
[3.0] Fix Service Discovery related concurrency issue

AlbumenJ:

- Fix when last ServiceInstancesChangedListener destroyed, service discovery registry notify new address will cause dead lock issue.
- Fix when reporting metadataInfo,  metadataInfo changed will cause metadataReport unexcepted content issue.


chickenlj (approved):
LGTM

Issue #9583, opened 2022-01-17
org.apache.dubbo.registry.client.ServiceDiscoveryRegistry.shouldRegister(URL)是不是有bug?


genvyday:
* Dubbo version: 3.0.5
* Operating System version: all
* Java version: 1.8

代码：
        String side = providerURL.getSide();
        boolean should = PROVIDER_SIDE.equals(side); // Only register the Provider.
但是调用此函数时，URL是发送到registry注册的URL，side字段已经被移除。
这导致了所有provider都不能在ServiceDiscoveryRegistry注册，从而导致如下的日志不停刷
WARN :20220117.164918010:Dubbo-shared-scheduler-thread-1: [DUBBO] Refreshing of service instance started, but instance hasn't been registered yet., dubbo version: 3.0.5, current host: 10.178.118.12:ServiceInstanceMetadataUtils.java:278

BurningCN:
为什么side字段被移除了？我本地尝试是正常的，side参数是有的
![image](https://user-images.githubusercontent.com/43363120/153330039-144055d4-49df-4ecd-b143-9b1054c45757.png)

genvyday:
我也很困惑，测试的配置如下：
<?xml version="1.0" encoding="UTF-8"?>
<beans xmlns="http://www.springframework.org/schema/beans"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xmlns:context="http://www.springframework.org/schema/context"
    xmlns:tx="http://www.springframework.org/schema/tx" 
    xmlns:aop="http://www.springframework.org/schema/aop" 
	xmlns:p="http://www.springframework.org/schema/p"
	xmlns:dubbo="http://dubbo.apache.org/schema/dubbo"     
	xsi:schemaLocation="
		http://www.springframework.org/schema/beans	
		http://www.springframework.org/schema/beans/spring-beans-3.0.xsd
		http://www.springframework.org/schema/context 
		http://www.springframework.org/schema/context/spring-context-3.0.xsd
		http://www.springframework.org/schema/aop 
		http://www.springframework.org/schema/aop/spring-aop-3.0.xsd 
		http://www.springframework.org/schema/tx 
		http://www.springframework.org/schema/tx/spring-tx-3.0.xsd
		http://dubbo.apache.org/schema/dubbo
		http://dubbo.apache.org/schema/dubbo/dubbo.xsd">
	<dubbo:application id="dubapp" name="dubtest">
[...]

PR #9671, opened 2022-02-10
Fix the issue that providerUrl could not be registered in the application-level registry when simplified is set to true

BurningCN:
issue #9583


wangchengming666 (approved):
LGTM

Issue #9647, opened 2022-02-01
注解的方式配置，@ Configuration @Bean @DubboReference（若配置methods则报错）


lifes:
env 
dubbo3.0.5 

1. 接口定义
```
public interface Service2 {

    String sayHello(String name);

    String sayHello(ClientInfo clientInfo, String name);

    default CompletableFuture<String> sayHelloAsync(String name){
       return CompletableFuture.completedFuture(sayHello(name));
    }

    default CompletableFuture<String> sayHelloAsync(ClientInfo clientInfo, String name){
        return CompletableFuture.completedFuture(sayHello(clientInfo, name));
    }
}
```
2.

```
@Configuration
public class DubboConsumerConfiguration {

    @Bean
    @DubboReference(
            check = false,
        methods = {
             // 如果删除这一行，则不报错
             //仅仅配置@Method(name = "sayHelloAsync"），报错
            //换成xml配置不报错
            //@DubboReference直接在filed上注入不报错
            @Method(name = "sayHelloAsync", onreturn = "service2_sayHelloAsync.onreturn")
        }
    )
    public ReferenceBean<Service2> service2() {
        return new ReferenceBean<>();
    }
}

```
3. 回调接口
 ```
  public interface Notify {

    void onreturn(String ret, String req);
    void onthrow(Throwable ex, String req);

}
```
```
@Component("service2_sayHelloAsync")
public class NotifyImpl implements Notify{
    @Override
    public void onreturn(String ret, String name) {
        System.out.println("callback调用");
        var clientAtt = RpcContext.getClientAttachment().getObjectAttachments();
        var serverAtt = RpcContext.getServerAttachment().getObjectAttachments();
        var serverCtx = RpcContext.getServerContext();
        var serviceCtx = RpcContext.getServiceContext();
        System.out.println("clientAtt" + JSON.toJSONString(clientAtt));
        System.out.println("serverAtt" + JSON.toJSONString(serverAtt));
        System.out.println("serverCtx" + JSON.toJSONString(serverCtx.getObjectAttachments()));
        System.out.println("serviceCtx" + JSON.toJSONString(serviceCtx.getObjectAttachments()));
        System.out.println("callback调用2");
    }

    @Override
    public void onthrow(Throwable ex, String name) {

    }
}
```
报错

[...]

PR #9693, opened 2022-02-17
Fix the problem that NPE may appear in ReferenceBeanSupport#convertToString method

BurningCN:

issue #9647

When processing all properties of `@Method`, the `parameters` property is an empty array, resulting in null return at `obj = DubboAnnotationUtils.convertParameters((String[]) obj);`, the following `else if (obj.getClass().isArray ())` NPE appears


wangchengming666 (approved):
lgtm

PR #9712, opened 2022-02-23
Fix not find server provider startup time

### LINKED ISSUE #9675: [3.0] Can't get timestamp cause loadbalance warm-up invalid
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.

`org.apache.dubbo.rpc.cluster.loadbalance.RandomLoadBalance#needWeightLoadBalance` get URL `timestamp` property is null, As a result, the weight cannot be calculated based on service start time.
image

xiaochangbai:

Fix the problem that the server provider startup time cannot be found on the consumer side

close #9675 

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/URLStrParser.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/URLStrParser.java
@@ -30,7 +30,6 @@ import static org.apache.dubbo.common.utils.Utf8Utils.decodeUtf8;
 
 public final class URLStrParser {
     public static final String ENCODED_QUESTION_MARK = "%3F";
-    public static final String ENCODED_TIMESTAMP_KEY = "timestamp%3D";
     public static final String ENCODED_PID_KEY = "pid%3D";
     public static final String ENCODED_AND_MARK = "%26";
     private static final char SPACE = 0x20;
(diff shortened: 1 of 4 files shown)


BurningCN (changes requested):
In fact, there is no need to do this. The provider has a `timestamp` parameter when registering, but when the registration goes to the `MetadataInfo#addService` method, the internal logic causes the parameter to be removed due to `KEYS_TO_REMOVE`. So the solution is to remove `timestamp` from `KEYS_TO_REMOVE`.

In addition, this modification is only for the application level, and the interface level also needs to be adjusted (`CacheableFailbackRegistry#toUrlsWithoutEmpty `also removes the `timestamp` of the provider url).

BurningCN (changes requested):
In fact, there is no need to do this. The provider has a `timestamp` parameter when registering, but when the registration goes to the `MetadataInfo#addService` method, the internal logic causes the parameter to be removed due to `KEYS_TO_REMOVE`. So the solution is to remove `timestamp` from `KEYS_TO_REMOVE`.

In addition, this modification is only for the application level, and the interface level also needs to be adjusted (`CacheableFailbackRegistry#toUrlsWithoutEmpty `also removes the `timestamp` of the provider url).

xiaochangbai:
I see. I didn't notice these two removal operations at that time

AlbumenJ:
TIMESTAMP key should not pass in metainfo, which is a cache for services' info for static information. BTW, it would be more proper to pass TIMESTAMP in instance info.

PR #9723, opened 2022-02-26
Fix the problem that the consumer stub event does not trigger

BurningCN:
When the `consumer` opens the `stub`, the exposure of the stub service (in the `StubProxyWrapper#getProxy` method) occurs after the connection between the client and the server is completed, so when the `connected` or `disconnected` event in the client is triggered(`DubboProtocol#requestHandler` ), the stub service cannot be found, so that the related methods of the consumer stub service cannot be called.


dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
```diff
@@ -181,13 +185,32 @@ private void invoke(Channel channel, String methodKey) {
...
                     received(channel, invocation);
                 } catch (Throwable t) {
                     logger.warn("Failed to invoke event method " + invocation.getMethodName() + "(), cause: " + t.getMessage(), t);
                 }
             }
         }
 
+        private void getOrWaitStubService(Channel channel, Invocation invocation) throws InterruptedException {
+            try {
+                Invoker<?> invoker = getInvoker(channel, invocation);
```
BurningCN, line 200: If getInvoker throws an exception, it means that the stub service has not been exposed, it will try to wait for the signal that the stub exposure is complete, and then call back the related methods of the stub service.

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
```diff
@@ -253,7 +276,8 @@ private boolean isClientSide(Channel channel) {
         // if it's callback service on client side
         isStubServiceInvoke = Boolean.TRUE.toString().equals(inv.getObjectAttachments().get(STUB_EVENT_KEY));
         if (isStubServiceInvoke) {
-            port = channel.getRemoteAddress().getPort();
+            // The port of the client exposed stub service is 0
+            port = 0;
```
BurningCN, line 280: When the consumer exposes the stub service, the port number is 0

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
```diff
@@ -702,6 +728,8 @@ public void destroy() {
             }
         }
         referenceClientMap.clear();
+        
+        stubServiceToLatch.clear();
```
chickenlj, line 732: I think we should take destroy of stubServiceToLatch on destroy of stub proxy too.

There might have lots of proxy invokers (reference config) create and destroy during the lifetime of DubboProtocol.

chickenlj (changes requested):

chickenlj (changes requested):

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
BurningCN: Consider the complexities of `latch.await()`, Can `stubServiceToLatch` be replaced with this similar structure, `Map<String,Boolean> stubService2Flag`.When the `connect` event occurs and the `stub invoker` is not found, execute `stubService2Flag.put(serviceKey, true)`.
When the stub service exposure is completed, check `if(stubService2Flag.get(serviceKey))`, then call `received(channel, invocation);`

chickenlj: `Map<String,Boolean> stubService2Flag` is still a variable that needs to be taken care of in case of a memory leak. I think a detailed log giving full information of possible causes would be fine.

BurningCN: ok, then there is no need to wait(`stubServiceToLatch`) or delay notification(`stubService2Flag`). If the `stub invoker `cannot be obtained, just print the detailed log.

chickenlj:
Let's change this pull request following this patch https://github.com/apache/dubbo/pull/9825.

Besides that, the only thing that needs to do is to add some extra error information.

BurningCN:
> Let's change this pull request following this patch #9825.
> 
> Besides that, the only thing that needs to do is to add some extra error information.

OK

PR #9732, opened 2022-03-03
Add test case for RouterChain

BurningCN:

Design test cases so that each type of `StateRouter` can effectively filter out non-compliant `invokers`


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/mesh/route/MeshRuleRouter.java
```diff
@@ -108,7 +108,8 @@ public MeshRuleRouter(URL url) {
                 }
             }
         }
-        result = result.or(ruleCache.getUnmatchedInvokers());
```
BurningCN, line 111: My understanding here is: Assuming the above operation makes the `result` non-empty (with a matching invoker), if the `UnmatchedInvoker` is non-empty, this line of code will cause the unmatched ones to be merged into the result, so I think this line of code should be removed

PR #9747, opened 2022-03-08
[3.0]Fix shutdown process, destroy registry and protocol server first before destroying of other resources

chickenlj:
https://github.com/apache/dubbo/issues/9422


YsoulM:
这个merge已经经过验证了么？打算把这个merge合并到3.0.6分支，自己打个版本使用，不知道是否存在风险

PR #9820, opened 2022-03-22
Fix 9806, check metadata protocol

chickenlj:
also see https://github.com/apache/dubbo/pull/9807

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/MetadataReportConfig.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/MetadataReportConfig.java
@@ -27,6 +27,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import static org.apache.dubbo.common.constants.CommonConstants.CYCLE_REPORT_KEY;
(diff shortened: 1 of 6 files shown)


dubbo-common/src/main/java/org/apache/dubbo/config/MetadataReportConfig.java
```diff
@@ -152,7 +153,11 @@ public URL toUrl() throws IllegalArgumentException {
         // Normalize the parameters
         map.putAll(convert(map, null));
         // put the protocol of URL as the "metadata"
-        map.put("metadata", url.getProtocol());
+        String metadataProtocol = isEmpty(url.getProtocol()) ? map.get(PROTOCOL_KEY) : url.getProtocol();
+        if (isEmpty(metadataProtocol)) {
+            throw new IllegalArgumentException("Please specify valid protocol for metadata report." + address);
+        }
```
CrazyHZM, line 159: It is recommended to be better as a config validation, and complete the logic in ConfigValidationUtils#validateMetadataConfig()

CrazyHZM (changes requested):

CrazyHZM (approved):
LGTM

PR #9850, opened 2022-03-29
[3.0] upgrade pom version

wangchengming666:
upgrade pom version

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ There's a [README](https://github.com/apache/dubbo-samples/tree/master/dubbo-sam
 
 ```xml
 <properties>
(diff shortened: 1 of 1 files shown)


guohao:
versions in dependency modules also need update :D

wangchengming666:
> versions in dependency modules also need update :D

已经更新过啦😄

guohao:
> > versions in dependency modules also need update :D
> 
> 已经更新过啦😄

Will be merged after 3.0.7 released

chickenlj:
Sorry, this patch will be conflicted with 3.0.7-release. So we have to close it.

PR #9933, opened 2022-04-18
 fix issue#9922, merge the provider side parameters exclude tag

zhaoguhong:
fix issue #9922 , merge the provider side parameters exclude tag

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
@@ -16,8 +16,10 @@
  */
 package org.apache.dubbo.registry.integration;
 
+import java.util.HashMap;
 import org.apache.dubbo.common.URL;
 import org.apache.dubbo.common.config.configcenter.DynamicConfiguration;
+import org.apache.dubbo.common.constants.CommonConstants;
 import org.apache.dubbo.common.extension.ExtensionLoader;
 import org.apache.dubbo.common.logger.Logger;
 import org.apache.dubbo.common.logger.LoggerFactory;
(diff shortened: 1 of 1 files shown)


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
```diff
@@ -354,8 +356,11 @@ private URL mergeUrl(URL providerUrl) {
 
         providerUrl = providerUrl.addParameter(Constants.CHECK_KEY, String.valueOf(false)); // Do not check whether the connection is successful or not, always create Invoker!
 
+        // fix issue#9922
+        Map<String, String> providerSideParameters = new HashMap<>(providerUrl.getParameters());
+        providerSideParameters.remove(CommonConstants.TAG_KEY);
         // The combination of directoryUrl and override is at the end of notify, which can't be handled here
-        this.overrideDirectoryUrl = this.overrideDirectoryUrl.addParametersIfAbsent(providerUrl.getParameters()); // Merge the provider side parameters
+        this.overrideDirectoryUrl = this.overrideDirectoryUrl.addParametersIfAbsent(providerSideParameters); // Merge the provider side parameters
```
CrazyHZM, line 363: `DefaultProviderURLMergeProcessor` has removed `dubbo.tag` , It's not the best way to fix it
.

CrazyHZM (changes requested):

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/integration/RegistryDirectory.java
zhaoguhong: `DefaultProviderURLMergeProcessor` only remove consumer side `dubbo.tag`, provder side `dubbo.tag` still exist

PR #9949, opened 2022-04-21
[3.0 Triple] Support Connection IDLE

wangchengming666:
Use netty IdleStateEvent to compensate idle. When the idle time is greater than three heartbeat intervals and the channel is inactive, the server adopts the reconnection strategy.


EarthChen:
`IdleStateHandler` already exists in `PortUnificationServer`

EarthChen:
IDLE should be implemented separately for each protocol. so close

Issue #9953, opened 2022-04-22
[2.7.15] AccessLog打印异常会导致异常日志重复输出


sunhk:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.15
* Operating System version: Linux
* Java version: 1.8

### Steps to reproduce this issue

若存在类使com.alibaba.fastjson.JSON.toJSONString调用异常，如：
```
public class Example implements Serializable {
    private static final long serialVersionUID = 493475700851136978L;
    private String name;
    public String getName() throws IllegalAccessException {
        if(name==null){
             throw new IllegalAccessException("name is undefined");
        }
        return name;
    }
    public static void main(String[] args) {
        System.out.println(JSON.toJSONString(new Example()));
    }
}
```
将Example作为provider方法参数。

### Expected Behavior

access log正常打印，或只报一次异常

### Actual Behavior

org.apache.dubbo.rpc.filter.AccessLogFilter#writeLogQueueToFile中processWithServiceLogger或processWithAccessKeyLogger方法调用异常，无法及时清理logQueue，使相同异常重复打印，阻断queue中后续日志打印，log queue也会随着使用逐渐充满。
dubbo 3.0未进行测试，但看代码逻辑是一样的

```
com.alibaba.fastjson.JSONException: write javaBean error, fastjson version 1.2.70, class test.Example, method : getName
	at com.alibaba.fastjson.serializer.JavaBeanSerializer.write(JavaBeanSerializer.java:539)
	at com.alibaba.fastjson.serializer.JavaBeanSerializer.write(JavaBeanSerializer.java:149)
	at com.alibaba.fastjson.serializer.JSONSerializer.write(JSONSerializer.java:285)
	at com.alibaba.fastjson.JSON.toJSONString(JSON.java:758)
	at com.alibaba.fastjson.JSON.toJSONString(JSON.java:696)
	at com.alibaba.fastjson.JSON.toJSONString(JSON.java:661)
	at org.apache.dubbo.rpc.support.AccessLogData.getLogMessage(AccessLogData.java:242)
	at org.apache.dubbo.rpc.filter.AccessLogFilter.processWithServiceLogger(AccessLogFilter.java:175)
	at org.apache.dubbo.rpc.filter.AccessLogFilter.writeLogQueueToFile(AccessLogFilter.java:133)
[...]

PR #9978, opened 2022-04-28
Fix issue#9953

106umao:

fix the issue #9953 

## Brief changelog

Modified the way the logQueue is traversed to ensure that items are removed from the queue before throwing an exception.
修改了logQueue的遍历方式，确保在抛出异常前将item从队列中移除。

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
@@ -141,7 +141,6 @@ public class AccessLogFilter implements Filter {
                 processWithAccessKeyLogger(logQueue, file);
             }
         } catch (Exception e) {
-            logQueue.poll();
             logger.error(e.getMessage(), e);
         }
     }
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
```diff
@@ -141,6 +141,7 @@ private void writeLogQueueToFile(String accessLog, Queue<AccessLogData> logQueue
                 processWithAccessKeyLogger(logQueue, file);
             }
         } catch (Exception e) {
+            logQueue.poll();
```
CrazyHZM, line 144: It will be better to modify the traversal method of the queue.

CrazyHZM (changes requested):

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/AccessLogFilter.java
106umao: > It will be better to modify the traversal method of the queue.

Yes, I think so too, but I'm afraid that changing too much code might not be acceptable

106umao: > It will be better to modify the traversal method of the queue.

And if you modify the queue traversal methods so processWithAccessKeyLogger two overloaded methods need to be modified

CrazyHZM: I'm sorry, maybe I didn't express it clearly, you should handle exceptions in processWithAccessKeyLogger, and you also need to dequeue operations. If the queue is not empty, the `while` operation will cause the thread to be blocked all the time.
@106umao

chickenlj (approved):
LGTM.

PR #10003, opened 2022-05-06
Fix interface-app mapping, recored mapping init status for each unique interface key.

chickenlj:
Related issue https://github.com/apache/dubbo/issues/9992


rainwolf01:
LGTM

PR #10006, opened 2022-05-06
Optimize configuration item key

chickenlj:


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/migration/model/MigrationRule.java
```diff
@@ -180,13 +181,21 @@ public MigrationStep getStep(URL consumerURL) {
...
         }
 
         return step;
     }
 
+    private String getDefaultStep(URL consumerURL, String defaultStep) {
+        String globalDefaultStep = ConfigurationUtils.getCachedDynamicProperty(consumerURL.getScopeModel(), DUBBO_SERVICEDISCOVERY_MIGRATION, null);
+        if (StringUtils.isEmpty(globalDefaultStep)) {
+            // check 'dubbo.application.service-discovery.step' for compatibility
+            globalDefaultStep = ConfigurationUtils.getCachedDynamicProperty(consumerURL.getScopeModel(), "dubbo.application.service-discovery.step", defaultStep);
```
CrazyHZM, line 194: Should be compatible with `dubbo.application.service-discovery.migration`

CrazyHZM (changes requested):

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/migration/model/MigrationRule.java
chickenlj: Sure, I will fix it now

Issue #9990, opened 2022-05-05
[3.0] warm-up feature don't work at instance register-mode


YoungHu:
- [x] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 3.0.7
* Operating System version: mac os
* Java version: JDK8

### Steps to reproduce this issue

1. set both provider and consumer side as `register-mode=instance` and ` service-discovery.migration=FORCE_APPLICATION`
2. debug at `AbstractLoadBalance.getWeight` method
3. start 2 providers and start 1 consumer to invoke.
4. `AbstractLoadBalance.getWeight` always return 100.

### Expected Behavior
new start node weight should less than 100

### Actual Behavior
got weight value=100

### What Reason
if want warn up feature work correct, the key parameter is `timestamp` which in `urlParam`.  `DubboServiceAddressURL` fill urlParam but 'InstanceAddressURL' not, so when dubbo register with instance mode, warn up can't work.

![image](https://user-images.githubusercontent.com/7418602/166858211-47224b65-9021-4fd3-9128-1438f59b036a.png)
![image](https://user-images.githubusercontent.com/7418602/166858350-6459156d-bfbc-4873-a0ce-cf0ec49a14ee.png)
![image](https://user-images.githubusercontent.com/7418602/166858399-df6354b2-3d6a-40c9-ab6b-1d2cf741f190.png)

chickenlj:
This is a known issue. `timestamp` key has currently been removed from previous 3.0 releases for performance seeking. We are considering keep `timestamp` as disabled by default and give users the way to enable it if necessary.

YoungHu:
> This is a known issue. `timestamp` key has currently been removed from previous 3.0 releases for performance seeking. We are considering keep `timestamp` as disabled by default and give users the way to enable it if necessary.

`timestamp` key how to effect performance?

chickenlj:
> > This is a known issue. `timestamp` key has currently been removed from previous 3.0 releases for performance seeking. We are considering keep `timestamp` as disabled by default and give users the way to enable it if necessary.
> 
> `timestamp` key how to effect performance?

The timestamp value will change every time one instance restarts.

chickenlj:
One way to do that is by extending `MetadataParamsFilter` SPI as below;
```java
@Activate
public class CustomizedParamsFilter implements MetadataParamsFilter {
    @Override
    public String[] instanceParamsIncluded() {
[...]

PR #10015, opened 2022-05-09
Add 'exclude' filtering policy for instance-level params

chickenlj:
related issue #9990

New definition of ParamsFilter is as below

```java
/**
 * This filter applies an either 'include' or 'exclude' policy with 'include' having higher priority.
 * That means if 'include' is specified then params specified in 'exclude' will be ignored
 *
 * If multiple Filter extensions are provided, then,
 * 1. All params specified as should be included within different Filter extension instances will determine the params that will finally be used.
 * 2. If none of the Filter extensions specified any params as should be included, then the final effective params would be those left after removed all the params specified as should be excluded.
 *
 * It is recommended for most users to use 'exclude' policy for service params and 'include' policy for instance params.
 * Please use 'params-filter=-default, -filterName1, filterName2' to activate or deactivate filter extensions.
 */
@SPI
public interface MetadataParamsFilter {

    */
   String[] serviceParamsIncluded();

    /**
     * params that need to be excluded before sending to metadata center
     *
     * @return arrays of keys
     */
    String[] serviceParamsExcluded();

   /**
    * params that need to be sent to registry center
    *
    * @return arrays of keys
    */
   String[] instanceParamsIncluded();

    /**
     * params that need to be excluded before sending to registry center
     *
     * @return arrays of keys
     */
    String[] instanceParamsExcluded();
}
```

### DIFF
--- a/dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/DefaultMetadataParamsFilter.java
+++ b/dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/DefaultMetadataParamsFilter.java
@@ -1,0 +1,54 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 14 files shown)


chickenlj:
Reviewed by Albumen

PR #10017, opened 2022-05-09
Followup of #10010, share proxy instance among References that have the same key.

chickenlj:
related to https://github.com/apache/dubbo/issues/10012


chickenlj:
Reviewed by Albumen

PR #10023, opened 2022-05-10
[optimization]RestProtocol opt, share client pool among services and avoid potential memory leak.

chickenlj:
An improved version of https://github.com/apache/dubbo/pull/10023

related issue https://github.com/apache/dubbo/issues/9629


dubbo-common/src/main/java/org/apache/dubbo/common/reference/ReferenceCountedResource.java
```diff
@@ -0,0 +1,77 @@
...
+
+import org.apache.dubbo.common.logger.Logger;
+import org.apache.dubbo.common.logger.LoggerFactory;
+
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+
+/**
+ * inspired by Netty
+ */
+public abstract class ReferenceCountedResource implements AutoCloseable {
```
guohao, line 27: Why not just use netty's `io.netty.util.ReferenceCounted`?

chickenlj: I think what we need is just a common and simplified version for components across the framework to use, so better not rely on Netty.

dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/ReferenceCountedClient.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public boolean isDestroyed() {
+        return resteasyClient.isClosed();
+    }
+
+    @Override
+    protected void destroy() {
+        try {
+            resteasyClient.close();
```
CrazyHZM, line 45: Add some log output, others look good to me.

CrazyHZM (approved):

AlbumenJ (approved):

dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/RestProtocol.java
```diff
@@ -136,65 +133,77 @@ protected <T> Runnable doExport(T impl, Class<T> type, URL url) throws RpcExcept
 
     @Override
     protected <T> T doRefer(Class<T> serviceType, URL url) throws RpcException {
+        ReferenceCountedClient referenceCountedClient = clients.computeIfAbsent(url.getAddress(), _key -> {
```
JasonKai9527, line 136: if the ip:port is the same,but httpclient config properties are different, how to fix th is situation?

PR #10099, opened 2022-05-28
Update fastjson to 1.2.83

kimmking:
Security issue: Update fastjson to 1.2.83


kevinkwok326:
When will this modification be released？

PR #10109, opened 2022-06-01
improve:speed up application start up time

qinliujie:

speed up application start up time when  application have many classloaders

## Brief changelog

some spi are implements by dubbo framework only and scan multi classloaders resources may cause application startup very slow,so add properties file to determine which spi should be load only dubbo framework classloader


AlbumenJ (approved):

chickenlj (approved):
LGTM.

PR #10157, opened 2022-06-14
Nacos registry, use 'group' as 'namespace' if not explicitly specified.

chickenlj:
Compatible with other registry configuration patterns, use 'group' as 'namespace' if not explicitly specified.

The following configuration `dubbo.registry.nacos.group=dev` will be interpreted as `dubbo.registry.nacos.namespace=dev`.


wangchengming666 (approved):
LGTM

chickenlj:
After discussing with Albumen, we agreed on not doing this interpretation for users for it might cause confusion for them. If they want to enable Nacos namespace then configure it explicitly.

PR #10164, opened 2022-06-15
Interrupt current thread when InterruptException occur

happytimor:
Interrupt current thread when InterruptException occur

org.apache.dubbo.rpc.protocol.AsyncToSyncInvoker.java
``` java
    @Override
    public Result invoke(Invocation invocation) throws RpcException {
        Result asyncResult = invoker.invoke(invocation);
        try {
            if (InvokeMode.SYNC == ((RpcInvocation) invocation).getInvokeMode()) {
                asyncResult.get(Integer.MAX_VALUE, TimeUnit.MILLISECONDS);
            }
        } catch (InterruptedException e) {
            // interrupt thread
            Thread.currentThread().interrupt();
            throw new RpcException("Interrupted unexpectedly while waiting for remote result to return!  method: " +
                    invocation.getMethodName() + ", provider: " + getUrl() + ", cause: " + e.getMessage(), e);
        } 
        return asyncResult;
    }
```

## Brief changelog
Interrupt current thread when InterruptException occur

## Checklist

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/FailoverClusterInvoker.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/FailoverClusterInvoker.java
@@ -93,6 +93,10 @@ public class FailoverClusterInvoker<T> extends AbstractClusterInvoker<T> {
                 if (e.isBiz()) { // biz exception.
                     throw e;
                 }
+                if (e.getCause() != null && InterruptedException.class.getName().equals(e.getCause().toString())) {
+                    // don`t catch interrupt exception
+                    throw new RuntimeException(e);
+                }
                 le = e;
             } catch (Throwable e) {
                 le = new RpcException(e.getMessage(), e);
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/FailoverClusterInvoker.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/FailoverClusterInvoker.java
@@ -93,10 +93,6 @@ public class FailoverClusterInvoker<T> extends AbstractClusterInvoker<T> {
                 if (e.isBiz()) { // biz exception.
                     throw e;
                 }
-                if (e.getCause() != null && InterruptedException.class.getName().equals(e.getCause().toString())) {
-                    // don`t catch interrupt exception
-                    throw new RuntimeException(e);
-                }
                 le = e;
             } catch (Throwable e) {
                 le = new RpcException(e.getMessage(), e);
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AsyncToSyncInvoker.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AsyncToSyncInvoker.java
@@ -61,6 +61,7 @@ public class AsyncToSyncInvoker<T> implements Invoker<T> {
                 asyncResult.get(Integer.MAX_VALUE, TimeUnit.MILLISECONDS);
             }
         } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
             throw new RpcException("Interrupted unexpectedly while waiting for remote result to return!  method: " +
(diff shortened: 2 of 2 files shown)


Kvicii (approved):

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/FailoverClusterInvoker.java
```diff
@@ -93,6 +93,10 @@ public Result doInvoke(Invocation invocation, final List<Invoker<T>> invokers, L
                 if (e.isBiz()) { // biz exception.
                     throw e;
                 }
+                if (e.getCause() != null && InterruptedException.class.getName().equals(e.getCause().toString())) {
+                    // don`t catch interrupt exception
+                    throw new RuntimeException(e);
+                }
```
CrazyHZM, line 99: You can choose other fault tolerance strategies, such as cluster="failfast"

CrazyHZM (changes requested):

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/FailoverClusterInvoker.java
happytimor: it is not confit, we need failover strategy at most time, but sometimes due to business needs, I need to interrupt thread pool tasks(which contains dubbo invoke)

CrazyHZM: Interrupting a thread is not a common scenario, and I don't approve of modifying non-generic logic.
Based on your scenario, I think you may need a custom Cluster implementation.

happytimor: my code is like this

``` java
Future<?> future =RUNNER_EXECUTOR.submit(() -> {
    pipelineService.startTask(taskId);
});
```
``` java
//because some logic error, i need find the future and  cancel task
future.cancel();
```

i think it is a normal scenario......

CrazyHZM: I suggest that you can first create an issue to discuss this requirement,  so that more people can see and participate in the discussion.

happytimor: maybe later, thanks for your patient reply.

AlbumenJ:
would you please add some test cases to verify this

happytimor:
> 



> would you please add some test cases to verify this

it is a bit difficult for me to setup the ut environment with mockito, i can only provide basic code

## consumer:
``` java
ThreadPoolExecutor threadPoolExecutor = new ThreadPoolExecutor(1, 1, 0, TimeUnit.SECONDS, new LinkedBlockingQueue<>());
String name = "world" + System.currentTimeMillis();
int initialFlag = 0, runningFlag = 1, normalReturn = 2, interruptReturn = 3;
AtomicInteger flag = new AtomicInteger(initialFlag);
Future<?> future = threadPoolExecutor.submit(() -> {
    try {
        flag.set(runningFlag);
        service.sayHello(name);
        flag.set(normalReturn);
    } catch (Exception e) {
        flag.set(interruptReturn);
    }
});
Thread.sleep(100);
future.cancel(true);
while (flag.get() == initialFlag || flag.get() == runningFlag) {
    Thread.yield();
}
Assertions.assertEquals(flag.get(), interruptReturn);
```

## provider：
``` java
String lastName = "";

@Override
    if (!lastName.equals(name)) {
        lastName = name;
        try {
            //will be interrupted at first time
            Thread.sleep(200);
        } catch (InterruptedException ignore) {

        }
    }
    return "Hello " + name + ", response from provider: " + RpcContext.getContext().getLocalAddress();
}

```

## important parameter:
retries=3&cluster=failover

------
flag will return normalReturn if we don`t add `Thread.currentThread().interrupt()`

AlbumenJ:
@happytimor Please apply this pr to 3.0 branch.

happytimor:
https://github.com/apache/dubbo/pull/10299  @AlbumenJ

PR #10229, opened 2022-06-28
refactor: optimize ExtensionLoader code style

linjianming02:
## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
@@ -100,7 +100,7 @@ public class ExtensionLoader<T> {
     private final Holder<Map<String, Class<?>>> cachedClasses = new Holder<>();
 
     private final Map<String, Object> cachedActivates = new ConcurrentHashMap<>();
-    private final ConcurrentMap<String, Holder<Object>> cachedInstances = new ConcurrentHashMap<>();
+    private final ConcurrentMap<String, T> cachedInstances = new ConcurrentHashMap<>();
     private final Holder<Object> cachedAdaptiveInstance = new Holder<>();
     private volatile Class<?> cachedAdaptiveClass = null;
     private String cachedDefaultName;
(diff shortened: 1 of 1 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
```diff
@@ -436,18 +425,8 @@ public T getExtension(String name, boolean wrap) {
...
-            synchronized (holder) {
-                instance = holder.get();
-                if (instance == null) {
-                    instance = createExtension(name, wrap);
-                    holder.set(instance);
-                }
-            }
-        }
-        return (T) instance;
+        final T instance = cachedInstances.get(name);
```
AlbumenJ, line 428: this line can be removed

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
linjianming02: Has been modified

dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
```diff
@@ -436,18 +425,7 @@ public T getExtension(String name, boolean wrap) {
...
-            synchronized (holder) {
-                instance = holder.get();
-                if (instance == null) {
-                    instance = createExtension(name, wrap);
-                    holder.set(instance);
-                }
-            }
-        }
-        return (T) instance;
+        return cachedInstances.computeIfAbsent(name, (k) -> createExtension(k, wrap));
```
AlbumenJ, line 428: I'm worried this might create a lot of class objects because of the lambda.

stonelion: Yes, Using a get-null prediction could be better.

AlbumenJ: rollback will make this action not atomic

stonelion: The method computeIfAbsent of ConcurrentHashMap is atomic. That function is applied at most once per key.

AlbumenJ: but lambda might create a lot of class object, and will lead to metadata space increase

stonelion: Yes，lambda capture the context variable of wrap as closure,so It renew a class for every invocation.

PR #10248, opened 2022-07-01
[2.7] Fix async call issues

yyykf:
## What is the purpose of the change

fix #10236 , asyncCall should not return a completed Future when protocol is not injvm.

when return type of method is primitive type, the proxy will convert null to default value of primitive type. It will cause user get wrong result.

## Checklist
- [x] Make sure there is a [GitHub_issue](https://github.com/apache/dubbo/issues) field for the change (usually before you start working on it). Trivial changes like typos do not require a GitHub issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [ ] Each commit in the pull request should have a meaningful subject line and body.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Check if is necessary to patch to Dubbo 3 if you are work on Dubbo 2.7
- [ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add sample in [dubbo samples](https://github.com/apache/dubbo-samples) project.
- [ ] Add some description to [dubbo-website](https://github.com/apache/dubbo-website) project if you are requesting to add a feature.
- [ ] GitHub Actions works fine on your own branch.
- [ ] If this contribution is large, please follow the [Software Donation Guide](https://github.com/apache/dubbo/wiki/Software-donation-guide).

PR #10258, opened 2022-07-02
fix RpcContext's asyncCall method invoke an interface that returns a value of type 'boolean' always returns 'false'

kefate:
fix RpcContext's asyncCall method invoke an interface that returns a value of type 'boolean' always returns 'false'

## Verifying this change
1. Write an mock DemoService as provider wiht an interface test() return 'boolean'
2. Set @Reference(injvm = false)
3. Call DemoService.test() with RpcContext.getContext().asyncCall()
4. No matter what the test() method returns, the result is always false

## Checklist


AlbumenJ:
This same issue like #10248
If use async call invoke a local service without remote call, this will break the result.

Issue #10257, opened 2022-07-02
启动时报java.lang.UnsupportedOperationException导致无法运行


applecyg:
org.apache.dubbo:dubbo:3.0.7及以上版本，程序运行会出现如下异常：
java.lang.IllegalStateException: Failed to override field value of config bean: <dubbo:service path="com.test.ApiDubboService" version="1.0.0" deprecated="false" protocolIds="dubbo" dynamic="true" id="ServiceBean:com.test.ApiDubboService:1.0.0" ref="com.test.ApiDubboServiceImpl@59e9bb43" interface="com.test.ApiDubboService" />
	at org.apache.dubbo.config.AbstractConfig.refresh(AbstractConfig.java:696) ~[dubbo-3.0.9.jar:3.0.9]
	at org.apache.dubbo.config.deploy.DefaultModuleDeployer.exportServiceInternal(DefaultModuleDeployer.java:320) ~[dubbo-3.0.9.jar:3.0.9]
	at org.apache.dubbo.config.deploy.DefaultModuleDeployer.exportServices(DefaultModuleDeployer.java:313) ~[dubbo-3.0.9.jar:3.0.9]
	at org.apache.dubbo.config.deploy.DefaultModuleDeployer.start(DefaultModuleDeployer.java:145) ~[dubbo-3.0.9.jar:3.0.9]
	at org.apache.dubbo.config.spring.context.DubboDeployApplicationListener.onContextRefreshedEvent(DubboDeployApplicationListener.java:111) ~[dubbo-3.0.9.jar:3.0.9]
	at org.apache.dubbo.config.spring.context.DubboDeployApplicationListener.onApplicationEvent(DubboDeployApplicationListener.java:100) ~[dubbo-3.0.9.jar:3.0.9]
	at org.apache.dubbo.config.spring.context.DubboDeployApplicationListener.onApplicationEvent(DubboDeployApplicationListener.java:45) ~[dubbo-3.0.9.jar:3.0.9]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:176) ~[spring-context-5.3.21.jar:5.3.21]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:169) ~[spring-context-5.3.21.jar:5.3.21]
[...]

PR #10262, opened 2022-07-03
fix UnsupportedOperationException in invokeSetParameters

cheese8:
#10257


wangchengming666 (approved):
LGTM

PR #10311, opened 2022-07-13
fix_metadatainfo_out_of_memory

owen200008:
fix_metadatainfo_out_of_memory

### DIFF
--- a/dubbo-metadata/dubbo-metadata-api/src/test/java/org/apache/dubbo/metadata/MetadataInfoTest.java
+++ b/dubbo-metadata/dubbo-metadata-api/src/test/java/org/apache/dubbo/metadata/MetadataInfoTest.java
@@ -16,6 +16,8 @@
  */
 package org.apache.dubbo.metadata;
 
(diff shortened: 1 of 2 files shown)


dubbo-metadata/dubbo-metadata-api/src/test/java/org/apache/dubbo/metadata/MetadataInfoTest.java
```diff
@@ -161,4 +163,26 @@ public void testJsonFormat() {
...
+
+        metadataInfo.addService(url2);
+
+        metadataInfo.calAndGetRevision();
+
+        metadataInfo.addService(url3);
+
+        metadataInfo.calAndGetRevision();
+
+        JSONObject object = JSON.parseObject(metadataInfo.getContent());
```
AlbumenJ, line 184: Use JsonUtil, prevent directly use fastjson

AlbumenJ (changes requested):

Issue #10312, opened 2022-07-13
Refactor a clean and testable dubbo


cheese8:
Refactor a clean and testable dubbo, to unify the code style:
1. use guard clause
2. no more than 2 depth if else
3. no more than 2 depth loop
4. avoid too big method
5. use final
6. remove useless pieces
7. use meaningful names
8. testcase coverage

PR #10313, opened 2022-07-13
refactor parameter on consumer url logic code

cheese8:
refactor parameter on consumer url logic code: #10312
1.using ParameterOnConsumerUrl, getParameterOnConsumerUrl and hasParameterOnConsumerUrl to refactor the multi-places similar code
2. using final

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/InstanceAddressURL.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/InstanceAddressURL.java
@@ -33,7 +33,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.Function;
+import java.util.function.Supplier;
 
 import static org.apache.dubbo.common.constants.CommonConstants.CONSUMER;
 import static org.apache.dubbo.common.constants.CommonConstants.CONSUMER_SIDE;
@@ -183,8 +183,8 @@ public class InstanceAddressURL extends URL {
         } else if (SIDE_KEY.equals(key)) {
             return getSide();
         }
-
-        Optional<String> parameterOnConsumerUrl = getParameterOnConsumerUrl(ParameterOnConsumerUrlEnum.Parameter, null, null, key);
+    
+        Optional<String> parameterOnConsumerUrl = new GetParameterOnConsumerUrlSupplier(ParameterOnConsumerUrlEnum.Parameter, null, null, key).get();
         if (parameterOnConsumerUrl.isPresent()) {
             return parameterOnConsumerUrl.get();
         }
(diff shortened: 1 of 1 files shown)


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/InstanceAddressURL.java
```diff
@@ -343,37 +312,91 @@ public boolean hasMethodParameter(String method, String key) {
...
+                return false;
+        }
+    }
+    
+    enum ParameterOnConsumerUrl {
+        Parameter,
+        ServiceParameter,
+        ServiceMethodParameter,
+        MethodParameter
+    }
```
AlbumenJ, line 399: Will this make code more unreadable

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/InstanceAddressURL.java
```diff
@@ -343,37 +312,91 @@ public boolean hasMethodParameter(String method, String key) {
...
         }
-
         String protocolServiceKey = getProtocolServiceKey();
         if (isEmpty(protocolServiceKey)) {
             return false;
         }
         return hasServiceMethodParameter(protocolServiceKey, method);
     }
+    
+    private String getParameterOnConsumerUrl(final ParameterOnConsumerUrl parameterOnConsumerUrl, final URL consumerUrl,
```
EarthChen, line 338: replace `if return` is better.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/InstanceAddressURL.java
AlbumenJ: just replaced `consumerUrl` operation with `Function<URL, String>`

cheese8: As `using Function to refactor the switch` way?

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/InstanceAddressURL.java
cheese8: How about this way? As the code in `using Function to refactor the switch`.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/InstanceAddressURL.java
AlbumenJ:
```
    public Optional<String> executeWithConsumerUrl (Function<URL, String> function) {
            URL consumerUrl = RpcContext.getServiceContext().getConsumerUrl();
            if (consumerUrl != null) {
                String v = function.apply(consumerUrl);
                if (StringUtils.isNotEmpty(v)) {
                    return Optional.of(v);
                }
                return Optional.empty();
            }
    }
```

chickenlj:
Thanks for the contribution.

This code seems to be more maintenance-friendly but harder to read, considering the fact that the logic in InstanceAddressURL is very likely to be changed in the future. I would prefer to keep the current code unchanged.

PR #10323, opened 2022-07-15
Multiple registries

chickenlj:
查看官网【[Java/参考手册/注册中心说明/多注册中心](https://dubbo.apache.org/zh/docs3-v2/java-sdk/reference-manual/registry/multiple-registry/#22-%E5%A4%9A%E6%B3%A8%E5%86%8C%E4%B8%AD%E5%BF%83%E5%9C%B0%E5%9D%80%E8%81%9A%E5%90%88)】了解配置详情。


dubbo-registry/dubbo-registry-multiple/src/main/java/org/apache/dubbo/registry/multiple/MultipleRegistry.java
```diff
@@ -292,15 +302,49 @@ public synchronized void notifySourceListener() {
...
+         * @param singleURLs single registry url list
+         * @param registryURL single registry configuration url
+         */
+        private void aggregateRegistryUrls(List<URL> notifyURLs, List<URL> singleURLs, URL registryURL) {
+            String registryAttachments = registryURL.getParameter("attachments");
+            if (StringUtils.isNotBlank(registryAttachments)) {
+                LOGGER.info("Registry attachments " + registryAttachments + " found, will append to provider urls, urls size " + singleURLs.size());
+                String[] pairs = registryAttachments.split(COMMA_SEPARATOR);
+                for (String rawPair : pairs) {
+                   String[] keyValuePair = rawPair.split("=");
```
24kpure, line 332: Here are my personal opinion，just some typo.
1.I find `split("="）` appear twice already in project.How about EQUAL_SEPARATOR?
2.Could we  remove `else` with the code like  below
 `if (StringUtils.isBlank(registryAttachments)) {
           //do somthing
            return;
        }`
3. Do some `for` in this method can be replace with stream?

chickenlj: Thanks. I will try to do following the first two suggestions.

PR #10348, opened 2022-07-21
same protocol listen on different ports

dadadadudu:
version: dubbo3.1
bug: https://github.com/apache/dubbo/issues/10238


dadadadudu:
close

PR #10354, opened 2022-07-22
Fix the problem of partial setXXX method injection of spi class

BurningCN:

When spiXXX implements ScopeModelAware, ExtensionAccessorAware, the setXXX of ScopeModelAware and ExtensionAccessorAware does not need to be injected(ExtensionDirectorTest#testInjection can be reproduced)

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
@@ -41,6 +41,7 @@ import org.apache.dubbo.rpc.model.FrameworkModel;
 import org.apache.dubbo.rpc.model.ModuleModel;
 import org.apache.dubbo.rpc.model.ScopeModel;
 import org.apache.dubbo.rpc.model.ScopeModelAccessor;
+import org.apache.dubbo.rpc.model.ScopeModelAware;
 
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -134,6 +135,8 @@ public class ExtensionLoader<T> {
 
     private static SoftReference<Map<java.net.URL,List<String>>> urlListMapCache = new SoftReference<>(new ConcurrentHashMap<>());
 
+    private static List<String> ignoredInjectMethodsDesc = getIgnoredInjectMethodsDesc();
+
     /**
      * Record all unacceptable exceptions when using SPI
      */
@@ -185,6 +188,13 @@ public class ExtensionLoader<T> {
      */
     public static List<LoadingStrategy> getLoadingStrategies() {
         return asList(strategies);
+    }
+
+    private static List<String> getIgnoredInjectMethodsDesc() {
+        List<String> ignoreInjectMethodsDesc = new ArrayList<>();
+        Arrays.stream(ScopeModelAware.class.getMethods()).map(ReflectUtils::getDesc).forEach(ignoreInjectMethodsDesc::add);
+        Arrays.stream(ExtensionAccessorAware.class.getMethods()).map(ReflectUtils::getDesc).forEach(ignoreInjectMethodsDesc::add);
+        return ignoreInjectMethodsDesc;
     }
 
     ExtensionLoader(Class<?> type, ExtensionDirector extensionDirector, ScopeModel scopeModel) {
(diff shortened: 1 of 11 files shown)


dubbo-common/src/main/java/org/apache/dubbo/rpc/model/ModuleModel.java
```diff
@@ -76,8 +73,6 @@ public ModuleModel(ApplicationModel applicationModel, boolean isInternal) {
     protected void initialize() {
         super.initialize();
         this.serviceRepository = new ModuleServiceRepository(this);
-        this.moduleConfigManager = new ModuleConfigManager(this);
-        this.moduleConfigManager.initialize();
```
BurningCN, line 80: Consistent with `applicationModel`, `ModuleConfigManager` is instantiated and initialized through spi.
和`applicationModel`保持一致，`ModuleConfigManager`通过spi的方式进行实例化以及初始化。

dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java
```diff
@@ -835,11 +845,23 @@ private T injectExtension(T instance) {
...
+                // the setXXX of ScopeModelAware and ExtensionAccessorAware does not need to be injected
+                if (method.getDeclaringClass() == ScopeModelAware.class) {
+                    continue;
+                }
+                if (instance instanceof ScopeModelAware || instance instanceof ExtensionAccessorAware) {
+                    if (ignoredInjectMethodsDesc.contains(ReflectUtils.getDesc(method))) {
+                        continue;
+                    }
+                }
+
```
BurningCN, line 864: 第一个if的作用是：一个`spi`实现类/接口 `implements ScopeModelAware`，但是没有重写`ScopeModelAware`的相关方法（直接继承`ScopeModelAware`的`setXXX`方法），则满足该if条件，不需要考虑注入，直接`continue`。
第二个if的作用是：如果`spi`实现类重写或实现了 `ScopeModelAware`或者`ExtensionAccessorAware` 的 `setXXX`方法，也不需要考虑注入，直接`continue`。

dubbo-common/src/main/java/org/apache/dubbo/rpc/model/FrameworkModel.java
```diff
@@ -123,7 +123,6 @@ protected void onDestroy() {
         // notify destroy and clean framework resources
         // see org.apache.dubbo.config.deploy.FrameworkModelCleaner
         notifyDestroy();
-        checkApplicationDestroy();
```
AlbumenJ, line 126: why remove this?

dubbo-common/src/main/java/org/apache/dubbo/rpc/model/ModuleModel.java
```diff
@@ -61,9 +61,6 @@ public ModuleModel(ApplicationModel applicationModel, boolean isInternal) {
         }
 
         initialize();
-        Assert.notNull(serviceRepository, "ModuleServiceRepository can not be null");
-        Assert.notNull(moduleConfigManager, "ModuleConfigManager can not be null");
-        Assert.assertTrue(moduleConfigManager.isInitialized(), "ModuleConfigManager can not be initialized");
```
AlbumenJ, line 66: why remove this line?

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/rpc/model/FrameworkModel.java
BurningCN: checkApplicationDestroy has been called on line 121
image

dubbo-common/src/main/java/org/apache/dubbo/rpc/model/ModuleModel.java
BurningCN: Of course, it can be reserved, but I just want to unify it with `ApplicationModule`.(`ApplicationModule` does not check its `configManager` and  `serviceRepository`)

AlbumenJ: It would be better that add this check to `ApplicationModel`

BurningCN: > It would be better that add this check to `ApplicationModel`

done

PR #10481, opened 2022-08-18
Add some test cases and optimizations

BurningCN:


dubbo-common/src/main/java/org/apache/dubbo/common/bytecode/ClassGenerator.java
```diff
@@ -398,7 +400,7 @@ private CtConstructor getCtConstructor(Constructor<?> c) throws NotFoundExceptio
         return getCtClass(c.getDeclaringClass()).getConstructor(ReflectUtils.getDesc(c));
     }
 
-    public static interface DC {
+    public interface DC {
```
AlbumenJ, line 403: do not modify this line

AlbumenJ (changes requested):

CrazyHZM (approved):

PR #10538, opened 2022-09-03
Add error code 0-3, and expose actual 'exception' of creating or deleting files in file cache merchanism.

win120a:


dubbo-registry/dubbo-registry-multicast/src/test/java/org/apache/dubbo/registry/multicast/MulticastRegistryTest.java
```diff
@@ -39,7 +39,7 @@
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-public class MulticastRegistryTest {
```
AlbumenJ, line 42: Why change this file

dubbo-common/src/test/java/org/apache/dubbo/common/utils/NetUtilsTest.java
```diff
@@ -42,51 +42,51 @@
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-public class NetUtilsTest {
```
AlbumenJ, line 45: Why change this file

AlbumenJ (changes requested):

dubbo-registry/dubbo-registry-multicast/src/test/java/org/apache/dubbo/registry/multicast/MulticastRegistryTest.java
win120a: Since Junit 5, it supports classes which has package private access, so it's redundant, and I have Sonarlint installed, it complains the modifier...

dubbo-common/src/test/java/org/apache/dubbo/common/utils/NetUtilsTest.java
win120a: Same as above...

win120a: Do I need to revert or ...?

PR #10598, opened 2022-09-09
Add error code 0-5, and more code of issuing error code 0-3, with code optimization.

win120a:


dkisser:
Test for api, don't reply

PR #10658, opened 2022-09-22
Provider supports thread pool isolation between services

BurningCN:

The description of the change will be added later....

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/transport/TripleHttp2FrameServerHandler.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/transport/TripleHttp2FrameServerHandler.java
@@ -48,17 +48,17 @@ public class TripleHttp2FrameServerHandler extends ChannelDuplexHandler {
         TripleHttp2FrameServerHandler.class);
     private final PathResolver pathResolver;
     private final FrameworkModel frameworkModel;
-    private final Executor executor;
     private final List<HeaderFilter> filters;
+    private final TripleExecutorSupport tripleExecutorSupport;
     private final String acceptEncoding;
 
     public TripleHttp2FrameServerHandler(
         FrameworkModel frameworkModel,
-        Executor executor,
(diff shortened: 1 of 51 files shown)


AlbumenJ (approved):
LGTM

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/transport/TripleIsolationExecutorSupport.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class TripleIsolationExecutorSupport extends AbstractIsolationExecutorSupport {
+    private static final ErrorTypeAwareLogger logger = LoggerFactory.getErrorTypeAwareLogger(TripleIsolationExecutorSupport.class);
+
+    public TripleIsolationExecutorSupport(URL url) {
+        super(url);
+    }
+
+    @Override
+    protected ServiceKey getServiceKey(Object data) {
```
EarthChen, line 35: The servicekey seems to have been obtained in onheader, can it be reused? But there is a problem here, onheader is already in the business thread pool, you may need to pay attention

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/transport/TripleHttp2FrameServerHandler.java
```diff
@@ -109,6 +107,8 @@ public void onDataRead(ChannelHandlerContext ctx, Http2DataFrame msg) throws Exc
     }
 
     public void onHeadersRead(ChannelHandlerContext ctx, Http2HeadersFrame msg) throws Exception {
+        Executor executor = executorSupport.getExecutor(msg);
```
EarthChen, line 110: I think that the calculation of the thread pool can be put into the onheader method, and the previous part of the logic can be executed in the eventloop. (Only basic judgment, no complex logic)

EarthChen (changes requested):

EarthChen (approved):
LGTM

PR #10752, opened 2022-10-15
Feature/self define exception body

aamingaa:
triple support selft define exception


AlbumenJ:
duplicated with #10667

PR #10781, opened 2022-10-22
remove useless code, the instanceof test always return true

zhangzq7:
remove useless code, the instanceof test will always return true

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/MockInvoker.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/MockInvoker.java
@@ -194,9 +194,11 @@ final public class MockInvoker<T> implements Invoker<T> {
                 + ", please check if there's mock class or instance implementing interface "
                 + serviceType.getName(), e);
         }
-        if (mockClass == null || !serviceType.isAssignableFrom(mockClass)) {
-            throw new IllegalStateException("The mock class " + mockClass.getName() +
-                " not implement interface " + serviceType.getName());
+        if (mockClass == null) {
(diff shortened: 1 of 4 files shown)


dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/MockInvoker.java
```diff
@@ -194,9 +194,11 @@ public static Object getMockObject(ExtensionDirector extensionDirector, String m
                 + ", please check if there's mock class or instance implementing interface "
                 + serviceType.getName(), e);
         }
-        if (mockClass == null || !serviceType.isAssignableFrom(mockClass)) {
-            throw new IllegalStateException("The mock class " + mockClass.getName() +
-                " not implement interface " + serviceType.getName());
+        if (mockClass == null) {
+            throw new IllegalStateException("not implement interface " + serviceType.getName());
```
AlbumenJ, line 198: log out `mockService` to help users find out why

AlbumenJ (changes requested):

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/support/MockInvoker.java
zhangzq7: mockService perhaps null

AlbumenJ: How about log out if is not null?

AlbumenJ (changes requested):

PR #10788, opened 2022-10-22
Update 'Actions' versions and 'Actions Runner' versions of GitHub Actions workflow file.

win120a:
With Node.js 12.x has been out of support <sup>[1]</sup>, it's recommended for us to upgrade all GitHub Actions to the version that uses Node.js 16.x.

[1] https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

## Brief changelog
Update some actions versions in workflow files.

_The codeconv actions will not be updated in this PR since I don't have the token to test. It will be updated by a separate PR._

### DIFF
--- a/.github/workflows/build-and-test-pr.yml
+++ b/.github/workflows/build-and-test-pr.yml
@@ -19,7 +19,7 @@ env:
 
 jobs:
   license:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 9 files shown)


AlbumenJ (changes requested):
Please fix conflicts

AlbumenJ (approved):
LGTM

PR #10797, opened 2022-10-23
fix no such extension by name netty4

icodening:
fix no such extension by name netty4

image


AlbumenJ (commented):
I wonder if user configure `netty 3` in properties what will happen?

AlbumenJ (approved):
Others look good to me

icodening:
> I wonder if user configure `netty 3` in properties what will happen?

`No such extension org.apache.dubbo.remoting.api.pu.PortUnificationTransporter by name netty3`
current version does not support `netty3`.  change `netty4` to be compatible with lower version

AlbumenJ:
> > I wonder if user configure `netty 3` in properties what will happen?
> 
> `No such extension org.apache.dubbo.remoting.api.pu.PortUnificationTransporter by name netty3` current version does not support `netty3`. change `netty4` to be compatible with lower version

I have create a issue (#10812) for this. We can continue track in that issue.

PR #10825, opened 2022-10-25
Fix assertion error and class cast exception in unit tests.

win120a:
Fix some errors in unit tests.

## Brief changelog
Correct the types in Mockito invocation and fix incorrect concurrent library uses.


AlbumenJ (approved):
LGTM

PR #10854, opened 2022-10-30
support encode in user thread, optimize decode in user thread

icodening:
支持用户线程处理encode，eventloop线程只负责发送消息，以此提高dubbo性能
参数: `encode.in.io=false`

## Brief changelog
1.`NettyChannel`encode message，降低`eventloop`上串行encode的压力
2.`DecodeableRpcInvocation`支持懒加载，以减少该对象在`eventloop`上初始化的耗时

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
@@ -603,5 +603,5 @@ public interface CommonConstants {
 
 
     String ENCODE_IN_IO_THREAD_KEY = "encode.in.io";
-    boolean DEFAULT_ENCODE_IN_IO_THREAD = true;
+    boolean DEFAULT_ENCODE_IN_IO_THREAD = false;
 }
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
@@ -600,4 +600,8 @@ public interface CommonConstants {
      *
      */
     String DUBBO_JSTACK_MAXLINE = "dubbo.jstack-dump.max-line";
+
+
+    String ENCODE_IN_IO_THREAD_KEY = "encode.in.io";
+    boolean DEFAULT_ENCODE_IN_IO_THREAD = true;
 }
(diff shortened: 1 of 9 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
```diff
@@ -600,4 +600,8 @@ public interface CommonConstants {
      *
      */
     String DUBBO_JSTACK_MAXLINE = "dubbo.jstack-dump.max-line";
+
+
+    String ENCODE_IN_IO_THREAD_KEY = "encode.in.io";
+    boolean DEFAULT_ENCODE_IN_IO_THREAD = true;
```
AlbumenJ, line 606: why not set `false` as default

AlbumenJ (changes requested):

AlbumenJ (changes requested):
Please fix uts

dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
icodening: I think this is a new feature, maybe use `false` as default

AlbumenJ: If this doesn't break usage for some users, we can enable it by default. By the way, this PR will be merged into the `3.2` branch, which can accept some breaking changes.

icodening: OK , get

icodening:
benchmark数据对比，可以看到`大报文`的情况下性能有极大幅度的提升
````
用户线程encode
Client.createUser  thrpt    3  87.469 ± 25.680  ops/ms
Client.existUser   thrpt    3  97.280 ±  9.747  ops/ms
Client.getUser     thrpt    3  85.073 ± 57.368  ops/ms
Client.listUser    thrpt    3  51.223 ± 22.005  ops/ms
````

````
未合并前的最新3.2版本
Benchmark           Mode  Cnt   Score    Error   Units
Client.createUser  thrpt    3  81.905 ± 20.098  ops/ms
Client.existUser   thrpt    3  96.708 ± 17.233  ops/ms
Client.getUser     thrpt    3  87.616 ±  8.268  ops/ms
Client.listUser    thrpt    3  34.187 ± 13.755  ops/ms
````

AlbumenJ (approved):
LGTM

PR #10921, opened 2022-11-11
fix NPE bug of URL

wxbty:
What is the purpose of the change

When the org.apache.dubbo.common.URL uses setHost, setPassword, there is an NPE risk of the urlAddress property, because the URL has an empty constructor. For example, when zk pushes the url to the listener, it will call the DefaultServiceInstance.toURL method to construct the URL

    @Override
        public InstanceAddressURL toURL(String protocol) {
            if (instanceAddressURL == null) {
                instanceAddressURL = new InstanceAddressURL(this, serviceMetadata, protocol);
            }
            return instanceAddressURL;
        }

The constructor of InstanceAddressURL here only sets its own properties, which is equivalent to calling the empty constructor of the parent class URL. The constructed URL instance urlAddress property is empty. When someone calls the setHost of this URL, an NPE error will be reported, such as this bug #issue https://github.com/apache/dubbo-spi-extensions/issues/166

The root cause of this error is the dubbo URL design problem, dubbo-spi-extensions did not judge, and then stepped into the trap.

Brief changelog

Verifying this change

Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
@@ -480,16 +480,12 @@ class URL implements Serializable {
         } else {
             host = address;
         }
-        URL url;
         if (urlAddress == null) {
-            url = new URL();
-            url.setHost(host);
-            url.setPort(port);
+            return new ServiceConfigURL(getProtocol(), host, port, getPath(), getParameters());
         } else {
             URLAddress newURLAddress = urlAddress.setAddress(host, port);
-            url = returnURL(newURLAddress);
+            return returnURL(newURLAddress);
         }
-        return url;
     }
 
     public String getIp() {
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
@@ -350,7 +350,7 @@ class URL implements Serializable {
 
     public URL setProtocol(String protocol) {
         if (urlAddress == null) {
-            return new ServiceConfigURL(getProtocol(), getHost(), getPort(), getPath(), getParameters());
+            return new ServiceConfigURL(protocol, getHost(), getPort(), getPath(), getParameters());
         } else {
             URLAddress newURLAddress = urlAddress.setProtocol(protocol);
             return returnURL(newURLAddress);
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
@@ -529,15 +529,12 @@ class URL implements Serializable {
     }
 
     public URL setPath(String path) {
-        URL url;
         if (urlAddress == null) {
-            url = new URL();
-            url.setPath(path);
+            return new ServiceConfigURL(getProtocol(), getHost(), getPort(), path, getParameters());
         } else {
             URLAddress newURLAddress = urlAddress.setPath(path);
-            url = returnURL(newURLAddress);
+            return returnURL(newURLAddress);
         }
-        return url;
     }
 
     public String getAbsolutePath() {
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
@@ -433,15 +433,13 @@ class URL implements Serializable {
     }
 
     public URL setHost(String host) {
-        URL url;
+```suggestion
         if (urlAddress == null) {
-            url = new URL();
-            url.setHost(host);
+            return new ServiceConfigURL(getProtocol(), host, getPort(), getPath(), getParameters());
         } else {
             URLAddress newURLAddress = urlAddress.setHost(host);
-            url = returnURL(newURLAddress);
+            return returnURL(newURLAddress);
         }
-        return url;
     }
 
 
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
@@ -349,8 +349,12 @@ class URL implements Serializable {
     }
 
     public URL setProtocol(String protocol) {
-        URLAddress newURLAddress = urlAddress.setProtocol(protocol);
-        return returnURL(newURLAddress);
+        if (urlAddress == null) {
+            return new ServiceConfigURL(getProtocol(), getHost(), getPort(), getPath(), getParameters());
+        } else {
+            URLAddress newURLAddress = urlAddress.setProtocol(protocol);
+            return returnURL(newURLAddress);
+        }
     }
 
     public String getUsername() {
@@ -358,8 +362,12 @@ class URL implements Serializable {
     }
 
     public URL setUsername(String username) {
-        URLAddress newURLAddress = urlAddress.setUsername(username);
-        return returnURL(newURLAddress);
+        if (urlAddress == null) {
+            return new ServiceConfigURL(getProtocol(), getHost(), getPort(), getPath(), getParameters()).setUsername(username);
+        } else {
+            URLAddress newURLAddress = urlAddress.setUsername(username);
+            return returnURL(newURLAddress);
+        }
     }
 
     public String getPassword() {
@@ -368,7 +376,7 @@ class URL implements Serializable {
 
     public URL setPassword(String password) {
         if (urlAddress == null) {
-            return new ServiceConfigURL(getProtocol(), getHost(), port, getPath(), getParameters()).setPassword(password);
+            return new ServiceConfigURL(getProtocol(), getHost(), getPort(), getPath(), getParameters()).setPassword(password);
         } else {
             URLAddress newURLAddress = urlAddress.setPassword(password);
             return returnURL(newURLAddress);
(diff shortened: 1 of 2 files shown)


AlbumenJ (commented):
Let `InstanceAddressURL` override these methods would be better

chickenlj:
Yeah, I agree with Albumen

wxbty:
> Let `InstanceAddressURL` override these methods would be better

If urlAddress is empty, does it simply return null? I think we need the necessary NPE check for the URL, which has other subclasses.

wxbty:
hi, @AlbumenJ
Do you think this modification is reasonable? The URL constructor does not use mandatory assignment attributes, so null checking is necessary, the base class should be more robust, and dubbo's extension project dependencies are confusing, and relying on subclasses may still occur frequently NPE.

AlbumenJ:
> > Let `InstanceAddressURL` override these methods would be better
> 
> If urlAddress is empty, does it simply return null? I think we need the necessary NPE check for the URL, which has other subclasses.

In `InstanceAddressURL`, a new URL instance should be created if there are some modifications. The new URL type can be `DubboServiceAddressURL`.

The case of `urlParam` is null is only happends in `InstanceAddressURL`, and override those origin method would fix.

wxbty:
> Do you think this modification is reasonable? The URL constructor does not use mandatory assignment attributes, so null checking is necessary, the base class should be more robust, and dubbo's extension project dependencies are confusing, and relying on subclasses may still occur frequently NPE.

If you think don't need to modify it, I will close it. But still feel that there is a risk of NPE like this, which looks badly designed

AlbumenJ:
> If you think don't need to modify it, I will close it. But still feel that there is a risk of NPE like this, which looks badly designed

Bugfix is needed. 

To make it clarify, I will give some suggest changes for it.

```java
public class InstanceAddressURL {
    @Override
    public URL setHost(String host) {
        // copy current URL into a new DubboServiceAddressURL
        DubboServiceAddressURL copyOf = createDubboServiceAddressURL(this);
      return copyOf.setHost(host);
    }
} 
```

AlbumenJ:
BTW, this issue may be already fixed in #10256.

AlbumenJ:
https://github.com/apache/dubbo/pull/10256/files#diff-8cea6b30b1f89caf0117eb66c7abca7e86e46ecb1f232a8508290ab40ca70ffd

AlbumenJ:
> I noticed this change, it will be fixed in version 3.1, what I want to show is that the parent URL needs to guarantee its NPE check, so that no matter what, Exception won't happen on the path of the URL. If URL defines a new subclass, the author will notice that setHost may return null, and he needs to handle this exception. With the current way of writing, he will think that the URL must be returned. I think, all base class methods should consider null scenarios.

The original solution can indeed solve the NPE problem to a certain extent. (You can reset this PR to the original commit, and I will merge it later.)

In the future, we hope to turn the URL object into an interface at an appropriate time(may be 3.2). The abstract class URL no longer provides attribute storage and implementation, but only provides interface definition.
Based on this, Dubbo's URL system is refactored to avoid the current need for various overriding methods. (Would you please help do this?)

wxbty:
> > I noticed this change, it will be fixed in version 3.1, what I want to show is that the parent URL needs to guarantee its NPE check, so that no matter what, Exception won't happen on the path of the URL. If URL defines a new subclass, the author will notice that setHost may return null, and he needs to handle this exception. With the current way of writing, he will think that the URL must be returned. I think, all base class methods should consider null scenarios.
> 
> The original solution can indeed solve the NPE problem to a certain extent. (You can reset this PR to the original commit, and I will merge it later.)
> 
> In the future, we hope to turn the URL object into an interface at an appropriate time(may be 3.2). The abstract class URL no longer provides attribute storage and implementation, but only provides interface definition. Based on this, Dubbo's URL system is refactored to avoid the current need for various overriding methods. (Would you please help do this?)

Ok, I'll be happy to do it

dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
```diff
@@ -499,8 +529,15 @@ public String getPath() {
...
-        return returnURL(newURLAddress);
+        URL url;
+        if (urlAddress == null) {
+            url = new URL();
+            url.setPath(path);
+        } else {
+            URLAddress newURLAddress = urlAddress.setPath(path);
+            url = returnURL(newURLAddress);
+        }
+        return url;
```
AlbumenJ, line 540:
```suggestion
        if (urlAddress == null) {
            return new ServiceConfigURL(getProtocol(), getHost(), getPort(), path, getParameters());
        } else {
            URLAddress newURLAddress = urlAddress.setPath(path);
            return returnURL(newURLAddress);
        }
```

dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
```diff
@@ -458,12 +480,20 @@ public URL setAddress(String address) {
...
+        URL url;
+        if (urlAddress == null) {
+            url = new URL();
+            url.setHost(host);
+            url.setPort(port);
+        } else {
+            URLAddress newURLAddress = urlAddress.setAddress(host, port);
+            url = returnURL(newURLAddress);
+        }
+        return url;
```
AlbumenJ, line 492:
```suggestion
        if (urlAddress == null) {
            return new ServiceConfigURL(getProtocol(), host, port, getPath(), getParameters());
        } else {
            URLAddress newURLAddress = urlAddress.setAddress(host, port);
            return returnURL(newURLAddress);
        }
```

dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
```diff
@@ -435,8 +450,15 @@ public int getPort() {
...
-        return returnURL(newURLAddress);
+        URL url;
+        if (urlAddress == null) {
+            url = new URL();
+            url.setPort(port);
+        } else {
+            URLAddress newURLAddress = urlAddress.setPort(port);
+            url = returnURL(newURLAddress);
+        }
+        return url;
```
AlbumenJ, line 461:
```suggestion
        if (urlAddress == null) {
            return new ServiceConfigURL(getProtocol(), getHost(), port, getPath(), getParameters());
        } else {
            URLAddress newURLAddress = urlAddress.setPort(port);
            return returnURL(newURLAddress);
        }
```

dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
```diff
@@ -425,8 +433,15 @@ public String getHost() {
...
-        return returnURL(newURLAddress);
+        URL url;
+        if (urlAddress == null) {
+            url = new URL();
+            url.setHost(host);
+        } else {
+            URLAddress newURLAddress = urlAddress.setHost(host);
+            url = returnURL(newURLAddress);
+        }
+        return url;
```
AlbumenJ, line 444:
```suggestion
```suggestion
        if (urlAddress == null) {
            return new ServiceConfigURL(getProtocol(), host, getPort(), getPath(), getParameters());
        } else {
            URLAddress newURLAddress = urlAddress.setHost(host);
            return returnURL(newURLAddress);
        }
```
```

**Review by AlbumenJ — CHANGES_REQUESTED** [review]

File: dubbo-common/src/main/java/org/apache/dubbo/common/URL.java [hunk 5]
```diff
@@ -425,8 +433,13 @@ public String getHost() {
     }
 
     public URL setHost(String host) {
-        URLAddress newURLAddress = urlAddress.setHost(host);
-        return returnURL(newURLAddress);
+```suggestion
```
AlbumenJ, line 436: remove this breaking line

dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
```diff
@@ -367,8 +367,16 @@ public String getPassword() {
...
+
+        URL url;
+        if (urlAddress == null) {
+            url = new URL();
+            url.setPassword(password);
+        } else {
+            URLAddress newURLAddress = urlAddress.setPassword(password);
+            url = returnURL(newURLAddress);
+        }
+        return url;
```
AlbumenJ, line 379:
```suggestion
        if (urlAddress == null) {
            return new ServiceConfigURL(getProtocol(), getHost(), port, getPath(), getParameters()).setPassword(password);
        } else {
            URLAddress newURLAddress = urlAddress.setPassword(password);
            return returnURL(newURLAddress);
        }
```

AlbumenJ: BTW, please fix `setUsername`, `setProtocol` and etc

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
```diff
@@ -349,26 +349,38 @@ public String getProtocol() {
     }
 
     public URL setProtocol(String protocol) {
-        URLAddress newURLAddress = urlAddress.setProtocol(protocol);
-        return returnURL(newURLAddress);
+        if (urlAddress == null) {
+            return new ServiceConfigURL(getProtocol(), getHost(), getPort(), getPath(), getParameters());
```
AlbumenJ, line 353:
```suggestion
            return new ServiceConfigURL(protocol, getHost(), getPort(), getPath(), getParameters());
```

AlbumenJ (changes requested):

AlbumenJ (approved):
LGTM

PR #10923, opened 2022-11-12
Distinguish demo for users and developers

chickenlj:
* `dubbo-demo` module is simplified for users
* developers please use  or add demos inside `dubbo-test`


AlbumenJ (commented):
Please submit to 3.1 branch

PR #10998, opened 2022-11-23
XSD file append prefer-serialization

tonycody:
XSD file append prefer-serialization fix


chickenlj:
Thanks. I will look into this PR

AlbumenJ (approved):
LGTM

PR #11045, opened 2022-11-28
fix issue:check whether invokers list has init before.

carlos-heshq:
fix bug:check whether invokers list has init before.
changelog:
check whether invokers list has init first.


AlbumenJ:
Already fixed in https://github.com/apache/dubbo/pull/11046

Issue #11059, opened 2022-12-01
Provider export tri protocol with declare context-path, consumer got a "Bad path format xxx" exception


twz007:
- [ ] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 3.1.1
* Operating System version: win10
* Java version: 1.8

### Steps to reproduce this issue

1. 提供者声明 `dubbo.protocols.tri.contextpath: ${spring.application.name}`
2. 消费者以tri协议调用 `@DubboReference(protocol = "tri")`
3. 消费者收到报错 `org.apache.dubbo.rpc.StatusRpcException: UNIMPLEMENTED : Bad path format:/context-path/com.a.b/c`

定位报错代码处：`org.apache.dubbo.rpc.protocol.tri.stream.TripleServerStream.ServerTransportObserver#processHeader`
```java
    // ......  省略
     //   path = /context-path/com.a.b/c
      String[] parts = path.split("/");
     // parts.length = 4 报错
      if (parts.length != 3) {
          responseErr(TriRpcStatus.UNIMPLEMENTED.withDescription("Bad path format:" + path));
          return;
      }
     // ...... 省略
     
```
上述代码限制了长度一定为3，实际可能为4

Pls. provide [GitHub address] to reproduce this issue.

### Expected Behavior
兼容掉声明了context-path的情况

### Actual Behavior
未兼容

If there is an exception, please attach the exception trace:

```
Just put your stack trace here!
```

guohao:
It is not a bug .
`Path`  is  defined here https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

EarthChen:
I would like to ask, what purpose do you want to use this configuration for?

twz007:
> It is not a bug . `Path` is defined here https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

It seems that the triple design does not support context-path at the beginning.

twz007:
> I would like to ask, what purpose do you want to use this configuration for?

Some of our old services initially only supported jsonrpc, http, and rest protocols. The simple way is to directly modify the protocol to triple without deleting the context-path.

EarthChen:
> > I would like to ask, what purpose do you want to use this configuration for?
> 
[...]

PR #11072, opened 2022-12-05
fixed tri protocol with declare context-path, consumer got a "Bad path format xxx" exception #11059

twz007:

#11059

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/stream/TripleServerStream.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/stream/TripleServerStream.java
@@ -336,12 +336,21 @@ public class TripleServerStream extends AbstractStream implements ServerStream {
             }
 
             String[] parts = path.split("/");
-            if (parts.length != 3) {
+            String serviceName;
+            String originalMethodName;
+            if (parts.length == 3) {
+                serviceName = parts[1];
+                originalMethodName = parts[2];
+            } else if (parts.length == 4) {
+                if (LOGGER.isWarnEnabled()) {
+                    LOGGER.warn(String.format("context-path: [%s] will be ignored in the triple protocol, please remove this configuration.", parts[1]));
+                }
+                serviceName = parts[2];
+                originalMethodName = parts[3];
+            } else {
                 responseErr(TriRpcStatus.UNIMPLEMENTED.withDescription("Bad path format:" + path));
                 return;
             }
-            String serviceName = parts[1];
-            String originalMethodName = parts[2];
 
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/stream/TripleServerStream.java
```diff
@@ -336,12 +336,21 @@ private void processHeader(Http2Headers headers, boolean endStream) {
...
+            if (parts.length == 3) {
+                serviceName = parts[1];
+                originalMethodName = parts[2];
+            } else if (parts.length == 4) {
+                if (LOGGER.isWarnEnabled()) {
+                    LOGGER.warn(String.format("context-path: [%s] will be ignored in the triple protocol, please remove this configuration.", parts[1]));
+                }
+                serviceName = parts[2];
+                originalMethodName = parts[3];
+            } else {
```
guohao, line 350: Do fallback here may looks work from remote side but failed on intermediate proxies.
May it be better to just log and `responseErr`?

EarthChen (changes requested):
It is better not to do compatibility at the server protocol layer, it is better to modify when reading the configuration in the client framework as explained in the issue

twz007:
> It is better not to do compatibility at the server protocol layer, it is better to modify when reading the configuration in the client framework as explained in the issue

What consumers read is the `context-path` registered by the provider. It seems a bit awkward for consumers to be compatible with the provider. What we want is to let the provider modify their configuration instead of hiding it, or manually to inform them that there is a problem with the configuration.

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/stream/TripleServerStream.java
twz007: I don't understand what the intermediate proxies refers to. Do you mean that it is better to modify it on the client side?

EarthChen:
> > It is better not to do compatibility at the server protocol layer, it is better to modify when reading the configuration in the client framework as explained in the issue
> 
> What consumers read is the `context-path` registered by the provider. It seems a bit awkward for consumers to be compatible with the provider. What we want is to let the provider modify their configuration instead of hiding it, or manually to inform them that there is a problem with the configuration.

Modifying from the provider side somewhat breaks the semantics of the protocol, and it doesn’t make sense for the provider side to print warn all the time, we hope consumers can find this problem by themselves.

CrazyHZM:
@twz007 Can you recommit to the 3.2 branch? 3.1 no longer makes this convenient change.

PR #11084, opened 2022-12-06
[DUBBO-11047][issue]Remove direct dependency of protobuf-java

wlazjr:

replace protobuf depency in triple rpc with customer encoding/decoding

## Brief changelog
add TripleCustomerProtocolWapper.java
 replace TripleRequestWrapper.TripleRequestWrapper with TripleCustomerProtocolWapper.TripleRequestWrapper
 replace TripleRequestWrapper.TripleResponseWrapper with TripleCustomerProtocolWapper.TripleResponseWrapper

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/service/TriBuiltinService.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/service/TriBuiltinService.java
@@ -48,6 +48,8 @@ public class TriBuiltinService {
 
     private Health healthService;
 
+    private FrameworkModel frameworkModel;
+
     private ReflectionV1AlphaService reflectionServiceV1Alpha;
     private HealthStatusManager healthStatusManager;
     private Configuration config = ConfigurationUtils.getGlobalConfiguration(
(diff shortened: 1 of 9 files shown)


EarthChen (changes requested):
fix ut & it

EarthChen:
If we remove the pb dependency, the built-in service also needs to be moved out to other modules, otherwise it will still fail to start due to the built-in service at startup.

wlazjr:
> ### This branch has no conflicts with the base branch



> If we remove the pb dependency, the built-in service also needs to be moved out to other modules, otherwise it will still fail to start due to the built-in service at startup.

ok，got it

guohao:
Changes:
1. Set `BuiltinService` to `disable` by default
2. Check the  `BuiltinService` if is `enable` in `TripleProtocol`
3. Add check condition at `Status` usages

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/service/TriBuiltinService.java
```diff
@@ -67,6 +74,11 @@ public void init() {
         }
     }
 
+    public boolean enable(){
+        return config.getBoolean(TRI_BUILTIN_SERVICE_INIT, false);
```
AlbumenJ, line 78: Can we check if users has add the `protobuf-java` dependency?

wlazjr: First of all, we do not know the real whole sets of versions that works fine, and in the future as protobuf-java update, we may need hava a glance at this; on the other hand, i think give the load permissions to user is a better choice.

guohao (approved):
LGTM

EarthChen (approved):
LGTM

PR #11060, opened 2022-12-01
[Feature] Provides a custom extension for service not found

wxbty:
## What is the purpose of the change
Provides a custom extension for service not found.
Issue Number: close #10950 

## Brief changelog
The spi extension ExceptionProcessor is provided to allow custom exception handling when an exception occurs during the decode phase. At the same time, it provides an all-round generalization of the server, which can accept any interface, thereby providing enhanced mock or gateway capabilities. The new server-side generalization relies on the default ExceptionProcessor implementation.
See class annotations for details.

Corresponding documents will be written before merge

## Verifying this change
Pass testcase and local mock service test.

## Checklist
- [x] Make sure there is a [GitHub_issue](https://github.com/apache/dubbo/issues) field for the change (usually before you start working on it). Trivial changes like typos do not require a GitHub issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [ ] Each commit in the pull request should have a meaningful subject line and body.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Check if is necessary to patch to Dubbo 3 if you are work on Dubbo 2.7
- [ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add sample in [dubbo samples](https://github.com/apache/dubbo-samples) project.
- [ ] Add some description to [dubbo-website](https://github.com/apache/dubbo-website) project if you are requesting to add a feature.
- [ ] GitHub Actions works fine on your own branch.
- [ ] If this contribution is large, please follow the [Software Donation Guide](https://github.com/apache/dubbo/wiki/Software-donation-guide).

PR #11088, opened 2022-12-07
Feature service not found

### LINKED ISSUE #10950: Dubbo lacks extensions for exceptions
We are working on a dubbo gateway and need to support mutual access between different network clusters. This should be a common scenario. This requires dubbo to do two things for us:

1. Dynamically modify the ip and change the access of the local cluster to remote. This extension can be done, although there are some bugs, https://dubbo.apache.org/zh/docs3-v2/java-sdk/advanced-features-and-usage/service/specify-ip/

2. Forward the dubbo request. When the gateway receives the access request from cluster A, it should forward it to cluster B (now it returns an exception).
  From the code execution process.

  current:
  gateway-dubbo-provider receives the request -> parses -> judges whether the request interface has a corresponding cache URL -> no -> throws IOException -> netty returns

  ```java
  void handleRequest(final ExchangeChannel channel, Request req) throws RemotingException {
          Response res = new Response(req.getId(), req.getVersion());
          if (req.isBroken()) {
              Object data = req.getData();
              String msg;
              if (data == null) {
                  msg = null;
              } else if (data instanceof Throwable) {
                  msg = StringUtils.toString((Throwable) data);
              } else {
                  msg = data.toString();
              }
              res.setErrorMessage("Fail to decode request due to: " + msg);
              res.setStatus(Response.BAD_REQUEST);

              channel.send(res);
              return;
          }
  }
  ```

  ​

  expect:
  gateway-dubbo-provider receives the request -> parses -> judges whether the request interface has a corresponding cache URL -> no -> obtains the exception handler through spi (the request can be forwarded here) -> netty returns

```java
  void handleRequest(final ExchangeChannel channel, Request req) throws RemotingException {
          Response res = new Response(req.getId(), req.getVersion());
          if (req.isBroken()) {
              Object data = req.getData();
           //expert begin
          List<ExceptionHandler> list = spiExtensions.listByType(ExceptionHandler); 
          list.forEach(handler.handle(path,version))
           //expert end
              String msg;
              if (data == null) {
                  msg = null;
              } else if (data instanceof Throwable) {
                  msg = StringUtils.toString((Throwable) data);
              } else {
                  msg = data.toString();
              }
              res.setErrorMessage("Fail to decode request due to: " + msg);
              res.setStatus(Response.BAD_REQUEST);

              channel.send(res);
              return;
          }
  }
```

Since dubbo.extension provides dynamic ip replacement, then we should consider whether the target ip can be accepted and processed. I think this exception handler is an extension function of the dubbo.extension package.
We can provide ExceptionHandler similar to springboot, which can be more flexible

wxbty:
Provides a custom extension for service not found.
Issue Number: close #10950 
rebase of #11060 

## Brief changelog
The spi extension ExceptionProcessor is provided to allow custom exception handling when an exception occurs during the decode phase. At the same time, it provides an all-round generalization of the server, which can accept any interface, thereby providing enhanced mock or gateway capabilities. The new server-side generalization relies on the default ExceptionProcessor implementation.
See class comments for details.

## Checklist

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/ExceptionProcessor.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/ExceptionProcessor.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.dubbo.remoting;
+
+import org.apache.dubbo.common.extension.SPI;
+import org.apache.dubbo.remoting.exchange.ExchangeChannel;
+import org.apache.dubbo.remoting.exchange.Request;
+
+/**
+ * <p>Interface for handling exceptions in specific spi-extended classes.
+ *
+ * <p>Mainly works in the decode phase {@link Decodeable#decode()},
+ * and can only handle exception types in this phase.
+ *
+ * <p>Allow three kinds of customized operations, customize abnormal results and return,
+ * customize normal results and return, interrupt and retry the process.
+ * @since 3.2.0
+ */
+@SPI("snf")
+public interface ExceptionProcessor {
+
+    /**
+     * Whether to return to the client immediately when an exception
+     * with the specified parameter is encountered
+     *
+     * @param throwable specified exception
+     * @since 3.2.0
+     */
+    boolean shouldReturnError(Throwable throwable);
+
+    /**
+     * <p>When encountering an exception in the decode phase, allow developer customize the behavior before return.
+     * <p>If exceptions are allowed and want to reprocess the request,
+     * you can add custom content and throw a retry exception{@link RetryHandleException} so that decode will continue processing.
+     * <p>Decode will not re-read the read content, so you need to save the decoded content, refer to default impl {snf}.
+     * <p>The number of times decode handle exceptions will not exceed 2.
+     * <p>If decode still cannot process the request, the error message of retry exception will be returned eventually,
+     * and the real reason that cannot be processed needs to be set instead of the literal exception information of retry.
+     * @return custom return information,
+     * normal or abnormal return is determined by
+     * {@link ExceptionProcessor#shouldReturnError(Throwable)}}
+     * @since 3.2.0
+     */
+    String wrapAndHandleException(ExchangeChannel channel, Request req) throws RetryHandleException;
+}
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/Request.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/Request.java
@@ -41,6 +41,8 @@ public class Request {
 
     private Object mData;
 
+    private Throwable mError;
+
     public Request() {
         mId = newId();
     }
@@ -113,6 +115,14 @@ public class Request {
 
     public void setData(Object msg) {
         mData = msg;
+    }
+
+    public Throwable getError() {
+        return mError;
+    }
+
+    public void setError(Throwable error) {
+        this.mError = error;
     }
 
     public boolean isHeartbeat() {
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeHandler.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeHandler.java
@@ -17,6 +17,7 @@
 package org.apache.dubbo.remoting.exchange.support.header;
 
 import org.apache.dubbo.common.URL;
+import org.apache.dubbo.common.extension.ExtensionLoader;
 import org.apache.dubbo.common.logger.ErrorTypeAwareLogger;
 import org.apache.dubbo.common.logger.LoggerFactory;
 import org.apache.dubbo.common.utils.NetUtils;
@@ -33,8 +34,11 @@ import org.apache.dubbo.remoting.exchange.Response;
 import org.apache.dubbo.remoting.exchange.support.DefaultFuture;
 import org.apache.dubbo.remoting.exchange.support.MultiMessage;
 import org.apache.dubbo.remoting.transport.ChannelHandlerDelegate;
+import org.apache.dubbo.remoting.ExceptionProcessor;
+import org.apache.dubbo.rpc.model.ApplicationModel;
 
 import java.net.InetSocketAddress;
+import java.util.Optional;
 import java.util.concurrent.CompletionStage;
 
 import static org.apache.dubbo.common.constants.CommonConstants.READONLY_EVENT;
@@ -82,20 +86,35 @@ public class HeaderExchangeHandler implements ChannelHandlerDelegate {
         Response res = new Response(req.getId(), req.getVersion());
         if (req.isBroken()) {
             Object data = req.getData();
+            Throwable error = req.getError();
 
             String msg;
+            boolean returnError = true;
             if (data == null) {
                 msg = null;
-            } else if (data instanceof Throwable) {
-                msg = StringUtils.toString((Throwable) data);
             } else {
                 msg = data.toString();
             }
-            res.setErrorMessage("Fail to decode request due to: " + msg);
-            res.setStatus(Response.BAD_REQUEST);
 
-            channel.send(res);
-            return;
+            if (error != null) {
+                // Give ExceptionProcessors a chance to retry request handle or custom exception information.
+                String exPs = ApplicationModel.defaultModel().getCurrentConfig().getParameters().get("exception.processor");
+                ExtensionLoader<ExceptionProcessor> extensionLoader = ApplicationModel.defaultModel().getDefaultModule().getExtensionLoader(ExceptionProcessor.class);
+                ExceptionProcessor expProcessor = exPs == null ? extensionLoader.getDefaultExtension() : extensionLoader.getOrDefaultExtension(exPs);
+                msg = StringUtils.toString(error);
+                returnError = expProcessor.shouldReturnError(error);
+                // If returnError is true, custom error message, return directly
+                // Or interrupt the error message, customize the process, process req, and continue to return normal information
+                msg = Optional.ofNullable(expProcessor.wrapAndHandleException(channel, req)).orElse(msg);
+            }
+
+            if (returnError) {
+                res.setErrorMessage("Fail to decode request due to: " + msg);
+                res.setStatus(Response.BAD_REQUEST);
+
+                channel.send(res);
+                return;
+            }
         }
         // find handler by message class.
         Object msg = req.getData();
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/transport/DecodeHandler.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/transport/DecodeHandler.java
@@ -23,6 +23,7 @@ import org.apache.dubbo.remoting.Channel;
 import org.apache.dubbo.remoting.ChannelHandler;
 import org.apache.dubbo.remoting.Decodeable;
 import org.apache.dubbo.remoting.RemotingException;
+import org.apache.dubbo.remoting.RetryHandleException;
 import org.apache.dubbo.remoting.exchange.Request;
 import org.apache.dubbo.remoting.exchange.Response;
 
@@ -38,19 +39,32 @@ public class DecodeHandler extends AbstractChannelHandlerDelegate {
 
     @Override
     public void received(Channel channel, Object message) throws RemotingException {
+        doRecveived(message);
+
+        try {
+            handler.received(channel, message);
+        } catch (RetryHandleException e) {
+            // Retry only once
+            doRecveived(message);
+            if (message instanceof Request) {
+                ((Request) message).setBroken(false);
+            } else {
+                throw new RemotingException(channel, "Unknown error encountered when retry handle: " + e.getMessage());
+            }
+            handler.received(channel, message);
+        }
+    }
+
+    private void doRecveived(Object message) {
         if (message instanceof Decodeable) {
             decode(message);
         }
-
         if (message instanceof Request) {
             decode(((Request) message).getData());
         }
-
         if (message instanceof Response) {
             decode(((Response) message).getResult());
         }
-
-        handler.received(channel, message);
     }
 
     private void decode(Object message) {
--- a/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DecodeableRpcInvocation.java
+++ b/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DecodeableRpcInvocation.java
@@ -29,6 +29,7 @@ import org.apache.dubbo.common.utils.StringUtils;
 import org.apache.dubbo.remoting.Channel;
 import org.apache.dubbo.remoting.Codec;
 import org.apache.dubbo.remoting.Decodeable;
+import org.apache.dubbo.remoting.ServiceNotFoundException;
 import org.apache.dubbo.remoting.exchange.Request;
 import org.apache.dubbo.remoting.transport.CodecSupport;
 import org.apache.dubbo.rpc.RpcInvocation;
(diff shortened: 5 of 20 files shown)


wxbty:
@AlbumenJ   I found a serious problem. After switching the address, if the passed parameter is a custom class (the previous test provider used the same api package as the consumer, but in gateway mode, the provider may not have any business class), it cannot be deserialized on the provider side change. I used a method similar to server-side generalization, changing the parameter type to JavaBeanDescriptor type on the consumer side, and the OmnipotentFilter on the provider side to judge and process,  the client uses the extension package, #https://github.com/apache/dubbo-spi-extensions/pull/189 ,  please take a look again

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeHandler.java
```diff
@@ -82,15 +87,31 @@ void handleRequest(final ExchangeChannel channel, Request req) throws RemotingEx
...
                 msg = data.toString();
             }
+
+            if (error != null) {
+                msg = StringUtils.toString(error);
+                // Give ExceptionProcessors a chance to retry request handle or custom exception information.
+                String exPs = ApplicationModel.defaultModel().getCurrentConfig().getParameters().get(EXCEPTION_PROCESSOR_KEY);
+                if (StringUtils.isNotBlank(exPs)) {
+                    ExtensionLoader<ExceptionProcessor> extensionLoader = ApplicationModel.defaultModel().getDefaultModule().getExtensionLoader(ExceptionProcessor.class);
+                    ExceptionProcessor expProcessor = extensionLoader.getOrDefaultExtension(exPs);
```
AlbumenJ, line 105: Prevent use `ApplicationModel.defaultModel()` here.

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DecodeableRpcInvocation.java
```diff
@@ -85,23 +91,68 @@ public DecodeableRpcInvocation(FrameworkModel frameworkModel, Channel channel, R
...
+                }
+                // Whether the process ends normally or abnormally (retry at most once), resources must be released
+            }
+        }
+    }
+
+    @Override
+    public void retry() {
+
+        String exPs = ApplicationModel.defaultModel().getCurrentConfig().getParameters().get(EXCEPTION_PROCESSOR_KEY);
```
AlbumenJ, line 136: `DecodeableRpcInvocation` here is a framework level class, and cannot get application model here.

AlbumenJ (changes requested):

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeHandler.java
wxbty: use System.getProperty instead

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/transport/DecodeHandler.java
```diff
@@ -38,19 +39,33 @@ public DecodeHandler(ChannelHandler handler) {
...
     public void received(Channel channel, Object message) throws RemotingException {
+        doRecveived(message);
+
+
+        try {
+            handler.received(channel, message);
+        } catch (RetryHandleException e) {
+            if (message instanceof Request) {
+                Request request = (Request) message;
+                retry(request.getData());
```
mytang0, line 50: The catch problem is the same as above

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/transport/DecodeHandler.java
```diff
@@ -38,19 +39,33 @@ public DecodeHandler(ChannelHandler handler) {
...
+            } else {
+                // Retry only once, and only Request will throw an RetryHandleException
+                throw new RemotingException(channel, "Unknown error encountered when retry handle: " + e.getMessage());
+            }
+            handler.received(channel, message);
```
mytang0, line 55: Should 'retry once handler.received' be part of the retry implementation, rather than a separate public part.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeHandler.java
```diff
@@ -82,15 +87,31 @@ void handleRequest(final ExchangeChannel channel, Request req) throws RemotingEx
...
+                // Give ExceptionProcessors a chance to retry request handle or custom exception information.
+                String exPs = System.getProperty(EXCEPTION_PROCESSOR_KEY);
+                if (StringUtils.isNotBlank(exPs)) {
+                    ExtensionLoader<ExceptionProcessor> extensionLoader = ApplicationModel.defaultModel().getDefaultModule().getExtensionLoader(ExceptionProcessor.class);
+                    ExceptionProcessor expProcessor = extensionLoader.getOrDefaultExtension(exPs);
+                    boolean handleError = expProcessor.shouldHandleError(error);
+                    if (handleError) {
+                        // Allow to custom error message, return directly
+                        // Or interrupt, reHandle the process, process req, and continue to return normal information
+                        msg = Optional.ofNullable(expProcessor.wrapAndHandleException(channel, req)).orElse(msg);
```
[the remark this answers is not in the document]
wxbty: RetryHandleException will be thrown up and caught here in DecodeHandler, which is specially designed

```java
public void received(Channel channel, Object message) throws RemotingException {
    doRecveived(message);
    try {
        handler.received(channel, message);
    } catch (RetryHandleException e) {
        if (message instanceof Request) {
            Request request = (Request) message;
            retry(request.getData());
        } else {
            // Retry only once, and only Request will throw an RetryHandleException
            throw new RemotingException(channel, "Unknown error encountered when retry handle: " + e.getMessage());
        }
        handler.received(channel, message);
    }
}
```

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/ExceptionProcessor.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * <p>Mainly works in the decode phase {@link Decodeable#decode()},
+ * and can only handle exception types in this phase.
+ *
+ * <p>Allow three kinds of customized operations, customize abnormal results and return,
+ * customize normal results and return, interrupt and retry the process.
+ *
+ * @since 3.2.0
+ */
+@SPI
```
AlbumenJ, line 37:
```suggestion
@SPI(scope = FRAMEWORK)
```

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeHandler.java
```diff
@@ -82,15 +87,31 @@ void handleRequest(final ExchangeChannel channel, Request req) throws RemotingEx
...
             } else {
                 msg = data.toString();
             }
+
+            if (error != null) {
+                msg = StringUtils.toString(error);
+                // Give ExceptionProcessors a chance to retry request handle or custom exception information.
+                String exPs = System.getProperty(EXCEPTION_PROCESSOR_KEY);
+                if (StringUtils.isNotBlank(exPs)) {
+                    ExtensionLoader<ExceptionProcessor> extensionLoader = ApplicationModel.defaultModel().getDefaultModule().getExtensionLoader(ExceptionProcessor.class);
```
AlbumenJ, line 104:
```suggestion
                    ExtensionLoader<ExceptionProcessor> extensionLoader = channel.getUrl().getOrDefaultFrameworkMode().getExtensionLoader(ExceptionProcessor.class);
```

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/Request.java
```diff
@@ -41,6 +41,9 @@ public class Request {
 
     private Object mData;
 
+    private Throwable mError;
+
+
```
AlbumenJ, line 46: Can we reuse data for error purpose

AlbumenJ (changes requested):

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/Request.java
wxbty: Removed the merror field.  PTAL

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboCodec.java
```diff
@@ -152,12 +154,31 @@ protected Object decodeBody(Channel channel, InputStream is, byte[] header) thro
                 } else {
                     req = new HeartBeatRequest(id);
                     DecodeableRpcInvocation inv;
+                    String exPs = System.getProperty(EXCEPTION_PROCESSOR_KEY);
+                    boolean existExpHandler = false;
+                    ExtensionLoader<ExceptionProcessor> extensionLoader;
+                    ExceptionProcessor expProcessor = null;
+                    if (StringUtils.isNotBlank(exPs)) {
+                        existExpHandler = true;
+                        extensionLoader = channel.getUrl().getApplicationModel().getExtensionLoader(ExceptionProcessor.class);
```
AlbumenJ, line 163:
```suggestion
                        extensionLoader = channel.getUrl().getFrameworkModel().getExtensionLoader(ExceptionProcessor.class);
```

AlbumenJ (changes requested):

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboCodec.java
wxbty: channel.getUrl() does not have a getApplicationModel method, use getOrDefault instead

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/ExceptionProcessor.java
```diff
@@ -0,0 +1,47 @@
...
+ * @since 3.2.0
+ */
+@SPI(scope = FRAMEWORK)
+public interface ExceptionProcessor {
+
+    /**
+     *  Get an enhanced DecodeableRpcInvocation subclass to allow custom decode.
+     *  The parameters are the same as {@link DecodeableRpcInvocation}
+     */
+    DecodeableRpcInvocation getRetryDecodeableRpcInvocation(FrameworkModel frameworkModel, Channel channel, Request req, InputStream is, byte proto);
```
AlbumenJ, line 46:
```suggestion
    DecodeableRpcInvocation getRetryDecodeableRpcInvocation(Channel channel, Request req, InputStream is, byte proto);
```
`frameworkModel` here is useless. ExceptionProcessor itself is a framework model SPI.

AlbumenJ (changes requested):

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/ExceptionProcessor.java
wxbty: indeed

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboCodec.java
```diff
@@ -152,12 +154,31 @@ protected Object decodeBody(Channel channel, InputStream is, byte[] header) thro
...
+                    String exPs = System.getProperty(EXCEPTION_PROCESSOR_KEY);
+                    boolean existExpHandler = false;
+                    ExtensionLoader<ExceptionProcessor> extensionLoader;
+                    ExceptionProcessor expProcessor = null;
+                    if (StringUtils.isNotBlank(exPs)) {
+                        existExpHandler = true;
+                        extensionLoader = channel.getUrl().getOrDefaultFrameworkModel().getExtensionLoader(ExceptionProcessor.class);
+                        expProcessor = extensionLoader.getOrDefaultExtension(exPs);
+                    }
+
```
AlbumenJ, line 166: Cached in constructor to prevent performance hit

AlbumenJ (changes requested):

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboCodec.java
wxbty: done

AlbumenJ (approved):
LGTM

AlbumenJ:
@chickenlj PTAL

chickenlj:
I didn't look into the details of the previous discussion. By just reviewing the code, I have the following concerns:

1. One more scattered SPI definition and its configuration item,  which in this case is `String exPs = System.getProperty(EXCEPTION_PROCESSOR_KEY);`.
2. The only functionality of `ExceptionFilter` is to provide users a way to customize `DecodeableRpcInvocation`, would it be better to include the current initialization of `DecodeableRpcInvocation` as a default implementation of `ExceptionFilter`? And in that way, we might want to rename `ExceptionFilter` as something like `DecodeableInvocationProvider`, so, for users who care about this SPI there will have:
  * A SPI definition `DecodeableInvocationProvider`
  * The default SPI implementation `DefaultDecodeableInvocationProvider`
  * The customized SPI implementation `ExceptionDecodeableInvocationProvider`

Besides the implementation, I care more about the use cases and the necessity of introducing this extension.

chickenlj:
I didn't look into the details of the previous discussion. By just reviewing the code, I have the following concerns:

1. One more scattered SPI definition and its configuration item,  which in this case is `String exPs = System.getProperty(EXCEPTION_PROCESSOR_KEY);`.
2. The only functionality of `ExceptionFilter` is to provide users a way to customize `DecodeableRpcInvocation`, would it be better to include the current initialization of `DecodeableRpcInvocation` as a default implementation of `ExceptionFilter`? And in that way, we might want to rename `ExceptionFilter` as something like `DecodeableInvocationProvider`, so, for users who care about this SPI there will have:
  * A SPI definition `DecodeableInvocationProvider`
  * The default SPI implementation `DefaultDecodeableInvocationProvider`
  * The customized SPI implementation `ExceptionDecodeableInvocationProvider`

Besides the implementation, I care more about the use cases and the necessity of introducing this extension.

wxbty:
> I didn't look into the details of the previous discussion. By just reviewing the code, I have the following concerns:
> 
> 1. One more scattered SPI definition and its configuration item,  which in this case is `String exPs = System.getProperty(EXCEPTION_PROCESSOR_KEY);`.
> 2. The only functionality of `ExceptionFilter` is to provide users a way to customize `DecodeableRpcInvocation`, would it be better to include the current initialization of `DecodeableRpcInvocation` as a default implementation of `ExceptionFilter`? And in that way, we might want to rename `ExceptionFilter` as something like `DecodeableInvocationProvider`, so, for users who care about this SPI there will have:
> 
> * A SPI definition `DecodeableInvocationProvider`
> * The default SPI implementation `DefaultDecodeableInvocationProvider`
> * The customized SPI implementation `ExceptionDecodeableInvocationProvider`
> 
> Besides the implementation, I care more about the use cases and the necessity of introducing this extension.

To put it simply, dubbo has very strong scalability throughout its life cycle, but it will throw a ServiceNotFound exception (404) when the decode period is not included. Compared with springmvc, the @ExceptionHandler annotation combined with the use of NoHandlerFoundException allows custom behavior for 404 scenarios.
If dubbo joins this exception extension, we can do a lot of things. For example, with the ip switching feature, a gateway for dubbo traffic can be built, which is very important to our current cross-network solution.

As for the implementation, I will modify it according to your suggestion. I understand that you want to use the concept of DecodeableInvocationProvider or Factory, and then provide the default DecodeableRpcInvocationProvider?

chickenlj:
> I understand that you want to use the concept of DecodeableInvocationProvider or Factory, and then provide the default DecodeableRpcInvocationProvider?

Yes, I think that would be a better design.

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboCodec.java
```diff
@@ -73,10 +75,13 @@ public class DubboCodec extends ExchangeCodec {
     private static final AtomicBoolean decodeInUserThreadLogged = new AtomicBoolean(false);
     private final CallbackServiceCodec callbackServiceCodec;
     private final FrameworkModel frameworkModel;
+    private ByteAccessor customByteAccessor;
 
     public DubboCodec(FrameworkModel frameworkModel) {
         this.frameworkModel = frameworkModel;
         callbackServiceCodec = new CallbackServiceCodec(frameworkModel);
+        Optional.ofNullable(System.getProperty(BYTE_ACCESSOR_KEY))
+            .ifPresent(spiKey -> customByteAccessor = frameworkModel.getExtensionLoader(ByteAccessor.class).getOrDefaultExtension(spiKey));
```
AlbumenJ, line 84:
```suggestion
        customByteAccessor = Optional.ofNullable(System.getProperty(BYTE_ACCESSOR_KEY))
            .filter(StringUtils::isNotBlank)
            .map(key -> frameworkModel.getExtensionLoader(ByteAccessor.class).getExtension(key))
            .orElse(null);
```

AlbumenJ (changes requested):

AlbumenJ (approved):

Issue #11101, opened 2022-12-09
SpringContainer does not print the exception stack when there is an exception during startup


g0715158:
Environment
Dubbo version: 2.7.6+
Operating System version: linux、windows
Java version: 1.8

Steps to reproduce this issue
Problem description:
org.apache.dubbo.container.spring.SpringContainer does not print the exception stack 
when there is an exception during startup, which is very unfavorable for troubleshooting。

CrazyHZM:
We have removed the container in 3.3.x.

PR #11104, opened 2022-12-09
SpringContainer does not print the exception stack when there is an exception during startup

g0715158:
SpringContainer does not print the exception stack when there is an exception during startup

fix issues #11101


dubbo-container/dubbo-container-spring/src/main/java/org/apache/dubbo/container/spring/SpringContainer.java
```diff
@@ -48,8 +48,16 @@ public void start() {
             configPath = DEFAULT_SPRING_CONFIG;
         }
         context = new ClassPathXmlApplicationContext(configPath.split("[,\\s]+"), false);
-        context.refresh();
-        context.start();
+        try {
+            context.refresh();
+        } catch (Throwable e) {
+            logger.error("refresh spring container failed!", e);
```
AlbumenJ, line 54: Should we throw the exception again?

AlbumenJ: If we swallow the exception, users could not know what happends.

AlbumenJ (changes requested):

PR #11177, opened 2022-12-20
1.test JSONObject to personMap pojo, the personMap extends HashMap

xieshouyu:
confirm that JSONObject can be converted to pojo extends Hashmap

## Brief changelog
1. When genericService is executed, the request object is JSONObject，Confirm that JSONObject can be converted to pojo extends Hashmap
## Verifying this change
run @Test testJSONObjectToPersonMapPojo  method

### DIFF
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/model/person/PersonMap.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/model/person/PersonMap.java
@@ -1,0 +1,29 @@
+package org.apache.dubbo.common.model.person;
+
+import java.util.HashMap;
+
+public class PersonMap extends HashMap<String, String> {
+
+    private static final String ID = "1";
+    private static final String NAME = "hand";
+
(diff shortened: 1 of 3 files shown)


dubbo-common/src/test/java/org/apache/dubbo/common/model/person/PersonMap.java
```diff
@@ -0,0 +1,29 @@
+package org.apache.dubbo.common.model.person;
```
AlbumenJ, line 1: Please add ASF license header

AlbumenJ (changes requested):

dubbo-common/src/test/java/org/apache/dubbo/common/model/person/PersonMap.java
xieshouyu: done

AlbumenJ (approved):
LGTM

PR #11207, opened 2022-12-26
Merge error code 1-36 to 99-0.

win120a:
Merge error code 1-36 to 99-0 since they have the same meaning.

## Brief changelog
Change constants and error codes in the corresponding code.


AlbumenJ (approved):
LGTM

PR #11266, opened 2023-01-10
update IPv6 annotation and filter invalid IPv6

HHoflittlefish777:
need discussion：
Invalid IPv6 address only filter  

- Any Local Address
- Link Local Address
- Loopback Address

but there still some invalid address need filter:

- Site Local Address, which was discarded, but some address may be deployed.
- Unique Local Address, use to replace Site Local Address, which same to private address in IPv4


AlbumenJ (approved):
LGTM

qinliujie:
LGTM

qinliujie (approved):

Issue #11494, opened 2023-02-07
 Modify the weight in the Nacos 2.1.0 console input box, which is not effective in Dubbo 2.7.21


November22:
### Environment

* Dubbo version: 2.7.21
* Operating System version: Widows
* Java version: 1.8
* Nacos：2.1.0

Modifying the weight does not take effect；
![image](https://user-images.githubusercontent.com/26109828/217158002-3f2b7795-e1ae-49d7-b80a-1c77c9bfda28.png)

The weight information delivered by Nacos is in the field weight，Dubbo 2.7.21 is not used when splicing URL information；

![image](https://user-images.githubusercontent.com/26109828/217158095-c26357ab-55dd-470f-8839-8c6c26a2adaf.png)
![image](https://user-images.githubusercontent.com/26109828/217158075-52dd4bac-84be-4651-a31e-370622c9464f.png)

![image](https://user-images.githubusercontent.com/26109828/217160626-e822ed5e-4eaa-493f-83af-3bc25ee42d51.png)

AlbumenJ:
Weight in nacos is not supported now. You can try to submit a PR to support it.

November22:
OK

November22:
I submit a PR [11498](https://github.com/apache/dubbo/pull/11498)

PR #11498, opened 2023-02-07
Enhanced support for Nacos weight configuration

November22:
support for Nacos weight configuration [#11494](https://github.com/apache/dubbo/issues/11494)

## Brief changelog
1. add method: org.apache.dubbo.registry.nacos.NacosRegistry#setWeight
2. upadte method: org.apache.dubbo.registry.nacos.NacosRegistry#buildURL

## Verifying this change
image

## Checklist

### DIFF
--- a/dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
+++ b/dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
@@ -628,6 +628,7 @@ public class NacosRegistry extends FailbackRegistry {
 
     private URL buildURL(URL consumerURL, Instance instance) {
         Map<String, String> metadata = instance.getMetadata();
+        setWeight(metadata, instance.getWeight());
         String protocol = metadata.get(PROTOCOL_KEY);
         String path = metadata.get(PATH_KEY);
         URL url = new ServiceConfigURL(protocol,
@@ -636,6 +637,12 @@ public class NacosRegistry extends FailbackRegistry {
             path,
             instance.getMetadata());
         return new DubboServiceAddressURL(url.getUrlAddress(), url.getUrlParam(), consumerURL, null);
+    }
+
+
+    private void setWeight(Map<String, String> metadata, double nacosWeight) {
+        long weight = Math.round(org.apache.dubbo.rpc.cluster.Constants.DEFAULT_WEIGHT * nacosWeight);
+        metadata.put(org.apache.dubbo.rpc.cluster.Constants.WEIGHT_KEY, String.valueOf(weight));
     }
 
     private Instance createInstance(URL url) {


dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
```diff
@@ -638,6 +639,12 @@ private URL buildURL(URL consumerURL, Instance instance) {
         return new DubboServiceAddressURL(url.getUrlAddress(), url.getUrlParam(), consumerURL, null);
     }
 
+
+    private void setWeight(Map<String, String> metadata, double nacosWeight) {
+        long weight = Math.round(org.apache.dubbo.rpc.cluster.Constants.DEFAULT_WEIGHT * nacosWeight);
```
AlbumenJ, line 644: 1. skip if nacos weight has not changed
2. We should consider if weight has been set in instance

AlbumenJ (changes requested):

dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
November22: 1. There is no flag to judge whether the weight of Nacos is modified，nacos will pass in the default weight of 1.0
image
2. If it is not set in the example, Nacos sends 1.0 by default, and the calculated result according to the code logic is 100, which is consistent with the default weight of dubbo;
image

AlbumenJ: > If it is not set in the instance, Nacos sends 1.0 by default, and the calculated result according to the code logic is 100, which is consistent with the default weight of dubbo;

If dubbo has set `200` in url, then set nacos weight as `3`, the weight should be `600` as expected.

November22: > > If it is not set in the instance, Nacos sends 1.0 by default, and the calculated result according to the code logic is 100, which is consistent with the default weight of dubbo;
> 
> If dubbo has set `200` in url, then set nacos weight as `3`, the weight should be `600` as expected.

Yes, because the weight field type of nacos is double, and the weight field type of dubbo is int, all weights set in nacos are multiplied by the default weight of 100, so that two decimal places are reserved; this logic is consistent in all instances;

AlbumenJ: `Math.round(org.apache.dubbo.rpc.cluster.Constants.DEFAULT_WEIGHT * nacosWeight)` will override the weight from Dubbo provider

AlbumenJ (changes requested):

dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
November22: > `Math.round(org.apache.dubbo.rpc.cluster.Constants.DEFAULT_WEIGHT * nacosWeight)` will override the weight from Dubbo provider


Yes, I may be able to modify it to judge whether Nacos is the default weight 1, if it is, the weight of Dubbo will not be updated, if not, the weight of Dubbo will be updated；
It’s just that the default weight 1 of Nacos is maintained in `com.alibaba.nacos.api.naming.pojo.Instance#weight`， which is not a constant, so a new constant value needs to be added to maintain the default value of Nacos weight;（what do you think of this way）

AlbumenJ: What I mean is that when the weight of nacos is not 1, it may conflict with the weight of dubbo.

November22: > What I mean is that when the weight of nacos is not 1, it may conflict with the weight of dubbo.

Or you can provide a variable to the user to specify the default weight of Nacos, but this feels heavy;
From my understanding, if Nacos is used, then the weight control should be controlled by Nacos;

November22: I feel that it is better to modify Nacos, so Nacos is needed to be compatible with Dubbo's weight configuration

AlbumenJ: It would be better to intoduce a new param like `registry_weight`. Set it in NacosRegistry and NacosServiceDiscoveryRegistry and get it in `org.apache.dubbo.rpc.cluster.loadbalance.AbstractLoadBalance#getWeight`.

November22: Are you referring to org.apache.dubbo.registry.nacos.NacosRegistry#createInstance? I see that when creating an instance here, the weight of the dubbo configuration is not set. Here, the weight of the dubbo configuration can be registered on Nacos, and the weight issued by Nacos shall prevail when obtaining it. In this way, there will be no scenarios covered by Nacos.
image

November22: The changes are as shown in the figure (default rounding for decimals)
image

AlbumenJ: > The changes are as shown in the figure (default rounding for double) image

This is an good idea. But I am afraid that it would not work in ServiceDiscovery model.

November22: > > The changes are as shown in the figure (default rounding for double) image
> 
> This is an good idea. But I am afraid that it would not work in ServiceDiscovery model.


I verified it, it works. When registering, set the weight of dubbo to Nacos; the weight issued by Nacos can be set to Dubbo;


image

AlbumenJ: > > > The changes are as shown in the figure (default rounding for double) image
> > 
> > 
> > This is an good idea. But I am afraid that it would not work in ServiceDiscovery model.
> 
> I verified it, it works. When registering, set the weight of dubbo to Nacos; the weight issued by Nacos can be set to Dubbo;
> 
> image

If different service has different weight, it may conflict.

November22: For example, for interface A, I start three instances and set three different weights respectively. The set weights are stored in the org.apache.dubbo.common.URL parameter. When reporting, the weight is obtained from the URL parameter, and the delivered weight Get it from com.alibaba.nacos.api.naming.pojo.Instance#weight;
————————
比如A接口，我启动三个实例，分别设置三个不同的权重，设置的权重存储在org.apache.dubbo.common.URL 参数中，上报的时候从URL参数中获取到权重，下发的权重的时候从com.alibaba.nacos.api.naming.pojo.Instance#weight 获取；

AlbumenJ: Dubbo 的应用级服务发现是把多个服务（接口）聚合为一个应用级的大服务注册到注册中心的，因此如果每个接口权重不一样的话，大服务注册到注册中心的权重就无法确认了

November22: 明白了，我新加的这个代码应该合并在2.7.x里面

JasonnLiu:
Is the feature still not finished and released？

Issue #11502, opened 2023-02-07
[Opt] Modifier optimization 


mytang0:
The attributes of some classes in the dubbo-rpc module lack the 'final' modifier, so optimize.

PR #11503, opened 2023-02-07
Modifier optimization, add 'final' modifier.

mytang0:
Solution to #11502. 
> Optimization of the problem that the attributes of some classes in the dubbo-rpc module lacks the 'final' modifier.


mytang0:
PTAL @AlbumenJ

AlbumenJ (approved):
LGTM

PR #11516, opened 2023-02-09
fix t-digest miss

wxbty:

When the metrics function is enabled, the t-digest package needs to be referenced separately, otherwise an exception will be thrown

image


songxiaosheng (approved):
LGTM

AlbumenJ (approved):
LGTM

PR #11517, opened 2023-02-09
Implement service-level payload

mytang0:
Implement service-level payload.  Issue #10773  

## Checklist


dubbo-metadata/dubbo-metadata-api/src/test/java/org/apache/dubbo/metadata/MetadataInfoTest.java
```diff
@@ -31,11 +31,7 @@
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 
-import static org.apache.dubbo.common.constants.CommonConstants.APPLICATION_KEY;
-import static org.apache.dubbo.common.constants.CommonConstants.GROUP_KEY;
-import static org.apache.dubbo.common.constants.CommonConstants.INTERFACE_KEY;
-import static org.apache.dubbo.common.constants.CommonConstants.TIMEOUT_KEY;
-import static org.apache.dubbo.common.constants.CommonConstants.VERSION_KEY;
+import static org.apache.dubbo.common.constants.CommonConstants.*;
```
AlbumenJ, line 34: Do not use comma import.
You can import ide code style config in `codestyle` folder.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeChannel.java
```diff
@@ -126,11 +126,16 @@ public CompletableFuture<Object> request(Object request, int timeout, ExecutorSe
...
-        req.setData(request);
+        Request req;
+        if (request instanceof Request) {
+            req = (Request) request;
+        } else {
+            // create request.
+            req = new Request();
+            req.setVersion(Version.getProtocolVersion());
+            req.setTwoWay(true);
+            req.setData(request);
```
AlbumenJ, line 137: What is the purpose of changing this?

AlbumenJ (changes requested):

dubbo-metadata/dubbo-metadata-api/src/test/java/org/apache/dubbo/metadata/MetadataInfoTest.java
mytang0: Ok, this problem has been fixed.

AlbumenJ: dubbo-metadata/dubbo-metadata-api/src/test/java/org/apache/dubbo/metadata/MetadataInfoTest.java

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeChannel.java
mytang0: > What is the purpose of changing this?

In order to get the payload on the provider side, I added an attribute to the Request. The value of payload will be set in DubboInvoker.


The differences are as follows:
image

dubbo-metadata/dubbo-metadata-api/src/test/java/org/apache/dubbo/metadata/MetadataInfoTest.java
mytang0: > MetadataInfoTest

Sorry, I'll fix it right away.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/header/HeaderExchangeChannel.java
mytang0: Done.

dubbo-metadata/dubbo-metadata-api/src/test/java/org/apache/dubbo/metadata/MetadataInfoTest.java
mytang0: Done, submitted.

dubbo-common/src/main/java/org/apache/dubbo/config/annotation/DubboService.java
```diff
@@ -323,4 +323,9 @@
      * @return
      */
     String executor() default "";
+
+    /**
+     * Payload max length.
+     */
+    String payload() default "";
```
wxbty, line 330: Why not use the int type, consistent with the previous one?

mytang0: > Why not use the int type, consistent with the previous one?

Thank you. Yes, I started using int and later made changes to be consistent with dubbo.xsd

wxbty (changes requested):
Adding a payload to the request at the communication layer is unnecessary and risky. The verification of the payload is in the following fragment. The payload itself is obtained through the url. We may modify the corresponding value in the current url through filter or other interception.

```java
protected static int getPayload(Channel channel) {
    if (channel != null && channel.getUrl() != null) {
        return channel.getUrl().getParameter(Constants.PAYLOAD_KEY, Constants.DEFAULT_PAYLOAD);
    }
    return Constants.DEFAULT_PAYLOAD;
}
```

mytang0:
> Adding a payload to the request at the communication layer is unnecessary and risky. The verification of the payload is in the following fragment. The payload itself is obtained through the url. We may modify the corresponding value in the current url through filter or other interception.
> 
> ```java
> protected static int getPayload(Channel channel) {
>     if (channel != null && channel.getUrl() != null) {
>         return channel.getUrl().getParameter(Constants.PAYLOAD_KEY, Constants.DEFAULT_PAYLOAD);
>     }
>     return Constants.DEFAULT_PAYLOAD;
> }
> ```

I am thinking about the following:
1. Channel is shared, so modifying it is complicated and risky
2. Channel's URL payload is a protocol payload
3. The payload property of Request is not passed over the network , so it will not increase network overhead and affect the packaging and unpacking of messages

wxbty:
> 

channel.getUrl() The attributes here come from app-config, service-config and other external configurations, so at least they are shared at the service level.
Since the parameters are obtained from the url, maybe we can achieve the payload configuration at the service level without doing anything. As the following example:

```java
@DubboService(parameters = {"payload", "3000"})
public class GameServiceImpl2 implements GameService {
    public String play(String name) {
        return "hello 2181:" + name;
    }
}
```


image

wxbty:
The configuration of parameters parameters may not be obvious to users, so it should be necessary to add attributes to @DubboService

mytang0:
Thank you very much for your suggestion. It is true that the solution you suggested is the best implementation when the channel is not shared. Consumers share channels through provider address (host + port), each channel holds the url of the service that created it. Therefore, in the sharing mode, it cannot be guaranteed that the payload obtained through the channel is accurate.

wxbty:
```java
/**
 * Get dubbo channel by netty channel through channel cache.
 * Put netty channel into it if dubbo channel don't exist in the cache.
 *
 * @param ch      netty channel
 * @param url
 * @param handler dubbo handler that contain netty's handler
 * @return
 */
static NettyChannel getOrAddChannel(Channel ch, URL url, ChannelHandler handler) {
    if (ch == null) {
        return null;
    }
    NettyChannel ret = CHANNEL_MAP.get(ch);
    if (ret == null) {
        NettyChannel nettyChannel = new NettyChannel(ch, url, handler);
        if (ch.isActive()) {
            nettyChannel.markActive(true);
            ret = CHANNEL_MAP.putIfAbsent(ch, nettyChannel);
        }
        if (ret == null) {
            ret = nettyChannel;
        }
    } else {
        ret.markActive(true);
    }
    return ret;
}
```

@AlbumenJ    NettyChannel seems to have a bug here. The key of CHANNEL_MAP is indeed the original channel of netty, and the url is not included, which means that when calling service2, the channel(created by first service) obtained contains the url of service1. The same goes for actual testing. Maybe we can solve it like this: the key of CHANNEL_MAP uses Channel+url instead of Channel

mytang0:
`private List<ReferenceCountExchangeClient> getSharedClient(URL url, int connectNum) {
        String key = url.getAddress();

        Object clients = referenceClientMap.get(key);
        if (clients instanceof List) {
            List<ReferenceCountExchangeClient> typedClients = (List<ReferenceCountExchangeClient>) clients;
            if (checkClientCanUse(typedClients)) {
                batchClientRefIncr(typedClients);
                return typedClients;
            }
        }
......
}
`

In this place, cause the url may already be inaccurate.  I think Url is used as a property container, and the remoting layer mainly transmits the configuration properties of the remoting layer.  Therefore, when dealing with sharing, Url seems to be less important.

AlbumenJ:
> ```java
> /**
>  * Get dubbo channel by netty channel through channel cache.
>  * Put netty channel into it if dubbo channel don't exist in the cache.
>  *
>  * @param ch      netty channel
>  * @param url
>  * @param handler dubbo handler that contain netty's handler
>  * @return
>  */
> static NettyChannel getOrAddChannel(Channel ch, URL url, ChannelHandler handler) {
>     if (ch == null) {
>         return null;
>     }
>     NettyChannel ret = CHANNEL_MAP.get(ch);
>     if (ret == null) {
>         NettyChannel nettyChannel = new NettyChannel(ch, url, handler);
>         if (ch.isActive()) {
>             nettyChannel.markActive(true);
>             ret = CHANNEL_MAP.putIfAbsent(ch, nettyChannel);
>         }
>         if (ret == null) {
>             ret = nettyChannel;
>         }
>     } else {
>         ret.markActive(true);
>     }
>     return ret;
> }
> ```
> 
> @AlbumenJ NettyChannel seems to have a bug here. The key of CHANNEL_MAP is indeed the original channel of netty, and the url is not included, which means that when calling service2, the channel(created by first service) obtained contains the url of service1. The same goes for actual testing. Maybe we can solve it like this: the key of CHANNEL_MAP uses Channel+url instead of Channel

We should prevent to get parameters from channel url. Channel should be shared among services.

dubbo-demo/dubbo-demo-xml/dubbo-demo-xml-provider/src/main/resources/spring/dubbo-provider.xml
```diff
@@ -44,5 +44,8 @@
                    ref="restDemoService" protocol="rest"/>
     <dubbo:service delay="5000" version="1.0.0" timeout="5000" interface="org.apache.dubbo.demo.TripleService"
                    ref="tripleService" protocol="tri"/>
+    <dubbo:service delay="5000" version="2.0.0" group="greeting" timeout="5000"
+                   interface="org.apache.dubbo.demo.GreetingService"
+                   ref="greetingService" protocol="dubbo" payload="1024"/>
```
CrazyHZM, line 49: the same as above.

CrazyHZM (changes requested):

CrazyHZM:
@mytang0 Regarding the payload configuration of this service level, is it possible to bind different services to different protocol configs to achieve the purpose?

dubbo-demo/dubbo-demo-annotation/dubbo-demo-annotation-provider/src/main/java/org/apache/dubbo/demo/provider/DemoServiceImpl.java
```diff
@@ -25,7 +25,7 @@
 
 import java.util.concurrent.CompletableFuture;
 
-@DubboService
+@DubboService(payload = "1024")
```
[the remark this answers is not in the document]
mytang0: Ok, i fix

mytang0:
> @mytang0 Regarding the payload configuration of this service level, is it possible to bind different services to different protocol configs to achieve the purpose?

Yes, it can be achieved by adding channels, but there is a little violation of channel sharing.

mytang0:
PTAL @chickenlj

chickenlj (approved):
LGTM.

AlbumenJ (approved):

PR #11538, opened 2023-02-12
Fix the bug in LFUCache#put() method

thisiswanghy:
We should not adopt the scheme of "first put new elements, then judge the capacity, and if it is full, trigger the proceedEviction() method".Instead, we should adopt the scheme of "first determine the capacity, then trigger the proceedEviction() method if it is full, and finally put in new elements".

```
@Test
void testPutAndGet() {
    LFUCache<String, Integer> cache = new LFUCache<>(5, 0.2f);
    for (int i = 0; i < 5; i++) {
        cache.put(i + "", i);
        cache.get(i + "");
    }
    cache.put("why", 18);
    Integer why = cache.get("why");
    System.out.println("why = " + why);
}
```

expect：18
but：null

before fix，when we put new element：first put new elements, then judge the capacity, and if it is full, trigger the proceedEviction() method.

so, "why" be eviction after proceedEviction() method.

Instead, we should adopt the scheme when put：first determine the capacity, then trigger the proceedEviction() method if it is full, and finally put in new elements.

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/LFUCache.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/LFUCache.java
@@ -85,12 +85,12 @@ public class LFUCache<K, V> {
                 freqTable[0].addLastNode(node);
                 map.put(key, node);
             } else {
-                node = freqTable[0].addLast(key, value);
-                map.put(key, node);
                 curSize++;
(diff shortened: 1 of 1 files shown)


shy-share (approved):

AlbumenJ (approved):
LGTM

PR #11576, opened 2023-02-15
metric package structure optimization

wxbty:
When burying points in the registration center, you need to use metrics-api, but there is an extension of clusterFilter in metrics-api, which will cause circular dependency problems. Move the metrics-api related filter to the default package to solve.


mytang0 (commented):
LGTM

AlbumenJ (approved):
LGTM

PR #11624, opened 2023-02-21
local mode service map return true

owen200008:
meta type local mode service maping map function need return true


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataServiceNameMapping.java
```diff
@@ -73,6 +76,15 @@ public boolean hasValidMetadataCenter() {
      */
     @Override
     public boolean map(URL url) {
+        Optional<ApplicationConfig> application = applicationModel.getApplicationConfigManager().getApplication();
+        if(application.isPresent()) {
+            String metadataType = application.get().getMetadataType();
+            if (!REMOTE_METADATA_STORAGE_TYPE.equals(metadataType)) {
```
AlbumenJ, line 82: Service mapping is required in local mode

PR #11658, opened 2023-02-25
fix context bug

wxbty:

Issue Number: close https://github.com/apache/dubbo/issues/11657

## Brief changelog

Pass the context returned by the provider, the largest order value makes the first call when returning,
 otherwise the subsequent filter will not get the context

## Verifying this change

The reproduction steps in issus get the expected results

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/support/ConsumerResponseContextFilter.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/support/ConsumerResponseContextFilter.java
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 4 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/support/ConsumerResponseContextFilter.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Pass the context returned by the provider, the largest order value makes the first call when returning,
+ * otherwise the subsequent filter will not get the context
+ *
+ * @see Filter
+ * @see RpcContext
+ */
+@Activate(group = CONSUMER, order = Integer.MAX_VALUE)
+public class ConsumerResponseContextFilter implements Filter, BaseFilter.Listener {
```
AlbumenJ, line 42: RpcContext  is not design for Filter developers.
All context operation in Filter should base on Invocation and Response.

If we support change RpcContext in Filter, achieving synchronization between RpcContext.clientAttachment and Invocation would be complex and inefficient.

The user's RpcContext will be converted into the initial attachment of Invocation at the entrance, and the attachment of Result will be converted into RpcContext at the exit and sent back to the user

AlbumenJ (changes requested):

PR #11665, opened 2023-02-26
improve NacosNamingServiceWrapper performance

sourcelliu:
Because the map is ConcurrentHashMap, so mapLock is not necessary to use.

## Brief changelog
improve NacosNamingServiceWrapper performance.

## Verifying this change
In my location benchmark test, the code like this:
```
    @Test
    void testRegisterInstance() throws NacosException, InterruptedException{
        NamingService namingService = new MockNamingService() {
            @Override
            public void registerInstance(String serviceName, String groupName, Instance instance) throws NacosException {

            }

            @Override
            public List<Instance> getAllInstances(String serviceName, String groupName) throws NacosException {
                return null;
            }
        };

        NacosNamingServiceWrapper nacosNamingServiceWrapper = new NacosNamingServiceWrapper(new NacosConnectionManager(namingService), 0, 0);
        CountDownLatch startLatch = new CountDownLatch(100);
        long startTime = System.currentTimeMillis();
        for(int i=0;i<100;i++){
            new Thread(()->{
                try {
                    for(int j=0;j<100000;j++){
                        Instance instance = new Instance();
                        nacosNamingServiceWrapper.registerInstance("service_name"+j, "test", instance);
                    }
                    startLatch.countDown();
                } catch (Exception e) {
                    throw new RuntimeException(e);
                }
            }).start();
        }
        startLatch.await();
        System.out.printf("the cost time is:%d ms\n",System.currentTimeMillis()-startTime);
    }
```
before change: the cost time is: 14564 ms
after change: the cost time is: 5587 ms
**the execution time reduced by 61.6%**

## Checklist


dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosNamingServiceWrapper.java
```diff
@@ -269,14 +244,9 @@ public void deregisterInstance(String serviceName, String group, Instance instan
...
-                mapLock.lock();
-                if (instancesInfo.getInstances().isEmpty()) {
-                    registerStatus.remove(new InstanceId(nacosServiceName, group));
-                    instancesInfo.setValid(false);
-                }
-            } finally {
-                mapLock.unlock();
+            if (instancesInfo.getInstances().isEmpty()) {
+                registerStatus.remove(new InstanceId(nacosServiceName, group));
+                instancesInfo.setValid(false);
```
AlbumenJ, line 249: Operation here is not atomic

sourcelliu: @AlbumenJ 
I think there is no need for atomicity guarantee here.
- because the code here has already got instancesInfo lock ```instancesInfo.lock()```; 
- and for the modification of registerStatus ```registerStatus.remove(new InstanceId(nacosServiceName, group));```, due to the protection of the lock mechanism inside concurrenthashmap , which has the same effect as the previous mapLock.lock()

AlbumenJ (approved):

PR #11670, opened 2023-02-27
replaced with EN assert

icodening:
replaced with EN assert

## Verifying this change
ut


EarthChen (approved):
LGTM

PR #11677, opened 2023-02-27
feat: generic timeout

aamingaa:
if config timeout method in generic call model,  it will throw exception. This pr is to ignore the verify logic

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
@@ -16,15 +16,6 @@
  */
 package org.apache.dubbo.config;
 
-import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.stream.Collectors;
-
 import org.apache.dubbo.common.URL;
 import org.apache.dubbo.common.Version;
 import org.apache.dubbo.common.config.ConfigurationUtils;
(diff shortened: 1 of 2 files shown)


AlbumenJ:
image

dubbo-common/src/main/java/org/apache/dubbo/config/AbstractInterfaceConfig.java
```diff
@@ -375,6 +375,9 @@ protected void processExtraRefresh(String preferredPrefix, InmemoryConfiguration
     }
 
     private boolean verifyMethodConfig(MethodConfig methodConfig, Class<?> interfaceClass, boolean ignoreInvalidMethodConfig) {
+        if (!isNeedCheckMethod()) {
+            return true;
+        }
```
AlbumenJ, line 380: It would be better to check anyway, but warn out only instead of throw a exception.

AlbumenJ (changes requested):

AlbumenJ:
Please add a ut to verify

dubbo-common/src/test/java/org/apache/dubbo/config/AbstractInterfaceConfigTest.java
```diff
@@ -236,7 +236,20 @@ void testScope() {
...
 
+    @Test
+    void testVerifyMethod() {
+        InterfaceConfig interfaceConfig = new InterfaceConfig();
+        MethodConfig methodConfig = new MethodConfig();
+        methodConfig.setTimeout(5000);
+        methodConfig.setName("sayHello");
+        Class<?> clazz = Greeting.class;
+        boolean verifyResult = interfaceConfig.verifyMethodConfig(methodConfig, clazz, false);
+        Assertions.assertEquals(true, verifyResult);
```
AlbumenJ, line 247:
```suggestion
        Assertions.assertTrue(verifyResult);
```

dubbo-common/src/test/java/org/apache/dubbo/config/AbstractInterfaceConfigTest.java
```diff
@@ -236,7 +236,20 @@ void testScope() {
...
+        boolean verifyResult2 = interfaceConfig.verifyMethodConfig(methodConfig, clazz, true);
+        Assertions.assertEquals(false, verifyResult2);
```
AlbumenJ, line 249:
```suggestion
        Assertions.assertFalse(verifyResult2);
```

dubbo-config/dubbo-config-api/src/test/java/org/apache/dubbo/config/MethodConfigTest.java
```diff
@@ -431,7 +431,6 @@ void testVerifyMethodConfigOfService() {
                 .application("demo-app")
                 .service(serviceConfig)
                 .initialize();
-            Assertions.fail("Method config verification should failed");
```
AlbumenJ, line 434: Revert this. Safe barrier

AlbumenJ (changes requested):

dubbo-config/dubbo-config-api/src/test/java/org/apache/dubbo/config/MethodConfigTest.java
aamingaa: has change to warn

PR #11698, opened 2023-03-01
tri create Http2StreamChannel async

icodening:
optimize tri protocol performance

## Brief changelog
create stream channel async, reduce the number of biz threads wait

## Verifying this change
ut


AlbumenJ (approved):

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/command/QueuedCommand.java
```diff
@@ -41,7 +41,13 @@ public void cancel() {
 
     public void run(Channel channel) {
         if (channel.isActive()) {
-            channel.write(this, promise);
+            channel.write(this).addListener(future -> {
```
guohao, line 44: The reason of changing is?

icodening: The stream channel has not been constructed when the header/data is sent, the `ChannelPromise` of header/data comes from the `parent channel`

guohao (approved):
LGTM

EarthChen (approved):
LGTM

PR #11723, opened 2023-03-05
triple server batch flush

icodening:
triple server batch flush

## Brief changelog
add FlushConsolidationHandler

#10558


icodening:
Flush will be called when the stream is closed, add `FlushConsolidationHandler` to reduce `flush`

EarthChen (approved):
LGTM.
But it should be noted that this is a temporary solution, and we need to wait for netty's PR 
 https://github.com/netty/netty/pull/13244 to consider the final solution

guohao (approved):
LGTM

PR #11842, opened 2023-03-15
fix gc bug when use micrometer

wxbty:

bugfix: In the case of metrics, the occasional index key is NaN.
reason: The Gauge of the micrometer use the application obj as a weak reference, and the EntryKey traversed by ConcurrentHashMap is not the same instance every time.

## Brief changelog

use an instance that will not be gc as the application obj

## Verifying this change

The consumer calls the provider interface 100 times in a loop, and the problem must occur. After modification, NaN no longer appears


songxiaosheng (approved):
LGTM

songxiaosheng:
@AlbumenJ PTAL

AlbumenJ (approved):

PR #11853, opened 2023-03-16
splits qos-api pkg

wxbty:
qos splits out qos-api to avoid extended circular dependency on qos


AlbumenJ (approved):
LGTM

PR #11855, opened 2023-03-17
weak gc bug fix

wxbty:

Similar to  #11842 , the cpu indicator also has a NaN bug
image

## Brief changelog

```
    static <T> Builder<T> builder(String name, @Nullable T obj, ToDoubleFunction<T> f) {
        return new Builder<>(name, obj, f);
    }
```
When using the above api, if T(obj) has no other strong references, gc will cause it to be recycled and print NaN.
We need to save a strong reference to the obj instance.


songxiaosheng (approved):
LGTM

AlbumenJ (approved):

PR #11857, opened 2023-03-17
Code opt

wxbty:
Code optimization
1. ServiceKeyMetric extracts the api package, and the metadata metrics can be reused.
2. skStats was originally used in the setKey scene, but now the num+1 scene is added, and it is changed to serviceNumStats and applicationNumStats to be more symmetrical
3. Delete redundant maps
4. add testcase and bugfix


AlbumenJ (approved):

dubbo-metrics/dubbo-metrics-api/src/main/java/org/apache/dubbo/metrics/model/MetricsKeyWrapper.java
```diff
@@ -17,17 +17,28 @@
...
 package org.apache.dubbo.metrics.model;
 
+import java.util.Map;
+
 /**
  * Let {@link MetricsKey MetricsKey}  output dynamic, custom string content
  */
 public class MetricsKeyWrapper {
 
+    // Register、subscribe、notify etc
```
songxiaosheng, line 27: can use doc comment such as /* */

songxiaosheng (changes requested):

dubbo-metrics/dubbo-metrics-api/src/main/java/org/apache/dubbo/metrics/model/MetricsKeyWrapper.java
wxbty: ok

songxiaosheng (approved):
LGTM  @AlbumenJ PATL

PR #11863, opened 2023-03-19
Simplify event publishing

wxbty:
Event publishing code optimization

## Brief changelog
1、Use EventBus to simplify the form of event publishing
2、rename some classes
3、Cohesive Event Time API usage, no longer exposed to biz caller
4、Delete unused ApplicationEvent and RequestEvent

## Verifying this change
can pass updated testcase


dubbo-metrics/dubbo-metrics-api/src/main/java/org/apache/dubbo/metrics/event/EventBus.java
```diff
@@ -0,0 +1,115 @@
...
+ */
+public class EventBus {
+
+    /**
+     * Posts an event to all registered subscribers and only once.
+     *
+     * @param event event to post.
+     */
+    public static void solo(MetricsEvent event) {
+        if (event.getSource() instanceof ApplicationModel) {
```
AlbumenJ, line 38: It is possible that source here is not a ApplicationModel? If so, we should change these usage. Otherwise, change the type of source

AlbumenJ (changes requested):

dubbo-metrics/dubbo-metrics-api/src/main/java/org/apache/dubbo/metrics/event/EventBus.java
wxbty: Modify it to only be used as ApplicationModel

AlbumenJ (approved):
LGTM

PR #11879, opened 2023-03-21
optimize performance. decode in user thread

icodening:
optimize performance. decode in user thread

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/call/TripleMessageProducer.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/call/TripleMessageProducer.java
@@ -26,35 +26,13 @@ class TripleMessageProducer implements ClientCall.MessageProducer {
 
     private final ThrowableSupplier<Object> throwableSupplier;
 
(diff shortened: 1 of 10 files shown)


dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/call/TripleMessageProducer.java
```diff
@@ -0,0 +1,63 @@
...
+        } catch (Throwable t) {
+            if (throwableConsumer != null) {
+                throwableConsumer.accept(t);
+            }
+            if (t instanceof RuntimeException) {
+                throw (RuntimeException) t;
+            }
+            if (t instanceof InvocationTargetException) {
+                Throwable targetException = ((InvocationTargetException) t).getTargetException();
+                throw new RuntimeException(targetException);
```
EarthChen, line 54: Why do you need to throw at the same time as consumer?

EarthChen: Why use runtime to wrap exceptions?

EarthChen (changes requested):

EarthChen (approved):
LGTM

guohao (approved):
LGTM

PR #11884, opened 2023-03-22
use invocation return support

owen200008:
add a switch，use invocation return type to deserialization（compatible with  rsp is generic class）


dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DecodeableRpcResult.java
```diff
@@ -150,7 +164,15 @@ public void decode() throws Exception {
 
     private void handleValue(ObjectInput in) throws IOException {
         try {
-            Type[] returnTypes = RpcUtils.getReturnTypes(invocation);
+            Type[] returnTypes = null;
+            if(useInvocationReturn) {
+                if(invocation instanceof RpcInvocation) {
+                    returnTypes = ((RpcInvocation) invocation).getReturnTypes();
+                }
+            }
```
AlbumenJ, line 172: When calling generically, the deserialization here should be implemented by GenericFilter. If there is a generic problem, PojoUtils should be modified to solve it.

AlbumenJ (changes requested):

owen200008:
https://github.com/apache/dubbo/pull/11914

PR #11914, opened 2023-03-24
pojoutil support generic

owen200008:
PojoUtils support generic class with return data no classname


AlbumenJ (changes requested):
Please check if 
```
public class A<T> {
    private T data;

    public T getData() {
        return data;
    }

    public void setData(T data) {
        this.data = data;
    }
}

public class B<T> {
    private T data;

    public T getData() {
        return data;
    }

    public void setData(T data) {
        this.data = data;
    }
}

public class C<T> {
    private T data1;
    private A<String> data2;

    public T getData() {
        return data;
    }

    public void setData(T data) {
        this.data = data;
    }
}
```

`new A<B<String>>()` and `new C<B<Number>>()` work proper

AlbumenJ (approved):

PR #11921, opened 2023-03-25
dubbo-starter  import metrics dep

wxbty:
The introduction of starter can use the metrics capability。
When you don't need observability, import the dubbp-starter package. If necessary, import the dubbo-spring-boot-observability-starter package


AlbumenJ (approved):
LGTM

PR #11944, opened 2023-03-28
feat: lowwer version

aamingaa:
lowwer grpc version


EarthChen (commented):
why downgrade grpc version?

aamingaa:
> why downgrade grpc version?

if use grpc version >= 1.41 in xds model, it will be fail to generate cert

AlbumenJ (approved):

PR #11973, opened 2023-03-30
Revert "feat: change default mode"

EarthChen:
Reverts apache/dubbo#11653

Serious performance degradation


EarthChen:
image


image

PR #11982, opened 2023-04-01
Metrics Event/Listener refactor

wxbty:
 The metric module needs  optimized refactoring to improve code reusability while looking more streamlined.
There are many redundant codes in the event mechanism, and the current PR optimizes the event code。           

## Brief changelog
The event object is created in a static way, so that the API is more streamlined.
Use the form of anonymous class to simplify the listener of metrics.
There are many modifications, try to optimize the registry first, and then optimize others later.


AlbumenJ (approved):
LGTM
@songxiaosheng PTAL

songxiaosheng (approved):
LGTM

PR #12039, opened 2023-04-09
feat: triple exception self

aamingaa:


AlbumenJ:
@aamingaa 
image

EarthChen (changes requested):

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/PbUnpack.java
```diff
@@ -35,4 +35,10 @@ public Object unpack(byte[] data) throws IOException {
         final ByteArrayInputStream bais = new ByteArrayInputStream(data);
         return SingleProtobufUtils.deserialize(bais, clz);
     }
+
+
+    public Object unpack(byte[] data, boolean isReturnTriException) throws IOException {
```
EarthChen, line 40: unused ?

EarthChen (changes requested):

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/PbUnpack.java
aamingaa: if remove it, the compiole with fail

Issue #11941, opened 2023-03-27
[Observability Task] Architectural Design Concept of Abstract Indicator System


songxiaosheng:
[Observability Task] Architectural Design Concept of Abstract Indicator System
In order for the indicator system to better serve other modules, we need to abstract and reconstruct the indicator system，Here are some current design ideas
Collectors - Event ->Data Aggregator (Optional) ->Indicator Samples ->Indicator Exporter
                              --------------------------------------->

[可观测性任务] 抽象指标体系的架构设计概念，
为了指标系统能更好的为其他模块提供服务，我们需要再对指标体系进行抽象重构，下面是目前的一些设计思想
采集器  ----事件 ---> 数据聚合器(可选）-->指标样本--->指标导出器
                       --------------------------------------->

songxiaosheng:
Refactoring code to support business metrics collection
Currently, the indicator collection function of Dubbo is mainly used inside Dubbo. In fact, every application has a requirement for indicator collection, and it is necessary to try to extend the concept of collector samples. At the same time, it is necessary to provide a certain API for the indicator registry that exports indicators to allow users to output their business indicators to the indicator registry

重构代码以支持使用Dubbo的应用程序的外部业务指标采集

目前Dubbo的指标采集功能主要用于Dubbo内部，其实每个应用程序都会有指标采集的需求，需要尝试将采集器样本的概念扩展出去，同时需要将指标导出的指标注册表也提供一定的API可以让用户的业务指标输出到指标注册表中

wxbty:
> Refactoring code to support business metrics collection Currently, the indicator collection function of Dubbo is mainly used inside Dubbo. In fact, every application has a requirement for indicator collection, and it is necessary to try to extend the concept of collector samples. At the same time, it is necessary to provide a certain API for the indicator registry that exports indicators to allow users to output their business indicators to the indicator registry
> 
> 重构代码以支持使用Dubbo的应用程序的外部业务指标采集
> 
> 目前Dubbo的指标采集功能主要用于Dubbo内部，其实每个应用程序都会有指标采集的需求，需要尝试将采集器样本的概念扩展出去，同时需要将指标导出的指标注册表也提供一定的API可以让用户的业务指标输出到指标注册表中

[...]

PR #12083, opened 2023-04-12
Refactor Metrics Key&Collector

wxbty:
1. Metricskey optimization, directly dependent on Metricskey enum, remove ApplicationType/ServiceType type
2. Listener/Collector abstracts the corresponding App/Service type interface and abstract class, reduces repeated code as much as possible, and has a clearer object model
link #11941


AlbumenJ (approved):
LGTM
@songxiaosheng PTAL

dubbo-metrics/dubbo-metrics-api/src/main/java/org/apache/dubbo/metrics/listener/MetricsServiceListener.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public static AbstractMetricsListener onErrorEventBuild(MetricsKey metricsKey, MetricsPlaceType placeType, ServiceMetricsCollector<TimeCounterEvent> collector) {
+        return AbstractMetricsListener.onError(metricsKey,
+            event -> incrAndAddRt(metricsKey, placeType, collector, event)
+        );
+    }
+
+    private static void incrAndAddRt(MetricsKey metricsKey, MetricsPlaceType placeType, ServiceMetricsCollector<TimeCounterEvent> collector, TimeCounterEvent event) {
+        collector.increment(event.appName(), event.getAttachmentValue(ATTACHMENT_KEY_SERVICE), metricsKey,1);
```
songxiaosheng, line 52: Can magic number 1 be modified to a constant

songxiaosheng (changes requested):
There are only two small places

songxiaosheng (changes requested):
There are only two small places

dubbo-metrics/dubbo-metrics-api/src/main/java/org/apache/dubbo/metrics/listener/MetricsServiceListener.java
wxbty: done

songxiaosheng (approved):
LGTM

songxiaosheng (approved):
LGTM

PR #12206, opened 2023-04-27
refactor metrics defaultCollector&&agg

wxbty:
Use the same API as the three centers
link https://github.com/apache/dubbo/issues/11941


wxbty:
@AlbumenJ @songxiaosheng  PTAL

dubbo-metrics/dubbo-metrics-api/src/main/java/org/apache/dubbo/metrics/model/MetricsSupport.java
```diff
@@ -56,4 +73,120 @@ public static Map<String, String> serviceTags(String appAndServiceName) {
...
+        if (e.isBiz()) {
+            targetKey = MetricsKey.METRIC_REQUEST_BUSINESS_FAILED;
+        }
+        if (e.isSerialization()) {
+            targetKey = MetricsKey.METRIC_REQUESTS_CODEC_FAILED;
+        }
+        if (e.isNetwork()) {
+            targetKey = MetricsKey.METRIC_REQUESTS_NETWORK_FAILED;
+        }
+        return targetKey;
```
songxiaosheng, line 106: in the future we also cloud use  map to  store  the  relation exception to metric key

songxiaosheng (approved):
LGTM

dubbo-metrics/dubbo-metrics-api/src/main/java/org/apache/dubbo/metrics/model/MetricsSupport.java
wxbty: 0

PR #12258, opened 2023-05-08
fix generic type deserialize bug ( #12257 )

CodePlayer:
fix fastjson2 generic type deserialize failed ( issue #12257 )

### DIFF
--- a/dubbo-serialization/dubbo-serialization-fastjson2/src/main/java/org/apache/dubbo/common/serialize/fastjson2/FastJson2ObjectInput.java
+++ b/dubbo-serialization/dubbo-serialization-fastjson2/src/main/java/org/apache/dubbo/common/serialize/fastjson2/FastJson2ObjectInput.java
@@ -118,6 +118,7 @@ public class FastJson2ObjectInput implements ObjectInput {
             result = JSONB.parseObject(bytes, cls, securityFilter,
                 JSONReader.Feature.UseDefaultConstructorAsPossible,
                 JSONReader.Feature.ErrorOnNoneSerializable,
+                JSONReader.Feature.SupportAutoType,
                 JSONReader.Feature.IgnoreAutoTypeNotMatch,
                 JSONReader.Feature.UseNativeObject,
                 JSONReader.Feature.FieldBased);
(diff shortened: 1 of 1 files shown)


AlbumenJ (changes requested):
Dubbo already use `ContextAutoTypeBeforeHandler` to indicate which classes can be deserialized. SupportAutoType should not be enabled.

CodePlayer:
> Dubbo already use `ContextAutoTypeBeforeHandler` to indicate which classes can be deserialized. SupportAutoType should not be enabled.


@AlbumenJ 
But, it **still has bugs**. Same code, using Hessian2 it works, switch to fastjson2 it goes wrong。

I think, Dubbo should remove `JSONReader.Feature.IgnoreAutoTypeNotMatch`, instead of `JSONReader.Feature.SupportAutoType`.

Because Dubbo is based on **interfaces** and is **strongly typed**, fastjson2 will judge whether the types match, and throw an exception if they do not match.

AlbumenJ:
> > Dubbo already use `ContextAutoTypeBeforeHandler` to indicate which classes can be deserialized. SupportAutoType should not be enabled.
> 
> @AlbumenJ But, it **still has bugs**. Same code, using Hessian2 it works, switch to fastjson2 it goes wrong。
> 
> I think, Dubbo should remove `JSONReader.Feature.IgnoreAutoTypeNotMatch`, instead of `JSONReader.Feature.SupportAutoType`.
> 
> Because Dubbo is based on **interfaces** and is **strongly typed**, fastjson2 will judge whether the types match, and throw an exception if they do not match.

For security purpose, we should check type if match. Otherwise, there may some arbitrary serialization issue.

CodePlayer:
I just tested it.
After I upgraded to fastjson **2.0.32-SNAPSHOT**,  it works fine.

The key reason is that fastjson has a bug.  After my feedback, it has been fixed in the new version.
But, the new version has not yet been officially released.

CodePlayer:
> 

Please note that, `JSONReader.Feature.IgnoreAutoTypeNotMatch` means that the types **can be** mismatched.

**It existed before**.

Issue #12279, opened 2023-05-10
When serialization fails, the code value in org.apache.dubbo.rpc.RpcException is set incorrectly.


xuziyang:
![image](https://github.com/apache/dubbo/assets/8465969/257189e3-4f71-4c9b-b34c-0f0fce780bac)

When serialization fails, the code value in org.apache.dubbo.rpc.RpcException should be SERIALIZATION_EXCEPTION.SERIALIZATION_EXCEPTION, but currently it is SERIALIZATION_EXCEPTION.NETWORK_EXCEPTION

PR #12280, opened 2023-05-10
Fix When serialization fails, the code value in org.apache.dubbo.rpc.RpcException is set incorrectly. (#12279)

xuziyang:
fix https://github.com/apache/dubbo/issues/12279

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
@@ -26,6 +26,7 @@ import org.apache.dubbo.common.timer.TimerTask;
 import org.apache.dubbo.common.utils.NamedThreadFactory;
 import org.apache.dubbo.remoting.Channel;
 import org.apache.dubbo.remoting.RemotingException;
+import org.apache.dubbo.remoting.SerializationException;
 import org.apache.dubbo.remoting.TimeoutException;
 import org.apache.dubbo.remoting.exchange.Request;
 import org.apache.dubbo.remoting.exchange.Response;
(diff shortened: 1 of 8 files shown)


AlbumenJ (commented):
Can SerializationException just a type of RpcException

xuziyang:
> RpcException

RpcException uses the internal code variable to indicate which type of exception has occurred. When a serialization exception occurs, the code value is set to NETWORK_EXCEPTION, which is incorrect. The correct value for code should be SERIALIZATION_EXCEPTION.

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
```diff
@@ -203,7 +204,9 @@ private void doReceived(Response res) {
             this.complete(res.getResult());
         } else if (res.getStatus() == Response.CLIENT_TIMEOUT || res.getStatus() == Response.SERVER_TIMEOUT) {
             this.completeExceptionally(new TimeoutException(res.getStatus() == Response.SERVER_TIMEOUT, channel, res.getErrorMessage()));
-        } else {
+        } else if(res.getStatus() == Response.SERIALIZATION_ERROR){
+            this.completeExceptionally(new SerializationException(channel, res.getErrorMessage()));
```
AlbumenJ, line 208: `new RpcException(SERIALIZATION_EXCEPTION, res.getErrorMessaeg())` would be enough?

AlbumenJ (changes requested):

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
xuziyang: As shown in the figure below, in the invoke method of AsyncToSyncInvoker, other exceptions are uniformly converted to RpcException. If you directly create a new RpcException(SERIALIZATION_EXCEPTION, res.getErrorMessage()) and throw it at the place you mentioned, would it conflict with the existing code logic?

Also, why did building and testing fail without any error messages?

image

AlbumenJ (approved):
LGTL

AlbumenJ:
@CrazyHZM @chickenlj PTAL

chickenlj (approved):
LGTM.

PR #12286, opened 2023-05-10
Fix ClassCastException caused by 2.x's filter in some condition

wegod:

When upgrading from dubbo 2.x to 3.x, we have many projects and need to call provider service using 2.x filter and return RpcResult in some case. 
So it's better to compatible of RpcResult when 3.x consumer calling 2.x provider

## Brief changelog
Fix ClassCastException caused by 2.x's filter in some condition and compatible of RpcResult in dubbo version 2.x

## Verifying this change

I test on my local env

## Checklist


dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/RpcResult.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public void setObjectAttachment(String key, Object value) {
+
+    }
+
+    public String toString() {
+        return "RpcResult [result=" + this.result + ", exception=" + this.exception + "]";
+    }
+}
+
```
AlbumenJ, line 115:
```suggestion
public class RpcResult extends AppResponse implements com.alibaba.dubbo.rpc.Result {
    public RpcResult() {
    }

    public RpcResult(Object result) {
        super(result);
    }

    public RpcResult(Throwable exception) {
        super(exception);
    }
}

```

AlbumenJ (changes requested):

dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/RpcResult.java
wegod: Thanks for your code review.
I tested on my local env, and add whenCompleteWithContext's empty implement which is called in FilterChainBuilder.

dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/RpcResult.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public RpcResult(Throwable exception) {
+        super(exception);
+    }
+    
+    @Override
+    public org.apache.dubbo.rpc.Result whenCompleteWithContext(BiConsumer<org.apache.dubbo.rpc.Result, Throwable> fn) {
+        return null;
+    }
```
AlbumenJ, line 38: What is the purpose of adding this?

AlbumenJ:
@wegod I have try to help you fix the test cases :)

dubbo-compatible/src/main/java/com/alibaba/dubbo/rpc/RpcResult.java
wegod: Wait a moment, I found when dubbo3 using dubbo2's Filter, RpcResult will be handle as AsyncRpcResult. 
This function whenCompleteWithContext will be changed later, more code need to fix.

wegod:
> @wegod I have try to help you fix the test cases :)

Thanks very much. 
Right now I'm using 3.1.10.  Will this branch merge into 3.1.11 or 12?

wegod:
@AlbumenJ My test is done, I found new change in Filter.java and CompatibleResult.java that's better than my first code change. 
All case works in both provider side and consumer side, no matter dubbo2.x or 3.x is provider or consumer.

AlbumenJ:
> > @wegod I have try to help you fix the test cases :)
> 
> Thanks very much. Right now I'm using 3.1.10. Will this branch merge into 3.1.11 or 12?

According to the roadmap, 3.1.x is only for security fix now. We can only maintian one branch due to limited working time (:

chickenlj (approved):
LGTM.

PR #12294, opened 2023-05-11
Add log for remote configuration.

chickenlj:
Print the remote configuration in the log file for possible troubleshooting

```java
 logger.info(String.format("Got application specific remote configuration from config center with key %s and group %s: \n %s", appConfigFile, appGroup, appConfigContent));

```

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/config/ConfigurationUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/config/ConfigurationUtils.java
@@ -173,25 +173,23 @@ public final class ConfigurationUtils {
 
     public static Map<String, String> parseProperties(String content) throws IOException {
         Map<String, String> map = new HashMap<>();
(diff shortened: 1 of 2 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/config/ConfigurationUtils.java
```diff
@@ -173,25 +172,23 @@ public static int get(ScopeModel scopeModel, String property, int defaultValue)
 
     public static Map<String, String> parseProperties(String content) throws IOException {
         Map<String, String> map = new HashMap<>();
-        if (StringUtils.isEmpty(content)) {
-            logger.warn(COMMON_UNEXPECTED_EXCEPTION, "", "", "Config center was specified, but no config item found.");
```
AlbumenJ, line 177: Why change this?

AlbumenJ (changes requested):

AlbumenJ (approved):

PR #12329, opened 2023-05-16
optimize performance for MetricsFilter

finefuture:
ReflectionUtils add event type cache, reduce the time consumption of MetricsFilter

## Checklist


songxiaosheng (approved):
LGTM @wxbty  PATL

wxbty (commented):
Good job,  write a unit test by the way？

finefuture:
Why did it affect the dubbo-samples-triple-reactor module？
image

wxbty (approved):
LGTM

AlbumenJ (changes requested):
image

How about cache in AbstractMetricsListener.

Allocate a map to cache class is not a good idea, and this may cause classloader unable to being gc.

finefuture:
> image
> 
> How about cache in AbstractMetricsListener.
> 
> Allocate a map to cache class is not a good idea, and this may cause classloader unable to being gc.

@AlbumenJ Good idea, Is this ok?
image

wxbty:
> > image
> > How about cache in AbstractMetricsListener.
> > Allocate a map to cache class is not a good idea, and this may cause classloader unable to being gc.
> 
> @AlbumenJ Good idea, Is this ok? image
> 
> Tested to improve performance by 2% than cache in ReflectionUtils

Would it be better to put the match method in util, and use Object for the event type? It is not the responsibility of the current class

finefuture:
> > > image
> > > How about cache in AbstractMetricsListener.
> > > Allocate a map to cache class is not a good idea, and this may cause classloader unable to being gc.
> > 
> > 
> > @AlbumenJ Good idea, Is this ok? image
> > Tested to improve performance by 2% than cache in ReflectionUtils
> 
> Would it be better to put the match method in util, and use Object for the event type? It is not the responsibility of the current class

Okay, Moved

wxbty (approved):

dubbo-metrics/dubbo-metrics-api/src/main/java/org/apache/dubbo/metrics/listener/AbstractMetricsListener.java
```diff
@@ -20,16 +20,19 @@
 import org.apache.dubbo.common.utils.ReflectionUtils;
 import org.apache.dubbo.metrics.event.MetricsEvent;
 
-import java.util.List;
+import java.util.HashMap;
+import java.util.Map;
 
 public abstract class AbstractMetricsListener<E extends MetricsEvent> implements MetricsListener<E> {
 
+    private final Map<Class<?>, Boolean> eventMatchCache = new HashMap<>();
```
AlbumenJ, line 28: ConcurrentHashMap here

AlbumenJ (changes requested):

dubbo-metrics/dubbo-metrics-api/src/main/java/org/apache/dubbo/metrics/listener/AbstractMetricsListener.java
finefuture: Fixed

AlbumenJ (approved):

PR #12354, opened 2023-05-19
When PropertyValue "parameters" is [null] or [EMPTY], return to HashMap that can be modified

YAGAMIL:
1. `org.apache.dubbo.config.spring.util.DubboAnnotationUtils.convertParameters`
When the `put` target is an empty String array, the empty `HashMap` instead of `java.util.Collections.EmptyMap` supports subsequent scalable `put` operations  #12350 

So that `@DubbboService#parameter`, even if empty, does not affect the extra put operation causing the program to throw an exception

```java
  if (ArrayUtils.isEmpty(parameters)) {
       return Collections.emptyMap();
   }
```

---->

```java
  if (ArrayUtils.isEmpty(parameters)) {
       return new HashMap<>();
  }
```

`Collections.emptyMap()`is a special implementation of EmptyMap inside Collections that doesn't support change, so I don't think it makes any sense to set a Map of that type here

```java
    public V put(K key, V value) {
        throw new UnsupportedOperationException();
    }

```

2. `org.apache.dubbo.config.AbstractMethodConfig#getParameters`

Prevents this function from returning null
```java
    public Map<String, String> getParameters() {
        return parameters;
    }

```
---->
```java
    public Map<String, String> getParameters() {
        this.parameters = Optional.ofNullable(this.parameters).orElseGet(HashMap::new);
        return this.parameters;
    }
```

## Verifying this change

......

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractMethodConfig.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractMethodConfig.java
@@ -23,7 +23,9 @@ import org.apache.dubbo.rpc.model.ModuleModel;
 import org.apache.dubbo.rpc.model.ScopeModel;
 
 import java.beans.Transient;
+import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * AbstractMethodConfig
@@ -240,7 +242,8 @@ public abstract class AbstractMethodConfig extends AbstractConfig {
     }
 
     public Map<String, String> getParameters() {
-        return parameters;
+        this.parameters = Optional.ofNullable(parameters).orElseGet(HashMap::new);
+        return this.parameters;
     }
 
     public void setParameters(Map<String, String> parameters) {
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractMethodConfig.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractMethodConfig.java
@@ -242,7 +242,7 @@ public abstract class AbstractMethodConfig extends AbstractConfig {
     }
 
     public Map<String, String> getParameters() {
-        this.parameters = Optional.ofNullable(parameters).orElseGet(HashMap::new);
+        this.parameters = Optional.ofNullable(this.parameters).orElseGet(HashMap::new);
         return this.parameters;
     }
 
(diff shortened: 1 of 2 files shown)


dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/util/DubboAnnotationUtils.java
```diff
@@ -25,11 +25,7 @@
 
 import org.springframework.util.Assert;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
AlbumenJ, line 28: Please do not use comma import

dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/util/DubboAnnotationUtils.java
```diff
@@ -154,7 +150,7 @@ public static String resolveInterfaceName(Reference reference, Class<?> defaultI
      */
     public static Map<String, String> convertParameters(String[] parameters) {
         if (ArrayUtils.isEmpty(parameters)) {
-            return Collections.emptyMap();
+            return new HashMap<>();
```
AlbumenJ, line 153: Null return in `org.apache.dubbo.config.AbstractMethodConfig#getParameters` is also need to be considered

AlbumenJ (changes requested):
BTW, pls replace the PR subject as a considerable one and comment in English.

dubbo-common/src/main/java/org/apache/dubbo/config/AbstractMethodConfig.java
```diff
@@ -240,7 +242,8 @@ public void setValidation(String validation) {
     }
 
     public Map<String, String> getParameters() {
-        return parameters;
+        this.parameters = Optional.ofNullable(parameters).orElseGet(HashMap::new);
```
AlbumenJ, line 245: Set the newly created map to `this`.

AlbumenJ (changes requested):

AlbumenJ (approved):

PR #12403, opened 2023-05-25
Metrics Code clean

wxbty:
After several refactorings, some old code is no longer needed, so clean it up


songxiaosheng (approved):
LGYM

PR #12446, opened 2023-06-01
support catch actual exception

icodening:
support catch actual exception

# case
When the provider is offline, the consumer continues to send requests, and it is expected that the `ClosedChannelException` will be caught

# after
image

# before
image


AlbumenJ (approved):

AlbumenJ:
@guohao @EarthChen PTAL

EarthChen (approved):
LGTM

PR #12528, opened 2023-06-14
Feature improvement, only register one url for port unification.

chickenlj:
Currently, [multiple protocol port-unification](https://cn.dubbo.apache.org/zh-cn/overview/tasks/protocols/multi-protocols/) feature works the following way:

With the following configuration:

```yaml
dubbo
  protocol
     name: tri
     port: -1
     ext-protocol: rest
```

It will create two provider urls in the registry:

* rest://demo-service
* tri://demo-service

But with this pull request, there will only have one url in the registry

`tri://demo-service?ext.protocol=rest`

Consumers can automatically tell the main protocol `tri` from the `ext` protocol,  those consumers without protocol set consume `tri` (the main protocol) while others with protocol set consume the protocol specified (for example `rest`). One consumer instance will always stick to one specific protocol for each specific service based on the protocol setting on its side, so this pull request also addresses the issue that consumers will randomly use all protocols (tri, rest in this case) for different invocations.

For application-based service discovery, the extra protocol url will still be registered to metadata, so the provide side process is not affected, what I do to this model in this patch is only to decide which protocol to use on the consumer side.


chickenlj:
For those who want multiple protocol support with multiple address URLs  should config like below:
```yaml
dubbo:
  protocols:
    - id: dubbo
       name: dubbo
    - id: rest
       name: rest
```

chickenlj:
With this change, the usage of Dubbo can be as simple as this:
```java
    private static void main(String[] args) {
        ServiceConfig<DemoService> service = new ServiceConfig<>();
        service.setInterface(DemoService.class);
        service.setRef(new DemoServiceImpl());

        DubboBootstrap.getInstance().service(service).start().await();
    }

```

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/deploy/DefaultApplicationDeployer.java
```diff
@@ -747,23 +747,27 @@ private void startModules() {
     }
 
     @Override
-    public void prepareApplicationInstance() {
+    public void prepareApplicationInstance(ModuleModel moduleModel) {
```
AlbumenJ, line 750: ApplicationDeployer depends on ModuleModel might not a good design.
BTW, if there are more than one ModuleModels, will this work not expected?

AlbumenJ (changes requested):

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -368,7 +370,17 @@ protected void exported() {
                 mapServiceName(url, serviceNameMapping, scheduledExecutor);
             }
         });
+
         onExported();
+
+        if (hasRegistrySpecified()) {
+            getScopeModel().getDeployer().getApplicationDeployer().exportMetadataService();
```
AlbumenJ, line 377: Depends on MetadataService here might not a good design

AlbumenJ (changes requested):

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/deploy/DefaultApplicationDeployer.java
chickenlj: DefaultApplicationDeployer.java
```java
    public void notifyModuleChanged(ModuleModel moduleModel, DeployState state) {
        checkState(moduleModel, state);
    }

    public void checkState(ModuleModel moduleModel, DeployState moduleState) {
        synchronized (stateLock) {
            if (!moduleModel.isInternal() && moduleState == DeployState.STARTED) {
                prepareApplicationInstance(moduleModel);
            }
}
```

It's currently designed to work this way: 
1. when the state of a ModuleModel changes, it will call `notifyModuleChanged`
2. `checkState` will then check if this ModelModel isInternal, if true, `prepareApplicationInstance(moduleModel);` will be called.

ApplicationDeployer depends on ModuleModel mainly querying for statuses such as 'STARTED', 'isInternal', and 'hasRegistryInteraction'. All these statuses can be wrapped into an Event or something, but I think this can be further optimized with this issue

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
chickenlj: Consider provide a solution for this issue with https://github.com/apache/dubbo/pull/12750

namelessssssssssss: It seems that the full service export process can move from ServiceConfig to a new ModuleLifecycle implement? This might solve this problem

chickenlj: Yes, I think so, this might need some extra efforts to merge this change with Lifecycle mechanism.

Issue #12125, opened 2023-04-18
[Observability] Error code information is added during exception statistics. 


songxiaosheng:
Observability: Error code information is added during exception statistics. Currently, there are error codes in Dubbo's exceptions. Attempt to collect both the exception and this error code information when an exception occurs.
The error code link is as follows:
https://cn.dubbo.apache.org/zh-cn/overview/mannual/java-sdk/faq/

【可观测性】异常统计时增加错误码信息，目前 dubbo的 异常中都存在错误码 ，尝试在发生异常时候采集异常 同时采集这个错误码信息。
错误码链接如下：
https://cn.dubbo.apache.org/zh-cn/overview/mannual/java-sdk/faq/

AlbumenJ:
agree

namelessssssssssss:
Please assign me.

PR #12661, opened 2023-07-04
Add support for error code metrics (#12125)

namelessssssssssss:
Add support for error code metrics.

## Brief changelog
* ​The error code metric is sampled when log is enabled.


namelessssssssssss:
@wxbty  @songxiaosheng  @AlbumenJ  PTAL  :)

wxbty (changes requested):
Whether to consider using a unified event api, such as EventBus

namelessssssssssss:
> Whether to consider using a unified event api, such as EventBus
dubbo-common doesn't import guava, I think it might be better to use a custom implementation in this simple scene

wxbty:
> > Whether to consider using a unified event api, such as EventBus
> 
> dubbo-common doesn't import guava, I think it might be better to use a custom implementation in this simple scene

MetricsEventBus，The rest of the metrics collection uses this uniformly

namelessssssssssss:
> > > Whether to consider using a unified event api, such as EventBus
> > 
> > 
> > dubbo-common doesn't import guava, I think it might be better to use a custom implementation in this simple scene
> 
> MetricsEventBus，The rest of the metrics collection uses this uniformly

That will import dubbo-metrics-api into dubbo-common.  But later maybe we can split out metrics-event from metrics-api to solve this problem, that can be a part of dubbo-core refactor.

AlbumenJ:
This is a feature, so I have change it to `3.3`

songxiaosheng (approved):
LGTM

AlbumenJ:
image

Please resolve conflicts

PR #12670, opened 2023-07-05
fix: zookeeper dependency conflict

sheiy:


sheiy:
> What is the purpose?

```xml
  <groupId>org.apache.curator</groupId>
  <artifactId>curator-x-discovery</artifactId>
```
include zookeeper conflict with the pom `dubbo-dependencies/dubbo-dependencies-zookeeper-curator5/pom.xml` include

AlbumenJ:
> > What is the purpose?
> 
> zookeeper version of
> 
> ```
>   <groupId>org.apache.curator</groupId>
>   <artifactId>curator-x-discovery</artifactId>
> ```
> 
> conflict with the pom `dubbo-dependencies/dubbo-dependencies-zookeeper-curator5/pom.xml` include

Maven will use the version defined in `dubbo-dependencies/dubbo-dependencies-zookeeper-curator5/pom.xml` instead of curator's version due to the shortest path.

PR #12730, opened 2023-07-14
Optimize performance for some basic APIs

wxbty:
Simply increase the cache for immutable beans or properties .
Can improve the performance of these frequently used APIs

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/beans/factory/ScopeBeanFactory.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/beans/factory/ScopeBeanFactory.java
@@ -30,6 +30,7 @@ import org.apache.dubbo.rpc.model.ScopeModelAccessor;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -53,7 +54,9 @@ public class ScopeBeanFactory {
     private final List<BeanInfo> registeredBeanInfos = new CopyOnWriteArrayList<>();
     private InstantiationStrategy instantiationStrategy;
     private final AtomicBoolean destroyed = new AtomicBoolean();
-    private List<Class<?>> registeredClasses = new ArrayList<>();
+    private final List<Class<?>> registeredClasses = new ArrayList<>();
+
+    private final Map<Class<?>, Object> registeredCacheBeanMap = new ConcurrentHashMap<>(64);
 
     public ScopeBeanFactory(ScopeBeanFactory parent, ExtensionAccessor extensionAccessor) {
         this.parent = parent;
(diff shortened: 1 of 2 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/beans/factory/ScopeBeanFactory.java
```diff
@@ -53,7 +54,9 @@ public class ScopeBeanFactory {
     private final List<BeanInfo> registeredBeanInfos = new CopyOnWriteArrayList<>();
     private InstantiationStrategy instantiationStrategy;
     private final AtomicBoolean destroyed = new AtomicBoolean();
-    private List<Class<?>> registeredClasses = new ArrayList<>();
+    private final List<Class<?>> registeredClasses = new ArrayList<>();
+
+    private final Map<Class<?>, Object> registeredCacheBeanMap = new ConcurrentHashMap<>(64);
```
AlbumenJ, line 59: Cache here should consider different bean with same class but different name

dubbo-common/src/main/java/org/apache/dubbo/rpc/model/ApplicationModel.java
```diff
@@ -234,12 +236,18 @@ public ApplicationConfig getCurrentConfig() {
     }
 
     public String getApplicationName() {
-        return getCurrentConfig().getName();
+        if (StringUtils.isBlank(this.name)) {
+            this.name = getCurrentConfig().getName();
+        }
+        return this.name;
```
AlbumenJ, line 242: ApplicationName and ApplicationModel's name are different

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/beans/factory/ScopeBeanFactory.java
wxbty: Only beans without names are cached here. If most beans have no names, is this enough? Maybe changing the name of the map to NoNameBeanMap is more appropriate

dubbo-common/src/main/java/org/apache/dubbo/rpc/model/ApplicationModel.java
wxbty: ok

dubbo-common/src/main/java/org/apache/dubbo/common/beans/factory/ScopeBeanFactory.java
wxbty: Aactually is the only unique Name

dubbo-common/src/main/java/org/apache/dubbo/rpc/model/ApplicationModel.java
```diff
@@ -234,12 +236,18 @@ public ApplicationConfig getCurrentConfig() {
     }
 
     public String getApplicationName() {
-        return getCurrentConfig().getName();
+        if (StringUtils.isBlank(this.applicationName)) {
+            this.applicationName = getCurrentConfig().getName();
+        }
+        return this.applicationName;
```
AlbumenJ, line 242: `ApplicationName` in config manager may changed in runtime.

dubbo-common/src/main/java/org/apache/dubbo/common/beans/factory/ScopeBeanFactory.java
```diff
@@ -197,8 +200,16 @@ public <T> List<T> getBeansOfType(Class<T> type) {
...
     public <T> T getBean(Class<T> type) {
-        return this.getBean(null, type);
+        if (registeredCacheBeansWithUniqueName.containsKey(type)) {
+            return (T) registeredCacheBeansWithUniqueName.get(type);
+        }
+        T bean = this.getBean(null, type);
+        if (bean != null) {
+            registeredCacheBeansWithUniqueName.put(type, bean);
+        }
+        return bean;
```
AlbumenJ, line 212: Is there a better way to improve the performance in `BeanFactory`

AlbumenJ (changes requested):

PR #12873, opened 2023-08-09
[feat.] new triple protocol implement

icodening:
[feat.] new triple protocol impl
tri protocol integration http12 module


chickenlj (approved):
LGTM.

Issue #12850, opened 2023-08-05
Adds metrics initialization metrics for the service interface


robin977:
Add the metrics initialization indicator of the service interface to 0
添加服务接口的metrics初始化指标数据为0

songxiaosheng:
@robin977 在处理了，最近会有pr

PR #12878, opened 2023-08-10
Adds metrics initialization metrics for the service interface(#12850)

robin977:
Adds metrics initialization metrics for the service interface(#12850)


songxiaosheng:
The branch is outdated. It is recommended to pull a cleaner branch

PR #13017, opened 2023-09-07
fix loading process cannot load environment variables.

chickenlj:
For example, environment variable `DUBBO_REGISTRY_ADDRESS=zookeeper://127.0.0.1:2182` should override same configuration in dubbo.properties `dubbo.registry.address=zookeeper://127.0.0.1:2181`.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/config/ConfigurationUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/config/ConfigurationUtils.java
@@ -317,7 +317,7 @@ public final class ConfigurationUtils {
         }
         for (Map.Entry<String, V> entry : copy.entrySet()) {
             String key = entry.getKey();
-            if (StringUtils.startsWithIgnoreCase(key, prefix)
+            if ((StringUtils.startsWithIgnoreCase(key, prefix) || StringUtils.startsWithIgnoreCase(key, StringUtils.toOSStyleKey(prefix)))
                 && key.length() > prefix.length()
                 && !ConfigurationUtils.isEmptyValue(entry.getValue())) {
                 return true;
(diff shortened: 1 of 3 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/config/ConfigurationUtils.java
```diff
@@ -277,7 +277,7 @@ private static <V extends Object> Map<String, V> getSubProperties(Map<String, V>
             for (Map.Entry<String, V> entry : copy.entrySet()) {
                 String key = entry.getKey();
                 V val = entry.getValue();
-                if (StringUtils.startsWithIgnoreCase(key, prefix)
+                if ((StringUtils.startsWithIgnoreCase(key, prefix) || StringUtils.startsWithIgnoreCase(key, StringUtils.toOSStyleKey(prefix)))
```
wxbty, line 280: There are similar judgments in other places. Should we also change them? for example:
```java
public static <V extends Object> boolean hasSubProperties(Map<String, V> configMap, String prefix) {
        if (!prefix.endsWith(".")) {
            prefix += ".";
        }
        Map<String, V> copy;
        synchronized (configMap) {
            copy = new HashMap<>(configMap);
        }
        for (Map.Entry<String, V> entry : copy.entrySet()) {
            String key = entry.getKey();
            if (StringUtils.startsWithIgnoreCase(key, prefix)
                && key.length() > prefix.length()
                && !ConfigurationUtils.isEmptyValue(entry.getValue())) {
                return true;
            }
        }
        return false;
    }
```

AlbumenJ (approved):

dubbo-common/src/main/java/org/apache/dubbo/common/config/ConfigurationUtils.java
chickenlj: Yes, I think so. I will check this place later.

PR #13066, opened 2023-09-18
feat: can set generic.include.class in a single call to control wheth…

Aresxue:

Can set 'generic.include.class' in a single call to control whether the class field is included in the response when generic invoke.

## Verifying this change

org.apache.dubbo.rpc.filter.GenericFilterTest#testInvokeRemoveClassField

## issue

https://github.com/apache/dubbo/issues/13048

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/PojoUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/PojoUtils.java
@@ -125,11 +125,18 @@ public class PojoUtils {
     }
 
     public static Object generalize(Object pojo) {
-        return generalize(pojo, new IdentityHashMap<>());
+        return generalize(pojo, null);
+    }
+
+    public static Object generalize(Object pojo, Boolean genericWithClz) {
+        if (null == genericWithClz) {
+            genericWithClz = GENERIC_WITH_CLZ;
+        }
+        return generalize(pojo, new IdentityHashMap<Object, Object>(), genericWithClz);
     }
 
     @SuppressWarnings("unchecked")
-    private static Object generalize(Object pojo, Map<Object, Object> history) {
+    private static Object generalize(Object pojo, Map<Object, Object> history, boolean genericWithClz){
         if (pojo == null) {
             return null;
         }
@@ -170,7 +177,7 @@ public class PojoUtils {
             history.put(pojo, dest);
             for (int i = 0; i < len; i++) {
                 Object obj = Array.get(pojo, i);
-                dest[i] = generalize(obj, history);
+                dest[i] = generalize(obj, history, genericWithClz);
             }
             return dest;
         }
@@ -180,7 +187,7 @@ public class PojoUtils {
             Collection<Object> dest = (pojo instanceof List<?>) ? new ArrayList<>(len) : new HashSet<>(len);
             history.put(pojo, dest);
             for (Object obj : src) {
-                dest.add(generalize(obj, history));
+                dest.add(generalize(obj, history, genericWithClz));
             }
             return dest;
         }
@@ -189,20 +196,20 @@ public class PojoUtils {
             Map<Object, Object> dest = createMap(src);
             history.put(pojo, dest);
             for (Map.Entry<Object, Object> obj : src.entrySet()) {
-                dest.put(generalize(obj.getKey(), history), generalize(obj.getValue(), history));
+                dest.put(generalize(obj.getKey(), history, genericWithClz), generalize(obj.getValue(), history, genericWithClz));
             }
             return dest;
         }
         Map<String, Object> map = new HashMap<>();
         history.put(pojo, map);
-        if (GENERIC_WITH_CLZ) {
+        if (genericWithClz) {
             map.put("class", pojo.getClass().getName());
         }
         for (Method method : pojo.getClass().getMethods()) {
             if (ReflectUtils.isBeanPropertyReadMethod(method)) {
                 ReflectUtils.makeAccessible(method);
                 try {
-                    map.put(ReflectUtils.getPropertyNameFromBeanReadMethod(method), generalize(method.invoke(pojo), history));
+                    map.put(ReflectUtils.getPropertyNameFromBeanReadMethod(method), generalize(method.invoke(pojo), history, genericWithClz));
                 } catch (Exception e) {
                     throw new RuntimeException(e.getMessage(), e);
                 }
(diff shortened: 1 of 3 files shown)


Aresxue:
Missing test cases

wxbty:
Is it possible to consider giving up this option?

```java
private static final boolean GENERIC_WITH_CLZ = Boolean.parseBoolean(
ConfigurationUtils.getProperty(ApplicationModel.defaultModel(),CommonConstants.GENERIC_WITH_CLZ_KEY, "true"));
```
There are some problems in getting it from DefaultModel (multiple application scenarios). Maybe we should get the corresponding app configuration from url.getApplicationModel.
In this way, we can simplify the operation and obtain the configuration directly from the URL. It will give priority to obtaining the interface-level configuration (specified by the user) and avoid complex judgments.

AlbumenJ:
> Is it possible to consider giving up this option?
> 
> ```java
> private static final boolean GENERIC_WITH_CLZ = Boolean.parseBoolean(
> ConfigurationUtils.getProperty(ApplicationModel.defaultModel(),CommonConstants.GENERIC_WITH_CLZ_KEY, "true"));
> ```
> 
> There are some problems in getting it from DefaultModel (multiple application scenarios). Maybe we should get the corresponding app configuration from url.getApplicationModel. In this way, we can simplify the operation and obtain the configuration directly from the URL. It will give priority to obtaining the interface-level configuration (specified by the user) and avoid complex judgments.
> 
> @AlbumenJ What do you think?

A question is that we should prevent performance degrade.

AlbumenJ:
How can the user specify it in **one invocation**

Aresxue:
> How can the user specify it in **one invocation**
set attachment in rpc context
'''java
RpcContext.getClientAttachment().setObjectAttachment(GENERIC_WITH_CLZ_KEY, "true");
'''

Aresxue:
> How can the user specify it in **one invocation**
set atachment in RpcContext
`RpcContext.getClientAttachment().setObjectAttachment(GENERIC_WITH_CLZ_KEY, "true");`

Aresxue:
> How can the user specify it in **one invocation**

set attachment in RpcContext
```java
RpcContext.getClientAttachment().setObjectAttachment(GENERIC_WITH_CLZ_KEY, "true");
```

AlbumenJ:
> > How can the user specify it in **one invocation**
> 
> set attachment in RpcContext
> 
> ```java
> RpcContext.getClientAttachment().setObjectAttachment(GENERIC_WITH_CLZ_KEY, "true");
> ```

This is a little wired. I wonder why we should support this future? Is there any scenarios?

Aresxue:
> > > How can the user specify it in **one invocation**
> > 
> > 
> > set attachment in RpcContext
> > ```java
> > RpcContext.getClientAttachment().setObjectAttachment(GENERIC_WITH_CLZ_KEY, "true");
> > ```
> 
> This is a little wired. I wonder why we should support this future? Is there any scenarios?

Because we don’t want to rely solely on the provider’s configuration to control whether to return the class field. Different consumers have different expectations for the class field. Our open platform will return the results to the third-party manufacturers. They want a result that does not contain class, but for In other scenarios, for example, I am doing dubbo traffic playback, then I want to get the class field when making a generalized call, because I need to restore the original untyped object to a java type object. In addition, we are based on dubbo Frameworks that implement extension point mechanisms also rely on classes to restore type information.

AlbumenJ:
> > > > How can the user specify it in **one invocation**
> > > 
> > > 
> > > set attachment in RpcContext
> > > ```java
> > > RpcContext.getClientAttachment().setObjectAttachment(GENERIC_WITH_CLZ_KEY, "true");
> > > ```
> > 
> > 
> > This is a little wired. I wonder why we should support this future? Is there any scenarios?
> 
> Because we don’t want to rely solely on the provider’s configuration to control whether to return the class field. Different consumers have different expectations for the class field. Our open platform will return the results to the third-party manufacturers. They want a result that does not contain class, but for In other scenarios, for example, I am doing dubbo traffic playback, then I want to get the class field when making a generalized call, because I need to restore the original untyped object to a java type object. In addition, we are based on dubbo Frameworks that implement extension point mechanisms also rely on classes to restore type information.

How about add a Filter to customize the response object? This would be more flexible.

Aresxue:
> > > > > How can the user specify it in **one invocation**
> > > > 
> > > > 
> > > > set attachment in RpcContext
> > > > ```java
> > > > RpcContext.getClientAttachment().setObjectAttachment(GENERIC_WITH_CLZ_KEY, "true");
> > > > ```
> > > 
> > > 
> > > This is a little wired. I wonder why we should support this future? Is there any scenarios?
> > 
> > 
> > Because we don’t want to rely solely on the provider’s configuration to control whether to return the class field. Different consumers have different expectations for the class field. Our open platform will return the results to the third-party manufacturers. They want a result that does not contain class, but for In other scenarios, for example, I am doing dubbo traffic playback, then I want to get the class field when making a generalized call, because I need to restore the original untyped object to a java type object. In addition, we are based on dubbo Frameworks that implement extension point mechanisms also rely on classes to restore type information.
> 
> How about add a Filter to customize the response object? This would be more flexible.

Sorry, I don't understand what filter has to do with this. If you mean to remove the redundant class field, consumers can traverse it in their own project code. For now, if the provider is configured with -Dgeneric.include.class=false, the consumer will not be able to obtain the class field anyway. Then I will not be able to restore the java object in some scenarios.

AlbumenJ:
> > > > > > How can the user specify it in **one invocation**
> > > > > 
> > > > > 
> > > > > set attachment in RpcContext
> > > > > ```java
> > > > > RpcContext.getClientAttachment().setObjectAttachment(GENERIC_WITH_CLZ_KEY, "true");
> > > > > ```
> > > > 
> > > > 
> > > > This is a little wired. I wonder why we should support this future? Is there any scenarios?
> > > 
> > > 
> > > Because we don’t want to rely solely on the provider’s configuration to control whether to return the class field. Different consumers have different expectations for the class field. Our open platform will return the results to the third-party manufacturers. They want a result that does not contain class, but for In other scenarios, for example, I am doing dubbo traffic playback, then I want to get the class field when making a generalized call, because I need to restore the original untyped object to a java type object. In addition, we are based on dubbo Frameworks that implement extension point mechanisms also rely on classes to restore type information.
> > 
> > 
> > How about add a Filter to customize the response object? This would be more flexible.
> 
> Sorry, I don't understand what filter has to do with this. If you mean to remove the redundant class field, consumers can traverse it in their own project code. For now, if the provider is configured with -Dgeneric.include.class=false, the consumer will not be able to obtain the class field anyway. Then I will not be able to restore the java object in some scenarios.

Design a customized Filter by yourself

PR #13139, opened 2023-09-27
fix https://github.com/apache/dubbo/issues/13135

andotorg:
add openjdk support javascript unit test jar pom

## Brief changelog
Fix the problem of maven build interruption due to lack of js script parsing jar package

## Verifying this change

build successful！

## Checklist
- [✅] fix  [GitHub_issue](https://github.com/apache/dubbo/issues/13135)


dubbo-cluster/pom.xml
```diff
@@ -91,5 +92,10 @@
             <artifactId>micrometer-tracing-integration-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.nashorn</groupId>
+            <artifactId>nashorn-core</artifactId>
+            <version>${nashorn-core.version}</version>
+        </dependency>
```
AlbumenJ, line 99: Test scope would be better. We should not import such engine by default.

andotorg: Okay, I'll change it to "testing".

AlbumenJ (approved):

PR #13217, opened 2023-10-17
Only throws exception when exporting service

AlbumenJ:
## Checklist

### DIFF
--- a/dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/protocol/QosProtocolWrapper.java
+++ b/dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/protocol/QosProtocolWrapper.java
@@ -75,13 +75,13 @@ public class QosProtocolWrapper implements Protocol, ScopeModelAware {
 
     @Override
     public <T> Exporter<T> export(Invoker<T> invoker) throws RpcException {
-        startQosServer(invoker.getUrl());
+        startQosServer(invoker.getUrl(), true);
         return protocol.export(invoker);
     }
 
     @Override
     public <T> Invoker<T> refer(Class<T> type, URL url) throws RpcException {
-        startQosServer(url);
+        startQosServer(url, false);
         return protocol.refer(type, url);
     }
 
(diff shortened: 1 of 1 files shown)


dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/protocol/QosProtocolWrapper.java
```diff
@@ -134,13 +134,18 @@ private void startQosServer(URL url) throws RpcException {
...
-                stopServer();
-            } catch (Throwable stop) {
-                logger.warn(QOS_FAILED_START_SERVER, "", "", "Fail to stop qos server: ", stop);
-            }
+
             if (qosCheck) {
-                throw new RpcException(throwable);
+                try {
+                    // Stop QoS Server to support re-start if Qos-Check is enabled
+                    stopServer();
```
chickenlj, line 141: This causes `stopServer()` only be called when `qosCheck=true` is met, is this expected?

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/protocol/QosProtocolWrapper.java
```diff
@@ -134,13 +134,18 @@ private void startQosServer(URL url) throws RpcException {
...
+                } catch (Throwable stop) {
+                    logger.warn(QOS_FAILED_START_SERVER, "", "", "Fail to stop qos server: ", stop);
+                }
+                if (isServer) {
```
chickenlj, line 145: Why do we need to tell 'export' from 'refer' and only throw exception for 'export', I don't get the  intention behind this change.

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/protocol/QosProtocolWrapper.java
AlbumenJ: Yes. If disable check, we should not retry start in the future.

dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/protocol/QosProtocolWrapper.java
AlbumenJ: In some scenarios, a reference to Dubbo is created when called. If we fail, this will cause the call to fail.

PR #13240, opened 2023-10-20
Set a threshold for instance event changes

finefuture:
Threshold=number of valid nodes after change/number of original valid nodes.

Assuming a threshold of 0.3:
1. The original 100 valid nodes received an event: 80 valid nodes offline, 20/100 less than the threshold of 0.3, discarding this change
2. The original 100 valid nodes have received an event: 50 valid nodes offline, with 50/100 exceeding the threshold of 0.3. Process this change
3. There are 100 original nodes, including 50 valid nodes and 50 unavailable nodes. Received event: 80 offline nodes (40 valid nodes+40 unavailable nodes), with 10/50 less than the threshold of 0.3. Discard this change
4. There are 100 original nodes, including 50 valid nodes and 50 unavailable nodes. Received event: 80 offline nodes (30 valid nodes+50 unavailable nodes), 20/50 is greater than the threshold of 0.3. Process this change
5. There are 100 original nodes, including 50 valid nodes and 50 unavailable nodes. Received event: 50 offline nodes (50 valid nodes), 0/50 is less than the threshold of 0.3, discard this change
6. There are 100 original nodes, including 50 valid nodes and 50 unavailable nodes. Received event: 50 offline nodes (10 valid nodes+40 unavailable nodes), with 40/50 exceeding the threshold of 0.3. Process this change

Scenario:
1. Provider publishing or scaling requires mass publishing, and the threshold will be lower than the normal threshold

> - It is necessary to have all upstream users change the threshold for this appkey

2. Due to the special nature of the business, the consumer service needs to adjust the threshold, such as the core service wanting to adjust the threshold higher to ensure the success rate of the call

> - It is necessary to change the default threshold for all appkeys in this service

Priority: provider>consumer>default

## Brief changelog
After the registration center recovers from a failure, all temporary nodes will be re registered, and a smaller number of nodes will be pushed to the consumer (such as events for one node or two nodes). If the consumer refreshes the list, it will result in only a few nodes receiving traffic, and in high traffic situations, a large number of reject exceptions will occur, causing traffic loss

## Checklist


AlbumenJ:
`org.apache.dubbo.registry.AddressListener#notify` may be useful for you. You can customize the url in the AddressListener, or the event returns the previous result.
For this PR, I think it is dangerous to merge. This may cause notifications to be lost in some cases.

finefuture:
> `org.apache.dubbo.registry.AddressListener#notify` may be useful for you. You can customize the url in the AddressListener, or the event returns the previous result. For this PR, I think it is dangerous to merge. This may cause notifications to be lost in some cases.

The AddressListener is OK, which is designed to improve the high availability of rpc, similar to the enableEmptyProtection parameter.
It can be temporarily held, don't merge yet, see if anyone else has the same problem in the future.

finefuture:
Perhaps providing an extension would be better, I will close this PR.

PR #13295, opened 2023-11-01
Revert "Protobuf method not allow override"

AlbumenJ:
Reverts apache/dubbo#13155


SpringStudent:
I am not sure can i check pb method override at ServiceConfig.postProcessRefresh.

PR #13309, opened 2023-11-06
Update com.alibaba.fastjson2:fastjson2: 2.0.41 -> 2.0.42

chaoyoung:

Upgrade fastsjon2 2.0.42


CrazyHZM:
@chaoyoung 
Thanks for your contribution, dubbo has configured a dependabot, which will upgrade automatically. We don't need to worry about dependent version upgrades

PR #13347, opened 2023-11-11
Fix 3.3 application name

chickenlj:


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/deploy/DefaultApplicationDeployer.java
```diff
@@ -779,6 +780,13 @@ public void prepareApplicationInstance(ModuleModel moduleModel) {
         // export MetricsService
         exportMetricsService();
 
+        if (moduleModel.getDeployer().hasRegistryInteraction()) {
+             ApplicationConfig applicationConfig = configManager.getApplicationOrElseThrow();
+             if (DEFAULT_APP_NAME.equals(applicationConfig.getName())) {
+                throw new IllegalStateException("Application name must be set when registry is enabled.");
+             }
```
AlbumenJ, line 787: I think it would be better to allow register interface mode if users have not specified the application name. It works in interface mode.

AlbumenJ (changes requested):

AlbumenJ (approved):

PR #13334, opened 2023-11-08
feat: fix #13327

Linchen-Xu:
## What is the purpose of the change
fix #13327 
- `dubbo.jstack-dump.max-line` can be used to indicate that all stack trace lines should be printed now by setting it to a negative number, -1 for example.
- write stack trace info directly into OutputStream instead of building a String

## Brief changelog
update org.apache.dubbo.common.utils.JVMUtil

## Checklist
- [x] Make sure there is a [GitHub_issue](https://github.com/apache/dubbo/issues) field for the change (usually before you start working on it). Trivial changes like typos do not require a GitHub issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [x] Each commit in the pull request should have a meaningful subject line and body.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Check if is necessary to patch to Dubbo 3 if you are work on Dubbo 2.7
- [ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add sample in [dubbo samples](https://github.com/apache/dubbo-samples) project.
- [ ] Add some description to [dubbo-website](https://github.com/apache/dubbo-website) project if you are requesting to add a feature.
- [ ] GitHub Actions works fine on your own branch.
- [ ] If this contribution is large, please follow the [Software Donation Guide](https://github.com/apache/dubbo/wiki/Software-donation-guide).

PR #13358, opened 2023-11-14
(reopened) feat: fix #13327

### LINKED ISSUE #13327: JVMUtil. jstack方法调用整体使用OutputStream作为输出可提升性能，而内部JVMUtil.getThreadDumpString使用String长字符串拼接反而降低了性能。
https://github.com/apache/dubbo/blob/3.0/dubbo-common/src/main/java/org/apache/dubbo/common/utils/JVMUtil.java

JVMUtil. jstack方法调用整体使用OutputStream作为输出可提升性能，而内部JVMUtil.getThreadDumpString使用String长字符串拼接反而降低了性能。
建议修改内部JVMUtil.getThreadDumpString方法同样为流式处理，减少字符串拼接的性能开销

同时，getThreadDumpString方法内部，可能还考虑性能影响，增加了栈深度控制（默认32）。
这个深度控制、对性能的提升，可能没有流式处理改造能带来的性能提升高。建议应先修改流式处理以提升性能。

Linchen-Xu:
Moved from #13334 due to branch operation error.

fix #13327 
- `dubbo.jstack-dump.max-line` can be used to indicate that all stack trace lines should be printed now by setting it to a negative number, -1 for example.
- write stack trace info directly into OutputStream instead of building a String.
- Add tests in `JVMUtilTest`.

## Brief changelog
- Support custom print stack depth and write error messages directly into OutputStream in `org.apache.dubbo.common.utils.JVMUtil`.
- Add tests in `org.apache.dubbo.common.utils.JVMUtilTest`.

## Checklist
- [ ] Add some description to [dubbo-website](https://github.com/apache/dubbo-website) project if you are requesting to add a feature. (will do after it is merged)

### DIFF
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/utils/JVMUtilTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/utils/JVMUtilTest.java
@@ -16,4 +16,60 @@
  */
 package org.apache.dubbo.common.utils;
 
-class JVMUtilTest {}
+import org.apache.dubbo.common.logger.ErrorTypeAwareLogger;
+import org.apache.dubbo.common.logger.LoggerFactory;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import org.junit.jupiter.api.Test;
+
+import static org.apache.dubbo.common.constants.CommonConstants.DUBBO_JSTACK_MAXLINE;
+import static org.apache.dubbo.common.constants.CommonConstants.OS_NAME_KEY;
+import static org.apache.dubbo.common.constants.CommonConstants.OS_WIN_PREFIX;
+import static org.apache.dubbo.common.constants.LoggerCodeConstants.COMMON_UNEXPECTED_CREATE_DUMP;
+import static org.apache.dubbo.common.utils.JVMUtil.jstack;
+
+class JVMUtilTest {
+
+    protected static final ErrorTypeAwareLogger logger = LoggerFactory.getErrorTypeAwareLogger(JVMUtilTest.class);
+
+    @Test
+    void testPrintStackTraceWithSpecifiedDepth() {
+        test(10);
+    }
+
+    @Test
+    void testPrintStackTraceWithUnlimitedDepth() {
+        test(-1);
+    }
+
+    private void test(Integer depth) {
+        SimpleDateFormat sdf;
+        String os = System.getProperty(OS_NAME_KEY).toLowerCase();
+        // window system don't support ":" in file name
+        if (os.contains(OS_WIN_PREFIX)) {
+            sdf = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss .SSS");
+        } else {
+            sdf = new SimpleDateFormat("yyyy-MM-dd_HH:mm:ss .SSS");
+        }
+        String dateStr = sdf.format(new Date());
+
+        String property = System.getProperty(DUBBO_JSTACK_MAXLINE);
+        System.setProperty(DUBBO_JSTACK_MAXLINE, depth.toString());
+        try (FileOutputStream jStackStream =
+                 new FileOutputStream(new File("/tmp", "Dubbo_JStack.log" + "." + dateStr))) {
+            jstack(jStackStream);
+        } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


dubbo-common/src/test/java/org/apache/dubbo/common/utils/JVMUtilTest.java
```diff
@@ -16,4 +16,60 @@
...
+        } else {
+            sdf = new SimpleDateFormat("yyyy-MM-dd_HH:mm:ss .SSS");
+        }
+        String dateStr = sdf.format(new Date());
+
+        String property = System.getProperty(DUBBO_JSTACK_MAXLINE);
+        System.setProperty(DUBBO_JSTACK_MAXLINE, depth.toString());
+        try (FileOutputStream jStackStream =
+                new FileOutputStream(new File("/tmp", "Dubbo_JStack.log" + "." + dateStr))) {
+            jstack(jStackStream);
```
AlbumenJ, line 64: Check the result which it print

AlbumenJ (changes requested):

dubbo-common/src/test/java/org/apache/dubbo/common/utils/JVMUtilTest.java
Linchen-Xu: Added the comparison between the depth printed by the old logic and the new logic. PTAL.
When it traces back to JVMUtil, the line number will be different, making purely comparing two outputs impractical, unless we ignore some lines in the output.

dubbo-common/src/test/java/org/apache/dubbo/common/utils/JVMUtilTest.java
```diff
@@ -16,4 +16,190 @@
...
+        } catch (Exception e) {
+            logger.error(COMMON_UNEXPECTED_CREATE_DUMP, "", "", "dump jStack error", e);
+        }
+
+        // compare the depth between new trace and old trace
+        try {
+            byte[] encoded = Files.readAllBytes(Paths.get(String.format("/tmp/%s", filename)));
+            String newOutput = new String(encoded, StandardCharsets.UTF_8);
+            int newStackDepth = calculateStackDepth(newOutput);
```
AlbumenJ, line 89: Maybe we can decode the stack file with some tools to verify if it match the standard output style.

AlbumenJ (changes requested):

dubbo-common/src/test/java/org/apache/dubbo/common/utils/JVMUtilTest.java
Linchen-Xu: I replaced it with `ByteArrayOutputStream` to get the result directly, there's no need to decode anymore.

By the way, I tried dumping threads and using the same thread info to generate the stack trace info in both the old way and the new way. Therefore, we can compare the stack trace directly, which turns out to be working correctly.

dubbo-common/src/test/java/org/apache/dubbo/common/utils/JVMUtilTest.java
```diff
@@ -16,4 +16,168 @@
...
+            ThreadInfo[] threadInfos = JVMUtil.dumpAllThreads();
+            // generate new stack trace info
+            JVMUtil.jstack(stream, threadInfos);
+            String newStackTrace = stream.toString();
+            // generate old stack trace info
+            String oldStackTrace = OldJVMUtil.jstack(threadInfos);
+            // calculate stack trace depth
+            int newStackDepth = calculateStackDepth(newStackTrace);
+            int oldStackDepth = calculateStackDepth(oldStackTrace);
+
```
AlbumenJ, line 67: I think we need to verify if `JVMUtil` work as expected, not work as previous version/

AlbumenJ (changes requested):

dubbo-common/src/test/java/org/apache/dubbo/common/utils/JVMUtilTest.java
Linchen-Xu: Added stack dump test code in `AbortPolicyWithReportTest.jStackDumpTest` to ensure that the new code is working as expected, but have to check the output file under /tmp.

The output file after running `AbortPolicyWithReportTest.jStackDumpTest` when setting `dubbo.jstack-dump.max-line=-1`:

AlbumenJ (approved):
LGTM

AlbumenJ:
@CrazyHZM PTAL

PR #13383, opened 2023-11-17
Provide a way to turn off the Spotless Plugin, and turn it off in non-style-checking phase of CI.

win120a:
Provide a way to turn off Spotless through Maven Profile. 

In [this workflow run result](https://github.com/apache/dubbo/actions/runs/6833017279/job/18584329187), the `error-code-inspecting` phase can't continue since the Spotless Plugin banned the process.

## Brief changelog
Add a Maven profile to turn off the Spotless Plugin, and turn on the profile in the `error-code-inspecting` phase. 

The `.editorconfig` is also changed to match the Spotless Plugin's rule.

### DIFF
--- a/dubbo-build-tools/pom.xml
+++ b/dubbo-build-tools/pom.xml
@@ -31,7 +31,7 @@
 
   <profiles>
     <profile>
-      <id>java11+</id>
+      <id>spotless</id>
       <activation>
         <jdk>[11,)</jdk>
       </activation>
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,6 +26,11 @@ end_of_line = lf
 insert_final_newline = true
 
 # 4 space indentation
-[*.{java,xml}]
+[*.java]
 indent_style = space
 indent_size = 4
+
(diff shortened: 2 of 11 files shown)


dubbo-build-tools/pom.xml
```diff
@@ -84,6 +84,13 @@
         </plugins>
       </build>
     </profile>
+
+    <profile>
+      <id>turn-off-spotless</id>
```
AlbumenJ, line 89: How about rename spotless profile to `<id>spotless</id>` and disable it with `-P-spotless`

AlbumenJ (changes requested):

dubbo-build-tools/pom.xml
win120a: I can't just rename the profile to 'spotless' since the Spotless Plugin is enabled by default. However, for a better understanding, I inverted the default action and re-introduce a profile called 'spotless' to enable the plugin.

AlbumenJ (commented):
image

Rename this profile id can satisfy

PR #13559, opened 2023-12-23
Fix proto `package` and ` java_package` issue

namelessssssssssss:
Fix issue https://github.com/apache/dubbo/issues/13496

* Add provider config `useJavaPackageAsPath`
* `useJavaPackageAsPath=ture` : Provider will export with java_package as path, which will cause exception when `package` and `java_package` are different.
* `useJavaPackageAsPath=false` (default) : Provider will export with package as path.

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
@@ -554,7 +554,7 @@ public class ServiceConfig<T> extends ServiceConfigBase<T> {
                     .orElse((ProviderConfig)
                             getScopeModel().getConfigManager().getProviders().toArray()[0]);
             if (!providerConfig.getUseJavaPackageAsPath()) {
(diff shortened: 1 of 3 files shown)


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ServiceConfig.java
```diff
@@ -548,6 +548,15 @@ private void doExportUrls(RegisterTypeEnum registerType) {
         final boolean serverService = ref instanceof ServerService;
         if (serverService) {
             serviceDescriptor = ((ServerService) ref).getServiceDescriptor();
+            ProviderConfig providerConfig = getScopeModel()
```
EarthChen, line 551: Is it possible to do this in a simpler way?

chickenlj: Has the `url` instance already been generated at this point?

EarthChen (approved):
LGTM

icodening (approved):
LGTM

PR #13582, opened 2023-12-26
Fix StubMethodDescriptor generate issue

namelessssssssssss:
One possible solution to fix issues mentioned in https://github.com/apache/dubbo/pull/13221


EarthChen (approved):
LGTM

PR #13585, opened 2023-12-26
Fix protobuf class deep copy issue

namelessssssssssss:
* fix https://github.com/apache/dubbo/issues/13497 
* fix https://github.com/apache/dubbo/issues/13498
* fix https://github.com/apache/dubbo/issues/13158


dubbo-rpc/dubbo-rpc-injvm/src/main/java/org/apache/dubbo/rpc/protocol/injvm/InjvmInvoker.java
```diff
@@ -82,7 +82,8 @@ public class InjvmInvoker<T> extends AbstractInvoker<T> {
         this.executorRepository = ExecutorRepository.getInstance(url.getOrDefaultApplicationModel());
         this.paramDeepCopyUtil = url.getOrDefaultFrameworkModel()
                 .getExtensionLoader(ParamDeepCopyUtil.class)
-                .getExtension(url.getParameter(CommonConstants.INJVM_COPY_UTIL_KEY, DefaultParamDeepCopyUtil.NAME));
+                .getActivateExtensions()
+                .get(0);
```
AlbumenJ, line 86: It might not a good idea to use `get(0)` here

AlbumenJ (changes requested):

PR #13592, opened 2023-12-30
Fix dubbo.tag override by consumer.tag in NacosRegistry

namelessssssssssss:
Fix https://github.com/apache/dubbo/issues/13184  
(`dubbo.tag` override by `consumer.tag` in `NacosRegistry`, which causes `TagStateRouter` not worked properly when provider using interface-level registry.)


namelessssssssssss:
@AlbumenJ PTAL

AlbumenJ (approved):

chickenlj (approved):
LGTM.

PR #13593, opened 2024-01-01
Fix MergeableClusterInvoker log issue

namelessssssssssss:
* fix https://github.com/apache/dubbo/issues/12614

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/MergeableClusterInvoker.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/MergeableClusterInvoker.java
@@ -183,7 +183,7 @@ public class MergeableClusterInvoker<T> extends AbstractClusterInvoker<T> {
 
     private void checkNoProvider(RpcException e, String invokerGroup) throws RpcException {
         if (e.isNoInvokerAvailableAfterFilter()) {
-            log.debug("No available provider for service" + getUrl().getServiceKey() + " on group " + invokerGroup
+            log.warn("No available provider for service" + getUrl().getServiceKey() + " on group " + invokerGroup
                     + ", will continue to try another group.");
         } else {
             throw e;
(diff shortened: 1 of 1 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/MergeableClusterInvoker.java
```diff
@@ -182,6 +181,15 @@ protected Result doInvoke(Invocation invocation, List<Invoker<T>> invokers, Load
         return AsyncRpcResult.newDefaultAsyncResult(result, invocation);
     }
 
+    private void checkNoProvider(RpcException e, String invokerGroup) throws RpcException {
+        if (e.isNoInvokerAvailableAfterFilter()) {
+            log.debug("No available provider for service" + getUrl().getServiceKey() + " on group " + invokerGroup
```
AlbumenJ, line 186: Should this log being updated?

AlbumenJ (changes requested):

CrazyHZM (approved):

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/MergeableClusterInvoker.java
CrazyHZM: @AlbumenJ @namelessssssssssss 
If there is no merger, it makes sense to debug because it will log every call, and if there is a merger, you can consider adjusting the level to warn/info.

AlbumenJ:
image

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/support/MergeableClusterInvoker.java
```diff
@@ -93,7 +94,20 @@ protected Result doInvoke(Invocation invocation, List<Invoker<T>> invokers, Load
...
+            try {
+                results.put(invoker.getUrl().getServiceKey(), invokeWithContext(invoker, subInvocation));
+            } catch (RpcException e) {
+                if (e.isNoInvokerAvailableAfterFilter()) {
+                    log.warn(
+                            LoggerCodeConstants.CLUSTER_NO_VALID_PROVIDER,
+                            "",
+                            "",
+                            "No available provider for service" + getUrl().getServiceKey() + " on group "
+                                    + invoker.getUrl().getGroup() + ", will continue to try another group.");
```
CrazyHZM, line 106: Add the exception stack as well.
@namelessssssssssss

CrazyHZM (changes requested):

PR #13609, opened 2024-01-04
fix AbortPolicyWithReport Semaphore lock not released

CycleMaker:
https://github.com/apache/dubbo/issues/13605
fix AbortPolicyWithReport Semaphore lock not released

## Brief changelog
image

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/support/AbortPolicyWithReport.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/support/AbortPolicyWithReport.java
@@ -156,13 +156,13 @@ public class AbortPolicyWithReport extends ThreadPoolExecutor.AbortPolicy {
         if (!guard.tryAcquire()) {
             return;
         }
-        // To avoid multiple dump, check again
-        if (System.currentTimeMillis() - lastPrintTime < TEN_MINUTES_MILLS) {
-            return;
-        }
-
-        ExecutorService pool = Executors.newSingleThreadExecutor();
+        ExecutorService pool = null;
         try {
+            // To avoid multiple dump, check again
+            if (System.currentTimeMillis() - lastPrintTime < TEN_MINUTES_MILLS) {
+                return;
+            }
+            pool = Executors.newSingleThreadExecutor();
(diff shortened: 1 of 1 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/support/AbortPolicyWithReport.java
```diff
@@ -158,6 +158,7 @@ private void dumpJStack() {
         }
         // To avoid multiple dump, check again
         if (System.currentTimeMillis() - lastPrintTime < TEN_MINUTES_MILLS) {
+            guard.release();
```
AlbumenJ, line 161: Maybe we can move 
```
        if (System.currentTimeMillis() - lastPrintTime < TEN_MINUTES_MILLS) {
            guard.release();
            return;
        }

        ExecutorService pool = Executors.newSingleThreadExecutor();
```
into try-finally block

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/support/AbortPolicyWithReport.java
CycleMaker: sorry,I didn't unstand.I haven't realized the advantages of moving these codes into try-finally block.
in this what you means? 

        if (!guard.tryAcquire()) {
            return;
        }
        
        ExecutorService pool;
        try {
            // To avoid multiple dump, check again
            if (System.currentTimeMillis() - lastPrintTime < TEN_MINUTES_MILLS) {
                guard.release();
                return;
            }
            pool = Executors.newSingleThreadExecutor();
            ...........

AlbumenJ: guard can be released in the finally block

CycleMaker: done, please take another look.

AlbumenJ (approved):

PR #13616, opened 2024-01-06
style(editorconfig): Maintain the source code format and spotless plu…

oxsean:
Maintain the source code format and spotless plugin format the same as much as possible through editorconfig


CrazyHZM (changes requested):
Please submit it to the 3.2 branch.

PR #13617, opened 2024-01-06
Hotfix rest Chinese encoding

MajaChen:
When utilizing Dubbo REST with the @QueryParam annotation on parameters, issues with Chinese character decoding may arise. Refer to the [13349 issue](https://github.com/apache/dubbo/issues/13349#issuecomment-1879523852) for a comprehensive and detailed description of the problem, including its cause, solution, and an explanation of why it works.

## Brief changelog
RequestFacade.initParameters will decode the request uri with UTF-8 before parsing uri to parameters.

## Verifying this change
- I add more unit tests , then can pass, and I recommend you refer the [13349 issue](https://github.com/apache/dubbo/issues/13349#issuecomment-1879523852) 

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/request/RequestFacade.java
+++ b/dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/request/RequestFacade.java
@@ -57,21 +57,21 @@ public abstract class RequestFacade<T> {
 
     protected void initParameters() {
         String requestURI = getRequestURI();
-        
-        String decodedRequestURI  = requestURI;
+
+        String decodedRequestURI = requestURI;
         try {
-            decodedRequestURI  = URLDecoder.decode(decodedRequestURI , "UTF-8");
-            if (StringUtils.isNotEmpty(decodedRequestURI )) {
-                requestURI = decodedRequestURI ;
+            decodedRequestURI = URLDecoder.decode(decodedRequestURI, "UTF-8");
+            if (StringUtils.isNotEmpty(decodedRequestURI)) {
+                requestURI = decodedRequestURI;
             }
         } catch (UnsupportedEncodingException e) {
             // do nothing
         }
-        
-        if (StringUtils.isNotEmpty(decodedRequestURI )) {
-            requestURI = decodedRequestURI ;
+
+        if (StringUtils.isNotEmpty(decodedRequestURI)) {
+            requestURI = decodedRequestURI;
         }
-        
+
         if (requestURI != null && requestURI.contains("?")) {
 
             String queryString = requestURI.substring(requestURI.indexOf("?") + 1);
--- a/dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/request/RequestFacade.java
+++ b/dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/request/RequestFacade.java
@@ -61,9 +61,6 @@ public abstract class RequestFacade<T> {
         String decodedRequestURI = requestURI;
         try {
             decodedRequestURI = URLDecoder.decode(decodedRequestURI, "UTF-8");
-            if (StringUtils.isNotEmpty(decodedRequestURI)) {
-                requestURI = decodedRequestURI;
-            }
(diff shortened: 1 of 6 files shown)


AlbumenJ:
@suncairong163 PTAL

MajaChen:
I'm wondering if anybody can review my code

dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/request/RequestFacade.java
```diff
@@ -56,6 +58,17 @@ protected void initHeaders() {}
     protected void initParameters() {
         String requestURI = getRequestURI();
 
+        String decodedRequestURI = requestURI;
+        try {
+            decodedRequestURI = URLDecoder.decode(decodedRequestURI, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            // do nothing
+        }
```
CrazyHZM: This is hardcoded, we can set utf-8 by default, but it must come from `Accept-Charset` first.

CrazyHZM (changes requested):

dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/request/RequestFacade.java
MajaChen: ok, I'll make a adjustment

MajaChen: I've made an adjustment according to your opinon, please take a look again, take your time

kimmking:
As json standard format, non-ansi chars are not allowed, so is nessary to do this?

MajaChen:
Take an example of SpringMVC framework, Chinese character encoding problems are very common among Chinese developers, and it offers solution. I think dubbo should also offer such solution to avoid such problems, even though it's not common to pass Chinese character as parameter. But in my situation,I does need to do so.

MajaChen:
And many users including me, are not familiar with  json standard format(sorry for that).So this bug can be a trap

kimmking:
I know your case now.
BTW: It need/maybe process q param to parse Accept-Charset field like:
```
Accept-Charset: <charset>

// Multiple types, weighted with the [quality value](https://developer.mozilla.org/zh-CN/docs/Glossary/Quality_values) syntax:
Accept-Charset: utf-8, iso-8859-1;q=0.5
```

refer:
https://developer.mozilla.org/zh-CN/docs/Web/HTTP/Headers/Accept-Charset

MajaChen:
I add a new method parseAcceptCharset, it can deal with multi values related with Accept-Charset.
It worked well as demonstrated by unit test.
I think it's ok to take the charset with the biggest weight alone.

I'm doubtful whether these code have pratical function in most case. As Accept-Charset is not recommended to use.
Most importantly, OkhttpClient don't take Accept-Charset into consideration when encoding url,see:
builder.url(requestTemplate.getURL());  (OKHttpRestClient.java line 55)

But I still think adding Accept-Charset make code more rigorous and compatible

MajaChen:
PATL, thanks

dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/request/RequestFacade.java
```diff
@@ -55,6 +58,24 @@ protected void initHeaders() {}
 
     protected void initParameters() {
         String requestURI = getRequestURI();
+        String decodedRequestURI = null;
+
+        try {
+            String enc = "UTF-8";
```
CrazyHZM: It should also be placed in `Constant`.

CrazyHZM (changes requested):

dubbo-rpc/dubbo-rpc-rest/src/main/java/org/apache/dubbo/rpc/protocol/rest/request/RequestFacade.java
MajaChen: fine

CrazyHZM (approved):

PR #13633, opened 2024-01-09
Fix  MultipleRegistry  cache file  error name #13603

shaowYe:

bugfix #13603


AlbumenJ:
LGTM
@chickenlj @CrazyHZM PTAL

CrazyHZM (approved):

PR #13644, opened 2024-01-12
New MetadataService with native stub

namelessssssssssss:
Migrate MetadataService to native stub to support https://github.com/apache/dubbo/issues/13584

* now MetadataService always use Triple protocol with native stub.


dubbo-metadata/dubbo-metadata-api/pom.xml
```diff
@@ -84,6 +84,52 @@
...
+        <executions>
+          <execution>
+            <goals>
+              <goal>compile</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
```
chickenlj, line 134: I think we should either put the generated source codes in the source directory or only rely on this plugin and generate code when packaging.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/metadata/ConfigurableMetadataServiceExporter.java
```diff
@@ -65,13 +66,12 @@ public synchronized ConfigurableMetadataServiceExporter export() {
                     .getInternalServiceExecutor();
             this.serviceConfig = InternalServiceConfigBuilder.<MetadataService>newBuilder(applicationModel)
                     .interfaceClass(MetadataService.class)
-                    .protocol(getApplicationConfig().getMetadataServiceProtocol(), METADATA_SERVICE_PROTOCOL_KEY)
+                    .protocol(TRIPLE, METADATA_SERVICE_PROTOCOL_KEY)
```
FoghostCn, line 69: so `application.metadataServiceProtocol ` will not work in the future

dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/MetadataService.java
FoghostCn, line 1: why delete `MetadataService.class`, the generated MetadataService can completely replace it?

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/metadata/ConfigurableMetadataServiceExporter.java
FoghostCn: If we always use tri protocol, how can we compatible with old version

namelessssssssssss: Yes. I will do some refactor later to make sure it can compatible with older version.

chickenlj: > Yes. I will do some refactor later to make sure it can compatible with older version.

I think it's desired to only support triple protocol in the new release. But we need to guarantee compatibility.

dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/MetadataService.java
chickenlj: I think so.  The service definition is as shown below:

```protobuf
package org.apache.dubbo.metadata;

option go_package = "dubbo.apache.org/dubbo-go/v3/metadata/triple_api";
option java_package = "org.apache.dubbo.metadata";
option java_multiple_files = true;
```

PR #13607, opened 2024-01-03
feat(triple): Dubbo triple&rest protocol convergense

oxsean:
## What is the purpose of the change
See #13427,only for code review.

## Brief changelog
1.  Refine Triple Transport layer code for the REST protocol support.
2.  Added initial support for the REST protocol.
image

## Checklist
- [x] Make sure there is a [GitHub_issue](https://github.com/apache/dubbo/issues) field for the change (usually before you start working on it). Trivial changes like typos do not require a GitHub issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [ ] Each commit in the pull request should have a meaningful subject line and body.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Check if is necessary to patch to Dubbo 3 if you are work on Dubbo 2.7
- [ ] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add sample in [dubbo samples](https://github.com/apache/dubbo-samples) project.
- [ ] Add some description to [dubbo-website](https://github.com/apache/dubbo-website) project if you are requesting to add a feature.
- [ ] GitHub Actions works fine on your own branch.
- [ ] If this contribution is large, please follow the [Software Donation Guide](https://github.com/apache/dubbo/wiki/Software-donation-guide).

PR #13698, opened 2024-01-29
Fix when the Accept header is "*/*" or null, adapt the Content-type to application/json

hhuang1231:

- When the Accept header is "\*/\*" or null, adapt the Content-type to application/json

### DIFF
--- a/dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/message/codec/CodecUtils.java
+++ b/dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/message/codec/CodecUtils.java
@@ -130,8 +130,8 @@ public class CodecUtils {
 
     public static String getEncodeMediaType(HttpHeaders headers) {
         String mediaType = headers.getAccept();
-        if (mediaType == null) {
-            mediaType = headers.getContentType();
+        if (mediaType == null || MediaType.ALL_VALUE.getName().equals(mediaType)) {
+            mediaType = MediaType.APPLICATION_JSON_VALUE.getName();
         }
         return mediaType;
     }
--- a/dubbo-remoting/dubbo-remoting-http12/src/test/java/org/apache/dubbo/remoting/http12/message/codec/CodeUtilsTest.java
+++ b/dubbo-remoting/dubbo-remoting-http12/src/test/java/org/apache/dubbo/remoting/http12/message/codec/CodeUtilsTest.java
@@ -18,7 +18,6 @@ package org.apache.dubbo.remoting.http12.message.codec;
 
 import org.apache.dubbo.remoting.http12.HttpHeaderNames;
 import org.apache.dubbo.remoting.http12.HttpHeaders;
-import org.apache.dubbo.remoting.http12.exception.UnsupportedMediaTypeException;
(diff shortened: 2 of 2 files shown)


namelessssssssssss (approved):
LGTM

yuluo-yx (approved):
LGTM.

CrazyHZM:
@hhuang1231 
Please fix the UT.

oxsean:
In the new REST support, the default codec has been set to json. This PR is about to be merged into 3.3. Thank you for your contribution.
https://github.com/apache/dubbo/pull/13607/
https://github.com/oxsean/dubbo/blob/feature-13427-triple-convergence-rest/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/HttpRequestHandlerMapping.java#L96

hhuang1231:
> In the new REST support, the default codec has been set to json. This PR is about to be merged into 3.3. Thank you for your contribution. #13607 https://github.com/oxsean/dubbo/blob/feature-13427-triple-convergence-rest/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/HttpRequestHandlerMapping.java#L96

@oxsean 
My PR actually changed two aspects:
- prevent the default encoder from being used when accept is \*/\*
- set the default encoder to the Json type.

And in your implement, why ignore accept and use content-type as the mediaType for decoding and encoding?

oxsean:
> > In the new REST support, the default codec has been set to json. This PR is about to be merged into 3.3. Thank you for your contribution. #13607 https://github.com/oxsean/dubbo/blob/feature-13427-triple-convergence-rest/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/HttpRequestHandlerMapping.java#L96
> 
> @oxsean My PR actually changed two aspects:
> 
> * prevent the default encoder from being used when accept is */*
> * set the default encoder to the Json type.
> 
> And in your implement, why ignore accept and use content-type as the mediaType for decoding and encoding?
The syntax of 'Accept' is quite complex, and merely judging by '/' does not properly handle it:
https://developer.mozilla.org/zh-CN/docs/Web/HTTP/Headers/Accept
Therefore, for default simple HTTP, the codec is determine based solely on the 'Content-Type'. For the new REST protocol, there is a specialized content negotiator to support Accept:
https://github.com/oxsean/dubbo/blob/feature-13427-triple-convergence-rest/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/rest/mapping/ContentNegotiator.java

hhuang1231:
@oxsean Does this mean that only the REST protocol will negotiate the response type?

oxsean:
> @oxsean Does this mean that only the REST protocol will negotiate the response type?

Yes, In the previous release version, dubbo simple http only support content-type. So I think it's sufficient for a simple protocol. If user need advanced features, the new REST can be used.

hhuang1231:
> > @oxsean Does this mean that only the REST protocol will negotiate the response type?
> 
> Yes, In the previous release version, dubbo simple http only support content-type. So I think it's sufficient for a simple protocol. If user need advanced features, the new REST can be used.

Why not add this feature to the triple protocol

PR #13777, opened 2024-02-22
feat: support read sacn packages from spring env

sheiy:


dubbo-config/dubbo-config-spring/src/main/java/org/apache/dubbo/config/spring/context/annotation/DubboComponentScanRegistrar.java
```diff
@@ -91,6 +97,18 @@ private Set<String> getPackagesToScan(AnnotationMetadata metadata) {
...
+            if (value.startsWith("${") && value.endsWith("}")) {
+                value = value.substring(2, value.length() - 1);
+                String scanPackage = environment.getProperty(value);
+                if (StringUtils.hasText(scanPackage)) {
+                    packagesToScan = Arrays.stream(scanPackage.split(","))
+                            .map(String::trim)
+                            .collect(Collectors.toSet());
+                }
+            }
+        }
```
AlbumenJ, line 111: Why should we support placeholder here?

AlbumenJ (changes requested):

sheiy:
I use yml config mapping to solve it

PR #13807, opened 2024-02-29
Fix URL issue 

namelessssssssssss:

URL#addAttributes used wrong `attributes` 

```java
//old
public URL addAttributes(Map<String, Object> attributes) {
        if (attributes != null) {
            attributes.putAll(attributes);
     return this;
}
//new
 public URL addAttributes(Map<String, Object> attributeMap) {
        if (attributeMap != null) {
            attributes.putAll(attributeMap);
        }
        return this;
}
```

## Brief changelog
* fix issues in URL#addAttributes
* code style improvement

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/URL.java
@@ -1061,7 +1061,7 @@ public /*final**/ class URL implements Serializable {
     }
 
     public URL addParameters(String... pairs) {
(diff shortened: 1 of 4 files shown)


AlbumenJ:
Are there any references of `addAttributes` in Dubbo?

namelessssssssssss:
> Are there any references of `addAttributes` in Dubbo?

Only in test cases

AlbumenJ (approved):

PR #13857, opened 2024-03-08
Netty Loom Adapt

921205zhangcheng:
reactor thread model based on virtual thread adaptation to netty

### DIFF
--- a/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyServer.java
+++ b/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyServer.java
@@ -153,13 +153,17 @@ public class NettyServer extends AbstractServer {
     }
 
     protected EventLoopGroup createBossGroup() {
-        return NettyEventLoopFactory.eventLoopGroup(1, EVENT_LOOP_BOSS_POOL_NAME);
+        return NettyEventLoopFactory.eventLoopGroup(true,1, EVENT_LOOP_BOSS_POOL_NAME);
     }
 
     protected EventLoopGroup createWorkerGroup() {
-        return NettyEventLoopFactory.eventLoopGroup(
-                getUrl().getPositiveParameter(IO_THREADS_KEY, Constants.DEFAULT_IO_THREADS),
-                EVENT_LOOP_WORKER_POOL_NAME);
+//        return NettyEventLoopFactory.eventLoopGroup(
+//                getUrl().getPositiveParameter(IO_THREADS_KEY, Constants.DEFAULT_IO_THREADS),
+//                EVENT_LOOP_WORKER_POOL_NAME);
+
+        return NettyEventLoopFactory.eventLoopGroup(true,
+            0,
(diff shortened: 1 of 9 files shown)


dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyServer.java
```diff
@@ -153,13 +153,17 @@ private boolean isSupportMetrics() {
     }
 
     protected EventLoopGroup createBossGroup() {
-        return NettyEventLoopFactory.eventLoopGroup(1, EVENT_LOOP_BOSS_POOL_NAME);
+        return NettyEventLoopFactory.eventLoopGroup(true,1, EVENT_LOOP_BOSS_POOL_NAME);
```
AlbumenJ, line 156: I am concerned about whether netty will pin the thread due to jni.

AlbumenJ (changes requested):

921205zhangcheng:
JNI is sometimes involved in low-level network operations for performance optimization purposes but it does not directly cause thread pinning. Netty's event loops are designed to handle I/O events and tasks efficiently, using non-blocking I/O and event-driven programming model to minimize thread blocking.
The "pinning" concept you might be referring to could mean whether a thread stays dedicated to specific tasks or gets stuck handling certain operations. In Netty, each event loop thread handles I/O events for its own set of channels and does not generally get blocked on these operations if they are implemented correctly as non-blocking (as most Netty components are).


However, if there is a bug or misuse of Netty where a JNI call is made in a way that causes a thread to block indefinitely, then yes, it could effectively "pin" that thread. But this would be an implementation issue rather than a characteristic of Netty itself.


The createBossGroup method returns an event loop thread group. The bossGroup is responsible for maintaining the long link between the client and server. It usually does not require too many threads because of the internal multiplexing mechanism and I have modified the thread group to be a virtual thread group Can NettyEventLoopFactory. EventLoopGroup view the virtual thread has not need to specify the initialize the number of threads when a thread pool Because virtual thread context switching is lightweight Completely controlled by the JVM initialize a thread takes less memory, I have integrated virtual threads for jdk21 into my rpc framework and have used them online





---Original---
From: "Albumen ***@***.***&gt;
Date: Mon, Mar 11, 2024 10:16 AM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [apache/dubbo] Netty Loom Adapt (PR #13857)




 
@AlbumenJ requested changes on this pull request.
 
 
In dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyServer.java:
 &gt; @@ -153,13 +153,17 @@ private boolean isSupportMetrics() {      }        protected EventLoopGroup createBossGroup() { -        return NettyEventLoopFactory.eventLoopGroup(1, EVENT_LOOP_BOSS_POOL_NAME); +        return NettyEventLoopFactory.eventLoopGroup(true,1, EVENT_LOOP_BOSS_POOL_NAME);  
I am concerned about whether netty will pin the thread due to jni.

PR #13947, opened 2024-03-19
Load the first zone detector

chickenlj:
Check all ZoneDetectors and pick the first one if `default` is not configured.


zrlw:
What are the benefits?

PR #13971, opened 2024-03-21
Fix the problem of multiple namespaces in MetadataReport

finefuture:

Key missing namespace information:
image

## Checklist


AlbumenJ (commented):
@chickenlj PTAL

dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/report/support/AbstractMetadataReportFactory.java
```diff
@@ -89,6 +91,12 @@ public MetadataReport getMetadataReport(URL url) {
     }
 
     protected String toMetadataReportKey(URL url) {
+        String namespace = url.getParameter(NAMESPACE_KEY);
+        if (!StringUtils.isEmpty(namespace)) {
+            return URL.valueOf(url.toServiceString())
+                    .addParameter(NAMESPACE_KEY, namespace)
+                    .toString();
+        }
```
chickenlj: Can you describe the difference between the new url and the old url?

chickenlj: How about making it a public util method to make calling it in other places easier? like here https://github.com/apache/dubbo/pull/13971/files#diff-17a0f234d88728ebc6f37959d84276704ca16846cd10cd0743616a8689949eabR33

chickenlj:
I think it would be better to just change the way how `REGISTRY_CLUSTER_KEY` is generated.

For now, `REGISTRY_CLUSTER_KEY` is the same as `id` (controlled by `@Parameter(key = REGISTRY_CLUSTER_KEY)`). If we want it to include `namespace` in `REGISTRY_CLUSTER_KEY`, we can change the code to make sure that `namespace` is included in `REGISTRY_CLUSTER_KEY` the first time the `registry url` is generated. 

Then the only place we need to change except for the `REGISTRY_CLUSTER_KEY` generating process (to include both id and namespace) is how to map the MetadataReport instance with the  `REGISTRY_CLUSTER_KEY`(also to have `namespace` included) https://github.com/apache/dubbo/pull/13971/files#diff-1e44977c9c4b360c444eb20fbc5599acc73533c783c897954e068e88ab339278R117.

finefuture:
> I think it would be better to just change the way how `REGISTRY_CLUSTER_KEY` is generated.
> 
> For now, `REGISTRY_CLUSTER_KEY` is the same as `id` (controlled by `@Parameter(key = REGISTRY_CLUSTER_KEY)`). If we want it to include `namespace` in `REGISTRY_CLUSTER_KEY`, we can change the code to make sure that `namespace` is included in `REGISTRY_CLUSTER_KEY` the first time the `registry url` is generated.
> 
> Then the only place we need to change except for the `REGISTRY_CLUSTER_KEY` generating process (to include both id and namespace) is how to map the MetadataReport instance with the `REGISTRY_CLUSTER_KEY`(also to have `namespace` included) https://github.com/apache/dubbo/pull/13971/files#diff-1e44977c9c4b360c444eb20fbc5599acc73533c783c897954e068e88ab339278R117.

Ok, I will try it.

finefuture:
@chickenlj PTAL

dubbo-common/src/main/java/org/apache/dubbo/config/RegistryConfig.java
```diff
@@ -220,12 +219,6 @@ public RegistryConfig(ApplicationModel applicationModel, String address, String
         setProtocol(protocol);
     }
 
-    @Override
-    @Parameter(key = REGISTRY_CLUSTER_KEY)
-    public String getId() {
-        return super.getId();
-    }
-
```
chickenlj: Please keep this method and only remove `@Parameter(key = REGISTRY_CLUSTER_KEY)`.

chickenlj (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/config/RegistryConfig.java
finefuture: Fixed.

dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/report/support/AbstractMetadataReportFactory.java
finefuture: Add a new method for extra parameters.
image

finefuture:
@chickenlj Does DynamicConfiguration need to support multiple namespaces?
image

chickenlj:
Sorry for the late response, I am working on this pull request now.

chickenlj (approved):
LGTM.

chickenlj:
@finefuture thanks for the patience and contribution, we do appreciate it.

AlbumenJ (approved):

PR #14057, opened 2024-04-10
Sync BitList for all the operations

AlbumenJ:
## Checklist

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
@@ -93,23 +93,23 @@ public class BitList<E> extends AbstractList<E> implements Cloneable {
     }
 
     // Provided by BitList only
-    public List<E> getOriginList() {
+    public synchronized List<E> getOriginList() {
         return originList;
     }
 
-    public void addIndex(int index) {
+    public synchronized void addIndex(int index) {
         this.rootSet.set(index);
     }
 
-    public int totalSetSize() {
+    public synchronized int totalSetSize() {
         return this.originList.size();
     }
 
-    public boolean indexExist(int index) {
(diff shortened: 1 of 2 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
```diff
@@ -93,23 +93,23 @@ public BitList(List<E> originList, BitSet rootSet, List<E> tailList) {
     }
 
     // Provided by BitList only
-    public List<E> getOriginList() {
+    public synchronized List<E> getOriginList() {
```
CrazyHZM, line 96: Read behavior doesn't need to lock.

CrazyHZM (changes requested):

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
He-Pin: Better to use a read and write lock?

oxsean: In general, reading without locking may lead to consistency issues.
@He-Pin If the operation itself is very short, using a read-write lock won't increase throughput. In fact, due to the internal complexity of the lock, it might even slow down. I've tested this with examples before.

chickenlj (approved):
LGTM.

Be aware of the potential performance drop caused by lock.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
He-Pin: Great, at least at this case, this is not needed:)

PR #14077, opened 2024-04-12
Add TripleConfig to ProtocolConfig as nest configuration

finefuture:
## Checklist


oxsean:
@AlbumenJ Could you take a look add TripleConfig into configManager is reasonable?, or is there a better option?

AlbumenJ:
Add a special protocol related config to `ConfigManager` is not a good idea. We need a more reasonable way to manage these special configs.

finefuture:
> Add a special protocol related config to `ConfigManager` is not a good idea. We need a more reasonable way to manage these special configs.

How about putting it in ProtocolConfig as nest configuration?

AlbumenJ:
> > Add a special protocol related config to `ConfigManager` is not a good idea. We need a more reasonable way to manage these special configs.
> 
> How about putting it in ProtocolConfig as nest configuration?

1. Still a little wired
2. This will change the full key from `dubbo.triple.xxx` to `dubbo.protocol.triple.xxx`

finefuture:
> > > Add a special protocol related config to `ConfigManager` is not a good idea. We need a more reasonable way to manage these special configs.
> > 
> > 
> > How about putting it in ProtocolConfig as nest configuration?
> 
> 1. Still a little wired
> 2. This will change the full key from `dubbo.triple.xxx` to `dubbo.protocol.triple.xxx`

Do you have any ideas? I think the second point is acceptable.

AlbumenJ:
> > > > Add a special protocol related config to `ConfigManager` is not a good idea. We need a more reasonable way to manage these special configs.
> > > 
> > > 
> > > How about putting it in ProtocolConfig as nest configuration?
> > 
> > 
> > 
> > 1. Still a little wired
> > 2. This will change the full key from `dubbo.triple.xxx` to `dubbo.protocol.triple.xxx`
> 
> Do you have any ideas? I think the second point is acceptable.

If we change the full key from `dubbo.triple.xxx` to `dubbo.protocol.triple.xxx`, all the users who have configured such configuration should change their configuration.

finefuture:
> > > > > Add a special protocol related config to `ConfigManager` is not a good idea. We need a more reasonable way to manage these special configs.
> > > > 
> > > > 
> > > > How about putting it in ProtocolConfig as nest configuration?
> > > 
> > > 
> > > 
> > > 1. Still a little wired
> > > 2. This will change the full key from `dubbo.triple.xxx` to `dubbo.protocol.triple.xxx`
> > 
> > 
> > Do you have any ideas? I think the second point is acceptable.
> 
> If we change the full key from `dubbo.triple.xxx` to `dubbo.protocol.triple.xxx`, all the users who have configured such configuration should change their configuration.

TripleConfig is newly added in version 3.3. Users should just add new configurations.

In version 3.2:
image

AlbumenJ:
> > > > > > Add a special protocol related config to `ConfigManager` is not a good idea. We need a more reasonable way to manage these special configs.
> > > > > 
> > > > > 
> > > > > How about putting it in ProtocolConfig as nest configuration?
> > > > 
> > > > 
> > > > 
> > > > 1. Still a little wired
> > > > 2. This will change the full key from `dubbo.triple.xxx` to `dubbo.protocol.triple.xxx`
> > > 
> > > 
> > > Do you have any ideas? I think the second point is acceptable.
> > 
> > 
> > If we change the full key from `dubbo.triple.xxx` to `dubbo.protocol.triple.xxx`, all the users who have configured such configuration should change their configuration.
> 
> TripleConfig is newly added in version 3.3. Users should just add new configurations.
> 
> In version 3.2: image

If so, we can change it

AlbumenJ (approved):

AlbumenJ:
@EarthChen @oxsean @icodening PTAL

dubbo-common/src/main/java/org/apache/dubbo/config/ProtocolConfig.java
```diff
@@ -236,6 +238,9 @@ public class ProtocolConfig extends AbstractConfig {
      */
     private String jsonCheckLevel;
 
+    @Nested
+    private TripleConfig triple;
```
oxsean, line 242: triple = new TripleConfig();
Initial value check can be simplified

dubbo-common/src/main/java/org/apache/dubbo/config/nested/TripleConfig.java
```diff
@@ -0,0 +1,254 @@
...
+     */
+    private Integer maxHeaderListSize;
+
+    /**
+     * Whether to pass through standard HTTP headers, default is false.
+     */
+    private Boolean passThroughStandardHttpHeaders;
+
+    public TripleConfig() {
+        init();
```
oxsean: The if statement will not be effective. Can we define a default value constant and initialize it directly in the field?

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/builders/TripleBuilder.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    protected TripleBuilder getThis() {
+        return this;
+    }
+
+    public TripleConfig build() {
+        TripleConfig triple = new TripleConfig();
+
+        triple.setMaxBodySize(maxBodySize);
```
oxsean: Need if condition,  otherwise, it will be overwritten as null

dubbo-common/src/main/java/org/apache/dubbo/config/nested/TripleConfig.java
```diff
@@ -0,0 +1,254 @@
...
+
+    /**
+     * Maximum header list size.
+     */
+    private Integer maxHeaderListSize;
+
+    /**
+     * Whether to pass through standard HTTP headers, default is false.
+     */
+    private Boolean passThroughStandardHttpHeaders;
```
oxsean: Unused.

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp2Protocol.java
```diff
@@ -204,21 +188,27 @@ private void configurerHttp2Handlers(URL url, List<ChannelHandler> handlers) {
...
+                        .maxHeaderListSize(tripleConfig.getMaxHeaderListSize()))
                 .frameLogger(SERVER_LOGGER)
                 .build();
     }
+
+    private TripleConfig getTripleConfig(URL url) {
+        return url.getOrDefaultApplicationModel()
+                .getApplicationConfigManager()
+                .getProtocol(url.getProtocol())
+                .flatMap(protocolConfig -> Optional.of(protocolConfig.getTriple()))
```
oxsean: If we use `triple = new TripleConfig()` in , it can be simplified to:
```
.getProtocol(url.getProtocol())
.getTriple()
```

oxsean (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/config/ProtocolConfig.java
finefuture: Okay, I'll fix it.

dubbo-common/src/main/java/org/apache/dubbo/config/nested/TripleConfig.java
finefuture: Okay, I'll fix it.

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/bootstrap/builders/TripleBuilder.java
finefuture: Okay, I'll fix it.

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp2Protocol.java
finefuture: The getProtocol method will return an Optional and the getTriple method cannot be called directly.

dubbo-common/src/main/java/org/apache/dubbo/config/nested/TripleConfig.java
finefuture: Removed.

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp2Protocol.java
finefuture: Updated:
```java
 .getProtocol(url.getProtocol())
 .orElseThrow(() -> new IllegalStateException("There's no ProtocolConfig specified."))
 .getTriple();
```

dubbo-config/dubbo-config-spring/src/main/resources/META-INF/dubbo.xsd
```diff
@@ -1134,6 +1134,86 @@
...
+            </xsd:annotation>
+        </xsd:attribute>
+
+        <xsd:attribute name="max-header-list-size" type="xsd:integer" default="32768">
+            <xsd:annotation>
+                <xsd:documentation><![CDATA[ Maximum header list size. ]]></xsd:documentation>
+            </xsd:annotation>
+        </xsd:attribute>
+
+        <xsd:attribute name="pass-through-standard-http-headers" type="xsd:boolean">
```
oxsean: Should remove

oxsean (approved):
@finefuture xsd should remove too.
@AlbumenJ LGTM.

finefuture:
There is a problem here. If the server-side and client-side protocols are inconsistent, configClientPipeline obtains the protocol name of the server side through url.getProtocol. At this time, ProtocolConfig will not be obtained from ConfigManager.

I'm fixing it.

dubbo-config/dubbo-config-spring/src/main/resources/META-INF/dubbo.xsd
finefuture: Removed.

finefuture:
> There is a problem here. If the server-side and client-side protocols are inconsistent, configClientPipeline obtains the protocol name of the server side through url.getProtocol. At this time, ProtocolConfig will not be obtained from ConfigManager.
> 
> I'm fixing it.

This problem can be solved by using the application's own ProtocolConfig. If the user wants to configure triple, he can set it directly under a single protocol. When using multiple protocols, he can set the protocol for which TripleConfig is configured as the default protocol.
And can handle the situation where ProtocolConfig is not set.

```java
return url.getOrDefaultApplicationModel().getApplicationConfigManager().getDefaultProtocols().stream()
        .findFirst()
        .flatMap(protocolConfig -> Optional.ofNullable(protocolConfig.getTriple()))
        .orElseGet(TripleConfig::new);
```

dubbo-common/src/main/java/org/apache/dubbo/config/ProtocolConfig.java
finefuture: image

image

@oxsean Maybe I should just put it in checkDefault and follow the old rules, do you think so?

AlbumenJ:
@EarthChen @icodening PTAL

dubbo-common/src/main/java/org/apache/dubbo/config/ProtocolConfig.java
oxsean: @finefuture make sense,  initialize the default values here can avoid the null check during in builder.
We can add a checkDefault in TripleConfig add call it in ProtocolConfig

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp2Protocol.java
```diff
@@ -204,10 +205,9 @@ private Http2FrameCodec buildHttp2FrameCodec(URL url) {
     }
 
     private TripleConfig getTripleConfig(URL url) {
-        return url.getOrDefaultApplicationModel()
-                .getApplicationConfigManager()
-                .getProtocol(url.getProtocol())
-                .orElseThrow(() -> new IllegalStateException("There's no ProtocolConfig specified."))
-                .getTriple();
+        return url.getOrDefaultApplicationModel().getApplicationConfigManager().getDefaultProtocols().stream()
```
oxsean: getApplicationConfigManager().getProtocol(url.getProtocol())?
Using findFirst is not reasonable. What if the first one is the Dubbo protocol? Also, if nothing is found, it should throw an exception; otherwise, how would the initial values be initialized?

dubbo-common/src/main/java/org/apache/dubbo/config/ProtocolConfig.java
finefuture: Okay，I will add it.

icodening:
use 'extends' maybe better ?
@AlbumenJ @oxsean

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp2Protocol.java
finefuture: If this application is a pure consumer, no ProtocolConfig will be configured, but if it needs to call the downstream triple service, configClientPipeline requires a TripleConfig.
Similarly, if the upstream is configured with the dubbo protocol, but want to call the downstream triple service, can only configure triple in the dubbo protocol.

finefuture: Updated:
```java
    private TripleConfig getTripleConfig(URL url) {
        return url.getOrDefaultApplicationModel()
                .getApplicationConfigManager()
                .getOrAddProtocol(url.getProtocol())
                .getTriple();
    }
```

finefuture:
> use 'extends' maybe better ? @AlbumenJ @oxsean

Embedded configuration may be better because it can be unified under the dubbo.protocol prefix. You can refer to MetricsConfig, and Prometheus configuration is also used as embedded configuration. If users do not want to use Prometheus, they can choose not to configure it.

dubbo-common/src/main/java/org/apache/dubbo/config/context/ConfigManager.java
```diff
@@ -197,6 +197,19 @@ public Optional<ProtocolConfig> getProtocol(String idOrName) {
         return getConfig(ProtocolConfig.class, idOrName);
     }
 
+    public ProtocolConfig getOrAddProtocol(String idOrName, boolean needRefresh) {
```
oxsean: When will needRefresh=false be used? I think it should always be refresh.

finefuture: Provides one more option, but it seems that no other place will call it at the moment, I will remove it.

finefuture: Already removed.

icodening (approved):
LGTM

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp2Protocol.java
```diff
@@ -204,21 +187,26 @@ private void configurerHttp2Handlers(URL url, List<ChannelHandler> handlers) {
...
+                        .maxFrameSize(tripleConfig.getMaxFrameSize())
+                        .maxHeaderListSize(tripleConfig.getMaxHeaderListSize()))
                 .frameLogger(SERVER_LOGGER)
                 .build();
     }
+
+    private TripleConfig getTripleConfig(URL url) {
+        return url.getOrDefaultApplicationModel()
+                .getApplicationConfigManager()
+                .getOrAddProtocol(url.getProtocol())
```
AlbumenJ, line 207: Why use `getOrAddProtocol`?

AlbumenJ (changes requested):

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp2Protocol.java
finefuture: Because the consumer side may not be configured with tri's ProtocolConfig, it needs to set a default tri's ProtocolConfig for it. 
For example: https://github.com/apache/dubbo/blob/3.2/dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/TripleProtocolTest.java

oxsean:
LGTM.

EarthChen (approved):
LGTM

Chenjp:
Feature request: Support for different level of "max-body-size":
1. Application;
2. Protocol;
3. Service;

Priority : Service>Protocol>Application.

Example:
1. Nginx:
```
Syntax: client_max_body_size size;
Context: http, server, location
```
2. Another Example: [Kong request-size-limiting](https://docs.konghq.com/hub/kong-inc/request-size-limiting/configuration/)

PR #14096, opened 2024-04-17
Compatible nacos service discovery, export noting suffix servicename

wangchengming666:
by default, dubbo export servicename is "${interface}:${version}:", this servicename with ':' suffix
for compatible nacos service discovery, export noting suffix servicename, eg: {interface}:${version}

### DIFF
--- a/dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
+++ b/dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
@@ -35,6 +35,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
```diff
@@ -697,14 +737,21 @@ private NacosServiceName createServiceName(URL url) {
...
+            return getServiceName0(url, url.getCategory(DEFAULT_CATEGORY));
+        }
         return getServiceName(url, url.getCategory(DEFAULT_CATEGORY));
     }
 
     private String getServiceName(URL url, String category) {
         return category + SERVICE_NAME_SEPARATOR + url.getColonSeparatedKey();
     }
 
+    private String getServiceName0(URL url, String category) {
```
chickenlj: I think `getCompatibleServiceName` would be easier to understand than `getServiceName0`. Please also use the `CompatibleXxx` naming pattern for other methods.

dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
```diff
@@ -174,18 +182,33 @@ public List<URL> lookup(final URL url) {
...
-                String serviceName = getServiceName(url);
                 Instance instance = createInstance(url);
+
+                Set<String> serviceNames = new HashSet<>();
+                // by default servicename is "org.apache.dubbo.xxService:1.0.0:"
+                String serviceName = getServiceName(url, false);
+                serviceNames.add(serviceName);
+
+                // in https://github.com/apache/dubbo/issues/14075
+                if (Boolean.parseBoolean(DUUBO_NACOS_EXPORT_NOTING_SUFFIX_SERVICENAME)) {
```
chickenlj: Using `getUrl().getParameter("nacos.register-compatible", false)` would be better. That way, users can configure in a more standard way supported by Dubbo Config component:

```yaml
dubbo:
   registry:
      parameters:
         nacos.register-compatible: true
```

or

```shell
-Ddubbo.registry.parameters.nacos.register-compatible=true
```

or

```properties
dubbo.registry.parameters.nacos.register-compatible=true
```

chickenlj (changes requested):

dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
chickenlj: Please add some unit tests to verify this change. Thanks.

chickenlj (approved):
LGTM.

PR #14137, opened 2024-04-28
Remove spring related transitive dependencies from dubbo-all

chickenlj:
1. spring xml related samples need to be adjusted to meet this change
2. documentation needs to be updated for non spring-boot users
3. spring-boot users would not be affected, but with dependency conflict chances reduced


AlbumenJ:
1. Test cases should be fixed first
2. Why add protobuf dependencies?

chickenlj:
> 1. Test cases should be fixed first

I am now working on it.

> 2. Why add protobuf dependencies?

As triple going to become the major protocol of Dubbo framework in the next release, I think we should have it by default, just like what we do with hessian2 before.

CrazyHZM (commented):
I agree to remove spring dependencies from dubbo-all. 
Dubbo is used in three ways:
1. dubbo API （without spring）
2. dubbo + dubbo-config-spring (without SpringBoot)
3. dubbo + SpringBoot (dubbo-spring-boot-starter)

wcy666103:
It's ready to merge

dubbo-distribution/dubbo-all/pom.xml
```diff
@@ -463,6 +447,18 @@
...
       <artifactId>fastjson2</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java-util</artifactId>
+    </dependency>
```
AlbumenJ: We may need some discussion about we add these dependencies or not

chickenlj: Sure, remember to bring it up on the next weekly meeting.

PR #9552, opened 2022-01-09
feature: support jsr-303 jakarta namespace

juzi214032:
## What is the purpose of the change
Java EE has been renamed from javax to jakarta, Related jar packages no longer use the package name starting with javax, including the jsr-303 api. To support Hibernate Validator 7.x, a new JValidatorNew file has been added for compatibility with version 7.x

close #9544
## Brief changelog
- Add JValidatorNew, use new api.
- ValidationFilter masks implementation class exceptions

## Verifying this change
I can't write new unit tests because Hibernate Validator 7.x and 6.x can't exist under the classpath at the same time and there would be a class conflict. But I have tested 7.x separately in my development environment and it works fine, so I hope someone else can test it again to make sure it works!

## Checklist
- [x] Make sure there is a [GitHub_issue](https://github.com/apache/dubbo/issues) field for the change (usually before you start working on it). Trivial changes like typos do not require a GitHub issue. Your pull request should address just this issue, without pulling in other changes - one PR resolves one issue.
- [x] Each commit in the pull request should have a meaningful subject line and body.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Check if is necessary to patch to Dubbo 3 if you are work on Dubbo 2.7
- [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exist. If the new feature or significant change is committed, please remember to add sample in [dubbo samples](https://github.com/apache/dubbo-samples) project.
- [x] Add some description to [dubbo-website](https://github.com/apache/dubbo-website) project if you are requesting to add a feature.
- [x] GitHub Actions works fine on your own branch.
- [x] If this contribution is large, please follow the [Software Donation Guide](https://github.com/apache/dubbo/wiki/Software-donation-guide).

PR #14194, opened 2024-05-16
feat: Validator supports automatic switch

aofall:

1. In the newer versions of Hibernate-validator, the package names of the relevant validation implementation classes have been migrated from `javax` to `jakarta`. 
Dubbo has been support new validator with jakarta in #9552 , but the setting of default validator is implement with javax, if the user is not configured to use a new validator under the SpringBoot3, an `NoClassDefFoundError` will be throw.
If the validator supports automatic switch, this will reduce a lot of configuration processes for users. So Add an adapter that it can automatically switch.

2. If user fill in any characters into `dubbo.provider/consumer.validation` configuration that the validation extension not exists, the validation will active the default validator. 
But in the [official documentation](https://cn.dubbo.apache.org/zh-cn/overview/mannual/java-sdk/reference-manual/config/properties/) the parameter type of `dubbo.provider.validation` and `dubbo.consumer.validation` is described as Boolean, it's unexpect that user fill in the `false` parameter. 
Support disabling validation when parameter is 'false' that it is expected the description in document the parameter type is boolean

## Brief changelog

1. Add an adapter for javax and jakarta that it can automatically switch

2. Support disabling validation when parameter is 'false' that it is expected the description in document the parameter type is boolean

## Checklist

### DIFF
--- a/dubbo-plugin/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/Validation.java
+++ b/dubbo-plugin/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/Validation.java
@@ -25,7 +25,7 @@ import static org.apache.dubbo.common.constants.FilterConstants.VALIDATION_KEY;
 /**
  * Instance of Validation interface provide instance of {@link Validator} based on the value of <b>validation</b> attribute.
  */
-@SPI("jvalidation")
+@SPI("adapter")
 public interface Validation {
 
     /**
--- a/dubbo-plugin/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/Validation.java
+++ b/dubbo-plugin/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/Validation.java
@@ -25,7 +25,7 @@ import static org.apache.dubbo.common.constants.FilterConstants.VALIDATION_KEY;
 /**
  * Instance of Validation interface provide instance of {@link Validator} based on the value of <b>validation</b> attribute.
  */
-@SPI("adapter")
+@SPI("jvalidation")
 public interface Validation {
 
     /**
--- a/dubbo-plugin/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/Validator.java
+++ b/dubbo-plugin/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/Validator.java
@@ -23,4 +23,6 @@ package org.apache.dubbo.validation;
 public interface Validator {
 
     void validate(String methodName, Class<?>[] parameterTypes, Object[] arguments) throws Exception;
+
+    boolean isSupport();
 }
--- a/dubbo-plugin/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/filter/ValidationFilter.java
+++ b/dubbo-plugin/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/filter/ValidationFilter.java
@@ -105,6 +105,7 @@ public class ValidationFilter implements Filter {
     private boolean needValidate(URL url, String methodName) {
         return validation != null
                 && !methodName.startsWith("$")
-                && ConfigUtils.isNotEmpty(url.getMethodParameter(methodName, VALIDATION_KEY));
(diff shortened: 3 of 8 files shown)


dubbo-plugin/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/Validation.java
```diff
@@ -25,7 +25,7 @@
 /**
  * Instance of Validation interface provide instance of {@link Validator} based on the value of <b>validation</b> attribute.
  */
-@SPI("jvalidation")
```
AlbumenJ: We'd better not to change the default name

AlbumenJ (changes requested):

dubbo-plugin/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/Validation.java
AlbumenJ: Also, we can use `jvalidation-javax` and `jvalidation-jaraka` for the real implementations.

aofall: > Also, we can use `jvalidation-javax` and `jvalidation-jaraka` for the real implementations.

OK, I will rename the `jvalidation` into `jvalidation-javax` and `jvalidationNew` into `jvalidation-jakarta`

> We'd better not to change the default name

Change the default extension is to use the adapter by default.

If not change the default extension for the adapter, while the parameter of `dubbo.provider/consumer.validation` fill in not exist extension name like `true` or other anything, the default extension `jvalidation-javax` will try to loaded and active. 

Unless the config parameter of `dubbo.provider/consumer.validation` is `adapter` or the newly adapter will not active if the parameter not `adapter`.

If change the default name is unexpected, also can merge the adapter logic into `jvalidation-javax`, but it could't use `@Active#onClass` on `jvalidation-javax` extension or it will not loaded while the env not exist `javax`, and adapter logic will also not active.

AlbumenJ: > If not change the default extension for the adapter, while the parameter of dubbo.provider/consumer.validation fill in not exist extension name like true or other anything, the default extension jvalidation-javax will try to loaded and active.

Still wondering why

aofall: Because of the `@Adaptive(VALIDATION_KEY)` in `org.apache.dubbo.validation.Validation`

https://github.com/apache/dubbo/blob/9968834e621e3cff492fed91c142aa176f1d883f/dubbo-filter/dubbo-filter-validation/src/main/java/org/apache/dubbo/validation/Validation.java#L36

Here is Some description of `@Adaptive(value)`

> Decide which target extension to be injected. The name of the target extension is decided by the parameter passed in the URL, and the parameter names are given by this method.
> 
> If the specified parameters are not found, then the default extension will be used for dependency injection (specified in its interface's @SPI(value) ).

Fill in the parameter of `dubbo.provider/consumer.validation` will load the extension what's the parameter named in URL. But `true` or other else charater (exclude jvalidation and jvalidationNew) is not an exist extension. So it will load the default extension.

And the `@SPI("value")` in `org.apache.dubbo.validation.Validation` was already set the `jvalidation` (now renamed to `jvalidation-javax`) as default.

AlbumenJ: 1. Configure`dubbo.provider/consumer.validation` with `true`
2. Dubbo use default extension
3. Default extension is `jvalidation` and its implementation is `org.apache.dubbo.validation.support.jvalidation.AdapterValidation`

Are there any error?

aofall: After the  commit([revert the default extension name](https://github.com/apache/dubbo/pull/14194/commits/d22e58eba602c4dbface26443d464ca95854c3ab)) it's right

AlbumenJ (approved):

PR #14210, opened 2024-05-20
New metadata service

namelessssssssssss:
## Purpose of the Change
This PR introduces a new version of `MetadataService` (V2) based on Protobuf, which provides better compatibility with dubbo-go.

See https://github.com/apache/dubbo/issues/13584

## Brief Changelog
To use `MetadataServiceV2`, the application must meet the following requirements:
* The application has imported Protobuf dependencies.
* `metadataServiceProtocol` is configured as 'tri', or at least one service is using the triple protocol.

Currently, `MetadataServiceV2` will automatically export along with `MetadataService` (V1, on the same port) if the above requirements are **satisfied**. If `onlyUseMetadataV2` is set to `true`, only `MetadataServiceV2` will be exported. If the above requirements are **not satisfied**, Dubbo will still only export `MetadataService` (V1).


chickenlj:
Well done. Generally looks good to me.

FoghostCn (commented):
Can we implement the tri protocol in a non-idl way, so that our interface definition can be reused, and we only need to set the protocol to tri when using it. I don't know if this is feasible.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/ServiceInstanceMetadataUtils.java
```diff
@@ -96,6 +96,8 @@ public class ServiceInstanceMetadataUtils {
      */
     public static final String METADATA_STORAGE_TYPE_PROPERTY_NAME = "dubbo.metadata.storage-type";
 
+    public static final String METADATA_SERVICE_VERSION_NAME = "dubbo.metadata.service-version";
```
chickenlj: Use as short a name as possible, for example, `meta-v`.

chickenlj (approved):
LGTM.

dubbo-metadata/dubbo-metadata-api/src/main/proto/metadata_service_v2.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
```
AlbumenJ, line 18: License header is required

AlbumenJ (changes requested):

dubbo-metadata/dubbo-metadata-api/src/main/proto/metadata_service_v2.proto
namelessssssssssss: OK. License are added now

PR #14223, opened 2024-05-20
[3.3] Fix triple rest http GET failed

oxsean:
HTTP/1 upgrade caused GET request to failed
Case:
https://github.com/apache/dubbo-samples/blob/master/2-advanced/dubbo-samples-triple-rest/dubbo-samples-triple-rest-springmvc/src/test/java/org/apache/dubbo/rest/demo/test/ConsumerIT.java#L47

```
curl.exe -v --http2 http://127.0.0.1:50052/demo/hello?name=world
```
Expect 'Hello world' but  respone empty body.
## Root cause
[HttpServerUpgradeHandler.java#L377](https://github.com/netty/netty/blob/netty-4.1.109.Final/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java#L377) already support write response, HttpServerAfterUpgradeHandler duplicate writes cause incorrect GET results.

## Brief changelog
Remove unnecessary HttpServerAfterUpgradeHandler


walklown:
I think you are right. I seem to be confused by another bug. Anyway 'HttpServerAfterUpgradeHandler' can be removed now, I will keep an eye on it.

AlbumenJ (approved):

PR #14317, opened 2024-06-14
Develop adaptive load balancing algorithm functions

Robin1987China:
The loadBalance algorithm is the core of the service governance framework, and dubbo defaults to the roundrobin algorithm, which is also the preferred and most used algorithm for the business system that currently uses dubbo, but each algorithm has its scope of application or limitations. It just doesn't take into account the error situation of a node in the provode terminal node and the corresponding performance problems; The SelfAdaptionLoadBaflance adaptive load balancing algorithm calculates the error rate and energy consumption of each client call to provide within a specific time window, and dynamically adjusts the call weight of the provide node to improve performance and reduce the error rate.

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/SelfAdaptionLoadBalance.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/SelfAdaptionLoadBalance.java
@@ -1,0 +1,157 @@
+package org.apache.dubbo.rpc.cluster.loadbalance;
+
+import org.apache.dubbo.common.URL;
+import org.apache.dubbo.common.utils.NamedThreadFactory;
+import org.apache.dubbo.rpc.Invocation;
+import org.apache.dubbo.rpc.Invoker;
+import org.apache.dubbo.rpc.RpcStatus;
+
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * @Author: Lee
+ */
+public class SelfAdaptionLoadBalance extends  AbstractLoadBalance{
+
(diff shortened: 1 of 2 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/SelfAdaptionLoadBalance.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * @Author: Lee
```
aofall, line 18: remove the author tag

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/SelfAdaptionLoadBalance.java
```diff
@@ -0,0 +1,157 @@
+package org.apache.dubbo.rpc.cluster.loadbalance;
```
aofall, line 1: Add the ASF License at the file start

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/SelfAdaptionLoadBalance.java
```diff
@@ -0,0 +1,157 @@
...
+        long leastFailed = Long.MAX_VALUE;
+
+        int leastFaiedCount = 0;
+
+        /**
+         *  the index of invokers having the same estimated shortest response time
+         */
+        int[] leastFailedIndexes = new int[length];
+        /**
+         * 每个服务提供者的权重
```
aofall, line 52: translate it into english, the same goes for other content

aofall (changes requested):
You can submit this Pull Request to the [apache/dubbo-spi-extensions](https://github.com/apache/dubbo-spi-extensions) repository

zrlw:
spotless:apply

Issue #14388, opened 2024-07-02
[Bug] NPE in RegistryProtocol.ExporterChangeableWrapper#unregister when shutting down application 


binfeiruci:
### Pre-check

- [X] I am sure that all the content I provide is in English.

### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo Java 3.2.5, OpenJDK 1.8

### Steps to reproduce this issue

Got the warn log below sometimes when shutting down our Dubbo-Spring integrated application.
```
 [DUBBO] null, dubbo version: 3.2.6-HA.1.0.5-RELEASE, current host: 10.232.232.201, error code: 99-0. This may be caused by unknown error in registry module, go to https://dubbo.apache.org/faq/99/0 to find instructions. 
java.lang.NullPointerException
  org.apache.dubbo.registry.integration.RegistryProtocol$ExporterChangeableWrapper.unregister(RegistryProtocol.java:1052)
  org.apache.dubbo.registry.integration.RegistryProtocol$DestroyableExporter.unregister(RegistryProtocol.java:760)
  org.apache.dubbo.config.ServiceConfig.unexport(ServiceConfig.java:208)
  ...
```
The corresponding thread name is SpringApplicationShutdownHook.

### What you expected to happen

DubboShutdownHook waited DubboDeployApplicationListener(within SpringApplicationShutdownHook thread) to destroy the applicationMode for 10 seconds(default), then it will continue executing.
If things happened in the order below, we will get this warn log.

1. RegistryProtocol.ExporterChangeableWrapper#unregister loop applicationModel.getPubModuleModels() in thread SpringApplicationShutdownHook
2. DubboShutdownHook destroy a moduleModel, set the moduleModel's serviceRepository to null.
3. RegistryProtocol.ExporterChangeableWrapper#unregister will get NPE when calling `moduleModel.getServiceRepository().getExportedServices()` 

### Are you willing to submit a pull request to fix on your own?

- [X] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wcy666103:
I think it is similar to this issue https://github.com/apache/dubbo/issues/10150#issuecomment-2201586582 , can you help us solve this problem through pr?

[...]

PR #14389, opened 2024-07-02
fix NPE in RegistryProtocol.ExporterChangeableWrapper#unregister when shutting down application

binfeiruci:
fix NPE in unregister

fix NPE

## Brief changelog
check whether moduleModel.getServiceRepository() is null

#14388


wcy666103:
use `mvn  spotless:apply` to format code firstly

PR #14395, opened 2024-07-05
Support asterisk

qingkaiyan:
I want consumer with * version to router diffrent version provider
there is two problem
1. nacos dataId not support * SYMBOL
2. consumer with * version invoke Null version provider will block by PermittedSerializationKeeper.checkSerializationPermitted.
Because the provider urlParam of version built by consumer has *.
Because the null value will not replace the exist urlParam.


wcy666103:
I think there are two issues that need to be discussed:
1. Don't commit any more feature changes to the `3.2 branch`
2.  \* does the semantics mean ` version! =null`

qingkaiyan:
it supported at 2.7

wcy666103:
use `mvn  spotless:apply` to format code

PR #14409, opened 2024-07-09
3.3

qingkaiyan:
I want consumer with * version to router diffrent version provider which dubbo 2.7 supported
there is two problem

1.nacos dataId not support * SYMBOL
2.consumer with * version invoke Null version provider will block by PermittedSerializationKeeper.checkSerializationPermitted
Because the provider urlParam of version built by consumer has *
Because the null value will not replace the exist urlParam

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/url/component/URLParam.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/url/component/URLParam.java
@@ -483,7 +483,7 @@ public class URLParam {
      * @return A new URLParam
      */
     public URLParam addParameter(String key, String value) {
-        if (StringUtils.isEmpty(key) || StringUtils.isEmpty(value)) {
+        if (StringUtils.isEmpty(key)) {
             return this;
         }
         return addParameters(Collections.singletonMap(key, value));
@@ -497,7 +497,7 @@ public class URLParam {
      * @return A new URLParam
      */
     public URLParam addParameterIfAbsent(String key, String value) {
-        if (StringUtils.isEmpty(key) || StringUtils.isEmpty(value)) {
+        if (StringUtils.isEmpty(key)) {
             return this;
         }
         if (hasParameter(key)) {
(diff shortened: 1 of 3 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/url/component/URLParam.java
```diff
@@ -483,7 +483,7 @@ public String getAnyMethodParameter(String key) {
      * @return A new URLParam
      */
     public URLParam addParameter(String key, String value) {
-        if (StringUtils.isEmpty(key) || StringUtils.isEmpty(value)) {
+        if (StringUtils.isEmpty(key)) {
```
AlbumenJ, line 486: Why change this? This may cause some usage mistakes

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/url/component/URLParam.java
qingkaiyan: When a consumer with a * version invoke a null version provider
Building a remote URL based on the consumer's URL will be contiains * version
It's not a good result and will blocked by PermittedSerializationKeeper.checkSerializationPermitted
This issue may only arise when the * symbol is used and The impact of the change will be widespread
But null valus has meanings seems makes scene and 10+ application running well when I made these changes
It's true that no one dares to say these is no other problem
I've been experimenting for over a month and haven't failed yet

qingkaiyan: And a consumer with a * version invoke a null version provider worked well at 2.7
I use different version and condition-router to achieve gray and business group
And like to make the dubbo:consumer version = *
It's not work when I update to 3.2 or 3.3

AlbumenJ: It would be better to change `PermittedSerializationKeeper.checkSerializationPermitted` in the scenarios.

qingkaiyan: Maybe It's not just this problom
It would be better when the * version consumer called the null version provider to build the null version URL

AlbumenJ: `URL` is a really really basic component for Dubbo, so I would prefer not to change it enless we have the conclusion that we can only change it to fix a issue.

PR #14700, opened 2024-09-20
bugfix: application/x-www-form-urlencoded There is a problem with parameter parsing

### LINKED ISSUE #14696: [Bug] application/x-www-form-urlencoded There is a problem with parameter parsing
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

dubbo 3.2.16

```
    @POST
    @Path("/test")
    public ResponseVO test(
        @FormParam("path") @QueryParam("path")  String path,@FormParam("body") @QueryParam("body")  String body) {
......
}
```

```
POST http://localhost:8087/test
Content-Type: application/x-www-form-urlencoded

body=a&
path=a
```
After the request, both path and body are null, because the following arraylist always reads only the 0th argument.
```
    public Object decode(byte[] body, Class<?> targetType, Type type) throws Exception {
        Object map = DataParseUtils.multipartFormConvert(body, targetType);
        Map valuesMap = (Map) map;
        if (Map.class.isAssignableFrom(targetType)) {
            return map;
        } else if (DataParseUtils.isTextType(targetType)) {

            // only fetch  first
            Set set = valuesMap.keySet();
            ArrayList arrayList = new ArrayList<>(set);
            Object key = arrayList.get(0);
            Object value = valuesMap.get(key);
            if (value == null) {
                return null;
            }
            return DataParseUtils.stringTypeConvert(targetType, String.valueOf(((List) value).get(0)));

        }
....
}
```
然后ParamProviderParamParser在解析为null时，会set下标为null，而在它之前的BodyProviderParamParser已经将值解析后进行set，却被前者给覆盖为null
Then, when ParamProviderParamParser parses to null, it sets the index to null, while the previous BodyProviderParamParser has already parsed and set the value, which gets overridden to null by the former

如果其中一个parser解析出null值，不应该进行set，然后应该将ArgInfo向下传入，而不是传入type，type中没有paramName，所以在BodyProviderParamParser中无法找到对应的Param
If one of the parsers parses a null value, it should not perform the set action, and the ArgInfo should be passed into the parser instead of just the type. The type does not contain the paramName, so the corresponding parameter cannot be found in BodyProviderParamParser

### Anything else

_No response_

### Code of Conduct

funky-eyes:

fixes #14696 

## Checklist

### DIFF
--- a/dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/rest/ArgInfo.java
+++ b/dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/rest/ArgInfo.java
@@ -160,6 +160,10 @@ public class ArgInfo {
         return actualType;
     }
 
+    public void setActualType(Type actualType) {
+        this.actualType = actualType;
+    }
+
     @Override
     public String toString() {
         return "ArgInfo{" + "index="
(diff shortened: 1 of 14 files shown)


suncairong163:
@CrazyHZM 
@AlbumenJ

PR #14702, opened 2024-09-20
Added support for configuring Tomcat HTTP2 max stream size in Spring boot

oxsean:
…Boot

## Checklist

### DIFF
--- a/dubbo-spring-boot/dubbo-spring-boot-3-autoconfigure/pom.xml
+++ b/dubbo-spring-boot/dubbo-spring-boot-3-autoconfigure/pom.xml
@@ -70,7 +70,12 @@
     <dependency>
       <groupId>jakarta.servlet</groupId>
       <artifactId>jakarta.servlet-api</artifactId>
-      <optional>true</optional>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.tomcat.embed</groupId>
+      <artifactId>tomcat-embed-core</artifactId>
+      <scope>provided</scope>
     </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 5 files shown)


dubbo-spring-boot/dubbo-spring-boot-3-autoconfigure/pom.xml
```diff
@@ -68,9 +68,9 @@
     </dependency>
 
     <dependency>
-      <groupId>jakarta.servlet</groupId>
-      <artifactId>jakarta.servlet-api</artifactId>
-      <optional>true</optional>
+      <groupId>org.apache.tomcat.embed</groupId>
+      <artifactId>tomcat-embed-core</artifactId>
+      <scope>provided</scope>
```
CrazyHZM, line 78: This change will cause users to have to use Tomcat.

oxsean: > This change will cause users to have to use Tomcat.

The scope is provided, which doesn't have any actual impact, just adds special support for Tomcat, as Tomcat already includes the servlet-api, so it can be replaced directly.
Add a ConditionalOnClass to avoid dependency on tomcat.

dubbo-spring-boot/dubbo-spring-boot-3-autoconfigure/src/main/java/org/apache/dubbo/spring/boot/autoconfigure/DubboTriple3AutoConfiguration.java
```diff
@@ -54,5 +59,23 @@ public FilterRegistrationBean<TripleFilter> tripleProtocolFilter(
...
+        @ConditionalOnClass(Http2Protocol.class)
+        @ConditionalOnProperty(prefix = PREFIX, name = "max-concurrent-streams")
+        public WebServerFactoryCustomizer<ConfigurableTomcatWebServerFactory> tripleTomcatHttp2Customizer(
+                @Value("${" + PREFIX + ".max-concurrent-streams}") int maxConcurrentStreams) {
+            return factory -> factory.addConnectorCustomizers(connector -> {
+                ProtocolHandler handler = connector.getProtocolHandler();
+                for (UpgradeProtocol upgradeProtocol : handler.findUpgradeProtocols()) {
+                    if (upgradeProtocol instanceof Http2Protocol) {
+                        Http2Protocol protocol = (Http2Protocol) upgradeProtocol;
+                        int value = maxConcurrentStreams < 0 ? Integer.MAX_VALUE : maxConcurrentStreams;
```
AlbumenJ:
```suggestion
                        int value = maxConcurrentStreams <= 0 ? Integer.MAX_VALUE : maxConcurrentStreams;
```

AlbumenJ (changes requested):

PR #14718, opened 2024-09-24
Fix triple client connection shareing race condition

### LINKED ISSUE #14716: [Bug] Potential memory leak in certain scenarios caused by MetadataServiceV2
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

3.3.0

Start a regular provider and consumer, enable application-level service discovery by adding the following configuration on provider side:

```yaml
dubbo:
  registry:
    register-mode: instance
```

Dump the heap on consumer side after it starts successfully, and find the metadata proxy instance doesn't destroy as expected.

The root reason is that `MetadataServiceV2` is a stub type proxy instead of a reflection proxy type, so it does not implement `Destroyable`. The following check `if (proxyV2 instanceof Destroyable) ` will never match.

```java
public void destroy() {
          if (proxy instanceof Destroyable) {
              ((Destroyable) proxy).$destroy();
          }

          if (proxyV2 instanceof Destroyable) {
              ((Destroyable) proxyV2).$destroy();
          }

          internalModel.getServiceRepository().unregisterConsumer(consumerModel);
}
```

Even if `MetadataServiceV2` instances failed to destroy as expected, the number of these instances won't increase with no limitation.  In most cases, the number of instances of `MetadataServiceV2` will stop increaseing when it reaches the number of the provider instances, that says, the maximum number depends on how many provider ips you have. 

However, in certain scenarios, `MetadataServiceV2` instances will keep growing as the provider instance redeploys with some Dubbo-related changes that can cause `revision` changes.

`MetadataServiceV2` destroyed every time.

### Anything else

_No response_

### Code of Conduct

chickenlj:
Fixes https://github.com/apache/dubbo/issues/14716 and https://github.com/apache/dubbo/issues/14717

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/api/connection/AbstractConnectionClient.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/api/connection/AbstractConnectionClient.java
@@ -66,6 +66,9 @@ public abstract class AbstractConnectionClient extends AbstractClient {
         if (oldCount <= 0) {
             COUNTER_UPDATER.getAndDecrement(this);
             logger.warn(
+                    PROTOCOL_ERROR_CLOSE_CLIENT,
+                    "",
+                    "",
                     "Retain failed, because connection {} has been destroyed but not yet removed, will create a new one instead."
                             + " Check logs below to confirm that this connection finally gets removed to make sure there's no potential memory leak!",
                     remote);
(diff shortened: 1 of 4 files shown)


dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/api/connection/AbstractConnectionClient.java
```diff
@@ -61,13 +61,17 @@ public final void increase() {
     /**
      * Increments the reference count by 1.
      */
-    public final AbstractConnectionClient retain() {
+    public final boolean retain() {
         long oldCount = COUNTER_UPDATER.getAndIncrement(this);
         if (oldCount <= 0) {
             COUNTER_UPDATER.getAndDecrement(this);
-            throw new AssertionError("This instance has been destroyed");
+            logger.warn(
```
oxsean: Should set error code

oxsean (approved):
LGTM.

AlbumenJ (approved):

PR #14730, opened 2024-09-29
Fix safeLock not run when interrupt & Fix partial notification being cancelled in ServiceInstancesChangedListener

AlbumenJ:
## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/LockUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/LockUtils.java
@@ -31,17 +31,20 @@ public class LockUtils {
 
     public static void safeLock(Lock lock, int timeout, Runnable runnable) {
         try {
-            if (!lock.tryLock(timeout, TimeUnit.MILLISECONDS)) {
-                logger.error(
-                        LoggerCodeConstants.INTERNAL_ERROR,
-                        "",
-                        "",
-                        "Try to lock failed, timeout: " + timeout,
-                        new TimeoutException());
+            boolean interrupted = false;
+            try {
+                if (!lock.tryLock(timeout, TimeUnit.MILLISECONDS)) {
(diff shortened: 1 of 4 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/utils/LockUtils.java
```diff
@@ -31,18 +31,24 @@ public class LockUtils {
...
+                            "",
+                            "",
+                            "Try to lock failed, timeout: " + timeout,
+                            new TimeoutException());
+                }
+            } catch (InterruptedException e) {
+                logger.warn(LoggerCodeConstants.INTERNAL_ERROR, "", "", "Try to lock failed", e);
+                interrupted = true;
             }
             runnable.run();
```
marmot-z, line 48: This may seem a bit strange, because generally the tryLock method is used like this:
```java
 Lock lock = ...;
  if (lock.tryLock()) {
      try {
          // do something with lock
          runnable.run();
      } finally {
          lock.unlock();
      }
  } 
```

Wouldn't this really lead to some race conditions?

AlbumenJ: `safeLock` here is to **try best** to **wait**.  In Dubbo, those actions needed `safeLock` can accept delay, but cannot accept skip.

AlbumenJ: Also, the patch in`dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/event/listener/ServiceInstancesChangedListener.java` is the real code to fix https://github.com/apache/dubbo/issues/14470. The original issue is once invoke failed, `submitRetryTask` won't being invoked because of the code order.

marmot-z: i got it, thanks for your patient explanation.

PR #14793, opened 2024-10-18
Fix NetUtils: should ignore link-local address

### LINKED ISSUE #14792: [Bug] NetUtils should ignore the local link address
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Any

In the K8S scenario, some applications may have two NICs, eth0 and veth0, and eth0 is configured with a real physical address. veth0 is configured with a local link address, which may randomly select the local link address to veth0, resulting in unavailability of the service.

When I look at the netUtils code, I see that netutils only ignores the lookback address, not the local link address.

https://github.com/mercyblitz/dubbo/blob/d8a4acd13c8270bdefe38f8b109dee457ee15593/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java#L148

NetUtils should ignore the local link address

### Anything else

_No response_

### Code of Conduct

cyclinder:

NetUtils: should be ignore link-local address

Fixes https://github.com/apache/dubbo/issues/14792

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
@@ -196,7 +196,7 @@ public final class NetUtils {
     }
 
     static boolean isValidV4Address(InetAddress address) {
-        if (address == null || address.isLoopbackAddress()) {
+        if (address == null || address.isLoopbackAddress() || address.isLinkLocalAddress()) {
             return false;
         }
 


heliang666s:
LGTM

dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
```diff
@@ -196,7 +196,7 @@ public static InetSocketAddress getLocalSocketAddress(String host, int port) {
     }
 
     static boolean isValidV4Address(InetAddress address) {
-        if (address == null || address.isLoopbackAddress()) {
+        if (address == null || address.isLoopbackAddress() || address.isLinkLocalAddress()) {
```
AlbumenJ, line 199: What would happen if there is no any non-local interface

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
yunmaoQu, line 199: @AlbumenJ Just suggesstion:add a parameter called  "allowLinkLocal(boolean)" in some related method( i find two extend point from original structure) .And user can config the value in system config file while  it's default value is true to  prevent the situation that though there is no any non-local interface and it also work ,  it's  ok?

yunmaoQu: i've done the work in local,if my idea is available and need my code ,i 'll push a request under this issue!

cyclinder: if there is no non-local address, `address.isLinkLocalAddress()` would return false, This function wouldn't return directly, Continue with the logic below, same as if it was a lookback address.

cyclinder: Refer to https://github.com/apache/dubbo/pull/14793/files#r1814182854, `address.isLinkLocalAddress()` shouldn't affect if there is no non-local address. I don't have a Dubbo environment locally, Can you help test the changes? thanks a lot.

PR #14872, opened 2024-11-07
add 'micrometer-core' as transitive dependency

chickenlj:
```xml
<dependency>
      <groupId>io.micrometer</groupId>
      <artifactId>micrometer-core</artifactId>
    </dependency>
```


songxiaosheng:
why add it

songxiaosheng:
let me fix it :
image

PR #14888, opened 2024-11-11
[Dubbo]: Fix the bug that isDefault configuration is not useful  #12961

### LINKED ISSUE #12961: protocols 的 default 属性配置不生效

Dubbo version: 3.2.5
Operating System version: windows
Java version: 1.8.0

dubbo配置文件里的 dubbo.protocols.rest.default=false 不生效，看了下应该是org.apache.dubbo.config.AbstractConfig#isDefault 方法有 @Parameter(excluded = true, attribute = false) 注解，导致被 org.apache.dubbo.config.AbstractConfig#isIgnoredAttribute 方法忽略了，这是有BUG，还是要准备用其它方式配置呢？
image
image

paradiseidler:
Fix #12961 
The tag with attribute=false for the isDefault property has been removed.

 

In the isIgnoredAttribute method, due to isDefault having attribute=false, the entire method returns a Boolean value of true.

In the assignProperties method of the AbstractConfig class, at the breakpoint, the final condition evaluates to false, preventing the default property assignment process from being executed, resulting in the default value remaining unprocessed as null.

## Checklist


oxsean:
It looks like this change also impacts the hashCode equals logic, possibly already existing configurations change from being equal to now not equal. 
If it's really necessary, may need to add a new annotation property

oxsean (changes requested):
It looks like this change also impacts the hashCode equals logic, possibly already existing configurations change from being equal to now not equal.
If it's really necessary, may need to add a new annotation property

PR #14900, opened 2024-11-17
fix flaky tests under DubboAppenderTest

ljmcr:
Two tests under module dubbo-common are detected as Order-Dependent flaky tests:
`org.apache.dubbo.common.utils.DubboAppenderTest.testAvailable`
`org.apache.dubbo.common.utils.DubboAppenderTest.testAppend`

In this PR, I modified the test setup to ensure that each unit test starts with an independent state by resetting the DubboAppender in the @BeforeEach method. Currently, all tests rely on the tearDown() method annotated with @AfterEach to clear the state. However, in unit testing, it is essential for each test to be independent and not depend on the state left by previous tests. If a test fails in a way that prevents tearDown() from running (e.g., due to an unexpected exception or forced termination), it can result in residual state in logList, causing dependencies between tests and breaking test isolation.

Additionally, I replaced assumeTrue() and assumeFalse() with assertThat() to ensure that tests fail with clear feedback when conditions are not met, rather than being skipped. Using assertions is better to report failures directly, rather than directly skipping a test when a condition is not met.

## Checklist


wcy666103:
use `mvn spotless:apply` to format the code.

PR #14950, opened 2024-11-28
Close-NodeCache-before-Curator

wcy666103:

https://github.com/apache/dubbo/issues/10150#issuecomment-2505167696

Close every NodeCache before close CuratorFramework.

## Checklist


wcy666103:
The `nodeCacheMap` is a cache, and each cache should not be closed when the current client is destroyed.

imfms:
> The `nodeCacheMap` is a cache, and each cache should not be closed when the current client is destroyed.
```java
private static Map<String, NodeCache> nodeCacheMap = new ConcurrentHashMap<>();
```
Global cache across multiple Dubbo instances in the same process?
It seems that Nodecache (CuratorFramework client, String path) holds a reference to the CuratorFramework-client, I don't know if there is a problem with different client-cache across multiple instances.

I'm not very familiar with the Dubbo code, or maybe I'm just overthinking it.
I've temporarily ignored NodeCache's ERROR log: Expected state [ STARTED ] was [ STOPPED]

Thanks for your work.

PR #15006, opened 2024-12-20
Check the md5 of the metadata cache file

finefuture:
The local cache file may be damaged for various reasons, resulting in a no provider exception, check the consistency of md5 can avoid this problem.

## Checklist


dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/MetadataInfo.java
```diff
@@ -212,6 +207,15 @@ public synchronized String calAndGetRevision() {
         return revision;
     }
 
+    public String calRevision() {
```
AlbumenJ: It would be better to add a lock here to prevent concurrent modification.

BTW, please check other operation will aquire lock

AlbumenJ (changes requested):

dubbo-metadata/dubbo-metadata-api/src/main/java/org/apache/dubbo/metadata/MetadataInfo.java
finefuture: Done.

AlbumenJ (approved):

PR #15014, opened 2024-12-24
remove unnecessary protocolRevisionsToUrls and build newServiceUrls d…

aruato:
…irectly

The primary role of **protocolRevisionsToUrls** in the overall code logic is to act as an intermediate cache, mapping **(protocol, port, revisions)** to **urls**. However, the ultimate goal is to construct **newServiceUrls**, and **protocolRevisionsToUrls** is not directly used elsewhere. This layer can be completely removed, and the results can be directly stored in **newServiceUrls**.

## Checklist


finefuture (commented):
If this logic is deleted, a lot of unnecessary repeated calculations will be added when the same revision is encountered.

Issue #15031, opened 2025-01-02
[Feature] Add a configuration to determine whether to enable the point-to-point network interface


finefuture:
### Pre-check

- [X] I am sure that all the content I provide is in English.

### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar feature requirement.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Descriptions

**The potential impact of point-to-point network interface on RPC**

> Interference communication mode: Point to point network interface are typically used to establish direct connections between two specific nodes, such as connecting two data centers through dedicated lines. In RPC, if not handled correctly, this point-to-point connection characteristic may interfere with normal communication modes. For example, RPC is typically expected to work in a more general network environment, which may involve multiple clients communicating with one or more servers, while point-to-point network interface may restrict communication between specific two endpoints, which does not meet the general communication requirements of RPC.

**What is the function I expect**

> Add a configuration to determine whether to enable the point-to-point network interface.

### Are you willing to submit a pull request to fix on your own?

- [X] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #15032, opened 2025-01-03
Add a configuration to determine whether to enable the point-to-point network interface

finefuture:
Related   #15031 

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
@@ -738,6 +738,15 @@ public interface CommonConstants {
          */
         String DUBBO_PREFERRED_NETWORK_INTERFACE = "dubbo.network.interface.preferred";
 
+        /**
+         * The property name for {@link NetworkInterface#isPointToPoint() return whether a network interface is a point
+         * to point interface} that the Dubbo application will determine whether to enable the point-to-point network
+         * interface
+         *
+         * @since 3.3
+         */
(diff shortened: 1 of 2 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
```diff
@@ -738,6 +738,15 @@ interface DubboProperty {
...
         String DUBBO_PREFERRED_NETWORK_INTERFACE = "dubbo.network.interface.preferred";
 
+        /**
+         * The property name for {@link NetworkInterface#isPointToPoint() return whether a network interface is a point
+         * to point interface} that the Dubbo application will determine whether to enable the point-to-point network
+         * interface
+         *
+         * @since 3.3
+         */
+        String DUBBO_NETWORK_INTERFACE_POINT_TO_POINT_ENABLE = "dubbo.network.interface.point-to-point.enable";
```
AlbumenJ:
```suggestion
        String DUBBO_NETWORK_INTERFACE_POINT_TO_POINT_IGNORED = "dubbo.network.interface.point-to-point.ignored";
```
would be better

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
finefuture: Done.

AlbumenJ (approved):

PR #15046, opened 2025-01-09
Remove datasource status checker

CrazyHZM:
This content check should not be placed in dubbo.


AlbumenJ:
How about move to 3.4?

CrazyHZM:
> How about move to 3.4?

Already changed it to 3.4. @AlbumenJ

PR #15115, opened 2025-02-09
Fix triple protocol for native image

CrazyHZM:
## Checklist

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataUtils.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataUtils.java
@@ -176,7 +176,7 @@ public class MetadataUtils {
         RemoteMetadataService remoteMetadataService;
         ProxyFactory proxyFactory =
                 applicationModel.getExtensionLoader(ProxyFactory.class).getAdaptiveExtension();
-        if (useV2) {
+        if (useV2 && !inNativeImage) {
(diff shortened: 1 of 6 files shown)


dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataUtils.java
```diff
@@ -176,7 +176,7 @@ public static RemoteMetadataService referMetadataService(ServiceInstance instanc
         RemoteMetadataService remoteMetadataService;
         ProxyFactory proxyFactory =
                 applicationModel.getExtensionLoader(ProxyFactory.class).getAdaptiveExtension();
-        if (useV2) {
+        if (useV2 && !inNativeImage) {
```
AlbumenJ, line 179: Why can't use v2 in native image here?

AlbumenJ (changes requested):

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/metadata/MetadataUtils.java
CrazyHZM: MetadataServiceV2 only supports `nativestub` proxy mode. This needs to be supported separately, and the original problem should be fixed first in this pr.

AlbumenJ:
image

CrazyHZM:
> image

It doesn't seem to be about the change.

FoghostCn (approved):
LGTM

PR #15122, opened 2025-02-10
Ignore RejectedExecutionException if executor is shutdown during submit task

AlbumenJ:
…it task

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/serial/SerializingExecutor.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/serial/SerializingExecutor.java
@@ -19,6 +19,7 @@ package org.apache.dubbo.common.threadpool.serial;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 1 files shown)


AlbumenJ:
@EarthChen @icodening @oxsean PTAL

dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/serial/SerializingExecutor.java
```diff
@@ -73,6 +74,11 @@ private void schedule(Runnable removable) {
                     executor.execute(this);
                     success = true;
                 }
+            } catch (RejectedExecutionException e) {
+                if (!isShutdown(executor)) {
```
oxsean, line 78: Should we add some logs when ignoring?

oxsean (approved):

icodening (approved):
LGTM

dubbo-common/src/main/java/org/apache/dubbo/common/threadpool/serial/SerializingExecutor.java
AlbumenJ: This may produce a lot of logs

Issue #15187, opened 2025-02-28
[Bug] http header Content-Type not be parsed correctly


brookqin:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo java 3.3.3 JDK 17

### Steps to reproduce this issue

When the HTTP header `Content-Type: multipart/form-data; charset=utf-8; boundary=__X_PAW_BOUNDARY__`, a parsing exception will occur.

```
java.nio.charset.IllegalCharsetNameException: utf-8; boundary=__X_PAW_BOUNDARY__
	at java.base/java.nio.charset.Charset.checkName(Charset.java:328) ~[na:na]
	at java.base/java.nio.charset.Charset.lookup2(Charset.java:512) ~[na:na]
	at java.base/java.nio.charset.Charset.lookup(Charset.java:492) ~[na:na]
	at java.base/java.nio.charset.Charset.forName(Charset.java:556) ~[na:na]
	at org.apache.dubbo.remoting.http12.message.DefaultHttpRequest.getDecoder(DefaultHttpRequest.java:611) ~[dubbo-3.3.3.jar:3.3.3]
	at org.apache.dubbo.remoting.http12.message.DefaultHttpRequest.queryParameterValues(DefaultHttpRequest.java:441) ~[dubbo-3.3.3.jar:3.3.3]
	at org.apache.dubbo.remoting.http12.message.DefaultHttpRequest.queryParameter(DefaultHttpRequest.java:436) ~[dubbo-3.3.3.jar:3.3.3]
	at org.apache.dubbo.rpc.protocol.tri.rest.mapping.ContentNegotiator.negotiate(ContentNegotiator.java:78) ~[dubbo-3.3.3.jar:3.3.3]
	at org.apache.dubbo.rpc.protocol.tri.rest.mapping.RestRequestHandlerMapping.getRequestHandler(RestRequestHandlerMapping.java:90) ~[dubbo-3.3.3.jar:3.3.3]
	at org.apache.dubbo.rpc.protocol.tri.route.DefaultRequestRouter.route(DefaultRequestRouter.java:50) ~[dubbo-3.3.3.jar:3.3.3]
[...]

PR #15188, opened 2025-02-28
Fix the issue where the charset from Content-Type is not parsed correctly, and refactor the related code.

RainYuY:
Fix the issue where the charset from Content-Type is not parsed correctly, and refactor the related code.
Ref:issue #15187

## Checklist

### DIFF
--- a/dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/HttpUtils.java
+++ b/dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/HttpUtils.java
@@ -81,6 +81,22 @@ public final class HttpUtils {
         return cookies;
     }
 
+    public static String getCharsetFromContentType(String contentType) {
+        String charset = null;
+        if (contentType == null) {
+            charset = StringUtils.EMPTY_STRING;
+        } else {
+            int index = contentType.lastIndexOf(CHARSET_PREFIX);
+            if (index == -1) {
+                charset = StringUtils.EMPTY_STRING;
+            } else {
+                charset = contentType.substring(index + CHARSET_PREFIX.length()).trim();
+                charset = charset.split(";")[0];
+            }
+        }
+        return charset;
+    }
+
     public static String encodeCookie(HttpCookie cookie) {
         DefaultCookie c = new DefaultCookie(cookie.name(), cookie.value());
         c.setPath(cookie.path());
--- a/dubbo-remoting/dubbo-remoting-http12/src/test/java/org/apache/dubbo/remoting/http12/message/codec/HttpUtilsTest.java
+++ b/dubbo-remoting/dubbo-remoting-http12/src/test/java/org/apache/dubbo/remoting/http12/message/codec/HttpUtilsTest.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.dubbo.remoting.http12.message.codec;
+
+import org.apache.dubbo.remoting.http12.HttpUtils;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class HttpUtilsTest {
+
+    @Test
+    void testGetCharsetFromContentType() {
+        String charset = HttpUtils.getCharsetFromContentType("text/html;charset=utf-8");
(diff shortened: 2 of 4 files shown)


RainYuY:
@AlbumenJ @oxsean PTAL

dubbo-remoting/dubbo-remoting-http12/src/test/java/org/apache/dubbo/remoting/http12/message/codec/HttpUtilsTest.java
```diff
@@ -0,0 +1,35 @@
...
+public class HttpUtilsTest {
+
+    @Test
+    void testGetCharsetFromContentType() {
+        String charset = HttpUtils.getCharsetFromContentType("text/html;charset=utf-8");
+        Assertions.assertEquals("utf-8", charset);
+        charset = HttpUtils.getCharsetFromContentType("text/html");
+        Assertions.assertEquals("", charset);
+        charset = HttpUtils.getCharsetFromContentType("application/json;charset=utf-8; boundary=__X_PAW_BOUNDARY__");
+        Assertions.assertEquals("utf-8", charset);
```
kimmking, line 33: should be constructing a multipart request to ensure it works well as expected.

kimmking (changes requested):

dubbo-remoting/dubbo-remoting-http12/src/test/java/org/apache/dubbo/remoting/http12/message/codec/HttpUtilsTest.java
RainYuY: I have added a multipart request. If you mean a completed HTTP connection test case, I think it would be better to add it to IT. I will add it after this PR is merged. Thanks for your CR.

brookqin:
@Stellar1999 
First of all, thank you for responding to this issue so promptly.

[rfc7231/3.1.1.1.  Media Type](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.1)
`media-type = type "/" subtype *( OWS ";" OWS parameter )`

According to the description in the RFC document, the `content-type` may include one or more parameters in addition to the media-type, with `charset` being just one of these parameters. There is no stipulated order for these parameters, so in extreme cases, relying on index positioning could potentially lead to issues.

Perhaps it is also necessary to define a new method: `String mediaTypeParam(String name);`, to retrieve the value of a specified parameter.
```
String mediaTypeParam(String name) {
  paramsStr = contentType.substring(X);
  params = paramsStr.split(";")
                    .filter(str -> str.contain("="))
                    .map(str -> str.split("="));
  return params[name];
}

var charset = mediaTypeParam("charset");
var boundary = mediaTypeParam("boundary");
```

RainYuY:
> @Stellar1999 First of all, thank you for responding to this issue so promptly.
> 
> [rfc7231/3.1.1.1. Media Type](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.1) `media-type = type "/" subtype *( OWS ";" OWS parameter )`
> 
> According to the description in the RFC document, the `content-type` may include one or more parameters in addition to the media-type, with `charset` being just one of these parameters. There is no stipulated order for these parameters, so in extreme cases, relying on index positioning could potentially lead to issues.
> 
> Perhaps it is also necessary to define a new method: `String mediaTypeParam(String name);`, to retrieve the value of a specified parameter.
> 
> ```
> String mediaTypeParam(String name) {
>   paramsStr = contentType.substring(X);
>   params = paramsStr.split(";")
>                     .filter(str -> str.contain("="))
>                     .map(str -> str.split("="));
>   return params[name];
> }
> 
> var charset = mediaTypeParam("charset");
> var boundary = mediaTypeParam("boundary");
> ```

Thanks for your review.
I think my new code does not rely on an index for positioning. (If you mean indexOf, it's only used to get the charset string.) Like you said, maybe we can find a better general method to retrieve more parameters. However, as I see in the Dubbo code, it only relies on the charset. So, I believe we just need a special method to obtain the charset.

dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/HttpUtils.java
```diff
@@ -81,6 +81,22 @@ public static List<HttpCookie> decodeCookies(String value) {
...
+        String charset = null;
+        if (contentType == null) {
+            charset = StringUtils.EMPTY_STRING;
+        } else {
+            int index = contentType.lastIndexOf(CHARSET_PREFIX);
+            if (index == -1) {
+                charset = StringUtils.EMPTY_STRING;
+            } else {
+                charset = contentType.substring(index + CHARSET_PREFIX.length()).trim();
+                charset = charset.split(";")[0];
```
oxsean, line 94: A tip, do not use split as it scans and slices the whole string and checks for regexp, use indexOf(char) for best performance.

dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/HttpUtils.java
```diff
@@ -81,6 +81,22 @@ public static List<HttpCookie> decodeCookies(String value) {
         return cookies;
     }
 
+    public static String getCharsetFromContentType(String contentType) {
```
oxsean, line 84: parse or resolve will be btter than get.
parseCharset(String contentType)
Clear enough.

dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/HttpUtils.java
RainYuY: OK, This PR has be merged. I will submit another PR to do this fix.
Thanks for your review.

PR #15189, opened 2025-03-01
check the network,Local site be the preferred one

RainYuY:
I found an issue: every time I enable the VPN, Dubbo's local address check always picks up the VPN address instead of my machine's actual address. This leads to an incorrect address selection, preventing normal startup.

I found a similar issue reported a long time ago. Here are two ideas I have:

1. When retrieving the address, check if it is a SiteLocalAddress. If it is, return it directly; otherwise, return a reachable alternative address. (After my development and testing, this solution at least resolves the issue on my computer.)
2.Add an IP parameter to specify the preferred local network interface instead of relying on the interface name parameter.

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
@@ -552,7 +552,11 @@ public final class NetUtils {
                     if (addressOp.isPresent()) {
                         try {
                             if (addressOp.get().isReachable(100)) {
-                                return networkInterface;
+                                if (addressOp.get().isSiteLocalAddress()) {
+                                    return networkInterface;
+                                } else {
+                                    result = networkInterface;
+                                }
                             }
                         } catch (IOException e) {
                             // ignore


zrlw:
<dubbo:protocol name="dubbo" port="20880" host="192.168.1.100"/>
you can use -D.... to choose special ip address.

zrlw:
see https://blog.csdn.net/tlcjbl/article/details/120708721

RainYuY:
> <dubbo:protocol name="dubbo" port="20880" host="192.168.1.100"/> you can use -D.... to choose special ip address.

You're right, @zrlw. This solves the protocol problem, but there are still some display issues and the need to consider whether there are other related issues. Also, I think we still need to consider the first tip.

zrlw:
There are many complex network scenarios besides VPN. for example, there might be no other way except to specify ip address and port bound to dubbo provider in docker container.

AlbumenJ:
How about register all the available ips to registry and let consumer try all the ips? But this might be a bad design:)

RainYuY:
> How about register all the available ips to registry and let consumer try all the ips? But this might be a bad design:)

I think so:)

PR #15190, opened 2025-03-01
FIX-15184 resolved NettyHttp Header add by set to empty String

dheerajaraj:
`io.netty.handler.codec.Headers.HttpMetadataAdapter` is adding a null value for `PseudoHeaderName.AUTHORITY` key. I checked the docs for `io.netty.handler.codec.Headers` and it might be linked to this [issue](https://github.com/spring-projects/spring-framework/issues/26274). So, I replaced with a default value of empty string "". All the client methods checking for the key `PseudoHeaderName.AUTHORITY` will do a null or empty value check using `StringUtils.isBlank()` and `StringUtils.isNotBlank`

## Checklist

### DIFF
--- a/dubbo-plugin/dubbo-triple-servlet/src/main/java/org/apache/dubbo/rpc/protocol/tri/servlet/HttpMetadataAdapter.java
+++ b/dubbo-plugin/dubbo-triple-servlet/src/main/java/org/apache/dubbo/rpc/protocol/tri/servlet/HttpMetadataAdapter.java
@@ -16,6 +16,7 @@
  */
 package org.apache.dubbo.rpc.protocol.tri.servlet;
 
+import org.apache.dubbo.common.utils.StringUtils;
 import org.apache.dubbo.remoting.http12.HttpHeaderNames;
 import org.apache.dubbo.remoting.http12.HttpHeaders;
 import org.apache.dubbo.remoting.http12.h2.Http2Header;
@@ -53,9 +54,10 @@ public final class HttpMetadataAdapter implements Http2Header {
                     headers.add(key, ven.nextElement());
                 }
             }
+            String hostName = request.getHeader(HttpHeaderNames.HOST.getName());
             headers.add(PseudoHeaderName.METHOD.value(), method());
             headers.add(PseudoHeaderName.SCHEME.value(), request.getScheme());
-            headers.add(PseudoHeaderName.AUTHORITY.value(), request.getHeader(HttpHeaderNames.HOST.getName()));
+            headers.add(PseudoHeaderName.AUTHORITY.value(), StringUtils.isBlank(hostName) ? "" : hostName);
             headers.add(PseudoHeaderName.PROTOCOL.value(), request.getProtocol());
             this.headers = headers;
         }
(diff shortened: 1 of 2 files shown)


RainYuY:
I think the target branch should be set to 3.3 instead of 3.4.

dheerajaraj:
hi @Stellar1999 actually the target branch was mentioned to be `3.3.3` in this [issue](https://github.com/apache/dubbo/issues/15184). The target class (`io.netty.handler.codec.Headers.HttpMetadataAdapter`) which is causing the Null Pointer Exception is occurring in `3.4` not in `3.3`.  Since I cannot find `3.3.3` in the [upstream repo](https://github.com/apache/dubbo/), I've set the target branch to `3.4`. Is this fine? Kindly advice 🙏

RainYuY:
> hi @Stellar1999 actually the target branch was mentioned to be `3.3.3` in this [issue](https://github.com/apache/dubbo/issues/15184). The target class (`io.netty.handler.codec.Headers.HttpMetadataAdapter`) which is causing the Null Pointer Exception is occurring in `3.4` not in `3.3`. Since I cannot find `3.3.3` in the [upstream repo](https://github.com/apache/dubbo/), I've set the target branch to `3.4`. Is this fine? Kindly advice 🙏

Version 3.3.3 is a subversion of 3.3, so I believe it should be merged into 3.3 first. After that, the Dubbo team usually merges it into 3.4 as well. I noticed you mentioned that this issue doesn’t affect 3.3. Does that mean you’re unable to reproduce this bug in version 3.3?

dheerajaraj:
Hi @Stellar1999, for version 3.3.3/3.4 the place where the error is happening (`org.apache.dubbo.rpc.protocol.tri.servlet.HttpMetadataAdapter.headers()#57`) looks like this as also mentioned in the original [issue](https://github.com/apache/dubbo/issues/15184): 
Screenshot 2025-03-02 at 15 46 35 

However, the same line in 3.3 looks like this: 
Screenshot 2025-03-02 at 15 47 16

Since you mentioned that 3.3.3 will be merged to 3.3, 
shall i change in 3.3 `request.getServerName()` to `request.getHeader(HttpHeaderNames.HOST.getName())` as how it's done in 3.3.3 & 3.4?

RainYuY:
> Hi @Stellar1999, for version 3.3.3/3.4 the place where the error is happening (`org.apache.dubbo.rpc.protocol.tri.servlet.HttpMetadataAdapter.headers()#57`) looks like this as also mentioned in the original [issue](https://github.com/apache/dubbo/issues/15184): Screenshot 2025-03-02 at 15 46 35
> 
> However, the same line in 3.3 looks like this: Screenshot 2025-03-02 at 15 47 16
> 
> Since you mentioned that 3.3.3 will be merged to 3.3, shall i change in 3.3 `request.getServerName()` to `request.getHeader(HttpHeaderNames.HOST.getName())` as how it's done in 3.3.3 & 3.4?

Oh, I see. I think this bug has already been fixed by PR [#15141](https://github.com/apache/dubbo/pull/15141). All we need to do now is merge 3.3 into 3.4. PTAL @zrlw @AlbumenJ

dheerajaraj:
Hi @Stellar1999 if i'm not wrong, the [bug](https://github.com/apache/dubbo/issues/15184) is caused because of this PR #15141. @robin977 could you kindly confirm if this the case?

AlbumenJ:
I will merge 3.3 into 3.4 later

AlbumenJ:
> I will merge 3.3 into 3.4 later

done

Issue #13666, opened 2024-01-16
Context loss when async invoke


bringwu:
provider service method：

```
	public CompletableFuture<Integer> newToken(NewTokenInfo tReq) {
		TokenInfo tokenInfo = new TokenInfo(JsonUtils.toJson(tReq), 60);	
		// 可以设置回应附加信息
//		RpcContext.getServerResponseContext().setAttachment("testInfo", JsonUtils.toJson(tokenInfo));
		// 可以设置回应附加信息
//		RpcContext.getServerContext().setAttachment("testInfo", JsonUtils.toJson(tokenInfo));

		CompletableFuture<Integer> completableFuture = new CompletableFuture<>();
		completableFuture = completableFuture.completeAsync(()->{
			logger.debug("tokenInfo:" + JsonUtils.toJson(tokenInfo));
			return 1;
		}, Executors.newFixedThreadPool(1));
		
		// 不可以设置回应附加信息
		RpcContextAttachment serverResponseContext = RpcContext.getServerResponseContext();
//		RpcContextAttachment serverResponseContext = RpcContext.getClientAttachment();
//		RpcContextAttachment serverResponseContext = RpcContext.getServerContext();
//		RpcServiceContext serverResponseContext = RpcContext.getServiceContext();
//		RpcServiceContext serverResponseContext = RpcContext.getCurrentServiceContext();
//		RpcContextAttachment serverResponseContext = RpcContext.getClientResponseContext();
//		RpcContextAttachment serverResponseContext = RpcContext.getServerAttachment();
		completableFuture = completableFuture.thenApply(ret->{
			// 不可以设置回应附加信息
//			RpcContextAttachment serverResponseContext = RpcContext.getServerResponseContext();
//			RpcContextAttachment serverResponseContext = RpcContext.getClientAttachment();
//			RpcContextAttachment serverResponseContext = RpcContext.getServerContext();
//			RpcServiceContext serverResponseContext = RpcContext.getServiceContext();
//			RpcServiceContext serverResponseContext = RpcContext.getCurrentServiceContext();
//			RpcContextAttachment serverResponseContext = RpcContext.getClientResponseContext();
//			RpcContextAttachment serverResponseContext = RpcContext.getServerAttachment();
			
			serverResponseContext.setAttachment("testInfo", JsonUtils.toJson(tokenInfo));
			return ret;
		});
		return completableFuture;
	}
```

consumer :
```
[...]

PR #15192, opened 2025-03-02
fix Context loss when async invoke issue#13666

MaxChen1123:
to fix issue [#13666](https://github.com/apache/dubbo/issues/13666) in Dubbo Protocol (still has this issue in Triple) 

## How

1.  make it not check whether the map in RpcContextAttachment is empty (cuz it may be changed)
2. write the ServerResponseContext in the future callback to make sure the operation to the context has already done

## Checklist


EarthChen:
Please write some tests to verify the current changes.

MaxChen1123:
@EarthChen PTAL

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncRpcResult.java
```diff
@@ -81,6 +82,16 @@ public AsyncRpcResult(CompletableFuture<AppResponse> future, Invocation invocati
                 && !future.isDone()) {
             async = true;
             this.storedContext = RpcContext.clearAndStoreContext();
+
+            // this is to fix https://github.com/apache/dubbo/issues/13666
+            this.responseFuture = this.responseFuture.thenApply((appResponse) -> {
+                RestoreContext tempStoredContext = RpcContext.clearAndStoreContext();
+                this.storedContext.restore();
```
AlbumenJ, line 89: Restore here might impact performance

AlbumenJ (changes requested):

dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/AsyncRpcResult.java
MaxChen1123: Should we just get the `serverResponseLocal` in `storedContext` directly? Or maybe we can have a flag in the RpcContext, once the user pass context into a thread, they should manually set this flag to true, then we will check the flag to decide whether add this `thenApply` or not. I'm not sure which one is better. @AlbumenJ

PR #15230, opened 2025-03-11
fix: getClassLoader() may return null

fayfox:
In the Java 8 + WAR + Tomcat runtime environment, when using CompletableFuture.runAsync() (which runs in ForkJoinPool), Thread.currentThread().getContextClassLoader() may return null.

Meanwhile, SerializerFactory#getClassFactory only creates a ClassFactory instance on its first invocation. If the first Dubbo call happens to be executed in ForkJoinPool, ClassFactory._loader will be initialized as null and will remain null permanently. As a result, all subsequent Dubbo calls will fail during deserialization (objects will be converted to HashMap).

Therefore, if classLoader is null at this point, it is better to throw an exception immediately to prevent ClassFactory from being initialized incorrectly.


AlbumenJ:
2.7.x has been EOL. In this case, hessian should support null classloader.

PR #15239, opened 2025-03-14
[3.3] Enable server socket reuseAddress option before binding

### LINKED ISSUE #15238: [Bug] sometimes the port got from NetUtils#getAvailablePort could not be bound directly
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3

Image

the port that NetUtils#getAvailablePort opened and closed should be reused instantly no matter the port status is TIME-WAIT or not.

### Anything else

_No response_

### Code of Conduct

zrlw:
try to fix #15238 
sometimes the port got from NetUtils#getAvailablePort could not be bound directly because the port status is still TIME-WAIT as getAvailablePort closed it without setting SO_REUSEADDR option.

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
@@ -105,6 +105,21 @@ public final class NetUtils {
      */
     private static BitSet USED_PORT = new BitSet(65536);
 
+    private static boolean reuseAddressSupported;
+
+    static {
+        try (ServerSocket serverSocket = new ServerSocket()) {
+            serverSocket.setReuseAddress(true);
+            reuseAddressSupported = true;
+        } catch (Throwable ignored) {
+            // ignore.
+        }
+    }
+
+    public static boolean isReuseAddressSupported() {
+        return reuseAddressSupported;
+    }
+
     public static int getRandomPort() {
         return RND_PORT_START + ThreadLocalRandom.current().nextInt(RND_PORT_RANGE);
     }
@@ -123,7 +138,12 @@ public final class NetUtils {
             if (USED_PORT.get(i)) {
                 continue;
             }
-            try (ServerSocket ignored = new ServerSocket(i)) {
+            try (ServerSocket serverSocket = new ServerSocket()) {
+                if (reuseAddressSupported) {
+                    // SO_REUSEADDR should be enabled before bind.
+                    serverSocket.setReuseAddress(true);
+                }
+                serverSocket.bind(new InetSocketAddress(i));
                 USED_PORT.set(i);
                 port = i;
                 break;
@@ -141,7 +161,12 @@ public final class NetUtils {
      * @return true if it's occupied
      */
     public static boolean isPortInUsed(int port) {
-        try (ServerSocket ignored = new ServerSocket(port)) {
+        try (ServerSocket serverSocket = new ServerSocket()) {
+            if (reuseAddressSupported) {
+                // SO_REUSEADDR should be enabled before bind.
+                serverSocket.setReuseAddress(true);
+            }
+            serverSocket.bind(new InetSocketAddress(port));
             return false;
         } catch (IOException e) {
             // continue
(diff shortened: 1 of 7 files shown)


oxsean:
Is it possible to add some tests?

zrlw:
> Is it possible to add some tests?

the results of these codes vary with os and jdk version,
```
ServerSocket serverSocket = new ServerSocket(port);
assertFalse(serverSocket.getReuseAddress());
```
passed if running on win10 + compiled by jdk8,
failed if running on linux + compiled by jdk17 as ```serverSocket.getReuseAddress()``` returns true.
i doubt there should have unified solution.

zrlw:
the reason is net.ipv4.tcp_tw_reuse was set to 1 on my linux, so the reuseAdd option of new serverSocket is true.

zrlw:
see ```SO_REUSEADDR``` at java.net.StandardSocketOptions,
```
Re-use address.
An implementation allows this socket option to be set before the socket is bound or connected. Changing the value of this socket option after the socket is bound has no effect. The default value of this socket option is system dependent.
```
SO_REUSEADDR
since the default value of SO_REUSEADDR is always true on some system, I didn't think of how to write appropriate testing case.
```
    @Test
    void testRepeatedStatusChecking() {
        int port = NetUtils.getAvailablePort();
        for (int i = 0; i < 1000; i++) {
            assertFalse(NetUtils.isPortInUsed(port));
        }
    } 
```

zrlw:
setReuseAddress
Enabling SO_REUSEADDR prior to binding the socket using bind(SocketAddress) allows the socket to be bound even though a previous connection is in a timeout state.

dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
```diff
@@ -123,7 +123,10 @@ public static synchronized int getAvailablePort(int port) {
             if (USED_PORT.get(i)) {
                 continue;
             }
-            try (ServerSocket ignored = new ServerSocket(i)) {
+            try (ServerSocket serverSocket = new ServerSocket()) {
+                // SO_REUSEADDR should be enabled before bind.
+                serverSocket.setReuseAddress(true);
```
AlbumenJ: Are there any systems which cannot support such configuration or will throw exception here?

zrlw: if an error occurs, or the socket is closed,  SocketException will be thrown here.

zrlw: i see. Embedded, real time system, or old Unix ( before 1980?) might not support SO_REUSEADDR. added condition check before setting it.

dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
```diff
@@ -105,6 +105,21 @@ DO NOT replace this logger to error type aware logger (or fail-safe logger), sin
      */
     private static BitSet USED_PORT = new BitSet(65536);
 
+    private static boolean reuseAddressSupported;
+
+    static {
+        try (ServerSocket serverSocket = new ServerSocket()) {
+            serverSocket.setReuseAddress(true);
+            reuseAddressSupported = true;
+        } catch (IOException e) {
```
AlbumenJ:
```suggestion
        } catch (Throwable ignore) {
```

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
zrlw: done.

zrlw:
there are still somethings need to be revised.
https://github.com/apache/dubbo/actions/runs/13937299977/job/39007599423
testConnection
test

zrlw:
@AlbumenJ PTAL

PR #15338, opened 2025-04-26
fix: toCommaDelimitedString returns "null,null" when inputs are null

ankitshokeen:
## Issue

https://github.com/apache/dubbo/issues/15321

This change corrects the behavior of StringUtils.toCommaDelimitedString(String one, String... others) to return null when either argument is null, as described in its Javadoc. Previously, the method would return a string like "null,null", which could lead to misleading outputs and test failures. This update ensures consistency between the method's implementation and its documented contract.

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
@@ -1250,6 +1250,12 @@ public final class StringUtils {
      * @since 2.7.8
      */
     public static String toCommaDelimitedString(String one, String... others) {
+        if (one == null) {
+            return null;
+        }
+        if (others == null) {
+            return one;
+        }
         String another = arrayToDelimitedString(others, COMMA_SEPARATOR);
         return isEmpty(another) ? one : one + COMMA_SEPARATOR + another;
     }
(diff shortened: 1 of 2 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
```diff
@@ -1242,14 +1242,17 @@ public static byte decodeHexByte(CharSequence s, int pos) {
...
-     * @param others others {@link String}
-     * @return <code>null</code> if <code>one</code> or <code>others</code> is <code>null</code>
+     * @param one    the first string value
+     * @param others additional string values
+     * @return a combined comma-delimited string, or <code>null</code> if {@code one} or {@code others} is {@code null}
      * @since 2.7.8
      */
     public static String toCommaDelimitedString(String one, String... others) {
+        if (one == null || others == null) {
+            return null;
```
AlbumenJ, line 1254: if one != null && others == null

should return one?

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
ankitshokeen: yes, it should return one

AlbumenJ (approved):

songxiaosheng (commented):
LGTM

PR #15347, opened 2025-04-28
Feature/triple http3 cleanup fix

ankitshokeen:

This pull request resolves the duplicate service mapping warnings observed during the execution of `TripleHttp3ProtocolTest`.
It ensures that each test is executed in a fully isolated environment by resetting both the `ApplicationModel` and `FrameworkModel` before initializing test logic.
By clearing previously registered services and protocol instances, this change guarantees cleaner test runs without interference or unexpected warnings.

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp3ProtocolTest.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp3ProtocolTest.java
@@ -32,6 +32,7 @@ import org.apache.dubbo.rpc.model.ModuleServiceRepository;
 import org.apache.dubbo.rpc.model.ProviderModel;
 import org.apache.dubbo.rpc.model.ServiceDescriptor;
 import org.apache.dubbo.rpc.model.ServiceMetadata;
+import org.apache.dubbo.rpc.protocol.tri.rest.mapping.DefaultRequestMappingRegistry;
 import org.apache.dubbo.rpc.protocol.tri.support.IGreeter;
 import org.apache.dubbo.rpc.protocol.tri.support.IGreeterImpl;
 import org.apache.dubbo.rpc.protocol.tri.support.MockStreamObserver;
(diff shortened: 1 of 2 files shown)


dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp3ProtocolTest.java
```diff
@@ -87,50 +100,70 @@ void testDemoProtocol() throws Exception {
...
+    void cleanup() {
+        if (applicationModel != null) {
+            try {
+                DefaultRequestMappingRegistry registry = applicationModel
+                        .getFrameworkModel()
+                        .getBeanFactory()
+                        .getBean(DefaultRequestMappingRegistry.class);
+                if (registry != null) {
+                    registry.destroy();
+                    System.out.println("DefaultRequestMappingRegistry destroyed after test");
```
AlbumenJ: System.out in ut is useless and please use logger instead

AlbumenJ (changes requested):

dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp3ProtocolTest.java
```diff
@@ -87,50 +100,70 @@ void testDemoProtocol() throws Exception {
...
+                }
+            } catch (Exception e) {
+                System.err.println("Cleanup error: " + e.getMessage());
+            }
```
mytang0: Exception should be thrown instead of just logging the error.

oxsean:
@ankitshokeen I think the root cause is that DefaultRequestMappingRegistry as a bean, is not properly destroyed. It needs to implement org.apache.dubbo.common.resource.Disposable so that it can be destroyed together when the ApplicationModel is destroyed.

ankitshokeen:
> @ankitshokeen I think the root cause is that DefaultRequestMappingRegistry as a bean, is not properly destroyed. It needs to implement org.apache.dubbo.common.resource.Disposable so that it can be destroyed together when the ApplicationModel is destroyed.

@oxsean Please let me know if there's anything else you'd like adjusted!

dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp3ProtocolTest.java
```diff
@@ -41,17 +41,33 @@
...
 
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 class TripleHttp3ProtocolTest {
 
+    private ApplicationModel applicationModel;
```
oxsean, line 52: Why hold applicationModel here? can we get it by `ApplicationModel.defaultModel()`?

dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/TripleHttp3ProtocolTest.java
```diff
@@ -41,17 +41,33 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(TripleHttp3ProtocolTest.class);
+
     @Test
     void testDemoProtocol() throws Exception {
+        // Full Environment Reset - Optimized
```
oxsean, line 58: Since AfterEach already destroys the applicationModel, why do we still need to destroy it here?

Issue #15355, opened 2025-05-08
[Bug] triple Bzip2 compressor will not properly close IO streams during exceptions


paxxie2:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.1 , JDK 17, mac  15.3.2

### Steps to reproduce this issue

```
   public byte[] compress(byte[] payloadByteArr) throws RpcException {
        if (null == payloadByteArr || 0 == payloadByteArr.length) {
            return new byte[0];
        }

        ByteArrayOutputStream out = new ByteArrayOutputStream();
        BZip2CompressorOutputStream cos;
        try {
            cos = new BZip2CompressorOutputStream(out);
            cos.write(payloadByteArr);
            cos.close();
        } catch (Exception e) {
            throw new IllegalStateException(e);
        }

        return out.toByteArray();
    }
```
If the cos.write() operation encounters an IO exception, the cos.close() method will not execute, which may result in incorrect resource release. Should execute close() in finally {} or use try-with-resources

### What you expected to happen

Using try-with-resources to manipulate BZip2CompressorOutputStream

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

oxsean:
@paxxie2 BZip2CompressorOutputStream does not hold any file(or io) handle here, and ByteArrayOutputStream itself will be release by gc. Therefore, even if it is not explicitly closed here, it should not cause any issue.

PR #15356, opened 2025-05-08
fix: Fix the issue of triple bzip2 compressor close ByteArrayOutputStream

paxxie2:
fix issue #15355

Correctly close BZip2CompressorOutputStream using try with sources

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/compressor/Bzip2.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/compressor/Bzip2.java
@@ -47,11 +47,8 @@ public class Bzip2 implements Compressor, DeCompressor {
         }
 
         ByteArrayOutputStream out = new ByteArrayOutputStream();
-        BZip2CompressorOutputStream cos;
-        try {
-            cos = new BZip2CompressorOutputStream(out);
+        try (BZip2CompressorOutputStream cos = new BZip2CompressorOutputStream(out)) {
             cos.write(payloadByteArr);
-            cos.close();
         } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/compressor/Bzip2.java
```diff
@@ -47,11 +47,8 @@ public byte[] compress(byte[] payloadByteArr) throws RpcException {
         }
 
         ByteArrayOutputStream out = new ByteArrayOutputStream();
-        BZip2CompressorOutputStream cos;
-        try {
-            cos = new BZip2CompressorOutputStream(out);
+        try (BZip2CompressorOutputStream cos = new BZip2CompressorOutputStream(out)) {
```
mytang0, line 50: 'Out' should also be fixed.

oxsean: As a FilterOutputStream, the underlying output stream should be closed together with it.

paxxie2: commons-io  ByteArrayOutputStream directly operates on memory, not network IO, and does not manipulate file handles, it's close() method is also an empty implementation, and it is not necessary to close。Just like a regular Java object. 
org.apache.commons.io.output.ByteArrayOutputStream#close: 
```
 public void close() throws IOException {
    }
```

oxsean: So I think we needn't do any change.

Issue #15357, opened 2025-05-08
[Bug] GenericFilter throws null pointer exception when executing proto related method calls


paxxie2:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo java 3.3.1 , jdk 17, mac os 15.3.2

### Steps to reproduce this issue

1. use dubbo-samples , https://github.com/apache/dubbo-samples/tree/master/2-advanced/dubbo-samples-generic/dubbo-samples-generic-call 
2. add proto file 
```
syntax = "proto3";

option java_multiple_files = true;

package org.apache.dubbo.samples.tri.grpc;

message GreeterRequest {
  string name = 1;
}

message GreeterReply {
  string message = 1;
}

service Greeter{

  rpc Greet(GreeterRequest) returns (GreeterReply);

}
```
3. Provider implement interface
```
public class GreeterImpl extends DubboGreeterTriple.GreeterImplBase{

    @Override
    public GreeterReply greet(GreeterRequest request) {
        System.out.println("接收来自客户端消息：" + request.getName());
        GreeterReply pax = GreeterReply.newBuilder().setMessage("pax").build();
        return pax;
    }

    public CompletableFuture<GreeterReply> greetAsync(GreeterRequest request) {
        System.out.println("接收来自客户端消息 async：" + request.getName());
        GreeterReply pax = GreeterReply.newBuilder().setMessage("pax").build();
        return CompletableFuture.completedFuture(pax);
    }

}

```
5. Inject beans into the provider. Update dubbo-samples-generic-call-provider resources/spring/generic-provider.xml
```
  <bean id="greeterimpl" class="org.apache.dubbo.samples.generic.call.impl.GreeterImpl"/>
  <dubbo:service interface="org.apache.dubbo.samples.tri.grpc.Greeter" ref="greeterimpl"/>
```
6. Update consumer. Update dubbo-samples-generic-call-consumer GenericCallConsumer
```
 public static void main(String[] args) throws Exception {
        ApplicationConfig applicationConfig = new ApplicationConfig();
        applicationConfig.setName("generic-call-consumer");
        RegistryConfig registryConfig = new RegistryConfig();
        registryConfig.setAddress("zookeeper://127.0.0.1:2181");
[...]

Issue #15361, opened 2025-05-08
[Bug] When new pull request about proto, Unit Test on ubuntu-latest sometime will throw IllegalStateException: Not found class TestService


paxxie2:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.1 , java 17, mac 15.3.2

### Steps to reproduce this issue

I new a pull request about proto,  Unit Test On ubuntu-latest execute failed, see https://github.com/apache/dubbo/actions/runs/14910545315/job/41883681545?pr=15360 .
Key log:
```
 [INFO] Running org.apache.dubbo.rpc.protocol.dubbo.MultiThreadTest
15:58:33.683 |-INFO  [main] bo.remoting.transport.netty4.NettyServer:69  -|  [DUBBO] Start NettyServer bind /0.0.0.0:31715, export /127.0.0.1:31715, dubbo version: 3.3.4-SNAPSHOT, current host: 10.1.0.186
15:58:33.687 |-INFO  [main] bo.remoting.transport.netty4.NettyClient:278 -|  [DUBBO] Successfully connect to server /10.1.0.186:31715 from NettyClient 10.1.0.186 using dubbo version 3.3.4-SNAPSHOT, channel is NettyChannel [channel=[id: 0xb9a4cf08, L:/10.1.0.186:54996 - R:/10.1.0.186:31715]], dubbo version: 3.3.4-SNAPSHOT, current host: 10.1.0.186
15:58:33.687 |-INFO  [main] bo.remoting.transport.netty4.NettyClient:95  -|  [DUBBO] Start NettyClient /10.1.0.186 connect to the server /10.1.0.186:31715, dubbo version: 3.3.4-SNAPSHOT, current host: 10.1.0.186
15:58:33.687 |-INFO  [NettyClientWorker-44-1] ting.transport.netty4.NettyClientHandler:64  -|  [DUBBO] The connection [id: 0xb9a4cf08, L:/10.1.0.186:54996 - R:/10.1.0.186:31715] of 10.1.0.186:54996 -> 10.1.0.186:31715 is established., dubbo version: 3.3.4-SNAPSHOT, current host: 10.1.0.186
[...]

Issue #15358, opened 2025-05-08
[Bug] When grpc client calls triple server, the first character of proto RPC method is not case sensitive


paxxie2:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

dubbo 3.3.1, jdk 17, mac os 15.3.2

### Steps to reproduce this issue

1. Define the grpc client proto file and declare the Greet method
```
syntax = "proto3";

option java_multiple_files = true;

package org.apache.dubbo.samples.tri.grpc;

message GreeterRequest {
  string name = 1;
}

message GreeterReply {
  string message = 1;
}

service Greeter{

  rpc Greet(GreeterRequest) returns (GreeterReply);

  rpc biStream(stream GreeterRequest) returns (stream GreeterReply);

  rpc serverStream(GreeterRequest) returns (stream GreeterReply);

}

```
2. Define the triple server proto file and declare the greet method
```
syntax = "proto3";

option java_multiple_files = true;

package org.apache.dubbo.samples.tri.grpc;

message GreeterRequest {
  string name = 1;
}

message GreeterReply {
  string message = 1;
}

service Greeter{

  rpc greet(GreeterRequest) returns (GreeterReply);

  rpc biStream(stream GreeterRequest) returns (stream GreeterReply);

  rpc serverStream(GreeterRequest) returns (stream GreeterReply);

}

```
3. Run triple server and grpc client, there are no exception and the results are returned normally
4. Update the grpc client proto file and declare the GReet method,
```
syntax = "proto3";

option java_multiple_files = true;

package org.apache.dubbo.samples.tri.grpc;

message GreeterRequest {
  string name = 1;
}

message GreeterReply {
  string message = 1;
}

service Greeter{

  rpc GReet(GreeterRequest) returns (GreeterReply);
//  rpc GREET(GreeterRequest) returns (GreeterReply);
  rpc Dreet(GreeterRequest) returns (GreeterReply);

  rpc biStream(stream GreeterRequest) returns (stream GreeterReply);

  rpc serverStream(GreeterRequest) returns (stream GreeterReply);

}
```
5. Run triple server and grpc client, there are excpetion,verify if the method name is not case sensitive
```
[...]

Issue #15291, opened 2025-04-01
[Bug] The triple service written using proto does not support capitalization of service methods when called on rest


paxxie2:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.4、 jdk 17

### Steps to reproduce this issue

1. use dubbo-sample-master
use dubbo-samples-triple-grpc,  the old protobuf file is follow:
```
syntax = "proto3";

option java_multiple_files = true;

package org.apache.dubbo.samples.tri.grpc;

message GreeterRequest {
  string name = 1;
}

message GreeterReply {
  string message = 1;
}

service Greeter{
 // key, the first letter is lowercase
  rpc greet(GreeterRequest) returns (GreeterReply);

  rpc biStream(stream GreeterRequest) returns (stream GreeterReply);

  rpc serverStream(GreeterRequest) returns (stream GreeterReply);

}

```

2. update protobuf file
```
syntax = "proto3";

option java_multiple_files = true;

package org.apache.dubbo.samples.tri.grpc;

message GreeterRequest {
  string name = 1;
}

message GreeterReply {
  string message = 1;
}

service Greeter{
 // key, Convert the first letter to uppercase
  rpc Greet(GreeterRequest) returns (GreeterReply);

  rpc biStream(stream GreeterRequest) returns (stream GreeterReply);

  rpc serverStream(GreeterRequest) returns (stream GreeterReply);

}
```
3. compile and run

4. Requests starting with uppercase or lowercase letters， call on http
The first http request detail (rpc method name is uppercase):
```
POST localhost:50053/org.apache.dubbo.samples.tri.grpc.Greeter/greet
Content-Type: application/json

{
  "name": "John"
}
```
Error details:
```
21:11:04.136 |-INFO  [NettyServerWorker-5-1] ing.transport.netty4.NettyChannelHandler:56  -|  [DUBBO] The connection [id: 0xade552f6, L:/127.0.0.1:50053 - R:/127.0.0.1:65057] of 127.0.0.1:65057 -> 127.0.0.1:50053 is established., dubbo version: 3.3.4, current host: 172.16.210.28
[...]

PR #15364, opened 2025-05-08
fix: fix some bugs about proto

paxxie2:
fix bug: #15357 #15358 #15361 #15291

## Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -147,7 +147,7 @@
 
     <!-- Maven plugins -->
     <maven_jar_version>3.4.2</maven_jar_version>
-    <maven_surefire_version>3.5.2</maven_surefire_version>
+    <maven_surefire_version>3.5.3</maven_surefire_version>
     <maven_deploy_version>3.1.4</maven_deploy_version>
     <maven_shade_version>3.6.0</maven_shade_version>
     <maven_compiler_version>3.14.0</maven_compiler_version>


dubbo-common/src/main/java/org/apache/dubbo/rpc/model/StubServiceDescriptor.java
```diff
@@ -96,6 +98,14 @@ public MethodDescriptor getMethod(String methodName, String params) {
      */
     public MethodDescriptor getMethod(String methodName, Class<?>[] paramTypes) {
         List<MethodDescriptor> methodModels = methods.get(methodName);
+        if (methodModels == null) {
+            for (Entry<String, List<MethodDescriptor>> entry : methods.entrySet()) {
+                if (methodName.equalsIgnoreCase(entry.getKey())) {
```
AlbumenJ: 1. Performance degradation
2. This is a breaking change for old-users

AlbumenJ (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/rpc/model/StubServiceDescriptor.java
paxxie2: For the 1st point, modifying and  the performance issue can avoid it.
For the 2nd point, does it mean case insensitive or does it mean that the getMethod() method has been modified?

Issue #15366, opened 2025-05-09
[Bug] @Mapping annotation cannot register rest RequestMapping when it is not a method declared by triple proto


paxxie2:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.1 , java 17, mac os 15.3.2

### Steps to reproduce this issue

1. proto
```
syntax = "proto3";

option java_multiple_files = true;

package org.apache.dubbo.samples.tri.grpc;

message GreeterRequest {
  string name = 1;
}

message GreeterReply {
  string message = 1;
}

service Greeter{

  rpc Greet(GreeterRequest) returns (GreeterReply);

  rpc BiStream(stream GreeterRequest) returns (stream GreeterReply);

  rpc ServerStream(GreeterRequest) returns (stream GreeterReply);

}
```
2. Define server implement, add extra method with Mapping annotation
```
public class TriGreeterImpl extends DubboGreeterTriple.GreeterImplBase {
    private static final Logger LOGGER = LoggerFactory.getLogger(org.apache.dubbo.samples.tri.grpc.interop.server.TriGreeterImpl.class);

    public TriGreeterImpl() {
    }

    @Override
    public GreeterReply greet(GreeterRequest request) {
        LOGGER.info("Server received greet request {}", request);
        return GreeterReply.newBuilder()
                .setMessage("hello," + request.getName())
                .build();
    }
    // extra add method
    @Mapping("/test")
    public String test(){
       return "success pax";
    }

}
```
3. Run server and http invoke
method: Post
url: 127.0.0.1:50051/org.apache.dubbo.samples.tri.grpc.Greeter/test
response: 404

![Image](https://github.com/user-attachments/assets/b97fd976-e999-45ff-8cab-6e557a791493)
4. When use java interface + pojo, it will be success whith http invoke
```
@DubboService
public class DemoServiceImpl implements DemoService {

    @Override
    public String sayHello(String name) {
        return "Hello " + name;
    }

    @Mapping("/test")
    public String test(){
       return "hello pax";
    }
}
```

### What you expected to happen

Comment  triple proto whether support additional methods to modify with mapping annotations

### Are you willing to submit a pull request to fix on your own?

[...]

PR #15370, opened 2025-05-12
Support automatic REST route registration and invocation for custom @Mapping methods in Triple stub service implementation classes.

heliang666s:
Support automatic REST route registration and invocation for custom @Mapping methods in Triple stub service implementation classes.
#15366

## Checklist


oxsean:
I don’t think we need to support this feature. Since the service is defined by proto, we should focus only on the methods exported by the proto. The reason Java reflection supports this is to make it easier for Spring MVC users to migrate.

Issue #15320, opened 2025-04-14
[Feature] support configuration key contains hyphen from environment.


btpka3:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar feature requirement.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Descriptions

Dubbo  java client version : 3.3.3
I'm using triple protocol to  send  http request  to dubbo service.
I will provide multiple  this service with different group or version, so it's required to specify group and version via http header:

```shell
curl -s -XPOST "http://localhost:50051/com.my.Service/request" \
-H "Content-Type: application/json" \
-H "rest-service-group: DUBBO" \
-H "rest-service-version: 1.0.0" \
-d '["firstStringParam",{"k1":"v1","k2":"v2"}]'
````
I found it required to specify environment  `DUBBO_RPC_TRI_RESOLVE-FALLBACK-TO-DEFAULT` to `false`  to work, or else the http header "rest-service-group",  "rest-service-version" has no effect.  

While  hyphen is not allowed as environment key: 
```shell
$ export DUBBO_RPC_TRI_RESOLVE-FALLBACK-TO-DEFAULT=false
-bash: export: `DUBBO_RPC_TRI_RESOLVE-FALLBACK-TO-DEFAULT=flase': not a valid identifier
```
Some k8s service (such as Aliyun EDAS can set such env key), some others can't.

Request : please support  allow specify such dubbo configuration keys from  dubbo.properties  and  enviroment keys , add explain it in dubbo reference .

links : 
  
-  org.apache.dubbo.rpc.protocol.tri.rest.mapping.condition.ServiceGroupVersionCondition#[match](https://github.com/apache/dubbo/blob/dubbo-3.3.3/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/rest/mapping/condition/ServiceGroupVersionCondition.java#L44)
- org.apache.dubbo.rpc.protocol.tri.TripleProtocol#[RESOLVE_FALLBACK_TO_DEFAULT](https://github.com/apache/dubbo/blob/dubbo-3.3.3/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TripleProtocol.java#L84C9-L84C36)
[...]

PR #15382, opened 2025-05-13
feat: support environment variable keys with hyphens and dots for Dubbo configuration, fully compatible with legacy and Spring Boot styles

heliang666s:
adds support for configuring Dubbo keys containing hyphens and dots via environment variables, fully compatible with legacy Dubbo and Spring Boot naming conventions. The new logic allows users to set configuration using various environment variable formats (e.g., underscores, uppercase, with or without hyphens), making Dubbo configuration more flexible and user-friendly. Comprehensive unit tests are included to ensure correctness and compatibility.

#15320

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/config/EnvironmentConfiguration.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/config/EnvironmentConfiguration.java
@@ -29,52 +29,45 @@ import java.util.Set;
  */
 public class EnvironmentConfiguration implements Configuration {
 
-    private final Map<String, String> envMap;
-
     @Override
     public Object getInternalProperty(String key) {
         if (StringUtils.isEmpty(key)) {
             return null;
         }
-        String value = envMap.get(key);
+        String value = getenv().get(key);
         if (value != null) {
             return value;
         }
         for (String candidateKey : generateCandidateEnvironmentKeys(key)) {
-            value = envMap.get(candidateKey);
+            value = getenv().get(candidateKey);
             if (value != null) {
                 return value;
             }
         }
 
         String osStyleKey = StringUtils.toOSStyleKey(key);
-        value = envMap.get(osStyleKey);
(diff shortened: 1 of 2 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/config/EnvironmentConfiguration.java
```diff
@@ -16,32 +16,81 @@
...
+        return candidates;
+    }
+
     public Map<String, String> getProperties() {
         return getenv();
     }
 
     // Adapt to System api, design for unit test
+    public EnvironmentConfiguration() {
+        this.envMap = System.getenv();
```
songxiaosheng: Previously, it was effective in real-time. Now, do we need to retrieve it in real-time when placing it in the constructor

oxsean: Suggest retrieving it in real time too. In one of my environment isolation feature in other project before, it replace the `System.env`. If it’s changed this way, it could cause issues.

dubbo-common/src/main/java/org/apache/dubbo/common/config/EnvironmentConfiguration.java
```diff
@@ -16,32 +16,81 @@
...
+                return value;
+            }
         }
+
+        String osStyleKey = StringUtils.toOSStyleKey(key);
+        value = envMap.get(osStyleKey);
         return value;
     }
 
+    private Set<String> generateCandidateEnvironmentKeys(String originalKey) {
```
oxsean, line 53: Although it’s not critical, could we optimize this piece of code? There’s no need to precompute all the keys — we can generate one, test it, and add it to the tested records. Also, the replacement of DOT_SEPARATOR doesn’t need to be performed multiple times.

oxsean (changes requested):

Issue #15392, opened 2025-05-17
[Bug] Void method return null object.


RainYuY:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

3.3.4-SNAPSHOT

### Steps to reproduce this issue

define a rest method like 
`void post(); 
`use tri-rest. after post request, response body will return null.

![Image](https://github.com/user-attachments/assets/229178ed-8f6b-4fc6-a7bb-66d218f66511)

### What you expected to happen

If the result is null, nothing should be written to the response.

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

oxsean:
It would be better to return HTTP 204 (No Content) in this case.

PR #15393, opened 2025-05-17
If the result is null, nothing should be written to the response.

RainYuY:

#15392 

## Checklist


RainYuY:
@oxsean PTAL

finefuture:
You can process the output content when the result is null by customizing responseEncoder.

finefuture (commented):
You can process the output content when the result is null by customizing responseEncoder.

Stellar1999:
> customizing

When I use a method that returns void, I don't think an additional customer process is necessary.

RainYuY:
> You can process the output content when the result is null by customizing responseEncoder.

When I use a method that returns void, I don't think an additional customer process is necessary.

finefuture:
> > You can process the output content when the result is null by customizing responseEncoder.
> 
> When I use a method that returns void, I don't think an additional customer process is necessary.

For this case, 204 status code may be a better choice.

Issue #15365, opened 2025-05-09
[Bug] Application-level service discovery fails when IDL `package` and `java_package` differ


lqscript:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo Java 3.3.1
JDK 17
Nacos 2.5.1

### Steps to reproduce this issue

[Application-level service discovery fails when IDL `package` and `java_package` differ](https://github.com/lqscript/dubbo-samples/tree/master/1-basic/dubbo-samples-spring-boot-idl)

org.apache.dubbo.rpc.RpcException: No provider available from registry RegistryDirectory(registry: 127.0.0.1:8848)-Directory(invokers: 0[], validInvokers: 0[], invokersToReconnect: 0[]) for service org.apache.dubbo.springboot.demo.idl.Greeter on consumer 10.94.201.88 use dubbo version 3.3.1, please check status of providers(disabled, not registered or in blocklist).
	at org.apache.dubbo.registry.integration.DynamicDirectory.doList(DynamicDirectory.java:204) ~[dubbo-3.3.1.jar:3.3.1]
	at org.apache.dubbo.rpc.cluster.directory.AbstractDirectory.list(AbstractDirectory.java:232) ~[dubbo-3.3.1.jar:3.3.1]
	at org.apache.dubbo.rpc.cluster.support.AbstractClusterInvoker.list(AbstractClusterInvoker.java:452) ~[dubbo-3.3.1.jar:3.3.1]
	at org.apache.dubbo.rpc.cluster.support.AbstractClusterInvoker.invoke(AbstractClusterInvoker.java:355) ~[dubbo-3.3.1.jar:3.3.1]
	at org.apache.dubbo.rpc.cluster.router.RouterSnapshotFilter.invoke(RouterSnapshotFilter.java:46) ~[dubbo-3.3.1.jar:3.3.1]
	at org.apache.dubbo.rpc.cluster.filter.FilterChainBuilder$CopyOfFilterChainNode.invoke(FilterChainBuilder.java:349) ~[dubbo-3.3.1.jar:3.3.1]
	at org.apache.dubbo.monitor.support.MonitorFilter.invoke(MonitorFilter.java:109) ~[dubbo-3.3.1.jar:3.3.1]
[...]

PR #15449, opened 2025-06-11
fix: For idl stub service, use idl service name for discovery (#15365)

lqscript:
fix: For idl stub service, use idl service name for discovery (#15365)

## Checklist

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ReferenceConfig.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/ReferenceConfig.java
@@ -349,10 +349,6 @@ public class ReferenceConfig<T> extends ReferenceConfigBase<T> {
                 initServiceMetadata(consumer);
 
                 serviceMetadata.setServiceType(getServiceInterfaceClass());
-                // TODO, uncomment this line once service key is unified
-                serviceMetadata.generateServiceKey();
-
-                Map<String, String> referenceParameters = appendConfig();
 
                 ModuleServiceRepository repository = getScopeModel().getServiceRepository();
                 ServiceDescriptor serviceDescriptor;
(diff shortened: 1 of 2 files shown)


zrlw:
could you submit a new pr at https://github.com/apache/dubbo-samples to add a new sample module  to confirm this pr?

lqscript:
> could you submit a new pr at https://github.com/apache/dubbo-samples to add a new sample module to confirm this pr?

https://github.com/apache/dubbo-samples/pull/1220

zrlw:
> > could you submit a new pr at https://github.com/apache/dubbo-samples to add a new sample module to confirm this pr?
> 
> [apache/dubbo-samples#1220](https://github.com/apache/dubbo-samples/pull/1220)

i will test it.

zrlw:
@heliang666s @oxsean PTAL
i tested https://github.com/apache/dubbo-samples/pull/1220 with this pr and got successful result.

zrlw:
version compatibility should also be considered.
|client|server|
|----|----|
|old | new |
|new | old |

lqscript:
> version compatibility should also be considered.
> 
> client	server
> old	new
> new	old

Given issue #15365, this incompatibility should only occur in interface-level service discovery scenarios where package and java_package differ. And when users want to migrate to application-level service discovery, they will also encounter this compatibility issue.

Based on the official documentation ([official documentation](https://cn.dubbo.apache.org/en/overview/mannual/java-sdk/tasks/protocols/triple/idl/#:~:text=The%20complete%20service%20name%20is%20greet.Greeter%2C%20which%20will%20be%20used%20during%20rpc%20calls%20and%20service%20discovery.)), it clearly states that package should be used for service discovery, which is why I believe this should be treated as a bug fix.

So maybe I should only change this in application-level service discovery, but leave the interface-level service discovery as it is to maintain compatibility with existing clients.

Since I'm relatively new to Dubbo, what would you recommend as the most appropriate way to handle this compatibility issue while still fixing the underlying problem?

zrlw:
In my opinion, It is best to provide users with an option, just like ```dubbo.registry.register-mode```, users could choose ```instance```(only new), ```interface```(only old) or ```all```(export both instance and interface for compatibility) according to their own situation.

lqscript:
I looked at the source code, and if i make this change, it would require modifications in many places...
I think if incompatibility issues are actually encountered, changing the package name would be a simpler workaround...

lqscript:
I found that the serviceKey is not only used for service discovery, but also for routing and circuit breaking. Therefore, using the `all` approach may not be appropriate, as some user configurations would also need to be updated accordingly.

How about adding a boolean value to indicate whether to use the IDL package name as the serviceKey? This way, existing services do not need to be changed, and new services can use the IDL package name as the serviceKey.

Issue #15465, opened 2025-06-17
[Bug] If a StreamObserver object calls onCompleted() without first calling onNext(), it will not necessarily throw an error


RainYuY:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

3.x

### Steps to reproduce this issue

If a StreamObserver object calls onCompleted() without first calling onNext(), it will not necessarily throw an error 

### What you expected to happen

close the connection.

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #15466, opened 2025-06-17
Fix StreamObserver object calls onCompleted() without first calling onNext() will fail.

RainYuY:

#15465 

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/http1/Http1SseServerChannelObserver.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/http1/Http1SseServerChannelObserver.java
@@ -36,6 +36,14 @@ public class Http1SseServerChannelObserver extends Http1ServerChannelObserver {
     }
 
     @Override
+    protected void doOnCompleted(Throwable throwable) {
+        if (!isHeaderSent()) {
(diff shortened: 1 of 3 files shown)


RainYuY:
PTAL @oxsean

oxsean (changes requested):
Http2SseServerChannelObserver should be update too.

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/http1/Http1SseServerChannelObserver.java
```diff
@@ -35,6 +35,12 @@ public void setResponseEncoder(HttpMessageEncoder responseEncoder) {
         super.setResponseEncoder(new ServerSentEventEncoder(responseEncoder));
     }
 
+    @Override
+    protected void doOnCompleted(Throwable throwable) {
+        sendMetadata(encodeHttpMetadata(true));
```
oxsean: Your previous approach was correct — a check is needed here to avoid an unnesssory encoding of Metadata

RainYuY: You're right.

RainYuY:
> Http2SseServerChannelObserver should be update too.

A little diff.PTAL @oxsean

oxsean (approved):

PR #15482, opened 2025-06-23
[GSoC 2025] Added event labels to Service Discovery logs

ankitshokeen:
https://github.com/apache/dubbo/issues/15212

Added event labels to Service Discovery logs using log4j2 for debugging.


ankitshokeen:
PTAL @AlbumenJ

PR #15486, opened 2025-06-25
[3.3] Declare clientSettings and tokenSettings as Object type to avoid compilation error with jdk8 or jdk11

zrlw:
1. the class file version of ClientSettings and TokenSettings is 61.0 (compiled by jdk17), it will cause compilation error when building dubbo using jdk8 or jdk11.
2. ```spring-security-oauth2-authorization-server``` is only used in the test cases,  depending on ```spring-security-oauth2-core``` is enough.

## Checklist


zrlw:
@wzkris

Issue #15501, opened 2025-06-29
[Bug] Demo dubbo-samples-triple-http3 consumer print org.apache.dubbo.rpc.StatusRpcException: CANCELLED


Joker-zc:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java Samples (apache/dubbo-samples)

### Dubbo Version

dubbo 3.3.1 jdk 17 windows/mac/centos7.9

### Steps to reproduce this issue

modify demo dubbo-samples-triple-http3  GreeterServiceImpl to this

`    @Override
    public HelloReply sayHello(HelloRequest request) {
        LOGGER.info("Received sayHello request: {}", request.getName());
        StringBuffer finalString = new StringBuffer();
        IntStream.range(0,10000).forEach(
              i->  finalString.append("Hello")
        );
        return toReply(finalString + request.getName());
    }
`

then run test
`   @Test
    public void sayHello() {
        HelloReply reply = greeterService.sayHello(buildRequest("world"));
        Assert.assertEquals("Hello world", reply.getMessage());
    }`

```then test log
 [fixed-42d43ca6-be10-4210-837d-7b50aadbfa45-172.16.57.34_8848] [subscribe] dubbo-springboot-triple-http3.MESHAPPRULE+dubbo+42d43ca6-be10-4210-837d-7b50aadbfa45
18:37:33.598 |-INFO  [main] ibaba.nacos.client.config.impl.CacheData:236 -| [fixed-42d43ca6-be10-4210-837d-7b50aadbfa45-172.16.57.34_8848] [add-listener] ok, tenant=42d43ca6-be10-4210-837d-7b50aadbfa45, dataId=dubbo-springboot-triple-http3.MESHAPPRULE, group=dubbo, cnt=1
18:37:33.598 |-INFO  [main] ba.nacos.client.config.impl.ClientWorker:418 -| [fixed-42d43ca6-be10-4210-837d-7b50aadbfa45-172.16.57.34_8848] [subscribe] dubbo-springboot-triple-http3.tag-router+dubbo+42d43ca6-be10-4210-837d-7b50aadbfa45
[...]

PR #15518, opened 2025-07-01
[3.3] Close channel handler context after the channel written operation is completed and Remove overflow check process to align with Netty http2 remote flow controller

zrlw:
#15501 

1. Close channel handler context after the channel written operation is completed.
gRPC or HTTP/2 sender should not close stream immediately after sending http output message or trailers, otherwise the receiver will encounter troubles during receiving packets due to ‌premature stream closure.
see details at https://github.com/apache/dubbo/issues/15520

2. Remove overflow check process to align with Netty http2 remote flow controller, refer to [DefaultHttp2RemoteFlowController](https://github.com/netty/netty/blob/4.2/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java)

if http2 sender throws StreamException when the window size of connectionState is zero:
dd8a27ae29eb78e4755824889c574115
the receiver might get ResourceAccessException caused by closed IOException during receiving response :
image

## Checklist

### DIFF
--- a/dubbo-remoting/dubbo-remoting-http3/src/main/java/org/apache/dubbo/remoting/http3/netty4/NettyHttp3FrameCodec.java
+++ b/dubbo-remoting/dubbo-remoting-http3/src/main/java/org/apache/dubbo/remoting/http3/netty4/NettyHttp3FrameCodec.java
@@ -97,36 +97,37 @@ public class NettyHttp3FrameCodec extends Http3RequestStreamInboundHandler imple
     public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
         if (msg instanceof Http2Header) {
             Http2Header headers = (Http2Header) msg;
-            ctx.write(
+            ChannelFuture future = ctx.write(
                     new DefaultHttp3HeadersFrame(((NettyHttpHeaders<Http3Headers>) headers.headers()).getHeaders()),
                     promise);
             if (headers.isEndStream()) {
-                if (promise.isDone()) {
+                if (future.isDone()) {
                     ctx.close();
                 } else {
-                    promise.addListener((ChannelFutureListener) future -> ctx.close());
+                    future.addListener((ChannelFutureListener) f -> ctx.close());
                 }
             }
         } else if (msg instanceof Http2OutputMessage) {
             Http2OutputMessage message = (Http2OutputMessage) msg;
+            ChannelFuture future = null;
             try {
                 OutputStream body = message.getBody();
                 if (body == null) {
                     Http3DataFrame frame = new DefaultHttp3DataFrame(Unpooled.EMPTY_BUFFER);
-                    ctx.write(frame, promise);
+                    future = ctx.write(frame, promise);
                     return;
                 }
                 if (body instanceof ByteBufOutputStream) {
                     Http3DataFrame frame = new DefaultHttp3DataFrame(((ByteBufOutputStream) body).buffer());
-                    ctx.write(frame, promise);
+                    future = ctx.write(frame, promise);
                     return;
                 }
             } finally {
                 if (message.isEndStream()) {
-                    if (promise.isDone()) {
+                    if (future == null || future.isDone()) {
                         ctx.close();
                     } else {
-                        promise.addListener((ChannelFutureListener) future -> ctx.close());
+                        future.addListener((ChannelFutureListener) f -> ctx.close());
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


dubbo-remoting/dubbo-remoting-http3/src/main/java/org/apache/dubbo/remoting/http3/netty4/NettyHttp3FrameCodec.java
```diff
@@ -95,7 +96,11 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
                     new DefaultHttp3HeadersFrame(((NettyHttpHeaders<Http3Headers>) headers.headers()).getHeaders()),
                     promise);
             if (headers.isEndStream()) {
-                ctx.close();
+                if (promise.isDone()) {
```
oxsean: When writing on the client side, does it also need to be handled this way?

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TriHttp2RemoteFlowController.java
```diff
@@ -778,11 +778,7 @@ private void checkConnectionThenStreamWritabilityChanged(FlowState state) throws
             } else if (isWritable(state) != state.markedWritability()) {
                 notifyWritabilityChanged(state);
             } else if (isOverFlowControl()) {
-                throw streamError(
-                        state.stream().id(),
-                        FLOW_CONTROL_ERROR,
-                        "TotalPendingBytes size overflow for stream: %d",
-                        state.stream().id());
+                logger.warn("TotalPendingBytes size overflow for stream: " + state.stream().id());
```
oxsean: I don't know much about flow control. Will this change have any negative impact?

dubbo-remoting/dubbo-remoting-http3/src/main/java/org/apache/dubbo/remoting/http3/netty4/NettyHttp3FrameCodec.java
zrlw: refer to io.netty.handler.codec.http2.Http2ConnectionHandler:
if the close operation causes the stream to be closed, it will add a hook to close the channel after the given future completes whether executing closeStreamLocal or closeStreamRemote method.
```
    @Override
    public void closeStream(final Http2Stream stream, ChannelFuture future) {
        if (future.isDone()) {
            doCloseStream(stream, future);
        } else {
            future.addListener(new ChannelFutureListener() {
                @Override
                public void operationComplete(ChannelFuture future) {
                    doCloseStream(stream, future);
                }
            });
        }
    }
```

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TriHttp2RemoteFlowController.java
zrlw: at server side, it should pause or slow down sending when overflow occurs, but it should not cancel or throw exception to break down the sending process.

dubbo-remoting/dubbo-remoting-http3/src/main/java/org/apache/dubbo/remoting/http3/netty4/NettyHttp3FrameCodec.java
zrlw: As for me, the client might close the context immediately after it determined that the FIN packet has been received,
the client should wait sending operation completion before doing stream close operation if it just sends ONE-WAY messages and does not get any stream closed notifications from the server, i might miss these codes because i am not familiar with h3 client processing.

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TriHttp2RemoteFlowController.java
RainYuY: Maybe log error will be better?

zrlw: When an overflow occurs, it is highly likely that the duration will not be very short, so it is better to follow the approach of Netty DefaultHttp2RemoteFlowController: remove these redundant codes completely.

dubbo-remoting/dubbo-remoting-http3/src/main/java/org/apache/dubbo/remoting/http3/netty4/NettyHttp3FrameCodec.java
zrlw: i checked codes and made same revision to the client side.

PR #15522, opened 2025-07-05
Remove System.out from the unit test cases.

little-kid-an:
Remove System.out from the unit test cases.

## What is the purpose of the change?
Closes #15499 

## Checklist
- [x] Make sure there is a [GitHub_issue](https://github.com/apache/dubbo/issues) field for the change.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Write necessary unit-test to verify your logic correction. If the new feature or significant change is committed, please remember to add sample in [dubbo samples](https://github.com/apache/dubbo-samples) project.
- [x] Make sure gitHub actions can pass. [Why the workflow is failing and how to fix it?](../CONTRIBUTING.md)

PR #15523, opened 2025-07-05
Remove System.out from the unit test cases.

little-kid-an:
Remove System.out from the unit test cases.

Close #15522 

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/rest/GeneralTypeConverterTest.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/rest/GeneralTypeConverterTest.java
@@ -23,6 +23,7 @@ import org.apache.dubbo.rpc.protocol.tri.rest.util.TypeUtils;
 
 import java.util.List;
 
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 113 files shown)


zrlw:
The reason of failed unit tests could be found by downloading the test log archive.

download-log-archive

little-kid-an:
> The reason of failed unit tests could be found by downloading the test log archive.
> 
> Screenshot_20250705_211450_com easy abroad_edit_580443459520803
> 

I've downloaded the log, but I'm not in front of the computer right now. I need to fix it tomorrow

dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/rest/GeneralTypeConverterTest.java
```diff
@@ -39,13 +42,13 @@ void convert() throws NoSuchFieldException {
     @Test
     void convert1() {
         Object convert = JsonUtils.toJavaObject("[1,\"aa\"]", List.class);
-        System.out.println(convert);
+        logger.info((String) convert);
```
zrlw: java.util.ArrayList cannot be cast to class java.lang.String.
list == null ? logger.info("empty list") : list.forEach(item -> logger.info(item));
but It's not worth doing this, you'd better delete all ```System.out.println``` statements and replace them by Assertions just like ```Assertions.assertTrue(((List) convert).size() == 2);```

little-kid-an: OK, I will check and revise them.

Issue #15024, opened 2024-12-26
[Feature] Graceful Shutdown Requirements for HTTP/3 Server Side


oxsean:
### Pre-check

- [X] I am sure that all the content I provide is in English.

### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar feature requirement.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Descriptions

**Objective:**
Implement a robust graceful shutdown mechanism for the HTTP/3 server connection handler using Netty's HTTP/3 incubator codec, focusing on leveraging the control stream for ping-pong and graceful shutdown (GOAWAY) logic.

Please refer to org.apache.dubbo.rpc.protocol.tri.transport.TripleServerConnectionHandler.

**Key Requirements:**

1. **Control Stream Utilization**
   - Utilize `io.netty.incubator.codec.http3.Http3.getLocalControlStream()` as the primary mechanism for managing connection state and initiating graceful shutdown.
   - Implement a bidirectional ping-pong mechanism to verify connection health and readiness for shutdown.

2. **Ping-Pong Mechanism**
   - Develop a lightweight ping-pong protocol on the control stream to:
     a) Validate active connections
     b) Detect connection responsiveness
     c) Prepare for graceful termination
   - Create a configurable ping interval and timeout strategy
   - Handle scenarios where ping responses are delayed or missing

3. **Graceful Shutdown (GOAWAY) Workflow**
   - Design a GOAWAY process that:
     a) Stops accepting new requests
     b) Allows existing in-flight requests to complete
     c) Provides a configurable grace period for request completion
   - Implement a systematic shutdown sequence using control stream signaling
   - Ensure all active streams are properly closed or terminated

4. **Error Handling and Resilience**
   - Implement comprehensive error handling for control stream operations
   - Provide logging and metrics for shutdown-related events
   - Handle potential network interruptions during the shutdown process

### Are you willing to submit a pull request to fix on your own?

- [X] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

aofall:
I'm interested with it, it's anyone work in progress?

PR #15529, opened 2025-07-08
Fire ping ack event when receiving `TRI_PING` Http3 Headers frame

zrlw:
```
private void pingAck(ChannelHandlerContext ctx) {
        ChannelPipeline pipeline = ctx.channel().parent().pipeline();
        pipeline.fireChannelRead(new DefaultHttp2PingFrame(0, true));
        pipeline.fireChannelReadComplete();
```
might break Netty event handling process, it might be better to use event notification to do the same thing.

Let Netty manage ```fireChannelReadComplete()``` as part of its internal event flow. Application should focus on implementing business logic in overridden ```channelReadComplete()``` handlers rather than manually triggering the event

In Netty, fireChannelReadComplete() should not be invoked directly due to its critical role in the framework's event processing pipeline:
1.‌ Event Propagation Mechanism‌
This method is designed to notify the ChannelPipeline when a read operation completes, automatically triggered by Netty's internal NioEventLoop after socket buffer processing. Direct invocation disrupts the event sequence and may cause data handling inconsistencies.
2. Coordination with channelRead()‌
channelRead(): Processes fully decoded application-layer messages (e.g., after protocol decoding). 
channelReadComplete(): Signals the end of a socket read cycle, potentially called multiple times due to packet fragmentation.
Manual calls might bypass essential message processing steps in channelRead()
3. Risks of Direct Invocation‌
‌Duplicate Processing‌: May trigger redundant business logic executions.
‌Resource Leaks‌: Could lead to buffer reference count mismanagement if not handled properly.

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TriplePingPongHandler.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TriplePingPongHandler.java
@@ -48,16 +48,31 @@ public class TriplePingPongHandler extends ChannelDuplexHandler {
 
     @Override
     public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+        if (evt instanceof PingAckEvent) {
+            if (pingAckTimeoutFuture != null) {
+                pingAckTimeoutFuture.cancel(true);
+                pingAckTimeoutFuture = null;
+            }
+            return;
+        }
         if (!(evt instanceof IdleStateEvent)) {
             ctx.fireUserEventTriggered(evt);
             return;
         }
-        ctx.writeAndFlush(new DefaultHttp2PingFrame(0));
         if (pingAckTimeoutFuture == null) {
             pingAckTimeoutFuture =
                     ctx.executor().schedule(new CloseChannelTask(ctx), pingAckTimeout, TimeUnit.MILLISECONDS);
         }
-        // not null means last ping ack not received
+        // pingAckTimeoutFuture not null means last ping ack not received
+        ctx.writeAndFlush(new DefaultHttp2PingFrame(0));
+    }
+
+    public static class PingAckEvent {
+
+        @SuppressWarnings("InstantiationOfUtilityClass")
+        public static final PingAckEvent INSTANCE = new PingAckEvent();
+
+        private PingAckEvent() {}
     }
 
     private static class CloseChannelTask implements Runnable {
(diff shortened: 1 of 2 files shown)


dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/TriplePingPongHandler.java
```diff
@@ -29,21 +30,28 @@ public class TriplePingPongHandler extends ChannelDuplexHandler {
 
     private final long pingAckTimeout;
 
-    private ScheduledFuture<?> pingAckTimeoutFuture;
+    private final AtomicReference<ScheduledFuture<?>> pingAckTimeoutFutureRef = new AtomicReference<>();
```
RainYuY: Why do we need to use AtomicReference? Doesn't Netty's single-threaded model guarantee that it won't be accessed concurrently?

zrlw: do you mean there is no scenario that two threads are running Http3ClientFrameCodec#channelRead (Triple Ping Headers frame) and TriplePingPongHandler#channelRead (HTTP/2 ping frame) concurrently?
i don't know how to trigger HTTP/2 ping frame, could you test it?

zrlw: the other reason is  the IdleStateEvent is triggered by ReaderIdleTimeoutTask of IdleStateHandler, the userEventTriggered  method of TriplePingPongHandler will be called by the task thread.

RainYuY: Yes, even if two threads use two separate connections, they should still be different objects. As far as I remember, this frame is only sent when the service goes offline. I can test it later.

zrlw: as for me, the Netty single-threaded model is: 
1. dealing connection by single reactor thread.
2. dealing logical process by multiple threads.

oxsean:
I’d prefer not change anything for now unless we run into issues. The way by me handle heartbeats isn’t great anyway, so maybe we can refactor it later.
https://github.com/apache/dubbo/issues/15024

zrlw:
> I’d prefer not change anything for now unless we run into issues. The way by me handle heartbeats isn’t great anyway, so maybe we can refactor it later. #15024

agree. it seemed that cross pipeline calling doesn't look elegant.

PR #15558, opened 2025-07-18
Fixed the issue of type loss in generalized call cases.

wuwen5:

The serialization module supports the Java8 date type, so here we should directly return the java date type to avoid type loss.

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/PojoUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/PojoUtils.java
@@ -160,7 +160,7 @@ public class PojoUtils {
         }
 
         if (pojo instanceof LocalDate || pojo instanceof LocalDateTime || pojo instanceof LocalTime) {
-            return pojo.toString();
+            return pojo;
         }
 
         if (pojo instanceof Class) {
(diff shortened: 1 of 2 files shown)


zrlw:
current branch is 3.3

oxsean (approved):
.

oxsean (changes requested):
Generalize means convert an object into a basic (primitive) type, you can't return the object directly.

wuwen5:
> Generalize means convert an object into a basic (primitive) type, you can't return the object directly.

@oxsean 
Simple types (including primitive types, `String`, `Date`, `Number`, and `Collections` thereof) 
Therefore, Java 8 time types should also be included as simple types in the processing logic.

@zrlw  `3.2` is still under maintenance, and I expect it to be released in version `3.2`, or I can also initiate modifications from the 3.3 branch #15562

zrlw:
> 3.2 is still under maintenance

As for me, it's not under maintenance, there are not  PRs had been backported  to 3.2 since 3.3 was released.

oxsean:
@wuwen5 
I disagree with you on this causal inference. Dubbo initially treated Date as a primitive type, but that doesn't mean its derived types should also be treated as primitives. Such a change is entirely incompatible. The user may rely on this output for logical processing, so this change would be a break change, and if you really want to do that, you'd have to at least add a feature toggle.

Issue #15267, opened 2025-03-23
[Bug] The dubbo-admin service uses nacos as the registration center. Due to frequent GC issues, it starts slowly or fails to start.


wcwssf:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo:3.2.16 
JVM: JDK 1.8
Server: Ubuntu 9.4.0-1ubuntu1~20.04.2

### Steps to reproduce this issue

When dubbo-admin uses nacos as the registration center, the registration center notifySubscriber process is very slow and consumes a lot of memory. Because of all the service info will be related to every single service , so that every single service changed , the Subscriber will process all the service, and cache file store the all related service info.

Step1: use nacos as the registration center
Step2: nacos need manage above 2000+ services
Step3: Dubbo admin server set JVM Xmx less than 6G

### What you expected to happen

We hope that when using nacos as the registration center, dubbo-admin startup is the same as other registration centers, and service information mapping is accurate

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zrlw:
dubbo-admin might be migrated to  https://github.com/apache/dubbo-kubernetes

zrlw:
https://github.com/apache/dubbo/issues/14987
```we have fixed this issue on our development branch and want to merge it into the main repository```

still in progress?

zrlw:
close it since the author hasn't responded for a long time.

wcwssf:
I will submit a MR later

PR #15561, opened 2025-07-19
NacosAggregateListener aggregates all instances into each service if dubbo-admin uses nacos as the registration center

wcwssf:
in dubbo-admin case , NacosAggregateListener will aggregate all instances into each service. because every instances in NacosAggregateListener will be related to the consumerURL URL of each service when build DubboServiceAddressURL from instance and DubboServiceAddressURL#getServiceInterface is realized as  ## return consumerURL.getServiceInterface() ##

PR of the issue : https://github.com/apache/dubbo/issues/15267

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/constants/RegistryConstants.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/constants/RegistryConstants.java
@@ -149,4 +149,9 @@ public interface RegistryConstants {
      * for compatible, we should export noting suffix servicename, eg: ${interface}:${version}
      */
     String NACOE_REGISTER_COMPATIBLE = "nacos.register-compatible";
+
+    /**
+     * protocol of admin
+     */
+    String ADMIN_PROTOCOL = "admin";
 }
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/url/component/ServiceAddressURL.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/url/component/ServiceAddressURL.java
@@ -29,9 +29,9 @@ import static org.apache.dubbo.common.constants.CommonConstants.CONSUMER_SIDE;
 import static org.apache.dubbo.common.constants.CommonConstants.GROUP_KEY;
 import static org.apache.dubbo.common.constants.CommonConstants.SIDE_KEY;
 import static org.apache.dubbo.common.constants.CommonConstants.VERSION_KEY;
+import static org.apache.dubbo.common.constants.RegistryConstants.ADMIN_PROTOCOL;
 import static org.apache.dubbo.common.constants.RegistryConstants.CATEGORY_KEY;
 import static org.apache.dubbo.common.constants.RegistryConstants.PROVIDERS_CATEGORY;
-import static org.apache.dubbo.common.constants.RegistryConstants.ADMIN_PROTOCOL;
 
 public abstract class ServiceAddressURL extends URL {
 
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/url/component/ServiceAddressURL.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/url/component/ServiceAddressURL.java
@@ -23,6 +23,7 @@ import org.apache.dubbo.rpc.model.ServiceModel;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 
 import static org.apache.dubbo.common.constants.CommonConstants.APPLICATION_KEY;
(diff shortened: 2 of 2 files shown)


zrlw:
Run `mvn spotless:apply`with jdk17+ to fix code format

zrlw:
if ```DubboServiceAddressURL#getServiceInterface``` has problems, why not add ```@Override``` method to it?

zrlw:
The content you expressed is a little difficult to understand, i try to explain it based on my own understanding.

> NacosAggregateListener will aggregate all instances into each service.

NacosAggregateListener aggregates instances to service instances map which grouped by service name.

> every instances in NacosAggregateListener will be related to the consumerURL URL of each service when build DubboServiceAddressURL from instance and DubboServiceAddressURL#getServiceInterface is realized as ## return consumerURL.getServiceInterface() ##

1. Every service instance that NacosAggregateListener aggregated is related to the consumerURL of the service.
2. Every DubboServiceAddressURL was built from the service instance.
3. ```DubboServiceAddressURL#getServiceInterface``` method took  the result of ```consumerURL#getServiceInterface``` as it's return value.

> every single service changed , the Subscriber will process all the service, and cache file store the all related service info.  #15267

The subscriber will be notified with all service info even just one service instance status is changed, e.g., 
there are 3 services which names are A,B and C,  let's say A has 100 instances, B has 200 instances, C has 300 instances, the subscribe will be notified with all A,B,C service instances info(100+200+300)  even just one instance of service A or B or C changed.

so you want change the result of ```DubboServiceAddressURL#getServiceInterface``` to limit the notification scope of dubbo service instance changing event?

zrlw:
you'd better provide test comparison report to show the effectiveness of this PR.

wcwssf:
image
image

wcwssf:
image

image

wcwssf:
image

wcwssf:
so the `notified` instance will be very big ,and if we use local file cache , it will wirte all the content to file when every time `notify` method invoked.

wcwssf:
in our test env, we have 1500+ service ,and 2200 + instance info. Dubbo-admin need 4G to save the URL info .it need more in pro env,the dubbo-admin start very slow.

zrlw:
> in our test env, we have 1500+ service ,and 2200 + instance info. Dubbo-admin need 4G to save the URL info .it need more ,the dubbo-admin start very slow.

i see.

zrlw (approved):

zrlw:
i doubt it's not only ```dubbo-admin``` that has this problem, ```getServiceInterface``` might need to get service interface from super when consumerURL.getParameter(SIDE_KEY) is ```PROVIDER_SIDE```

zrlw:
if the issue only existed in nacos, maybe the simplest way is using zk as your register center because it's difficult to determine this pr might affect existed applications or not. 
According to other member saying, more than one big issues occurred due to nacos getServiceInterface codes modification in the history.  He prefer you get atomic object directly and deal it at dubbo-admin.

wcwssf:
We have old services using old versions of dubbo (zk 3.4), springboot, and JDK, and some new services using higher versions of dubbo, Springboot, and JDK. So the solution we use is dual registration of zk and nacos, and all services will be registered with nacos. After all old services are upgraded, we can use the higher version of zk as a registration center, but this process will be longer.

wcwssf:
You are right. Directly modifying the ServiceAddress URL will have a wide impact. This problem is mainly caused by the dubbo-admin scenario and nacos' compatibility with the old service format. Do we have plans to remove nacos' support for the old service format?

zrlw:
> You are right. Directly modifying the ServiceAddress URL will have a wide impact. This problem is mainly caused by the dubbo-admin scenario and nacos' compatibility with the old service format. Do we have plans to remove nacos' support for the old service format?

i don't know. As for me, you might customize‌ these codes to meet your requirements. it's not easy to remove or modify any existed algorithm of community edition due to compatibility reason.

wcwssf:
> i doubt it's not only `dubbo-admin` that has this problem, `getServiceInterface` might need to get service interface from super when consumerURL.getParameter(SIDE_KEY) is `PROVIDER_SIDE`

This PR is limited to the admin protocol and will not affect other features.

zrlw (changes requested):
it seemed that we should find a more elegant way to solve this problem.

wcwssf:
What do you plan to do? In the dubbo-admin scenario, create a separate NacosAggregateListener object instance for each service?

zrlw:
> In the dubbo-admin scenario, create a separate NacosAggregateListener object instance for each service?

As for me, it might be better to customerize your dubbo-admin. By now, there are no plan to modify Dubbo for this scenario.

PR #15562, opened 2025-07-19
Fixed the issue of type loss in generalized call cases.

wuwen5:
## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/PojoUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/PojoUtils.java
@@ -160,7 +160,7 @@ public class PojoUtils {
         }
 
         if (pojo instanceof LocalDate || pojo instanceof LocalDateTime || pojo instanceof LocalTime) {
-            return pojo.toString();
+            return pojo;
         }
 
         if (pojo instanceof Class) {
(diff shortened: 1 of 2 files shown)


zrlw:
As for me, you'd better add LocalDate,LocalDateTime,LocalTime to ReflectUtils.isPrimitive, just like ```Date```
```
    public static boolean isPrimitive(Class<?> cls) {
        return cls.isPrimitive()
                || cls == String.class
                || cls == Boolean.class
                || cls == Character.class
                || Number.class.isAssignableFrom(cls)
                || Date.class.isAssignableFrom(cls);
    }
```
@wuwen5

zrlw:
another problem: performance regression might be introduced because current dubbo 3.3 take back hessian as it's default serialization protocol that will call ```Class.forName``` for each LocalDate,LocalDateTime,LocalTime deserialization.
see details at
https://github.com/apache/dubbo-hessian-lite/blob/master/hessian-lite/src/main/java/com/alibaba/com/caucho/hessian/io/java8/LocalDateHandle.java

https://blog.csdn.net/chuohuanghan7401/article/details/101053165

zrlw:
@wuwen5 hessian-lite might be optimized before applying this PR.
see details at  #15572

wuwen5:
In fact, this is not a conflict, it is just a matter of generalized call scenarios.
The performance concern with `LocalDateTime` is independent of this and still exists regardless of this PR. It is part of a broader optimization concern within the serialization module.

zrlw:
> In fact, this is not a conflict, it is just a matter of generalized call scenarios. The performance concern with `LocalDateTime` is independent of this and still exists regardless of this PR. It is part of a broader optimization concern within the serialization module.

not only performance, 
```there are existed alibaba codes which deal LocalDate as String type with hessian deserialization, this PR will make them in trouble.``` said by other member.

PR #15564, opened 2025-07-19
set the opt 'nacos.check' default val is 'false'

### LINKED ISSUE #15271: [Feature] Refactor Registry Implement
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Descriptions

1. When the **Registry** is wrapped and created by **ListenerRegistryWrapper**, there is a possibility that the **Registry** is null (due to connection failure), which may result in a **NullPointerException (NPE)** in subsequent operations.  
   Should we consider handling this error during initialization? If the **Registry** is null, an error can be thrown, allowing the upper-layer caller to handle it accordingly.  

2. When obtaining the **ProviderUrl** and the registration mode is **ALL** (default value), a **ServiceDiscoveryRegistry URL** will be generated synchronously (**org.apache.dubbo.config.utils.ConfigValidationUtils#genCompatibleRegistries**).  
   However, if multiple registries exist and one of them goes down, the **ServiceDiscoveryRegistry** fails to generate, causing the startup process to fail.  

   Should we consider:  
   - 1. Implementing functionality in **ServiceDiscoveryRegistry** similar to **MultipleRegistry**, supporting **check**.  
   - 2. Lowering the dependency level of **ServiceDiscoveryRegistry**, making it an attribute of **Registry**, and allowing other registries to handle its operations.  

### Related issues

#15003 

### Code of Conduct

fantiq:

resolve #15271 

set the `nacos.check` option default value is `false`, so that the NPE will be resolved. this nacos client will reconnect to the server when the server is ok.

## Checklist

### DIFF
--- a/dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/util/NacosNamingServiceUtils.java
+++ b/dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/util/NacosNamingServiceUtils.java
@@ -111,7 +111,7 @@ public class NacosNamingServiceUtils {
      * @since 2.7.5
      */
     public static NacosNamingServiceWrapper createNamingService(URL connectionURL) {
-        boolean check = connectionURL.getParameter(NACOS_CHECK_KEY, true);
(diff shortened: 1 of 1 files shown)


zrlw:
it's breaking change, both existed application and test cases are affected.

Issue #15557, opened 2025-07-18
attachment parameter transparency behavior


imjyz:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

<img width="1597" height="257" alt="dubbo-attachment-issues" src="https://github.com/user-attachments/assets/6cb7d312-9817-470f-9341-0a8632e5e2dc" />

The documentation claims that Dubbo 3 has fixed the attachment parameter transparency issue, but the code still silently propagates these parameters. The root cause is in ConsumerContextFilter, where supportedSelectors defaults to an empty set.

```java
if (CollectionUtils.isNotEmpty(supportedSelectors)) {
    xxxxx
} else {
    ((RpcInvocation) invocation)
            .addObjectAttachments(RpcContext.getServerAttachment().getObjectAttachments());
}
```

### Steps to reproduce this issue

Prepare three applications with the following relationships:
A ---> B ---> C

A-->B: RpcContext.getClientAttachment().setAttachment("jyz", "pass-through");

### What you expected to happen

The attachment received by C should not include `{"jyz":"pass-through"}`.

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

zrlw:
```
The attachment received by C includes {"jyz":"pass-through"}.
```
Do you expect to happen this??

imjyz:
> Do you expect to happen this??

@zrlw Yes, this is my test result.

zrlw:
> > Do you expect to happen this??
> 
> [@zrlw](https://github.com/zrlw) Yes, this is my test result.

if all the result are your expected, why you submit it as a issue?

imjyz:
Could you take a look at the description in my issue?

> https://cn.dubbo.apache.org/en/overview/mannual/java-sdk/tasks/framework/attachment/#reading-implicit-parameters-1
[...]

PR #15580, opened 2025-07-23
3.2 fix: attachment parameter transparency (#15557)

imjyz:

fix: attachment parameter transparency issue #15557


imjyz:
rebase 3.3

Issue #15003, opened 2024-12-16
[Bug] Registry check does not work in the case of multiple registrations and subscriptions at the application level, and provider export error causes Dubbo to fail to start


huangqy17:
### Pre-check

- [X] I am sure that all the content I provide is in English.

### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

dubbo3.2.16;
jdk openjdk21

### Steps to reproduce this issue

Dubbo3.2.16 uses two Nacos clusters as registration centers to implement application level multi subscription and multi subscription, and configure regirty.check=false. When one of the Nacos clusters is unavailable, dubbo serviceConfig. export reports an error causing startup failure, with the error message 'this.registry is null'

### What you expected to happen

When a Nacos cluster is unavailable, it should be possible to export services normally on another Nacos cluster and ignore the other Nacos cluster

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

walkinggo:
Can you show the complete error message?

huangqy17:
ServiceConfig#doExportUrl -> RegistryProtocol#export方法中，this.getRegistry(registryUrl)获得的Registry 是null,后面的逻辑没有判断这个值是否null，在RegistryProtocol#register方法中报空指针异常

huangqy17:
其他场景也有类似的问题，有的场景会判断Registry是否为null，而正常结束； 有的就类似现在启动export时报空指针异常

huangqy17:
我这个是多注册中心的高可用场景，如果多注册支持高可用的话，这里循环其中一个注册中心不应该异常退出

huangqy17:
![Uploading dubbo配置.jpg…]()
我的配置如图

huangqy17:
![Uploading dubbo配置.jpg…]()
配置如图

huangqy17:
![dubbo配置](https://github.com/user-attachments/assets/09131de5-c812-4a38-903c-909422d549ef)

walkinggo:
是的，我查看了这个问题，并确定它是存在的。我们应该提交一个合并请求来修复它。 @AlbumenJ

[...]

PR #15594, opened 2025-07-29
feat(registry): support the  option, whether to check the connection status

### LINKED ISSUE #15271: [Feature] Refactor Registry Implement

fantiq:

resolve #15271 #15003 , this issues are related to the config option `registry.check`. 

When `registry.check=true` (default value is `true`) is set, the connection status will be checked when creating the registry client. if the check fails, an exception will be thrown. 
when `registry.check=false` is set, no check is performed and no exception is ever thrown.

The reason for NEP is caused by setting the option `registry.check` to `false` and throwning an exception.
so here when `registry.check=false` no exception should be thrown. that is, the connection status should not be checked.

Currently, in the nacos client, the option `nacos.check` is used to determine whether to check the connection status or not.
the zookeeper client does not yet support not checking the connection status.

This PR add an new option `zookeeper.check`, it is similar to option `nacos.check`.

When creating a registry client, it will first be created using the default behavior. if an exception is thrown and `registry.check=false`, 
it will try to create the registry client again by turning on the option (`nacos.check=false` or `zookeeper.check=false`) to not check the connection status.

## Checklist


fantiq:
@RainYuY @zrlw PTAL

dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosConnectionManager.java
```diff
@@ -116,7 +120,7 @@ public synchronized void shutdownAll() {
      * @return {@link NamingService}
      */
     protected NamingService createNamingService() {
-        Properties nacosProperties = buildNacosProperties(this.connectionURL);
+        // Properties nacosProperties = buildNacosProperties(this.connectionURL);
```
zrlw: why not delete it directly?

dubbo-registry/dubbo-registry-zookeeper/src/main/java/org/apache/dubbo/registry/zookeeper/ZookeeperServiceDiscovery.java
```diff
@@ -80,8 +82,19 @@ public class ZookeeperServiceDiscovery extends AbstractServiceDiscovery {
...
         try {
-            this.curatorFramework = buildCuratorFramework(registryURL, this);
+            try {
+                curatorFramework = buildCuratorFramework(registryURL, this);
+            } catch (IllegalStateException e) {
+                if (check) {
+                    throw e;
+                }
+                curatorFramework = buildCuratorFramework(
+                        registryURL.addParameter(ZookeeperClient.ZOOKEEPER_CHECK_KEY, false), this);
```
RainYuY: I don't think we need a special key for ZK. How about using CHECK_KEY?

RainYuY: I think this logic can be moved to the abstract class, since both ZK and Nacos share the same logic.

dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/util/NacosNamingServiceUtils.java
```diff
@@ -114,8 +115,17 @@ public static NacosNamingServiceWrapper createNamingService(URL connectionURL) {
...
+        NacosConnectionManager nacosConnectionManager;
+        try {
+            nacosConnectionManager =
+                    new NacosConnectionManager(connectionURL, check, retryTimes, sleepMsBetweenRetries);
+        } catch (IllegalStateException e) {
+            if (RegistryFactory.isCheck(connectionURL)) {
+                throw e;
+            }
+            nacosConnectionManager =
+                    new NacosConnectionManager(connectionURL, false, retryTimes, sleepMsBetweenRetries);
```
RainYuY: I think this change might cause some problems. This registry isn’t functioning normally, but it's still added to the RegistryManager. As a result, subsequent actions might use this registry, which could lead to more errors. I'm not sure exactly what will happen after this PR is merged, but I’ll try to reproduce the issue later.

RainYuY: After trying it out, I got the result.
```
23:41:19.981 |-WARN  [DubboRegistryRetryTimer-thread-1] registry.nacos.NacosNamingServiceWrapper:    -|  [DUBBO] Failed to request nacos naming server. Dubbo will try to retry in 10. Try times: 6, dubbo version: 3.3.6-SNAPSHOT, current host: 192.168.31.224, error code: 1-37. This may be caused by , go to https://dubbo.apache.org/faq/1/37 to find instructions. 
com.alibaba.nacos.api.exception.NacosException: Client not connected, current status:STARTING
	at com.alibaba.nacos.common.remote.client.RpcClient.request(RpcClient.java:645) ~[nacos-client-2.5.1.jar:?]
	at com.alibaba.nacos.common.remote.client.RpcClient.request(RpcClient.java:624) ~[nacos-client-2.5.1.jar:?]
	at com.alibaba.nacos.client.naming.remote.gprc.NamingGrpcClientProxy.requestToServer(NamingGrpcClientProxy.java:449) ~[nacos-client-2.5.1.jar:?]
	at com.alibaba.nacos.client.naming.remote.gprc.NamingGrpcClientProxy.doRegisterService(NamingGrpcClientProxy.java:252) ~[nacos-client-2.5.1.jar:?]
	at com.alibaba.nacos.client.naming.remote.gprc.NamingGrpcClientProxy.registerServiceForEphemeral(NamingGrpcClientProxy.java:147) ~[nacos-client-2.5.1.jar:?]
	at com.alibaba.nacos.client.naming.remote.gprc.NamingGrpcClientProxy.registerService(NamingGrpcClientProxy.java:138) ~[nacos-client-2.5.1.jar:?]
	at com.alibaba.nacos.client.naming.remote.NamingClientProxyDelegate.registerService(NamingClientProxyDelegate.java:96) ~[nacos-client-2.5.1.jar:?]
	at com.alibaba.nacos.client.naming.NacosNamingService.registerInstance(NacosNamingService.java:161) ~[nacos-client-2.5.1.jar:?]
	at org.apache.dubbo.registry.nacos.NacosNamingServiceWrapper.lambda$registerInstance$5(NacosNamingServiceWrapper.java:133) ~[dubbo-3.3.6-SNAPSHOT.jar:3.3.6-SNAPSHOT]
	at org.apache.dubbo.registry.nacos.NacosNamingServiceWrapper.accept(NacosNamingServiceWrapper.java:495) ~[dubbo-3.3.6-SNAPSHOT.jar:3.3.6-SNAPSHOT]
	at org.apache.dubbo.registry.nacos.NacosNamingServiceWrapper.registerInstance(NacosNamingServiceWrapper.java:133) ~[dubbo-3.3.6-SNAPSHOT.jar:3.3.6-SNAPSHOT]
	at org.apache.dubbo.registry.nacos.NacosRegistry.doRegister(NacosRegistry.java:205) ~[dubbo-3.3.6-SNAPSHOT.jar:3.3.6-SNAPSHOT]
	at org.apache.dubbo.registry.retry.FailedRegisteredTask.doRetry(FailedRegisteredTask.java:33) ~[dubbo-3.3.6-SNAPSHOT.jar:3.3.6-SNAPSHOT]
	at org.apache.dubbo.registry.retry.AbstractRetryTask.run(AbstractRetryTask.java:127) ~[dubbo-3.3.6-SNAPSHOT.jar:3.3.6-SNAPSHOT]
	at org.apache.dubbo.common.timer.HashedWheelTimer$HashedWheelTimeout.expire(HashedWheelTimer.java:654) ~[dubbo-3.3.6-SNAPSHOT.jar:3.3.6-SNAPSHOT]
	at org.apache.dubbo.common.timer.HashedWheelTimer$HashedWheelBucket.expireTimeouts(HashedWheelTimer.java:733) ~[dubbo-3.3.6-SNAPSHOT.jar:3.3.6-SNAPSHOT]
	at org.apache.dubbo.common.timer.HashedWheelTimer$Worker.run(HashedWheelTimer.java:455) ~[dubbo-3.3.6-SNAPSHOT.jar:3.3.6-SNAPSHOT]
	at java.base/java.lang.Thread.run(Thread.java:1583) [?:?]

```
It will return to normal once the Nacos server is back online.
It's different from metadata or the config center — I think they should follow the same behavior.
This needs some discussion.

@zrlw @AlbumenJ @oxsean @heliang666s

RainYuY (changes requested):
Great job! Please take a look at these comments.

dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosConnectionManager.java
fantiq: > why not delete it directly?

i missed it there, i'll remove it later.

dubbo-registry/dubbo-registry-zookeeper/src/main/java/org/apache/dubbo/registry/zookeeper/ZookeeperServiceDiscovery.java
fantiq: > I don't think we need a special key for ZK. How about using CHECK_KEY?

agree with u

fantiq:
Thank you very mush for your preview.

In the zookeeper registry client, i have modified it to use `registry.check` directly instead of introducing new specific options.

Can it be considered that `nacos.check` actually replaces the function of `registry.check`? 
So here i only need to be compatible with this situation which led to #15003.
When `registry.check = false` and `nacos.check = true`, the value of `nacos.check` should be `false` actually.

And then i will evaluate the impact of this change on the config center and metadata center.

@RainYuY @zrlw  need more review, thx

zrlw (dismissed):

RainYuY:
> Thank you very mush for your preview.
> 
> In the zookeeper registry client, i have modified it to use `registry.check` directly instead of introducing new specific options.
> 
> Can it be considered that `nacos.check` actually replaces the function of `registry.check`? So here i only need to be compatible with this situation which led to #15003. When `registry.check = false` and `nacos.check = true`, the value of `nacos.check` should be `false` actually.
> 
> And then i will evaluate the impact of this change on the config center and metadata center.
> 
> @RainYuY @zrlw need more review, thx


The modification is OK, and I even think the nacos.check configuration item could be removed in the future.

Also, what I mentioned earlier was not that this PR affects the config center or metadata, but that their behavior is inconsistent. In the current implementation of this PR, even if the check fails, the registry is still initialized. As a result, the service keeps trying to register to an unavailable registry.

I think a better approach would be not to repeatedly attempt registration when the registry is unavailable. Instead, we should first check the availability of the registry, and only proceed with registration once it becomes available.

fantiq:
> Also, what I mentioned earlier was not that this PR affects the config center or metadata, but that their behavior is inconsistent

@RainYuY  Your focus is right. I have checked the startup process of the config-center and metadata. they both support the `check` option and need to be modified to support the logic of not checking the connection status. 
I will continue to modify these later.

> the service keeps trying to register to an unavailable registry

About this situation, is it possible to add connection status check logic before register?And the retry task should be retained.

RainYuY:
> @RainYuY Your focus is right. I have checked the startup process of the config-center and metadata. they both support the `check` option and need to be modified to support the logic of not checking the connection status. I will continue to modify these later.

I think there will be a lot of work to do. The best practice might be to open a new PR to reproduce it.

>  the service keeps trying to register to an unavailable registry
> 
> About this situation, is it possible to add connection status check logic before register?And the retry task should be retained.

Yes, I think this is the correct action.And it should be solved at this PR.

fantiq:
> I think there will be a lot of work to do. The best practice might be to open a new PR to reproduce it.

@RainYuY  It's a wise decision to create a new issue. In addition to making the ConfigCenter and MetadataCenter support `check=false`, we also need to consider the processing logic when the connection is recovery, this reqires more discussion.
I will create a new issue later.

dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/retry/AbstractRetryTask.java
```diff
@@ -124,6 +124,9 @@ public void run(Timeout timeout) throws Exception {
             logger.info(taskName + " : " + url);
         }
         try {
+            if (!registry.isAvailable()) {
+                throw new IllegalStateException("Registry is not available.");
```
RainYuY, line 128: Please use log.error not exception. Because the Exception will be catched.

RainYuY:
LGTM

RainYuY (dismissed):

zrlw:
duplicated with #15639

RainYuY:
@fantiq cc conflicts

RainYuY (dismissed):
LGTM

RainYuY:
@AlbumenJ @oxsean @zrlw PTAL

dubbo-remoting/dubbo-remoting-zookeeper-curator5/src/test/java/org/apache/dubbo/remoting/zookeeper/curator5/Curator5ZookeeperClientManagerTest.java
```diff
@@ -18,36 +18,55 @@
...
-        Curator5ZookeeperClient mockCurator5ZookeeperClient = mock(Curator5ZookeeperClient.class);
-        mockedCurator5ZookeeperClientConstruction =
-                mockConstructionWithAnswer(Curator5ZookeeperClient.class, invocationOnMock -> invocationOnMock
-                        .getMethod()
-                        .invoke(mockCurator5ZookeeperClient, invocationOnMock.getArguments()));
+        String zookeeperConnectionAddress1 = "zookeeper://127.0.0.1:2181";
+        zookeeperUrl = URL.valueOf(zookeeperConnectionAddress1 + "/service");
+
+        CuratorFrameworkFactory.Builder realBuilder = CuratorFrameworkFactory.builder();
+        CuratorFrameworkFactory.Builder spyBuilder = spy(realBuilder);
```
zrlw, line 55: it might conflict with line 78: CuratorFrameworkFactory.Builder spyBuilder = CuratorFrameworkFactory.builder();

zrlw (approved):
LGTM

RainYuY (approved):

Issue #6588, opened 2020-08-12
Dubbo多虚拟网卡情况下 怎么配置消费者指定真实ip


wxywizard:
- [ ] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.
- [ ] I have checked the [FAQ](https://github.com/apache/dubbo/blob/master/FAQ.md) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: v2.7.6
* Operating System version: window10
* Java version: 1.8

### Steps to reproduce this issue

1. 本机存在多个虚拟网卡
2. 采用nacos注册中心，服务提供者配置dubbo.protocol.host之后可以注册本地真实ip但是服务消费者一直找不到这个提供者
3. 禁用虚拟网卡后一切都会好
尝试启动参数也无效
-DDUBBO_IP_TO_BIND和DDUBBO_IP_TO_REGISTRY在idea中vm options 配置都没有效果

mercyblitz:
Please use `-Ddubbo.network.interface.preferred=${your-network-interface-name}` to filter network interface what you want.

598572:
如果确实就想在application.ymal文件中配 那么在启动时 拦截到ymal文件中的配置 然后通过 
 System.setProperty(DUBBO_PREFERRED_NETWORK_INTERFACE_CONSTANT, preferredInterface);
System.setProperty(DUBBO_NETWORK_IGNORED_INTERFACE_CONSTANT, ignoredInterface);
设置到系统变量中去，也是一个可行的方法，亲测有效，这里给出代码：

public class DubboNetworkInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {

    private static final String DUBBO_PREFERRED_NETWORK_INTERFACE_CONSTANT = "dubbo.network.interface.preferred";
    private static final String DUBBO_NETWORK_IGNORED_INTERFACE_CONSTANT = "dubbo.network.interface.ignored";

    @Override
    public void initialize(ConfigurableApplicationContext applicationContext) {
        String preferredInterface = applicationContext.getEnvironment().getProperty(DUBBO_PREFERRED_NETWORK_INTERFACE_CONSTANT);
        String ignoredInterface = applicationContext.getEnvironment().getProperty(DUBBO_NETWORK_IGNORED_INTERFACE_CONSTANT);

        if (StringUtils.isNotBlank(preferredInterface)) {
[...]

PR #15601, opened 2025-07-31
Enhance Network Interface Configuration via Spring Environment

fu-cheng1:
#6588 

This change allows the dubbo.network.interface.preferred configuration to be set via Spring configuration files, rather than requiring it to be specified as a system property at startup.

Currently, Dubbo determines the preferred network interface too early in the application lifecycle, due to logging mechanisms like FailsafeLogger that trigger host resolution (via NetUtils) before Spring configuration is available.

To address this, a new listener DubboNetInterfaceConfigApplicationListener is introduced. It listens for the ApplicationPreparedEvent, reads the network interface configuration from the Spring Environment, and sets it into system properties if not already configured. The listener is intentionally not registered as an ApplicationContextInitializer to avoid running before PropertySourceBootstrapConfiguration in Spring Cloud environments. Additionally, since the current Dubbo version does not yet depend on a Spring Boot version that supports ApplicationContextInitializedEvent, the ApplicationPreparedEvent is chosen as a safe extension point.

This change improves flexibility and integration with Spring-based configuration management, especially in Spring Cloud environments.

### DIFF
--- a/dubbo-spring-boot/dubbo-spring-boot-compatible/autoconfigure/src/test/java/org/apache/dubbo/spring/boot/context/DubboNetInterfaceConfigApplicationContextInitializerTest.java
+++ b/dubbo-spring-boot/dubbo-spring-boot-compatible/autoconfigure/src/test/java/org/apache/dubbo/spring/boot/context/DubboNetInterfaceConfigApplicationContextInitializerTest.java
@@ -1,0 +1,59 @@
+package org.apache.dubbo.spring.boot.context;
+
+import org.apache.dubbo.common.constants.CommonConstants;
+import org.apache.dubbo.spring.boot.context.event.DubboNetInterfaceConfigApplicationListener;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.context.event.ApplicationEnvironmentPreparedEvent;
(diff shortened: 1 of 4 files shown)


heliang666s:
please run `mvn spotless:apply`

dubbo-spring-boot/dubbo-spring-boot-compatible/autoconfigure/src/test/java/org/apache/dubbo/spring/boot/context/DubboNetInterfaceConfigApplicationContextInitializerTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.context.ApplicationListener;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.MapPropertySource;
+import org.springframework.core.env.MutablePropertySources;
+
+import static org.apache.dubbo.common.constants.CommonConstants.DUBBO_NETWORK_IGNORED_INTERFACE;
+import static org.apache.dubbo.common.constants.CommonConstants.DUBBO_PREFERRED_NETWORK_INTERFACE;
+
+/**
+ * @since 3.2
```
heliang666s, line 22: if you want to develop or enhance sone features, please merge into 3.3 branch, 3.2 branch is only bugfix now.

heliang666s (changes requested):

PR #15603, opened 2025-08-01
Add serviceConfigV2 null check at metadata service export

zrlw:
The condition ```serviceConfig == null``` at ```ConfigurableMetadataServiceExporter#export``` is always true since it will not be assgined when ```onlyUseMetadataV2``` is true, which will cause ```ConfigurableMetadataServiceExporter#export``` exported check failure. 

```
    public synchronized ConfigurableMetadataServiceExporter export() {
        if (serviceConfig == null || !isExported()) {
            if (MetadataServiceVersionUtils.needExportV1(applicationModel)) {
                exportV1();
            }
            if (MetadataServiceVersionUtils.needExportV2(applicationModel)) {
                exportV2();
            }
        } else {
            if (logger.isWarnEnabled()) {
                logger.warn(
                        CONFIG_METADATA_SERVICE_EXPORTED,
                        "",
                        "",
                        "The MetadataService has been exported : " + serviceConfig.getExportedUrls());
            }
        }

        return this;
    }
```

## Checklist

### DIFF
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/metadata/ConfigurableMetadataServiceExporter.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/metadata/ConfigurableMetadataServiceExporter.java
@@ -73,7 +73,7 @@ public class ConfigurableMetadataServiceExporter {
     }
 
     public synchronized ConfigurableMetadataServiceExporter export() {
-        if ((serviceConfig == null && serviceConfigV2 == null) || !isExported()) {
+        if (!isExported()) {
             if (MetadataServiceVersionUtils.needExportV1(applicationModel)) {
                 exportV1();
             }
--- a/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/metadata/ConfigurableMetadataServiceExporter.java
+++ b/dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/metadata/ConfigurableMetadataServiceExporter.java
@@ -16,6 +16,7 @@
  */
 package org.apache.dubbo.config.metadata;
 
+import org.apache.dubbo.common.URL;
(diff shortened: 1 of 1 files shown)


dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/metadata/ConfigurableMetadataServiceExporter.java
```diff
@@ -71,7 +73,7 @@ public ConfigurableMetadataServiceExporter(
     }
 
     public synchronized ConfigurableMetadataServiceExporter export() {
-        if (serviceConfig == null || !isExported()) {
+        if ((serviceConfig == null && serviceConfigV2 == null) || !isExported()) {
```
FoghostCn: I think `!isExported()` is enough,since it will check serviceConfig and serviceConfigV2

FoghostCn (changes requested):

dubbo-config/dubbo-config-api/src/main/java/org/apache/dubbo/config/metadata/ConfigurableMetadataServiceExporter.java
zrlw: it will not call isExported() when the app only exported v2 because servieConfig is always null

zrlw: you are right, ```(serviceConfig == null && serviceConfigV2 == null)``` is included in ```!isExported()```

zrlw: removed ``serviceConfig == null```

FoghostCn (approved):
LGTM

oxsean (approved):
LGTM.

PR #15604, opened 2025-08-01
Support Preferred Network Interface via Spring Environment and Fix Early Host Resolution

fu-cheng1:
This change removes the logic in FailsafeLogger that appends the local host information to log messages. The motivation is that FailsafeLogger is used by the WelcomeLogoApplicationListener to print logs very early during application startup—before Spring’s configuration files are fully loaded. As a result, NetUtils resolves and caches the local host information prematurely. This prevents the application from re-evaluating the preferred network interface later based on user-defined configuration. To avoid this issue, the host resolution in FailsafeLogger has been removed.

A new class, DubboNetInterfaceConfigApplicationListener, has been introduced to retrieve the preferred network interface configuration from Spring’s Environment and set it as a system property—since NetUtils reads this value from environment variables. This listener listens to the ApplicationContextInitializedEvent, which ensures that the Spring Environment (including values from configuration centers) is fully prepared when the listener executes.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/logger/support/FailsafeLogger.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/logger/support/FailsafeLogger.java
@@ -18,6 +18,7 @@ package org.apache.dubbo.common.logger.support;
 
 import org.apache.dubbo.common.Version;
 import org.apache.dubbo.common.logger.Logger;
+import org.apache.dubbo.common.utils.NetUtils;
 
 public class FailsafeLogger implements Logger {
 
@@ -46,7 +47,8 @@ public class FailsafeLogger implements Logger {
     }
 
     private String appendContextMessage(String msg) {
-        return " [DUBBO] " + msg + ", dubbo version: " + Version.getVersion();
+        return " [DUBBO] " + msg + ", dubbo version: " + Version.getVersion() + ", current host: "
+                + NetUtils.getLocalHost();
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


dubbo-common/src/main/java/org/apache/dubbo/common/logger/support/FailsafeLogger.java
```diff
@@ -47,8 +46,7 @@ public void setLogger(Logger logger) {
     }
 
     private String appendContextMessage(String msg) {
-        return " [DUBBO] " + msg + ", dubbo version: " + Version.getVersion() + ", current host: "
-                + NetUtils.getLocalHost();
+        return " [DUBBO] " + msg + ", dubbo version: " + Version.getVersion();
```
zrlw: if apply this pr, the current host information will be logged at where? could you provide your log information about it?

fu-cheng1: However, if the printing of this host information is not removed, it will cause premature caching of the network interface card (NIC) that should be used in `WelcomeLogoApplicationListener`. The execution timing of `WelcomeLogoApplicationListener` may occur *before* Spring loads the configuration files, resulting in the NIC configuration specified in the configuration files failing to take effect in `NetUtils`.

Therefore, if host information logging is mandatory:  
1. Could the execution of `WelcomeLogoApplicationListener` be deferred?  
2. Alternatively, would it be acceptable to omit logging the host information?

zrlw: the host information should be logged for existed exception checking apps, you'd better log it at other location instead of omitting it directly.

fu-cheng1: I have updated the execution timing of WelcomeLogoApplicationListener and re-enabled the logging of local host information.

zrlw (approved):

RainYuY (approved):
LGTM

PR #15616, opened 2025-08-06
Implement full zero-copy output for Dubbo Triple protocol 

heliang666s:
The patch implements zero-copy encoding for protobuf responses on the Triple server side. When needWrapper() is false and the stream is a ByteBufOutputStream, the encoder writes the 5-byte gRPC header directly into the buffer and streams the
  Message without creating an intermediate byte[]. To enable that path, ReflectionPackableMethod now exposes a PbArrayPacker for protobuf responses. All non-protobuf/wrapper or compressed cases continue to use the original byte-array fallback.

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/ReflectionPackableMethod.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/ReflectionPackableMethod.java
@@ -50,7 +50,6 @@ public class ReflectionPackableMethod implements PackableMethod {
     private static final String REACTOR_RETURN_CLASS = "reactor.core.publisher.Mono";
     private static final String RX_RETURN_CLASS = "io.reactivex.Single";
     private static final String GRPC_STREAM_CLASS = "io.grpc.stub.StreamObserver";
-    private static final Pack PB_PACK = o -> ((Message) o).toByteArray();
 
     private final Pack requestPack;
     private final Pack responsePack;
@@ -89,7 +88,7 @@ public class ReflectionPackableMethod implements PackableMethod {
         this.needWrapper = needWrap(method, actualRequestTypes, actualResponseType);
         if (!needWrapper) {
             requestPack = new PbArrayPacker(singleArgument);
-            responsePack = PB_PACK;
+            responsePack = new PbArrayPacker(true);
             requestUnpack = new PbUnpack<>(actualRequestTypes[0]);
             responseUnpack = new PbUnpack<>(actualResponseType);
         } else {
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/ReflectionPackableMethod.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/ReflectionPackableMethod.java
@@ -38,8 +38,6 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.stream.Stream;
 
-import com.google.protobuf.Message;
-
 import static org.apache.dubbo.common.constants.CommonConstants.$ECHO;
 import static org.apache.dubbo.common.utils.ProtobufUtils.isProtobufClass;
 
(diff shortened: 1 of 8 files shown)


dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2Serialization.java
```diff
@@ -83,4 +83,71 @@ public ObjectInput deserialize(URL url, InputStream is) throws IOException {
...
+        boolean isStreamingMode = isStreamingMode(url);
+
+        // Use Hessian2Output directly for zero-copy serialization
+        com.alibaba.com.caucho.hessian.io.Hessian2Output hessianOutput =
+                new com.alibaba.com.caucho.hessian.io.Hessian2Output(outputStream);
+        hessianOutput.setSerializerFactory(hessian2FactoryManager.getSerializerFactory(
+                Thread.currentThread().getContextClassLoader()));
+
+        try {
+            if (isStreamingMode) {
```
RainYuY: It looks like the same code is in both the if and else branches. Why?

dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2Serialization.java
```diff
@@ -83,4 +83,71 @@ public ObjectInput deserialize(URL url, InputStream is) throws IOException {
...
+                .map(URL::getOrDefaultFrameworkModel)
+                .orElseGet(FrameworkModel::defaultModel)
+                .getBeanFactory()
+                .getBean(Hessian2FactoryManager.class);
+
+        // Check if we're in streaming mode (used with StreamingLengthEncoder)
+        boolean isStreamingMode = isStreamingMode(url);
+
+        // Use Hessian2Output directly for zero-copy serialization
+        com.alibaba.com.caucho.hessian.io.Hessian2Output hessianOutput =
```
RainYuY: Why is there not org.apache.dubbo.common.serialize.hessian2.Hessian2ObjectOutput?

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/ReflectionPackableMethod.java
```diff
@@ -86,23 +87,51 @@ public ReflectionPackableMethod(
...
 
         boolean singleArgument = method.getRpcType() != MethodDescriptor.RpcType.UNARY;
-        this.needWrapper = needWrap(method, actualRequestTypes, actualResponseType);
+
+        // Check protocol requirements for wrapper - this is mainly for legacy compatibility
+        boolean protocolNeedsWrapper = needWrap(method, actualRequestTypes, actualResponseType);
+        boolean serializationSupportsZeroCopy = isZeroCopyCapable(serializeName, url);
+
+        // Zero-copy should only be used when protocol allows it AND serialization supports it
+        // If protocol needs wrapper OR serialization doesn't support zero-copy, use wrapper mode
```
RainYuY: Why not use and?

RainYuY (changes requested):

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcCompositeCodec.java
```diff
@@ -78,12 +85,42 @@ public void loadPackableMethod(MethodDescriptor methodDescriptor) {
...
+
+                Pack responsePack = packableMethod.getResponsePack();
+
+                if (responsePack instanceof PbArrayPacker) {
+                    PbArrayPacker pbPacker = (PbArrayPacker) responsePack;
+                    ByteBufOutputStream bbos = (ByteBufOutputStream) outputStream;
+                    ByteBuf buffer = bbos.buffer();
+
+                    try {
+                        com.google.protobuf.Message message = extractProtobufMessage(data, pbPacker);
```
RainYuY: Just use Message？

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcCompositeCodec.java
```diff
@@ -78,12 +85,42 @@ public void loadPackableMethod(MethodDescriptor methodDescriptor) {
 
     @Override
     public void encode(OutputStream outputStream, Object data, Charset charset) throws EncodeException {
-        // protobuf
-        // TODO int compressed = Identity.MESSAGE_ENCODING.equals(requestMetadata.compressor.getMessageEncoding()) ? 0 :
-        // 1;
         try {
-            int compressed = 0;
```
RainYuY, line 85: I think still use compressed would be better.

RainYuY (changes requested):
Because code generate,I think the code will not be executed..

PR #15646, opened 2025-08-19
[ISSUE 15635] Add support for Zookeeper-based config center and metadata reporting in Demo

lingyufei:
Currently, dubbo-demo-api-consumer and dubbo-demo-api-provider doesn't have the metadata-report and config-center dependencies like dubbo-metadata-report-zookeeper.

Because of this, when testing the dubbo.application.register-mode = instance mode in dubbo-demo-api-provider, the metadata would be missing in the zookeeper. This would result in the following error when starting dubbo-demo-api-consumer.

It would be better to add dubbo-metadata-report-zookeeper and dubbo-metadata-report-zookeeper to the demo as default. This will be more convenient for debugging.

Please refer to ISSUE [15635](https://github.com/apache/dubbo/issues/15635) for more details.

### DIFF
--- a/dubbo-demo/dubbo-demo-api/dubbo-demo-api-consumer/pom.xml
+++ b/dubbo-demo/dubbo-demo-api/dubbo-demo-api-consumer/pom.xml
@@ -48,6 +48,16 @@
     </dependency>
     <dependency>
       <groupId>org.apache.dubbo</groupId>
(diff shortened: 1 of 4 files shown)


zrlw (approved):

RainYuY (approved):
LGTM

PR #15657, opened 2025-08-30
feat(common): add optional sign support to StringUtils.isNumeric

6mmLIU:
Introduce an overload `StringUtils.isNumeric(String str, boolean allowDot, boolean allowSign)` to optionally validate leading `+`/`-` signs while keeping the existing API and behavior unchanged. Callers that do not need sign support can continue using the original method without any behavior change.

## Background & motivation
Several call sites need numeric validation for signed integers/decimals. Today they either re-implement ad-hoc checks or reject valid inputs with a leading sign. Centralizing the logic in `StringUtils` reduces duplication and subtle bugs, while remaining strictly backward compatible.

## Brief changelog
- Add `isNumeric(String, boolean allowDot, boolean allowSign)` overload.
- Keep the original signature and semantics intact.
- Leading sign is accepted only at position 0 when `allowSign == true`; a lone sign (e.g. `"+"`, `"-"`) is still invalid.
- Dot handling is identical to the current implementation: at most one dot if `allowDot == true`; otherwise any dot fails.
- Only ASCII digits are considered numeric (no scientific notation like `1e3`).

## Verifying this change
Unit tests cover positive and negative paths, including:
- `"-1"` with `allowSign=true` passes; `"+"` / `"-"` alone fail.
- `"+1.23"` passes when `allowDot=true && allowSign=true`.
- Multiple dots, embedded non-digits, and null/empty strings fail.
- Original method keeps returning `false` for signed inputs (backward compatibility check).

## Compatibility, risk and docs
- **Backward compatible**: no changes to the existing method’s behavior.
- JavaDoc added for the new overload; no external docs need updating.
- Non-goals: scientific notation, locale digits, and whitespace trimming.

## Issue
Fixes #<#15658>

## Checklist
- [x] A GitHub issue is linked (`Fixes #<issue-id>`).
- [x] Clear description of motivation, design, and behavior boundaries.
- [x] Unit tests added for signed/unsigned and edge cases.
- [x] No breaking changes for existing callers.
- [x] CI: run `mvn -q -pl dubbo-common -am test` (ensure a networked environment so parent POM and dependencies resolve).

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
@@ -606,17 +606,26 @@ public final class StringUtils {
         if (str == null || str.isEmpty()) {
             return false;
         }
+        int start = 0;
+        // Allow leading '+' or '-' sign
+        if (str.charAt(0) == '-' || str.charAt(0) == '+') {
+            if (str.length() == 1) {
+                return false;
+            }
+            start = 1;
+        }
         boolean hasDot = false;
         int sz = str.length();
-        for (int i = 0; i < sz; i++) {
-            if (str.charAt(i) == '.') {
+        for (int i = start; i < sz; i++) {
+            char ch = str.charAt(i);
+            if (ch == '.') {
                 if (hasDot || !allowDot) {
                     return false;
                 }
                 hasDot = true;
                 continue;
             }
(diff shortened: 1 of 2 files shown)


oxsean (changes requested):
This is a class used internally by Dubbo, could you give a case where this feature is necessary? I don’t think we need to make the change.

PR #15666, opened 2025-09-02
End the 'read/write' combination operation early to improve performance

frozenNoob:
Fix the issue：In multi-threaded parallel execution, a large number of read only operations waste a lot of unnecessary time because the combination of ```read and write``` operations requires ```serialization```

issue link: https://github.com/apache/dubbo/issues/15665

## My Solution
Issue 1 must be solved, Issue 2 can be tolerated.
My proposed solution is as follows: allow a large number of read-only operations to proceed in parallel (memory visibility is guaranteed because the `value` field in the inner class `Node` of `ConcurrentHashMap` is `volatile`), and terminate early via detection.
Related code is as follows:
```java
@Override
protected <T> Invoker<T> doSelect(List<Invoker<T>> invokers, URL url, Invocation invocation) {
    String methodName = RpcUtils.getMethodName(invocation);
    String key = invokers.get(0).getUrl().getServiceKey() + "." + methodName;
    int invokersHashCode = invokers.hashCode();

    // If the detection is successful, return in advance. it may be different from selector, but it doesn't matter
    ConsistentHashSelector<T> oldSelector0;
        if((oldSelector0 = (ConsistentHashSelector<T>) selectors.get(key)) != null && oldSelector0.identityHashCode == invokersHashCode){
            return oldSelector0.select(invocation);
        }

    // using the hashcode of invoker list to create consistent selector by atomic computation.
    ConsistentHashSelector<T> selector = (ConsistentHashSelector<T>) selectors.compute(
            key,
            (k, oldSelector) -> (oldSelector == null || oldSelector.identityHashCode != invokersHashCode)
                    ? new ConsistentHashSelector<>(invokers, methodName, invokersHashCode)
                    : oldSelector);
    return selector.select(invocation);
}
```

## Checklist
- [ ] Write necessary unit-test to verify your logic correction. If the new feature or significant change is committed, please remember to add sample in [dubbo samples](https://github.com/apache/dubbo-samples) project.

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/ConsistentHashLoadBalance.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/loadbalance/ConsistentHashLoadBalance.java
@@ -51,6 +51,13 @@ public class ConsistentHashLoadBalance extends AbstractLoadBalance {
         String methodName = RpcUtils.getMethodName(invocation);
         String key = invokers.get(0).getUrl().getServiceKey() + "." + methodName;
         int invokersHashCode = invokers.hashCode();
+        // If the detection is successful, return in advance. it may be different from selector, but it doesn't matter
+        ConsistentHashSelector<T> oldSelector0;
+        if ((oldSelector0 = (ConsistentHashSelector<T>) selectors.get(key)) != null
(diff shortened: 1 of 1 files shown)


zrlw (approved):

zrlw:
LGTM

EarthChen (approved):
LGTM

PR #15667, opened 2025-09-02
Unset `DEBIAN_MIRROR` as the test image of Debian was already upgraded from buster to bullseye

zrlw:
Unset ```DEBIAN_MIRROR``` which was set  in the workflow files.
```DEBIAN_MIRROR``` should not be set in dubbo workflows because the docker image name of ```dubbo-samples``` and ```dubbo-integration-cases``` was upgraded to ```bullseye```.
Debian changed the archive URL again, we don't know whether current Debian archive URL is temporary or not,  so it might be better to upgrade the test docker image version to the lastest - ```bullseye```. 

## Checklist


oxsean (approved):
LGTM.

zrlw:
Current github action will always fail if we do not take any action since Debian changed its archive site address.

EarthChen (approved):
LGTM

PR #15680, opened 2025-09-09
Specify the jsr310 time format based on the configuration when generic invoke

Aresxue:
Specify the jsr310 time format based on the configuration when generic invoke, the handling of jsr310 times (package names start with java.time) is opened to customize with dubbo's own spi implementation, which provides a default implementation in spring that allows you to configure date formats.

## Verifying this change
org.apache.dubbo.spring.boot.context.event.Jsr310ConverterApplicationListenerTest

## Old pull request
https://github.com/apache/dubbo/pull/13067https://github.com/apache/dubbo/pull/13067

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/convert/jsr310/AbstractTemporalAccessorConverter.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/convert/jsr310/AbstractTemporalAccessorConverter.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.dubbo.common.convert.jsr310;
+
+import org.apache.dubbo.common.utils.CollectionUtils;
+
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.TemporalAccessor;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.BiFunction;
+
+public abstract class AbstractTemporalAccessorConverter {
+
+    private final List<DateTimeFormatter> dateTimeFormatterList;
+
+    public AbstractTemporalAccessorConverter() {
+        this.dateTimeFormatterList = Collections.emptyList();
+    }
+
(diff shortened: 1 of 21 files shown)


Aresxue:
@zrlw PTAL

zrlw (approved):
[code format](https://github.com/apache/dubbo/actions/runs/17574217290)

zrlw:
run ```mvn spotless:apply```  with jdk21 to format codes

zrlw (changes requested):
see details at https://github.com/apache/dubbo/actions

zrlw (approved):
LGTM

oxsean (changes requested):
If it’s just to support generic invoke, wouldn’t modifying only the PojoUtils class be enough?
how to config date format, you can refer to org.apache.dubbo.common.utils.PojoUtils#GENERIC_WITH_CLZ.

zrlw (changes requested):
Configure date time format at generic invocation for what?

Aresxue:
> 如果只是为了支持泛型调用，只修改 PojoUtils 类不就可以了吗？ 如何配置日期格式，可以参考 org.apache.dubbo.common.utils.PojoUtils#GENERIC_WITH_CLZ。



> 如果只是为了支持泛型调用，只修改 PojoUtils 类不就可以了吗？ 如何配置日期格式，可以参考 org.apache.dubbo.common.utils.PojoUtils#GENERIC_WITH_CLZ。

Because the format required by each user for JSR310 time is different, the default format of JDK is java.time.format.DateTimeFormatter#ISO_LOCAL_DATE_TIME, such as '2025-01-01T00:00:01', but traditional Chinese companies prefer to use the format of '2025-01-01 00:00:01'. At the same time, sometimes it is necessary to be compatible with multiple time formats such as '2025-01-01 00:00:01' and '2025-01-01' during deserialization. Therefore, SPI allows users to define their own processing strategies.

Aresxue:
> Configure date time format at generic invocation for what?

In a classic gateway scenario with generalized calls, the time returned by the downstream RPC service is serialized in the format of '2025-01-01T00:00:01'. Many users are dissatisfied with this format and prefer the format of '2025-01-01 00:00:01'. However, because the time is converted to a String, it lacks Java type information. External format conversion requires checking whether each String is a time, which incurs significant overhead and is cumbersome to implement.

oxsean:
This introduces too much complexity. Since it’s a generic invoke, it should follow the format requirements for generic invoke. If you need flexibility in formatting, you should wrap it as a REST API instead of exposing it through Dubbo. Triple REST also supports this kind of customization.

Aresxue:
> This introduces too much complexity. Since it’s a generic invoke, it should follow the format requirements for generic invoke. If you need flexibility in formatting, you should wrap it as a REST API instead of exposing it through Dubbo. Triple REST also supports this kind of customization.

I apologize for not explaining this clearly enough. This generalized call occurs at the gateway layer, and I don't have permission to modify the downstream RPC interface implementation. I believe generalized calls and flexibility don't conflict. For example, consider grpc's DynamicMessage. A classic use case for generalized calls is the gateway, but in practice, Dubbo's flexibility in this area is far from sufficient.

oxsean:
I don’t recommend merging this change, or at least it should be made simpler.

Aresxue:
> 我不建议合并此更改，或者至少应该使其更简单。

I suggest you refer to grpc to do some reconstruction of the generalized call. PojoUtils is really a bad design.

PR #15695, opened 2025-09-17
Fixes #15671: remove Future from FUTURES after timeout and add RejectedExecution test fix 

### LINKED ISSUE #15671: [Bug] MemoryLeak Issue In DefaultFuture
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

when i set threadPool with my custom threadPool . 
it may be  executor reject exception when submit processTask in DefaultFuture.   

The map of Futures cache may can not be clear for long time.  

Is it necessary to try this kind of exception.

```java
private static final Map<Long, DefaultFuture> FUTURES = new ConcurrentHashMap<>();
  @Override
        public void run(Timeout timeout) {
            DefaultFuture future = DefaultFuture.getFuture(requestID);
            if (future == null || future.isDone()) {
                return;
            }

            ExecutorService executor = future.getExecutor();
            if (executor != null && !executor.isShutdown()) {
                executor.execute(() -> notifyTimeout(future));
            } else {
                notifyTimeout(future);
            }
        }
```

1. set custom thread pool in consumer side 
2.  set core thread and max thread and use limited queue
3. when process task can not be processed the future remain in the memory 

the dump of application, I use MAT  to search futureMap. It  keeps a
future which associate with request send 3 hours ago.

### Anything else

only when certain situation happened.  

### Code of Conduct

qxggg:
Fixes #15671
to fix:
When I set threadPool with my custom threadPool,
it may be executor reject exception when submit processTask in DefaultFuture.
The map of Futures cache may can not be clear for long time.
see #15671

## MySolution
Wrap the executor.execute call in a try-catch.
If the thread pool is full, notifyTimeout is called manually after a RejectedExecutionException to clean up and prevent memory leaks.
```java
@Override
public void run(Timeout timeout) {
    DefaultFuture future = DefaultFuture.getFuture(requestID);
    if (future == null || future.isDone()) {
        return;
    }
    ExecutorService executor = future.getExecutor();
    if (executor != null && !executor.isShutdown()) {
        try {
            executor.execute(() -> notifyTimeout(future));
        } catch (RejectedExecutionException e) {
            notifyTimeout(future);
        }
    } else {
        notifyTimeout(future);
    }
}
```

## Test
This unit test testTimeoutWithRejectedExecution verifies the behavior of DefaultFuture when using a custom ThreadPoolExecutor that may reject tasks due to full capacity.
```java
  @Test
  void testTimeoutWithRejectedExecution() throws Exception {
      // Create a ThreadPoolExecutor with a queue capacity of 1
      ThreadPoolExecutor customExecutor = new ThreadPoolExecutor(
              1, // corePoolSize
              1, // maxPoolSize
              60L,
              TimeUnit.SECONDS,
              new ArrayBlockingQueue<>(1), // queue capacity is 1
              new ThreadPoolExecutor.AbortPolicy() // default rejection policy: throws exception
              );
      // Submit two tasks to occupy the thread and the queue
      customExecutor.submit(() -> {
          try {
              Thread.sleep(500); // occupy the thread for a while
          } catch (InterruptedException ignored) {
          }
      });
      customExecutor.submit(() -> {
          try {
              Thread.sleep(500); // occupy the queue
          } catch (InterruptedException ignored) {
          }
      });
      // Create a Dubbo Mock Channel and a request
      Channel channel = new MockedChannel();
      Request request = new Request(999);
      // Use Dubbo's newFuture and pass in the custom thread pool
      DefaultFuture future = DefaultFuture.newFuture(channel, request, 100, customExecutor);
      // Mark the request as sent
      DefaultFuture.sent(channel, request);
      // Wait for the timeout task to trigger
      Thread.sleep(300);
      Assertions.assertNull(DefaultFuture.getFuture(999), "Future should be removed from FUTURES after timeout");
      customExecutor.shutdown();
  }
```
## Checklist

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
@@ -38,6 +38,7 @@ import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 
 import static org.apache.dubbo.common.constants.CommonConstants.DEFAULT_TIMEOUT;
@@ -324,7 +325,12 @@ public class DefaultFuture extends CompletableFuture<Object> {
 
             ExecutorService executor = future.getExecutor();
             if (executor != null && !executor.isShutdown()) {
-                executor.execute(() -> notifyTimeout(future));
+                try {
+                    executor.execute(() -> notifyTimeout(future));
+                } catch (RejectedExecutionException e) {
+                    notifyTimeout(future);
+                    throw e;
+                }
             } else {
                 notifyTimeout(future);
             }
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
@@ -38,7 +38,6 @@ import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 
 import static org.apache.dubbo.common.constants.CommonConstants.DEFAULT_TIMEOUT;
@@ -322,13 +321,10 @@ public class DefaultFuture extends CompletableFuture<Object> {
             if (future == null || future.isDone()) {
                 return;
             }
+
             ExecutorService executor = future.getExecutor();
             if (executor != null && !executor.isShutdown()) {
-                try {
-                    executor.execute(() -> notifyTimeout(future));
-                } catch (RejectedExecutionException e) {
-                    notifyTimeout(future);
-                }
+                executor.execute(() -> notifyTimeout(future));
             } else {
                 notifyTimeout(future);
             }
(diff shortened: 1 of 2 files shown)


zrlw (dismissed):
LGTM

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
```diff
@@ -321,10 +322,13 @@ public void run(Timeout timeout) {
...
                 return;
             }
-
             ExecutorService executor = future.getExecutor();
             if (executor != null && !executor.isShutdown()) {
-                executor.execute(() -> notifyTimeout(future));
+                try {
+                    executor.execute(() -> notifyTimeout(future));
+                } catch (RejectedExecutionException e) {
+                    notifyTimeout(future);
```
RainYuY, line 331: I think it should throw an exception. Also, I think we need another PR to change it so that it does not call notifyTimeout, because it’s not actually a timeout.

RainYuY (changes requested):

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
qxggg: Hi, just to confirm:
Thanks for your reply. This is my first time contributing, so I want to make sure I fully understand.
Should I change the code now to simply let the RejectedExecutionException be thrown and then update this PR?
Or would you prefer me to leave this PR as is and wait until we open another PR for handling the cleanup logic?

qxggg: Hi @RainYuY, I've updated the code according to your feedback.
I've kept a TODO test case in place—should we leave it in, or remove it?

RainYuY: Your first version of the code had the right idea, but it missed throwing the RejectedExecutionException after the notifyTimeout call. Actually, we should not call notifyTimeout here because this is not a timeout. You can fix the notifyTimeout issue in this PR or in a follow-up PR, but for this PR please revert to your original implementation and add the exception throw.

qxggg: OK, I get it. 
Thanks for your patience! Could you please tell me if this is correct now?

RainYuY (approved):

RainYuY:
LGTM

qxggg:
@zrlw  I’ve updated the code based on Rain's feedback.
Could you please take another look when you have time? Thanks!

zrlw (approved):

Issue #15732, opened 2025-10-21
[Bug] BatchExecutorQueue performance risk


fxli4:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.6 JDK21

### Steps to reproduce this issue

**Root Cause:**

The issue occurs in a BatchExecutorQueue implementation where:

- Multiple worker threads produce messages into a ConcurrentLinkedQueue

- A single netty EventLoop thread consumes messages by copying all elements from ConcurrentLinkedQueue to a LinkedList in one operation

**Problem:**
In extreme scenes, when multiple worker threads generate large volumes of messages simultaneously, the bulk copy operation causes:

1. Long polling/blocking: The entire copy process blocks the EventLoop thread

2. Availability impact: Extended blocking reduces system responsiveness

3. Network I/O disruption: Since this uses Netty's EventLoop thread, the blocking prevents timely reading of incoming messages from other services, creating a cascading failure effect

**SourceCode:** [BatchExecutorQueue.run()](https://github.com/apache/dubbo/blob/dea02069082226b4677579474db4d156dbb8fb0e/dubbo-common/src/main/java/org/apache/dubbo/common/BatchExecutorQueue.java#L53)

<img width="576" height="867" alt="Image" src="https://github.com/user-attachments/assets/fa665eb1-a501-4c82-8d65-af3f3566c7a4" />

### What you expected to happen

**Impact:**

- EventLoop thread starvation

- Delayed message processing

- Potential service unavailability during high load

- Impaired ability to handle incoming requests from other services

The core issue is the synchronous bulk copy operation that doesn't yield the EventLoop thread, violating the non-blocking principle essential for proper EventLoop performance.

### Anything else

I conducted a comprehensive performance test to evaluate and compare three distinct implementation strategies for [BatchExecutorQueue.run()](https://github.com/apache/dubbo/blob/dea02069082226b4677579474db4d156dbb8fb0e/dubbo-common/src/main/java/org/apache/dubbo/common/BatchExecutorQueue.java#L53). 

1. Old Implemention
<img width="406" height="1010" alt="Image" src="https://github.com/user-attachments/assets/ab4aa053-dbca-4fbe-a88c-b3ec342db955" />

[...]

PR #15734, opened 2025-10-21
BatchExecutorQueue performance risk #15732

fxli4:
# Reference Issue: https://github.com/apache/dubbo/issues/15732

# Double-Queue Alternating Mechanism

## Core Design
The **double-queue alternating mechanism** employs two queues that switch roles between producer and consumer operations:

- **`queue` (Write Queue)**: Active queue where multiple producer threads continuously enqueue new messages
- **`readQueue` (Read Queue)**: Standby queue that serves as the processing buffer for the consumer

### Queue Swapping Operation
```java
private Queue<T> swapQueue() {
    Queue<T> snapshot = queue;  // Capture current write queue
    queue = readQueue;          // Swap: write queue ← read queue  
    readQueue = snapshot;       // Swap: read queue ← captured snapshot
    return snapshot;            // Return accumulated messages for processing
}
```

### Mechanism Flow
1. **Parallel Production**: Multiple producers safely add messages to the volatile `queue` (ConcurrentLinkedQueue)
2. **Atomic Swap**: The single consumer atomically swaps queue references, transferring all accumulated messages
3. **Batch Processing**: Consumer processes the entire batch from the swapped-out queue
4. **Continuous Operation**: Producers continue writing to the fresh `queue` without blocking

## Key Benefits

- **Lock-free Producers**: Writers never block—they always have an active queue available
- **Batch Efficiency**: Consumer processes messages in chunks rather than one-by-one
- **Memory Visibility**: `volatile` modifiers ensure immediate reference update visibility across threads
- **Reduced Contention**: Separation of write and read operations minimizes synchronization overhead

## Thread Safety Guarantees

- **Single Consumer**: `scheduled` AtomicBoolean ensures only one consumer thread processes batches
- **Safe Publication**: Volatile variables provide safe publication of queue references between threads
- **Non-blocking**: Producers continue operating during consumer processing

This design effectively decouples production from consumption while maintaining high throughput under concurrent load.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/BatchExecutorQueue.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/BatchExecutorQueue.java
@@ -22,9 +22,9 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class BatchExecutorQueue<T> {
+
     static final int DEFAULT_QUEUE_SIZE = 128;
-    private volatile Queue<T> queue;
-    private volatile Queue<T> readQueue;
+    private final Queue<T> queue;
     private final AtomicBoolean scheduled;
     private final int chunkSize;
 
@@ -34,7 +34,6 @@ public class BatchExecutorQueue<T> {
 
     public BatchExecutorQueue(int chunkSize) {
         this.queue = new ConcurrentLinkedQueue<>();
-        this.readQueue = new ConcurrentLinkedQueue<>();
         this.scheduled = new AtomicBoolean(false);
         this.chunkSize = chunkSize;
     }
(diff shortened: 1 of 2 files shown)


zrlw:
you'd better submit pr after testing it to ensure everything is ok.
image

fxli4:
> you'd better submit pr after testing it to ensure everything is ok. image

Unit test data samples being too large. It's timeout.

I've using CountDownLatch to rewritten the test unit code.

How should I correct other 'failing check'?

zrlw:
> 
> How should I correct other 'failing check'?

it's none of your business because it might be due to `AWS` network failure since yesterday.
https://health.aws.amazon.com/health/status

zrlw:
it might be more difficult than you thought, the sending operation should be stopped before doing `swap` , otherwise the exception might be thrown:
341c74f8a50f106ac15e3004c708ea4f

fxli4:
> it might be more difficult than you thought, the sending operation should be stopped before doing `swap` , otherwise the exception might be thrown: 341c74f8a50f106ac15e3004c708ea4f

Oh, No, I did't thought Triple Protocol using Http2 Stream Frame.

This PR should probably be closed.



But this code really looks a bit redundant.
```java
            Queue<T> snapshot = new LinkedList<>();
            T item;
            while ((item = queue.poll()) != null) {
                snapshot.add(item);
            }
```

EarthChen (changes requested):
It seems that concurrency issues may arise during the swap process

EarthChen:
> Oh, No, I did't thought Triple Protocol using Http2 Stream Frame.
> 
> This PR should probably be closed.
> 
> But this code really looks a bit redundant.

This copy is to avoid lock contention.

PR #15738, opened 2025-10-23
Fix issue #15698 Distinguish between timeout and task rejection in DefaultFuture

### LINKED ISSUE #15698: Distinguish between timeout and task rejection in DefaultFuture
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Descriptions

see #15671 and #15695
```java
@Override
public void run(Timeout timeout) {
    DefaultFuture future = DefaultFuture.getFuture(requestID);
    if (future == null || future.isDone()) {
        return;
    }
    ExecutorService executor = future.getExecutor();
    if (executor != null && !executor.isShutdown()) {
        try {
            executor.execute(() -> notifyTimeout(future));
        } catch (RejectedExecutionException e) {
            notifyTimeout(future);
            throw e;
        }
    } else {
        notifyTimeout(future);
    }
}
```

We have already solved #15671, but in PR #15695, @RainYuY thinks that we need another PR to change it so that it does not call notifyTimeout, because it’s not actually a timeout.(When error was caught).

Do you think we should do that?
If so, I can work on it.

### Related issues

#15671  
#15695

### Code of Conduct

qxggg:
What is the purpose of the change?
To fix #15698 
This pull request improves the handling of Timeout tasks in DefaultFuture when the associated executor rejects the task due to being full or shut down.

Previously, if the executor rejected the timeout task, a RejectedExecutionException was thrown, which could interrupt the processing flow and was not easily consumable by the user. This change:
	1.	Captures the rejection and generates a proper Response with status SERVER_THREADPOOL_EXHAUSTED_ERROR (100).
	2.	Sends the response back to the caller via DefaultFuture.received(), ensuring the user can detect that the timeout task was not executed.
	3.	Removes the future from FUTURES to prevent resource leakage.

This improves both reliability and user visibility of executor rejection scenarios without throwing exceptions.


qxggg:
Hi, I was wondering if you have some time to review my PR. I’d really appreciate your feedback. Thanks a lot! @zrlw

zrlw:
@RainYuY

dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/support/DefaultFuture.java
```diff
@@ -328,8 +328,7 @@ public void run(Timeout timeout) {
                 try {
                     executor.execute(() -> notifyTimeout(future));
                 } catch (RejectedExecutionException e) {
-                    notifyTimeout(future);
-                    throw e;
+                    notifyExecutionError(future, e);
```
RainYuY, line 331: why not throw e?

qxggg: We catch RejectedExecutionException so that we can handle it internally with notifyExecutionError. This prevents unhandled exceptions from propagating unnecessarily.

Do you think we should still rethrow it to the caller? If so, I can add throw e immediately.

qxggg: @RainYuY  I have added throw e. Thank you!

RainYuY (approved):
LGTM

zrlw (approved):
LGTM

PR #15757, opened 2025-11-06
Fix: Stabilize `URLParam.hashCode()` to prevent Non-Dex nondeterminism in `URLTest.testHashcode`

Anshul-creator:

This PR stabilizes `URLParam.hashCode()` to eliminate Non-Dex flakiness in tests (notably `URLTest.hashCode`).

## Root Cause

The previous implementation multiplied `hashCodeCache` by 31 for each map entry.
Because iteration order over `EXTRA_PARAMS` (a `HashMap`) is nondeterministic, the resulting hash code varied between runs, breaking consistency with `equals()` and causing nondeterministic test failures.

## Changes Made

- Replaced the order-dependent multiplication-based accumulation with an **order-independent summation** of entry.hashCode() values.
- Kept the same handling for `VALUE`, `KEY`, and ignored `TIMESTAMP_KEY`.
- Preserved the existing structure and caching logic.

## Verification

```bash
./mvnw -pl dubbo-common edu.illinois:nondex-maven-plugin:2.2.1:nondex \
  -DnondexRuns=20 \
  -Dtest=org.apache.dubbo.common.URLTest#testHashcode
```

NonDex run logs will be available under the `dubbo-common/.nondex` directory.

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/url/component/URLParam.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/url/component/URLParam.java
@@ -890,15 +890,17 @@ public class URLParam {
     @Override
     public int hashCode() {
         if (hashCodeCache == -1) {
+            int result = 1;
             for (Map.Entry<String, String> entry : EXTRA_PARAMS.entrySet()) {
                 if (!TIMESTAMP_KEY.equals(entry.getKey())) {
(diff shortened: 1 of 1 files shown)


heliang666s:
LTGM

heliang666s (approved):

zrlw (approved):
LGTM

PR #15758, opened 2025-11-06
Fix: Stabilize `ConfigManagerTest.testGetConfigByIdOrName` to prevent Non-Dex nondeterminism 

Anshul-creator:

This PR stabilizes `testGetConfigByIdOrName` by fixing an order-dependent assertion that caused flakiness under randomized execution orders.

## Root Cause

When multiple `ProtocolConfig` objects share the same name, `ConfigManager.getConfig()` throws an `IllegalStateException` whose message lists all matching configs. The order of these instances in the message depends on the iteration order of internal collections, which is non-deterministic. The previous test compared the entire exception message with a hard-coded string, making it fail intermittently when the order of entries changed.

## Changes Made

- Replaced strict `assertEquals` on the exception message with **content-based checks** (`startsWith`, `contains`, `endsWith`).
- Verified that both protocol entries appear in the message, regardless of their order.
- Kept the rest of the test logic unchanged.

## Verification

```bash
./mvnw -q -pl dubbo-common edu.illinois:nondex-maven-plugin:2.2.1:nondex \
  -DnondexRuns=50 \
  -Dtest=org.apache.dubbo.config.context.ConfigManagerTest#testGetConfigByIdOrName
```

NonDex run logs will be available under the `dubbo-common/.nondex` directory.

## Checklist

### DIFF
--- a/dubbo-common/src/test/java/org/apache/dubbo/config/context/ConfigManagerTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/config/context/ConfigManagerTest.java
@@ -410,11 +410,20 @@ class ConfigManagerTest {
             Assertions.fail();
         } catch (Exception e) {
             Assertions.assertTrue(e instanceof IllegalStateException);
-            Assertions.assertEquals(
-                    e.getMessage(),
-                    "Found more than one config by name: dubbo, instances: "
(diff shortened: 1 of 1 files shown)


heliang666s:
LGTM

heliang666s (approved):

zrlw (approved):
LGTM

PR #15766, opened 2025-11-10
Fix: Stabilize `ReferenceConfigTest.testSearchReferences` to prevent Non-Dex nondeterminism in `ReferenceConfigTest.testLargeReferences` 

Anshul-creator:

This PR stabilizes `ReferenceConfigTest.testSearchReferences` to prevent Non-Dex nondeterministic failures in the method `ReferenceConfigTest.testLargeReferences` by making the equality check over the reference collection deterministic.

## Root Cause

The previous implementation was unstable because:
- `references` is backed by an unordered collection.
- `references.iterator().next()` was called inside the stream filter, so each element in the stream could be compared against a potentially different “first” element if the iteration order changed.
- Under NonDex, this led to `results.size()` sometimes being 0, sometimes > 1, causing intermittent failures in `testSearchReferences`, and therefore in `testLargeReferences`.

## Changes Made

Only testSearchReferences was modified:

- Capture the “first” element once, then reuse it throughout the stream
- No longer depend on calling `iterator().next()` repeatedly on an unordered collection inside the filter.
- The timing logic and assertions in both `testLargeReferences` and `testSearchReferences` are left unchanged
- Still asserting that exactly one reference in the large collection is equal to the first one (the intended behavior).

## Verification

```bash
./mvnw -pl dubbo-config/dubbo-config-api \
  -DnondexRuns=50 \
  -Dtest=org.apache.dubbo.config.ReferenceConfigTest#testLargeReferences
```

NonDex run logs will be available under the `dubbo-config/dubbo-config-api/.nondex` directory.

## Checklist

### DIFF
--- a/dubbo-config/dubbo-config-api/src/test/java/org/apache/dubbo/config/ReferenceConfigTest.java
+++ b/dubbo-config/dubbo-config-api/src/test/java/org/apache/dubbo/config/ReferenceConfigTest.java
@@ -58,6 +58,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


zrlw (approved):
LGTM

Anshul-creator:
@RainYuY do you wanna take a look at this one too

RainYuY (approved):
LGTM

PR #15767, opened 2025-11-11
refactor: Avoid calling overridable methods in constructors

redoom:
## Motivation

Avoid Calling Other Methods in Constructors https://github.com/apache/dubbo/issues/15600

## The Problem

```
public abstract class AbstractServer {
    public AbstractServer(URL url, ChannelHandler handler) {
        super(url, handler);
        doOpen();
    }
    protected abstract void doOpen() throws Throwable;
}

public class NettyServer extends AbstractServer {
    private ServerBootstrap bootstrap;  // Still null when doOpen() called

    @Override
    protected void doOpen() {
        bootstrap.group(...);
    }
}
```

## Solution

Two-phase initialization with static factory methods:

```
// ✅ After: Separate construction from initialization
protected AbstractServer(URL url, ChannelHandler handler) {
    super(url, handler);
    // Only field initialization
}

protected final void init() throws RemotingException {
    doOpen();  // object fully constructed
}

public static NettyServer create(URL url, ChannelHandler handler) {
    NettyServer server = new NettyServer(url, handler);
    server.init();
    return server;
}

private NettyServer(URL url, ChannelHandler handler) {
    super(url, ChannelHandlers.wrap(handler, url));
}
```

## Changes

- Split initialization: construction (fields only) + init() (resources)
- Private constructors + static factory methods
- Final init() methods prevent subclass override

```
// ✅
NettyServer server = NettyServer.create(url, handler);

// ❌
// NettyServer server = new NettyServer(url, handler);
```

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/AffinityStateRouter.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/AffinityStateRouter.java
@@ -68,7 +68,15 @@ public class AffinityStateRouter<T> extends AbstractStateRouter<T> {
 
     private final boolean enabled;
 
-    public AffinityStateRouter(URL url) {
+    public static <T> AffinityStateRouter<T> create(URL url) {
+        return new AffinityStateRouter<>(url);
+    }
+
+    public static <T> AffinityStateRouter<T> create(URL url, String affinityKey, Double ratio, boolean enabled) {
+        return new AffinityStateRouter<>(url, affinityKey, ratio, enabled);
+    }
+
+    private AffinityStateRouter(URL url) {
         super(url);
         this.enabled = url.getParameter(ENABLED_KEY, true);
         this.affinityKey = url.getParameter(AFFINITY_KEY, "");
@@ -80,7 +88,7 @@ public class AffinityStateRouter<T> extends AbstractStateRouter<T> {
         }
     }
 
-    public AffinityStateRouter(URL url, String affinityKey, Double ratio, boolean enabled) {
+    private AffinityStateRouter(URL url, String affinityKey, Double ratio, boolean enabled) {
         super(url);
         this.enabled = enabled;
         this.affinityKey = affinityKey;
@@ -92,7 +100,7 @@ public class AffinityStateRouter<T> extends AbstractStateRouter<T> {
         }
     }
 
-    public void init(String rule) {
+    private final void init(String rule) {
         try {
             if (rule == null || rule.trim().isEmpty()) {
                 throw new IllegalArgumentException("Illegal affinity rule!");
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/AffinityStateRouter.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/AffinityStateRouter.java
@@ -69,11 +69,19 @@ public class AffinityStateRouter<T> extends AbstractStateRouter<T> {
     private final boolean enabled;
 
     public static <T> AffinityStateRouter<T> create(URL url) {
-        return new AffinityStateRouter<>(url);
+        AffinityStateRouter<T> router = new AffinityStateRouter<>(url);
+        if (router.enabled) {
+            router.init(router.affinityKey);
+        }
+        return router;
     }
 
     public static <T> AffinityStateRouter<T> create(URL url, String affinityKey, Double ratio, boolean enabled) {
-        return new AffinityStateRouter<>(url, affinityKey, ratio, enabled);
+        AffinityStateRouter<T> router = new AffinityStateRouter<>(url, affinityKey, ratio, enabled);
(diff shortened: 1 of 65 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/AffinityStateRouter.java
```diff
@@ -80,7 +88,7 @@ public AffinityStateRouter(URL url) {
         }
```
zrlw:
```
        if (this.enabled) {
            this.init(affinityKey);  <== called at consturction method
        }
```

zrlw (dismissed):
LGTM

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/config/AffinityListenableStateRouter.java
```diff
@@ -53,15 +53,14 @@ public abstract class AffinityListenableStateRouter<T> extends AbstractStateRout
     private volatile AffinityStateRouter<T> affinityRouter;
     private final String ruleKey;
 
-    public AffinityListenableStateRouter(URL url, String ruleKey) {
+    protected AffinityListenableStateRouter(URL url, String ruleKey) {
         super(url);
         this.setForce(false);
-        this.init(ruleKey);
```
RainYuY, line 59: why not use the same style like AffinityStateRouter.java, I don't think init at AffinityServiceStateRouter is a better way....

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/config/AffinityListenableStateRouter.java
```diff
@@ -85,8 +84,12 @@ public synchronized void process(ConfigChangedEvent event) {
                         e);
             }
         }
+
+        onConfigChange(event);
```
RainYuY: why?

RainYuY (changes requested):
I really don't think putting a mandatory action like init into another class is a good practice.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/config/AffinityListenableStateRouter.java
redoom: It's an abstract class, so I didn't make a factory. I thought about adding an abstract method back then, but I'll  make it a concrete method now.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/config/AffinityListenableStateRouter.java
redoom: I left an extension point because I wasn't sure if flexibility would be needed after making the method final. Would you prefer I remove it？

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/config/AffinityListenableStateRouter.java
redoom: It's an abstract class, so I didn't make a factory.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/config/AffinityListenableStateRouter.java
RainYuY: yes. Now I don't think we need this.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/affinity/config/AffinityListenableStateRouter.java
RainYuY: Maybe we don't need to modify init? Because you've already decorated it with the `final`.

redoom: I didn't quite get what you meant,AffinityServiceStateRouter.java is abstract class , so I put init in the subclasses instead.

> Maybe we don't need to modify init? Because you've already decorated it with the `final`.

RainYuY: I mean I don't know why it's better to put init in the subclasses. Because you've marked init as final.

RainYuY:
I think we shouldn’t initialize objects in their subclasses. The better way, like the code you modified, is to use final or private to restrict them so they can only be initialized within their own class.

redoom:
> I think we shouldn’t initialize objects in their subclasses. The better way, like the code you modified, is to use final or private to restrict them so they can only be initialized within their own class.

Got it

PR #15776, opened 2025-11-17
Fix: Stabilize `XmlSafetyTest` by filtering child `sleep 60` processes using baseline diff

Anshul-creator:

This PR stabilizes `XmlSafetyTest` by removing environment-dependent flakiness in its process-inspection logic, while fully preserving the original test intention of detecting unsafe external command execution during XML deserialization.

## Root Cause

`XmlSafetyTest` verifies that calling `XmlCodec.decode(...)` does not spawn external commands such as `sleep 60`.  
However, the original validation logic had two issues:

1. It checked for `"sleep"` in the entire `ps -e` output *after* execution, without using the `processesBefore` baseline captured in `prepare()`.
2. It monitored **all system processes**, meaning the test failed if any unrelated `sleep` process was created by the OS, CI infrastructure, or other tests running concurrently.

As a result, both `testSafe1` and `testSafe2` intermittently failed under randomized execution orders (NonDex) and on shared machines, even when `XmlCodec` did not spawn a process.

## Changes Made

Only the `ProcessChecker` implementation was modified:

- The current JVM PID is captured in `prepare()`.
- `getProcesses()` was updated to only record processes where:
  - The **parent PID matches the JVM**, ensuring they are child processes, and
  - The command contains `"sleep"` **and** `"60"`, matching the attack payload.
- The baseline comparison remains (`processesBefore` vs `processesAfter`), ensuring the test still fails if XML deserialization launches a new `sleep 60` process.

## Verification

```bash
./mvnw -pl dubbo-remoting/dubbo-remoting-http12 \
  -DnondexRuns=50 \
  -Dtest=org.apache.dubbo.remoting.http12.message.codec.XmlSafetyTest
```

NonDex run logs will be available under the `dubbo-remoting/dubbo-remoting-http12/.nondex` directory.

## Checklist

### DIFF
--- a/dubbo-remoting/dubbo-remoting-http12/src/test/java/org/apache/dubbo/remoting/http12/message/codec/XmlSafetyTest.java
+++ b/dubbo-remoting/dubbo-remoting-http12/src/test/java/org/apache/dubbo/remoting/http12/message/codec/XmlSafetyTest.java
@@ -82,15 +82,41 @@ public class XmlSafetyTest {
 
     static class ProcessChecker {
         Set<String> processesBefore;
+        String currentPid;
 
         public Set<String> getProcesses() {
+            if (currentPid == null) {
+                return Collections.emptySet();
+            }
             try {
                 Set<String> processes = new HashSet<>();
-                Process process = Runtime.getRuntime().exec("ps -e");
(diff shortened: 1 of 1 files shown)


zrlw (approved):
LGTM

heliang666s:
LGTM

heliang666s (approved):

Issue #15783, opened 2025-11-20
[Bug] In Dubbo 3.2.11 and all latest Dubbo 3 versions, the interface-level timeout and retry configurations become invalid after the caller application is restarted.


renyuweighub:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo Java 3.2.11，jdk 1.8，mac os 14.2

### Steps to reproduce this issue

1.First, define and deploy the provider of the RPC interface com.demo.framework.dubbo.demo.api.UdubboDemoService
<br class="Apple-interchange-newline">

<img width="338" height="602" alt="Image" src="https://github.com/user-attachments/assets/da3a6e46-524f-4c6d-93c4-333a7ec9d336" />

2.Then modify the subscriber’s interface-level dynamic configuration for timeout and retries

<img width="792" height="413" alt="Image" src="https://github.com/user-attachments/assets/5936b852-9409-4586-a0a1-6ddde1485f1d" />

3.Use the FORCE_INTERFACE invocation strategy

 dubbo.application.migration.step=FORCE_INTERFACE

4.Start the subscriber application to invoke the provider, and observe the actual timeout and retry configuration values used during the invocation

<img width="1355" height="475" alt="Image" src="https://github.com/user-attachments/assets/f11d677f-ab05-4e03-9fe6-5b88bb11e489" />

### What you expected to happen

I expect the timeout and retry configuration values to match the interface-level dynamic configuration, where the timeout should be 3001 and the retry count should be 4. However, the actual behavior does not meet this expectation. In principle, the caller’s interface-level dynamic configuration should have the highest priority, but after restarting the application, the configuration does not take effect. This issue is consistently reproducible. The configuration only becomes effective if I modify the interface-level dynamic configuration again after the application starts.

### Anything else

This issue of configuration loss upon restart only occurs at the interface-level call. For the consuming application (caller), it is 100% reproducible

The root cause identified through self-investigation is that the configuration becomes invalid because the referenceConfigurationListener is null/empty after starting the subscriber application
[...]

PR #15784, opened 2025-11-20
Fix: the interface-level timeout and retry configurations become invalid after the caller application is restarted

renyuweighub:
Details: By moving the subscription logic after initializing referenceConfigurationListener in the dynamic configuration
Addresses: #15783

### What's the problem

In Dubbo 3.2.11 and all latest Dubbo 3 versions, the interface-level timeout and retry configurations become invalid after the caller application is restarted

### Root cause

The investigation into why the referenceConfigurationListener was null points to a potential timing issue in the RegistryDirectory class's subscribe method: providers are being subscribed to before the referenceConfigurationListener is initialized.

### How to fix

I solve this problem by placing the referenceConfigurationListener initialization logic before the call to super.subscribe(url).

## Checklist


zrlw (approved):
LGTM

RainYuY (approved):
LGTM

PR #15796, opened 2025-11-25
Fix: Override Guava to 32.0.0-jre in parent POM to remediate CVE-2020-8908 and CVE-2023-2976

guptas6est:
This PR overrides **com.google.guava:guava** to 32.0.0-jre in the parent dependencyManagement section.
The goal is to remediate the following reported CVEs:

CVE-2020-8908 – Temporary directory creation vulnerability

CVE-2023-2976 – Insecure temporary directory handling

These CVEs were detected via security scanners in multiple Dubbo modules.
Overriding Guava in the parent POM ensures a consistent, safe version is applied across the entire multi-module build without altering individual submodules.

## Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -186,6 +186,12 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>32.0.0-jre</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
 


pom.xml
```diff
@@ -186,6 +186,12 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+
+      <dependency>
```
zrlw: all dependencies version should be set at dubbo-dependencies-bom/pom.xml

pom.xml
```diff
@@ -186,6 +186,12 @@
...
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>32.0.0-jre</version>
```
zrlw: by the way, current dubbo3 uses protobuf-java-util 3.25.8 which depends on guava 32.0.1-jre.

PR #15806, opened 2025-11-27
fix:openapi

redoom:
## Summary

Fix two critical issues in Dubbo REST OpenAPI schema resolution for generic nested types:

1. **Generic nested type resolution incomplete** - Schemas for intermediate generic types (e.g., `PageResponse` in `ApiResponse<PageResponse<AccessLogRes>>`) were completely missing

2. **Annotated fields disappearing or losing annotation metadata** - Fields with `@Schema` annotations were filtered out, and `description`/`example` properties were not displayed

### Before Fix

```json
{
  "AccessLogRes": {
    "properties": {
      "userId": { "type": "string" },
      "createTime": { "format": "date-time", "type": "string" }
    }
  }
}
```

Missing `id` and `action` fields
No description or example in any field

---

### After Fix

```json
{
  "AccessLogRes": {
    "properties": {
      "id": {
        "type": "string",
        "description": "操作ID",
        "example": "123456"
      },
      "userId": { "type": "string" },
      "action": {
        "type": "string",
        "description": "操作类型",
        "example": "LOGIN"
      },
      "createTime": { "format": "date-time", "type": "string" }
    }
  }
}
```

All 4 fields present
Annotation metadata (description, example) properly displayed

---

## Related Issues

https://github.com/apache/dubbo/issues/15718

### DIFF
--- a/dubbo-plugin/dubbo-rest-openapi/src/main/java/org/apache/dubbo/rpc/protocol/tri/rest/support/swagger/SwaggerOpenAPIDefinitionResolver.java
+++ b/dubbo-plugin/dubbo-rest-openapi/src/main/java/org/apache/dubbo/rpc/protocol/tri/rest/support/swagger/SwaggerOpenAPIDefinitionResolver.java
@@ -242,7 +242,7 @@ public final class SwaggerOpenAPIDefinitionResolver
         schema.setNullable(anno.nullable() ? Boolean.TRUE : null);
         schema.setDeprecated(anno.deprecated() ? Boolean.TRUE : null);
 
-        return chain.resolve(parameter, context);
+        return schema;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


RainYuY:
@heliang666s PTAL

dubbo-plugin/dubbo-rest-openapi/src/main/java/org/apache/dubbo/rpc/protocol/tri/rest/support/swagger/SwaggerOpenAPIDefinitionResolver.java
```diff
@@ -242,7 +242,7 @@ public Schema resolve(ParameterMeta parameter, SchemaContext context, SchemaChai
         schema.setNullable(anno.nullable() ? Boolean.TRUE : null);
         schema.setDeprecated(anno.deprecated() ? Boolean.TRUE : null);
 
-        return chain.resolve(parameter, context);
+        return schema;
```
oxsean, line 245: Why remove it?

dubbo-plugin/dubbo-rest-openapi/src/main/java/org/apache/dubbo/rpc/protocol/tri/rest/support/swagger/SwaggerOpenAPIDefinitionResolver.java
redoom, line 1: Because it has already been called once here, if it is called again, it will be called twice. So I directly returned the schema. I believe the  schema annotation is failing because it's being called an extra time at the end, which prevents the preceding set methods from taking effect.
@oxsean

PR #15821, opened 2025-12-02
fix:the Configuration Cluster and retries

redoom:
## Problem
Previously, Dubbo's cluster fault-tolerance strategy could only be configured on the consumer side via `@DubboReference(cluster = "...")`. Provider-side configuration via `@DubboService(cluster = "...")` was ignored.

## Solution
Modified `RegistryProtocol` to support provider-side cluster configuration with consumer-side override capability:
- Check if consumer has explicitly configured cluster parameter
- If not, fetch cluster configuration from provider's URL metadata
- Consumer config takes precedence over provider config

### Provider side
```java
@DubboService(cluster = "failsafe")
public class GreetingServiceImpl implements GreetingService {
    @Override
        return "Hello " + name;
    }
}
```

### Consumer side (will use provider's failsafe strategy)
```java
@DubboReference
private GreetingService greetingService;
```

### Consumer override (explicit config takes precedence)
```java
@DubboReference(cluster = "failover")
private GreetingService greetingService;
```

## Priority Order
1. Consumer-side explicit configuration (highest priority)
2. Provider-side configuration (fallback)
3. Default strategy: failover

## Related Issues
https://github.com/apache/dubbo/issues/15668


zrlw:
it might be better to add new test cases for this PR.

redoom:
@zrlw  Sorry for the late reply.
I’ve reviewed the three added tests. They cover:

  - provider cluster used when the consumer doesn’t set one,
  - consumer cluster honored when explicitly set,
  - consumer cluster overriding provider hints.

Issue #15742, opened 2025-10-28
[Bug] Found some resources seem not handled properly


AROP-REPLICATION:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubble 3.3.5, java 17

### Steps to reproduce this issue

dubbo-plugin/dubbo-security/src/main/java/org/apache/dubbo/security/cert/DubboCertManager.java#generatePemKey
dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/compressor/Bzip2.java#decompress
dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/compressor/Bzip2.java#compress

### What you expected to happen

close the resource properly

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

RishiMenpara:
Hello, I’m a new contributor to this project and I’m interested in working on this issue. Kindly assign it to me so I can get started. Thank you!

PR #15874, opened 2025-12-15
fix: resolve resource leaks using try-with-resources

heliang666s:
fix: resolve resource leaks using try-with-resources #15742


dubbo-common/src/main/java/org/apache/dubbo/common/io/Bytes.java
```diff
@@ -786,12 +785,8 @@ public static byte[] base642bytes(final String str, final int off, final int len
      */
     public static byte[] zip(byte[] bytes) throws IOException {
         UnsafeByteArrayOutputStream bos = new UnsafeByteArrayOutputStream();
-        OutputStream os = new DeflaterOutputStream(bos);
-        try {
+        try (DeflaterOutputStream os = new DeflaterOutputStream(bos)) {
             os.write(bytes);
-        } finally {
-            os.close();
-            bos.close();
```
RainYuY, line 794: BOS need close too.

RainYuY (changes requested):

RainYuY (dismissed):

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/compressor/Bzip2.java
```diff
@@ -58,15 +58,11 @@ public byte[] compress(byte[] payloadByteArr) throws RpcException {
         }
 
         ByteArrayOutputStream out = new ByteArrayOutputStream();
```
zrlw: ```out``` should be included in try-with-resources block.

zrlw (approved):
LGTM

heliang666s:
@RainYuY please review again!

Issue #15850, opened 2025-12-15
[Bug] Dubbo3.3.2 ，Can not Use broadcast and Tag same time


HanlyL:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.2

### Steps to reproduce this issue

Start 4 service provider instances, each configured with different dubbo.provider.tag (e.g., t-01, t-02...)
Consumer uses the following Reference configuration:
@DubboReference(
    interfaceClass = BpRcvMsgApi.class,
    version = "1.0.0",
    group = "all-t",
    cluster = "broadcast"
)
private ApiTest api;
Invoke api.test() method
Exception thrown: No provider available for the service

### What you expected to happen

The broadcast call should succeed and send requests to all 4 provider instances.

### Anything else

TagStateRouter filters out all providers in the router chain and returns an empty list, causing the call to fail.
The issue is in org.apache.dubbo.rpc.cluster.router.tag.TagStateRouter.filterUsingStaticTag(),if tag is Null:
            result = filterInvoker(
                    invokers, invoker -> StringUtils.isEmpty(invoker.getUrl().getParameter(TAG_KEY)));
When tagRouterRule is not configured, this method performs static tag filtering with exact string matching. Since providers have tag=t-01 and broadcast requests don't specify a tag, the matching fails and eventually returns an empty list.
Can at least one wildcard * be added?like:
result = filterInvoker(invokers, invoker -> 
    "*".equals(tag) || tag.equals(invoker.getUrl().getParameter(TAG_KEY))
);

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

coder-jeffery:
Hi, I’d like to work on this issue. Before submitting a fix, could you confirm whether broadcast invocation is expected to bypass static tag filtering, or whether an explicit wildcard tag should be propagated by the broadcast cluster?

PR #15876, opened 2025-12-15
fix(cluster): support broadcast and wildcard tag routing

heliang666s:
fix(cluster): support broadcast and wildcard tag routing #15850


zrlw (dismissed):
LGTM

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/tag/TagStateRouter.java
```diff
@@ -105,6 +108,17 @@ public BitList<Invoker<T>> doRoute(
             return invokers;
         }
 
+        String tag = StringUtils.isEmpty(invocation.getAttachment(TAG_KEY))
+                ? url.getParameter(TAG_KEY)
+                : invocation.getAttachment(TAG_KEY);
+        if (ANY_VALUE.equals(tag)
+                || (StringUtils.isEmpty(tag) && BROADCAST_CLUSTER.equals(url.getParameter(CLUSTER_KEY)))) {
```
RainYuY: No, we shouldn’t assume that null tags or empty tags can be used to route all invokers. This will cause some incompatible issues, and I believe it’s not in line with common sense.

RainYuY (changes requested):

zrlw (approved):

RainYuY (approved):
Looks fine for me.

PR #15891, opened 2025-12-21
Reuse Nacos NamingService connections across registry groups

### LINKED ISSUE #15624: [Bug] Use Nacos as registry center, every diffrent registry.group eatablished it's own NamingService connection, that causes too many connections to be established
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo Java 3.2.13, JDK 8, Mac OS/Linux

In one Dubbo project use Nacos as registry center, there are several interfaces, and set different `registry.group` for each interface, that will established one connection to Nacos for each interface(registry.group), so it causes too many connections to be established.

I check the code in class `org.apache.dubbo.registry.nacos.NacosRegistryFactory`, in the method `createRegistry` will create a `NamingService` every time.

***Is it neccessary to create a new `NamingService` every time for each different `registry.group` ?***

some related code:

Image

Image

Image

Establish only one NamingService Connection for each `Nacos Host`，not for each `registry.group`

### Anything else

_No response_

### Code of Conduct

somiljain2006:
This change prevents the creation of multiple physical Nacos NamingService connections when a single Dubbo application uses multiple registry groups pointing to the same Nacos server. It introduces a reference-counted cache that reuses a single NamingService instance per unique Nacos connection identity, ensuring connections are only closed when no registries remain.

Fixes #15624 

## Checklist
- [x] Write necessary unit tests to verify your logic correction. If the new feature or significant change is committed, please remember to add a sample in the [dubbo samples](https://github.com/apache/dubbo-samples) project.
- [x] Make sure GitHub actions can pass. [Why the workflow is failing and how to fix it?](../CONTRIBUTING.md)

### DIFF
--- a/dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
+++ b/dubbo-registry/dubbo-registry-nacos/src/main/java/org/apache/dubbo/registry/nacos/NacosRegistry.java
@@ -196,6 +196,13 @@ public class NacosRegistry extends FailbackRegistry {
                     serviceNames.add(compatibleServiceName);
                 }
 
+                /**
+                 *  namingService.registerInstance with
+                 *  {@link org.apache.dubbo.registry.support.AbstractRegistry#registryUrl}
+                 *  default {@link DEFAULT_GROUP}
+                 *
+                 * in https://github.com/apache/dubbo/issues/5978
+                 */
                 for (String service : serviceNames) {
                     namingService.registerInstance(service, getUrl().getGroup(Constants.DEFAULT_GROUP), instance);
                 }
@@ -256,6 +263,16 @@ public class NacosRegistry extends FailbackRegistry {
         try {
             if (isServiceNamesWithCompatibleMode(url)) {
 
+                /**
+                 * Get all instances with serviceNames to avoid instance overwrite and but with empty instance mentioned
+                 * in https://github.com/apache/dubbo/issues/5885 and https://github.com/apache/dubbo/issues/5899
+                 *
+                 * namingService.getAllInstances with
+                 * {@link org.apache.dubbo.registry.support.AbstractRegistry#registryUrl}
+                 * default {@link DEFAULT_GROUP}
+                 *
+                 * in https://github.com/apache/dubbo/issues/5978
+                 */
                 for (String serviceName : serviceNames) {
                     List<Instance> instances = namingService.getAllInstancesWithoutSubscription(
                             serviceName, getUrl().getGroup(Constants.DEFAULT_GROUP));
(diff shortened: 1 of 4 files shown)


RainYuY:
Hmm. I’m wondering if we just remove the call to addParameter(CONFIG_NAMESPACE_KEY) from org.apache.dubbo.registry.nacos.NacosRegistryFactory#createRegistryCacheKey, will we get the same result?

somiljain2006:
No, that would cause different namespaces to collapse into the same cache key. Also, relying solely on RegistryFactory caching is unsafe when multiple services share a Registry instance; one service calling destroy() would physically kill the connection for everyone else. We need the reference counting to manage that lifecycle.

RainYuY:
> No, that would cause different namespaces to collapse into the same cache key. Also, relying solely on RegistryFactory caching is unsafe when multiple services share a Registry instance; one service calling destroy() would physically kill the connection for everyone else. We need the reference counting to manage that lifecycle.

Regarding the first point, I don’t know what problems the same cache key will cause. In fact, I think the same cache key should yield the correct result—maybe there’s some reason behind the special logic. However, I don’t think the second scenario exists. Because as for the registry, I believe we only destroy it when destroying all components; nor do I think there’s any scenario that requires destroying a nacos group. Moreover, the logic should be that one Nacos server corresponds to one registry instance.

RainYuY:
> > No, that would cause different namespaces to collapse into the same cache key. Also, relying solely on RegistryFactory caching is unsafe when multiple services share a Registry instance; one service calling destroy() would physically kill the connection for everyone else. We need the reference counting to manage that lifecycle.
> 
> Regarding the first point, I don’t know what problems the same cache key will cause. In fact, I think the same cache key should yield the correct result—maybe there’s some reason behind the special logic. However, I don’t think the second scenario exists. Because as for the registry, I believe we only destroy it when destroying all components; nor do I think there’s any scenario that requires destroying a nacos group. Moreover, the logic should be that one Nacos server corresponds to one registry instance.

I don’t mean I’m correct. But I think we must find out why Nacos has this special group-related logic, and what scenarios would trigger the destroy logic.

somiljain2006:
That makes sense. I agree we should understand the intent behind Nacos’s group-related logic and the lifecycle assumptions more clearly.

From what I’ve seen, Nacos treats groups as a naming-level concern, while the NamingService itself is bound to a specific server + namespace. That’s why the Nacos API exposes a group on registration/subscription methods rather than requiring a separate NamingService per group.

On the destroy side, the primary scenario guards against is the shutdown race condition. When multiple registry wrappers share a single NamingService, they are destroyed in sequence during application shutdown. Without reference counting, the first registry to close would call shutdown() on the shared connection, instantly killing it.

PR #15926, opened 2025-12-26
Unify graceful shutdown interface,add Triple support

EarthChen:
The read-only logic in the hook is removed due to two reasons: first, the read-only operation and graceful shutdown are already executed during connection closure; second, in the case of multiple protocols sharing a single port, invalid frames will be transmitted, leading to error reports on the peer end.

## Checklist


RainYuY (dismissed):
Please add ut

RainYuY (approved):

zrlw (approved):
LGTM

PR #15934, opened 2025-12-28
[Bug] Fix RpcContext compatibility between Object and String attachments

CuSO41108:
This PR fixes the compatibility issue in `RpcContext` where non-String objects stored via `setObjectAttachment` could not be retrieved via the legacy `getAttachment(String)` method.

## Brief changelog
- Modified `RpcContext#setObjectAttachment` to automatically store a String representation of the object for legacy API compatibility.
- Addressed the TODO: "compatible with previous" in RpcContext.java.

## Verifying this change
I have added a unit test `RpcContextCompatibilityTest` to verify that:
1. `getObjectAttachment` returns the original object (e.g., Long).
2. `getAttachment` returns the String representation (e.g., "12345") instead of null.

I have already sent my ICLA to secretary@apache.org

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcContext.java
+++ b/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcContext.java
@@ -574,6 +574,9 @@ public class RpcContext {
     public RpcContext setObjectAttachment(String key, Object value) {
         // TODO compatible with previous
         CLIENT_ATTACHMENT.get().setObjectAttachment(key, value);
+        if (value != null && !(value instanceof String)) {
+            CLIENT_ATTACHMENT.get().setAttachment(key, String.valueOf(value));
+        }
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


wangchengming666 (approved):
LGTM

RainYuY:
I’m a bit concerned that this PR will alter the existing behavior. Additionally, it’s odd that when I set an Integer value as an attachment and retrieve it using getAttachment (instead of getStringAttachment), I end up getting a String.

CuSO41108:
> I’m a bit concerned that this PR will alter the existing behavior. Additionally, it’s odd that when I set an Integer value as an attachment and retrieve it using getAttachment (instead of getStringAttachment), I end up getting a String.

Hi @RainYuY, thank you for your feedback! 

Regarding your concern, this PR is specifically intended to address the `// TODO: compatible with previous` comment in `RpcContext.java`. 

Currently, the legacy `getAttachment(String key)` method returns `null` or may cause type conflicts when a non-String object is stored via the new `setObjectAttachment` API. This breaks backward compatibility for users who still rely on the legacy String-based API but interact with services using the newer Object-based attachments.

By performing a "read-time conversion" in `getAttachment`:
1. **Backward Compatibility**: Legacy callers get a safe String representation instead of `null`.
2. **Type Safety**: The new `getObjectAttachment` still returns the original type (e.g., Integer/Long) without any loss of information, as the underlying storage remains untouched.
3. **Consistency**: It aligns the behavior of `getAttachment` with the existing `AttachmentsAdapter.ObjectToStringMap` logic.

Do you think we should maintain the current behavior (returning null for non-String objects in the legacy API), or is this dynamic conversion a more robust way to fulfill the "compatible with previous" TODO?

Issue #11157, opened 2022-12-15
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String


asimplepeople:
- [ ] I have searched the [issues](https://github.com/apache/dubbo/issues) of this repository and believe that this is not a duplicate.

### Environment

* Dubbo version: 2.7.3
* Operating System version: xxx
* Java version: xxx

### Steps to reproduce this issue

1. 编写request

public class xxxxxxRequest implements Serializable {
    private static final long serialVersionUID = 1L;
    private List<String> ids;
}
泛化调用
"request": {
      "ids": [1]
    }
2. ![image](https://user-images.githubusercontent.com/26264536/207911883-73482c62-9987-45c7-98fe-5875de59ad90.png)
4. ![image](https://user-images.githubusercontent.com/26264536/207912387-a73c4ab2-480c-4062-ab7b-1aeb920b6a6f.png)

Pls. provide [GitHub address] to reproduce this issue.

### Expected Behavior
正确的抛出类型转换错误而不是在调用for循环的时候抛类型转换错误

### Actual Behavior

If there is an exception, please attach the exception trace:

```
Just put your stack trace here!
```

EarthChen:
plz update latest version to try it.

EarthChen:
plz update latest version to try it.

samyScott:
please assign to me

CrazyHZM:
@samyScott 
Is there any progerss?

ghostly72:
hi, 
I added a pr, trying to resolve this. Can u take a look at it? @CrazyHZM 
PojoUtils.realize() ignored the generic parameter when deserializing Collection fields. As a result, it always converted elements as Object instead of using the declared element type from ParameterizedType. So i added an else block to correct generic element type from ParameterizedType and throw an early IllegalArgumentException if type conversion fails (instead of deferring it to user code). 

And tested it using a unit test, mimicking this situation and got an illegal argument error as expected in the deserialization stage

happy to hear suggestions and improvements
thanks

ghostly72:
hi,
I added a check for early type-validation inside `PojoUtils.realize()` to fix the case where List<String> fields were being deserialized as List<Integer>, leading to ClassCastException at runtime. The fix works for that specific case (now throws illegalargumentexception early), but it broke existing tests like:
```
Map<String, Object> map = PojoUtils.generalize(person);
PojoUtils.realize(map, Person.class);

```
[...]

PR #15963, opened 2026-01-04
 #11157: Throw early type mismatch for List<T> deserialization

ghostly72:
## Checklist
- [x] Make sure gitHub actions can pass. [Why the workflow is failing and how to fix it?](../CONTRIBUTING.md

#11157

During generic invocation deserialization, when a request field is declared as a List<String> but the incoming payload contains numeric elements (e.g., [1]), Dubbo incorrectly realizes it as List<Integer>. This causes a late ClassCastException when iterating over the list inside user code. PojoUtils.realize() ignored the generic parameter when deserializing Collection fields. As a result, it always converted elements as Object instead of using the declared element type from ParameterizedType.

Fix proposed: 
-Added an else block to correct generic element type from ParameterizedType and realize elements using that type
-Throw an early IllegalArgumentException if type conversion fails (instead of deferring it to user code)
- Also added a test case testRealizeListGenericElementType() in PojoUtilsTest.java to verify that invalid element types trigger an early exception.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/PojoUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/PojoUtils.java
@@ -443,14 +443,14 @@ public class PojoUtils {
                     Class<?> keyClazz = obj == null ? null : obj.getClass();
                     if (keyType instanceof Class) {
                         keyClazz = (Class<?>) keyType;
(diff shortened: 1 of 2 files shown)


zrlw:
this approach fundamentally conflicts with Dubbo’s design: Map is taken as generialized type for any java Object.

PR #15970, opened 2026-01-06
Fix AwaitingNonWebApplicationListener await with parent context

### LINKED ISSUE #13722: AwaitingNonWebApplicationListener bug

### Environment

* Dubbo version: 3.2.6
* Operating System version: xxx
* Java version: 21

1. 依赖加入 spring-cloud-starter-bootstrap，没有任何的容器时

Pls. provide [GitHub address] to reproduce this issue.

### Expected Behavior
AwaitingNonWebApplicationListener ，await()方法生效。

### Actual Behavior
AwaitingNonWebApplicationListener ，await()方法不生效。

由于 加入spring-cloud-starter-bootstrap 包时 isRootApplicationContext(applicationContext)存在parent contenxt
If there is an exception, please attach the exception trace:

```
Just put your stack trace here!
```

ishankhemani:
Description

This PR fixes the await logic in AwaitingNonWebApplicationListener for
non-web Spring Boot applications using parent–child ApplicationContext
hierarchies.

Previously, the listener could incorrectly participate in the lifecycle when a
child context was present. The listener should only operate on the root,
non-web ApplicationContext, and must ignore child contexts.

Changes

Ensure onApplicationReadyEvent returns early for:

non-root ApplicationContexts

web ApplicationContexts

Add a regression test to verify correct behavior when a non-web application
has a parent–child context hierarchy

Testing

Added testMultipleContextNonWebApplication

Verified that the listener does not enter await state for child contexts

Fixes #13722

### DIFF
--- a/dubbo-spring-boot-project/dubbo-spring-boot/src/main/java/org/apache/dubbo/spring/boot/context/event/AwaitingNonWebApplicationListener.java
+++ b/dubbo-spring-boot-project/dubbo-spring-boot/src/main/java/org/apache/dubbo/spring/boot/context/event/AwaitingNonWebApplicationListener.java
@@ -112,7 +112,7 @@ public class AwaitingNonWebApplicationListener implements SmartApplicationListen
 
         final ConfigurableApplicationContext applicationContext = event.getApplicationContext();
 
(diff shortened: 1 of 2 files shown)


zrlw:
duplicated with https://github.com/apache/dubbo/pull/15971

PR #15979, opened 2026-01-08
Optimize Zookeeper startup wait in StartZookeeperWindowsProcessor

LI123456mo:
This PR addresses the `// TODO: Help me to optimize the ugly sleep` in `StartZookeeperWindowsProcessor.java`. 

The previous implementation used a hardcoded `TimeUnit.SECONDS.sleep(3)`, which was inefficient and potentially flaky. I have replaced it with a socket-based polling mechanism that:
1. Checks the specific `clientPorts` from the context.
2. Polls every 500ms to see if the port is open.
3. Includes a 10-second timeout to prevent infinite hangs.

## Brief changelog
- Removed the static 3-second sleep.
- Added a polling loop using `java.net.Socket` to verify Zookeeper is ready.
- Maintained existing logging patterns to keep the console output consistent.

## Verifying this change
- Since this is a Windows-specific processor, I have verified the change on a Windows environment.
- Ran `mvn clean compile -pl dubbo-test/dubbo-test-check` to ensure no regressions in the module.

### DIFF
--- a/dubbo-test/dubbo-test-check/src/main/java/org/apache/dubbo/test/check/registrycenter/processor/StartZookeeperWindowsProcessor.java
+++ b/dubbo-test/dubbo-test-check/src/main/java/org/apache/dubbo/test/check/registrycenter/processor/StartZookeeperWindowsProcessor.java
@@ -70,15 +70,32 @@ public class StartZookeeperWindowsProcessor extends ZookeeperWindowsProcessor {
                     .toString());
             context.getExecutorService().submit(() -> executor.execute(cmdLine));
         }
-        try {
-            // TODO: Help me to optimize the ugly sleep.
-            // sleep to wait all of zookeeper instances are started successfully.
-            // The best way is to check the output log with the specified keywords,
-            // however, there maybe keep waiting for check when any exception occurred,
-            // because the output stream will be blocked to wait for continuous data without any break
-            TimeUnit.SECONDS.sleep(3);
-        } catch (InterruptedException e) {
-            // ignored
+
+        // optimized the sleep by pulling the ports untill they are ready.
+        // This avoids waiting the full 3 seconds if the instances start quickly,
+        // and provide a safer timeout mechanism than blocking on log output streams
+        for (int clientPort : context.getClientPorts()) {
+            boolean started = false;
+            long timeout = System.currentTimeMillis() + 10000; // 10 seconds max wait
+
+            while (System.currentTimeMillis() < timeout) {
+                try (java.net.Socket socket = new java.net.Socket("127.0.0.1", clientPort)) {
+                    started = true;
+                    break;
+                } catch (Exception e) {
+                    try {
+                        TimeUnit.MILLISECONDS.sleep(500);
+                    } catch (InterruptedException ie) {
+                        Thread.currentThread().interrupt();
+                        break;
+                    }
+                }
+            }
+            if (started) {
+                logger.info("Zookeeper on port {} is successfully started.", clientPort);
+            } else {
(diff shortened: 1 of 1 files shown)


dubbo-test/dubbo-test-check/src/main/java/org/apache/dubbo/test/check/registrycenter/processor/StartZookeeperWindowsProcessor.java
```diff
@@ -70,15 +70,32 @@ protected void doProcess(ZookeeperWindowsContext context) throws DubboTestExcept
...
-            // ignored
+
+        // optimized the sleep by pulling the ports untill they are ready.
+        // This avoids waiting the full 3 seconds if the instances start quickly,
+        // and provide a safer timeout mechanism than blocking on log output streams
+        for (int clientPort : context.getClientPorts()) {
+            boolean started = false;
+            long timeout = System.currentTimeMillis() + 10000; // 10 seconds max wait
+
+            while (System.currentTimeMillis() < timeout) {
```
zrlw, line 81: it might be better to use wakeup mode, e.g., Condition#await with timeout parameter

LI123456mo: Hi @zrlw, thank you for the feedback! I used the Socket polling approach because we are waiting for an external Zookeeper process to open its port.

Could you clarify if there is a specific Dubbo utility or a shared Condition in ZookeeperWindowsContext I should be using instead? I'm happy to refactor it to match the project's preferred pattern!

zrlw: Since it's just a testcase,  using java.net.Socket#connect with timeout parameter to check port status will be fine if you aren't familiar with asynchronous programming which wakeup mechanism depends on.

dubbo-test/dubbo-test-check/src/main/java/org/apache/dubbo/test/check/registrycenter/processor/StartZookeeperWindowsProcessor.java
```diff
@@ -70,15 +70,32 @@ protected void doProcess(ZookeeperWindowsContext context) throws DubboTestExcept
...
+                try (java.net.Socket socket = new java.net.Socket("127.0.0.1", clientPort)) {
```
zrlw, line 82: Timeout control mechanism is totally meaningless because the ```Socket(host,port)```  method will not return by default until the connection is established or network error occurs.

zrlw:
Thank you very much for your efforts, but before submitting a PR, I think you might need to be more familar with Dubbo processing.

LI123456mo:
Thank you for the honest feedback, @zrlw. I appreciate you being direct with me. As a student, I realize I have gaps in my foundational knowledge that made this task more difficult than it should have been.
Before I go back to study, could you point out the specific areas or concepts I should focus on to meet the standards of the Dubbo project? You mentioned asynchronous programming and wakeup mechanisms—are there other core Java or networking topics you'd recommend I master before contributing again? I want to make sure my next attempt is much more informed.

________________________________

zrlw:
@LI123456mo As for me, ```Dubbo``` isn't beginner-friendly as it might be too complex. There are many ways to start by laying a solid foundation, you might be familar with Dubbo framework by browsing through https://cn.dubbo.apache.org/en/

LI123456mo:
Thank you for clarifying
________________________________

Issue #15484, opened 2025-06-24
[Bug] Use http to call Dubbo Triple Service Returns 404 When Extending Stub Base Class Due to Method Name Case Mismatch


lqscript:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo Java 3.3.5
Jdk 1.8

### Steps to reproduce this issue

Let service extends the stub base class (e.g., DubboGreeterServiceTriple.GreeterServiceImplBase), which is the method described in the [official documentation](https://cn.dubbo.apache.org/en/overview/mannual/java-sdk/tasks/protocols/triple/idl/#:~:text=extend%20the%20generated%20base%20class%20DubboGreeterTriple.GreeterImplBase) for Dubbo Triple services

![Image](https://github.com/user-attachments/assets/acd20e4b-0b49-4db0-a6c4-1f76f5308e14)

### What you expected to happen

I found that this problem does not occur if the service class directly implements the service interface (e.g., `implements GreeterService`) in[ dubbo-demo](https://github.com/apache/dubbo/blob/3.3/dubbo-demo/dubbo-demo-spring-boot-idl/dubbo-demo-spring-boot-idl-provider/src/main/java/org/apache/dubbo/springboot/idl/demo/provider/GreeterServiceImpl.java). 

However, if the service extends the stub base class (e.g., `DubboGreeterServiceTriple.GreeterServiceImplBase`), which is the method described in the [official documentation](https://cn.dubbo.apache.org/en/overview/mannual/java-sdk/tasks/protocols/triple/idl/#:~:text=extend%20the%20generated%20base%20class%20DubboGreeterTriple.GreeterImplBase) for Dubbo Triple services, it returns a 404 error whenever a lowerCamelCase method or an UpperCamelCase method is called.

<img width="771" alt="Image" src="https://github.com/user-attachments/assets/71833d20-b377-4509-abcc-9c6851666d34" />

After debugging the source code, I found that the root cause is related to how Dubbo registers HTTP routes for Triple services.

When a service extends the generated base class, Dubbo uses a `StubServiceDescriptor`. 
org/apache/dubbo/config/ServiceConfig.java:565
[...]

PR #15999, opened 2026-01-13
patch for case mismatch error

ghostly72:
#15484 

When a Dubbo Triple service extends the generated stub base class, HTTP-based requests could return 404 if the service method name uses lowerCamelCase (sayHello) while the generated StubServiceDescriptor stores method keys in UpperCamelCase (SayHello) because StubServiceDescriptor stores entries using method names in UpperCamelCase while DefaultRequestMappingRegistry  returns method names in lowerCamelCase.
As a result, lookups like getMethod("sayHello", ...) returned null, and no HTTP route was registered.

Fix:
Added a fallback lookup to try the name with the first letter capitalized when the initial lookup fails:


zrlw:
https://github.com/apache/dubbo/issues/15484 had been fixed by https://github.com/apache/dubbo/pull/15825

PR #16010, opened 2026-01-17
Fix duplicate reconnect scheduling when provider goes offline

Rafiaminhaj:
What was the problem?
When a provider goes offline, the reconnect timer could be scheduled multiple times without cancelling the previous task. This resulted in duplicate reconnect attempts and unnecessary resource usage.

What is changed?

Improved ReconnectTimerTask to cancel any existing timeout before scheduling a new reconnect.

Ensured reconnect scheduling is properly controlled to avoid duplicate timers.

Updated AbstractClient to align with the corrected reconnect behavior.

Why is this change needed?
This fix prevents repeated reconnect executions, reduces redundant network attempts, and improves stability when handling offline providers.

How was it tested?

Built the project successfully using:
mvnw clean install -DskipTests

Verified no Spotless or formatting violations


zrlw:
1. don't modify file mode
2. don't sumit any PR written totally by AI tools.

PR #16021, opened 2026-01-19
Implement tags filtering logic in TagStateRouter and add unit test for tag routing

RainYuY:
https://github.com/apache/dubbo/issues/4912

## Checklist


RainYuY:
This is designed for a specific scenario: a provider can accept requests from consumers with tag a and tag b (where tags may represent regions), but cannot accept requests from consumers with tag c. In such cases, users can configure the provider to support both tag a and tag b simultaneously to achieve this requirement.

zrlw:
```tags.support``` might be superfluous‌. e.g., multiple groups function has been supported  by setting multiple groups with seperator comma directly in the attribute ```group``` of ```@DubboReference```.

PR #16028, opened 2026-01-20
refactor: enhance backpressure handling with byte tracking 

EarthChen:
Key changes:
- Use numSentBytesQueued counter (AtomicLong) with 32KB threshold
- isReady() returns true when numSentBytesQueued < ON_READY_THRESHOLD
- Trigger onReady when crossing threshold from not-ready to ready
- Add messageSize() method to HttpOutputMessage interface
- Server-side onReadyHandler now executes in business thread via executor

This fixes the issue where initial onReadyHandler was triggered with
isReady()=false, preventing client stream data transmission.

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/stream/AbstractTripleClientStream.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/stream/AbstractTripleClientStream.java
@@ -49,6 +49,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicLong;
 
 import com.google.protobuf.Any;
 import com.google.rpc.DebugInfo;
@@ -90,8 +91,7 @@ public abstract class AbstractTripleClientStream extends AbstractStream implemen
      * Number of bytes currently queued, waiting to be sent.
      * When this falls below ON_READY_THRESHOLD, onReady will be triggered.
      */
-    private final java.util.concurrent.atomic.AtomicLong numSentBytesQueued =
-            new java.util.concurrent.atomic.AtomicLong(0);
+    private final AtomicLong numSentBytesQueued = new AtomicLong(0);
 
     /**
      * The threshold below which isReady() returns true (32KB).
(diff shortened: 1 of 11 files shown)


EarthChen:
# Why change from `channel.isWritable()` to byte counting?

### Background

The original implementation used Netty's `channel.isWritable()` to determine `isReady()` state, which had the following issues:

1. **Timing issue**: When a stream is just created, `channel.isWritable()` may return `false`, causing the initial `onReadyHandler` to fire with `isReady()=false`, preventing the client from sending data.

2. **Unreliable events**: The `channelWritabilityChanged` event only fires when Netty channel's high/low watermarks are crossed, which is unrelated to HTTP/2 flow control windows, potentially causing missed `onReady` callbacks.

3. **gRPC incompatibility**: gRPC uses byte counting approach. This difference may cause interoperability issues with gRPC clients/servers.

### New Implementation

Uses a byte counting approach consistent with gRPC:

- Track pending bytes with `numSentBytesQueued`
- Threshold `ON_READY_THRESHOLD = 32KB` (same as gRPC)
- Initial `numSentBytesQueued=0` ensures `isReady()=true`
- Increment count when sending, decrement after send completes
- Trigger `onReady` when count drops from `>=32KB` to `<32KB`

### Advantages of New Implementation

1. **Deterministic**: Initial state `numSentBytesQueued=0`, `isReady()` is guaranteed to be `true`, avoiding timing issues

2. **Reliable**: `onReady` triggering is fully controlled by byte counting, not dependent on external events, no missed callbacks

3. **Precise Control**: Based on actual bytes sent, not channel state, providing more accurate flow control

4. **Thread-safe**: Uses `AtomicLong` for atomic operations, supports high-concurrency scenarios

5. **Predictable**: Fixed threshold at 32KB, predictable behavior, easier to debug and test

### Limitations and Considerations

1. **Memory overhead**: Each stream maintains an `AtomicLong` counter, but the overhead is minimal (8 bytes)

2. **Network-unaware**: Byte counting only tracks send queue size, not actual network congestion. In extreme cases (e.g., network disconnected), may continue allowing sends until buffer is full

3. **Future callback dependency**: `onSentBytes` relies on write future callbacks. If Netty's future handling is delayed, it may affect `isReady()` accuracy

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/stream/AbstractTripleClientStream.java
```diff
@@ -256,43 +307,23 @@ public boolean isReady() {
...
-        boolean isNowReady = isReady();
-        lastReadyState = isNowReady;
-
-        // Trigger onReady if:
-        // 1. forceNotify is true (initial notification, spurious is OK), or
-        // 2. state changes from "not ready" to "ready"
-        if (forceNotify || (!wasReady && isNowReady)) {
-            listener.onReady();
-        }
+        listener.onReady();
```
RainYuY, line 326: should we check listener would be null?

RainYuY (approved):
LGTM

zrlw (approved):
LGTM

PR #16033, opened 2026-01-22
Fix timestamp filtering in ZooKeeper registry causing pre-warmup failure

funky-eyes:

原代码逻辑是想去除timestamp，pid这种可变的值，剥离后生成出一串不可变的url作为key，进行缓存，避免该实例重启或重新注册到zk，可变值产生变化导致缓存失效，stringUrls中的缓存膨胀的问题，但是这样去除后，在loadbalance进行预热时，由于timestamp被过滤从而无法获取timestamp，导致预热功能失效。
以及修复源代码逻辑存在缺点，比如老的dubbox，dubbo2.6左右的版本存在remote.timestamp，default.remote.timestamp,该逻辑会导致假设url为remote.timestamp=xxxx&revision=xxxx
被过滤后就变成了remote.revision=xxxx，导致revision这个参数直接消失了。

The original logic intended to strip out variable fields like timestamp and pid from the URL, generating a stable, immutable string as the cache key. This prevents cache invalidation when an instance restarts or re-registers to ZooKeeper (due to changing values), and avoids cache bloat in stringUrls.
However, after removing these fields, during load balancing pre-warming, the absence of timestamp means it can no longer be retrieved, causing the pre-warming functionality to fail completely.
Additionally, the original logic has a flaw: in older Dubbo versions (e.g., DubboX or Dubbo 2.6), parameters such as remote.timestamp and default.remote.timestamp exist. When filtering removes timestamp, a URL like remote.timestamp=xxxx&revision=xxxx becomes remote.revision=xxxx, causing the revision parameter to disappear entirely.

## Checklist

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/CacheableFailbackRegistry.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/support/CacheableFailbackRegistry.java
@@ -37,6 +37,7 @@ import org.apache.dubbo.rpc.model.ScopeModel;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


RainYuY (approved):
LGTM

EarthChen (approved):
LGTM

PR #16041, opened 2026-01-22
Unify the deframer of the client and server sides for the Triple protocol

EarthChen:
## Checklist


EarthChen:
The pack and unpack logic still remain to be made compatible in the next PR.

zrlw (dismissed):
LGTM

zrlw (approved):
LGTM

dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/message/StreamingDecoder.java
```diff
@@ -59,6 +62,6 @@ private NoopFragmentListener() {}
         public void bytesRead(int numBytes) {}
 
         @Override
-        public void onFragmentMessage(InputStream rawMessage) {}
+        public void onFragmentMessage(InputStream rawMessage, int messageLength) {}
```
RainYuY, line 65: Why do we need this message length parameter? Shouldn’t we just use the stream’s available length instead? I think this approach would be better and also avoid some security issues.

RainYuY (approved):
LGTM

PR #16055, opened 2026-01-27
feat: enhance VirtualThreadPool with configurable minimum thread count

### LINKED ISSUE #16042: [Feature] Refactor VirtualThreadPool to use a pooled virtual thread model
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Descriptions

Currently, several third-party libraries rely on ThreadLocal-based caching pools for their internal buffers. If Dubbo's thread pool is switched to virtual threads without any pooling mechanism (i.e., creating a new virtual thread for each task), these libraries — such as FastJSON, Aerospike Java client, etc. — will allocate several KB-sized byte buffers in the ThreadLocal of each new virtual thread.
Since virtual threads are not pooled in this scenario, the ThreadLocal caches become ineffective: each virtual thread gets its own fresh cache instance instead of reusing pooled ones. This leads to a massive increase in short-lived object allocations, causing significantly higher GC pressure — even with generational ZGC, the increased GC frequency still puts noticeable load on the CPU.

example：
https://github.com/aerospike/aerospike-client-java/blob/master/client/src/com/aerospike/client/util/ThreadLocalData.java

### Related issues

current
```
public class VirtualThreadPool implements ThreadPool {
    @Override
    public Executor getExecutor(URL url) {
        String name =
                url.getParameter(THREAD_NAME_KEY, (String) url.getAttribute(THREAD_NAME_KEY, DEFAULT_THREAD_NAME));
        return Executors.newThreadPerTaskExecutor(
                Thread.ofVirtual().name(name, 1).factory());
    }
}
```
after
```
public class VirtualThreadPool implements ThreadPool {
    @Override
    public Executor getExecutor(URL url) {
        String name =
                url.getParameter(THREAD_NAME_KEY, (String) url.getAttribute(THREAD_NAME_KEY, DEFAULT_THREAD_NAME));
        return new ThreadPoolExecutor(200, Integer.MAX_VALUE, 60, java.util.concurrent.TimeUnit.SECONDS,
            new SynchronousQueue<>(), Thread.ofVirtual().name(name, 1).factory());
    }
}
```
We can take inspiration from the design of FixedThreadPool, but with the following adjustments for virtual threads:

Set the maximum thread count to Integer.MAX_VALUE
Use a SynchronousQueue as the work queue

This way, when the number of threads is insufficient, the pool can still automatically expand by creating new virtual threads — behaving just like an unpooled virtual thread executor.
At the same time, the keepAliveTime mechanism ensures that excess virtual threads are eventually reclaimed, effectively limiting how many threads get created and preventing unbounded growth.
Since virtual threads are extremely lightweight, pooling a reasonable number of them has negligible impact on memory usage.

### Code of Conduct

funky-eyes:
fixes #16042

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
@@ -126,7 +126,7 @@ public interface CommonConstants {
 
     String THREADS_KEY = "threads";
 
-    String VT_THREADS_KEY = "vt-threads";
+    String THREADS_VIRTUAL_CORE = "threads.virtual.core";
 
     String QUEUES_KEY = "queues";
 
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/constants/CommonConstants.java
@@ -126,6 +126,8 @@ public interface CommonConstants {
 
     String THREADS_KEY = "threads";
 
+    String VT_THREADS_KEY = "vt-threads";
+
     String QUEUES_KEY = "queues";
 
     String ALIVE_KEY = "alive";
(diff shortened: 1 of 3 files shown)


heliang666s (dismissed):
LGTM

RainYuY (approved):
LGTM

zrlw (approved):
LGTM

PR #16056, opened 2026-01-27
test: add unit test for TypeUtils

LI123456mo:
This PR introduces unit test coverage for the TypeUtils interface. Since this utility is fundamental for handling Java Generics and ParameterizedTypes within Dubbo, ensuring its reliability is crucial for data mapping and serialization.

Key Coverage Areas:

Type Validation: Verified isParameterizedType correctly distinguishes between standard classes and generic types.

Class Extraction: Tested getRawClass to ensure base classes are correctly extracted from complex generic signatures.

Generics Discovery: Used static inner classes (e.g., StringList, StringIntegerMap) to verify that findActualTypeArguments accurately identifies generic arguments like <String> across interfaces and superclasses.

Metadata: Validated getClassName for accurate type string representation.

Verification:

Successfully ran mvn clean test -Dtest=TypeUtilsTest -pl dubbo-common.

Verified code style via mvn spotless:apply -pl dubbo-common.

### DIFF
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/utils/TypeUtilsTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/utils/TypeUtilsTest.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyConnectionHandler.java
```diff
@@ -105,6 +105,7 @@ public void channelActive(ChannelHandlerContext ctx) {
     @Override
     public void channelInactive(ChannelHandlerContext ctx) throws Exception {
         super.channelInactive(ctx);
+        // creates the pipeline
```
heliang666s: why add this comment?

heliang666s (changes requested):

dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyConnectionHandler.java
LI123456mo: Hi @heliang666s, sorry about that! I was studying the Netty transport logic and accidentally left that comment in while I was taking notes on the code flow. It doesn't add value to the PR, so I'll remove it. Thanks for catching that!

Issue #15968, opened 2026-01-05
[Bug] Cannot directly call APISIX gateway via @DubboReference annotation (Triple protocol)


jockercheng91:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

- Dubbo Version: 3.3.0
- APISIX Version: Unknown (tested with HTTP/1.1 enabled)

- JDK Version: 1.8.0_452

- Framework: Spring Boot 2.7.18

- Protocol: Triple (tri)

### Steps to reproduce this issue

step1: Deploy APISIX Gateway and configure service discovery via Nacos
Follow the official documentation to deploy APISIX and complete service discovery with Nacos: https://cn.dubbo.apache.org/zh-cn/blog/2024/04/22/%E4%BD%BF%E7%94%A8-apache-apisix-%E4%BB%A3%E7%90%86-dubbo-%E6%9C%8D%E5%8A%A1triple%E5%8D%8F%E8%AE%AE/
Key configurations:
- APISIX listens on 10.216.23.40:8090.
- Route rule: Proxy requests to /com.example.edu.initializer.samples.dubbo3.DemoService/* to the backend Dubbo Triple service (discovered via Nacos).

step2:
Verify APISIX proxy works via curl
Execute the following curl command and get a successful response:
```
curl http://10.216.23.40:8090/com.example.edu.initializer.samples.dubbo3.DemoService/sayHello -i
```
Curl response (successful):
```
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 12
Connection: keep-alive
alt-svc: h2=":24395"
Server: APISIX/3.2.2

"Hello null"
```

step3:Call via @DubboReference annotation and trigger exception
Consumer code with @DubboReference:
```
@DubboReference(
        interfaceClass = DemoService.class,
        protocol = "tri",
        url = "tri://10.216.23.40:8090/com.example.edu.initializer.samples.dubbo3.DemoService/sayHello"
)
private DemoService demoService;
```
Trigger the call in consumer application:
```
String result = demoService.sayHello("test");
```
exception happened：
```
Caused by: org.apache.dubbo.rpc.RpcException: java.util.concurrent.ExecutionException: org.apache.dubbo.rpc.StatusRpcException: CANCELLED
[...]

PR #16058, opened 2026-01-29
Fix Triple client to stop deframing when response headers are invalid

SURYAS1306:

When using Triple through gateways like APISIX, the client may receive a non-gRPC HTTP response (for example `text/plain` like `"Hello"`).  
Previously, the Triple client still initialized the deframer and attempted to decode the payload, which led to parsing errors and `CANCELLED` exceptions.

This change makes the Triple client **fail fast** when response headers are invalid:
- immediately cancels the stream
- prevents deframer creation
- ensures non-gRPC HTTP bodies never enter the gRPC decoder

This avoids incorrect decoding and provides a clearer failure path.

Related issue: #15968

---

## What was changed?

- Added early cancellation when header validation fails in `onHeaderReceived`
- Ensured invalid responses never reach the deframer
- Kept existing DATA-frame guard as a second safety layer

---

## How was this tested?

- `mvn -pl dubbo-rpc/dubbo-rpc-triple -am test`
- Full build on branch `3.3`

---

## Checklist

- [x] Make sure there is a GitHub issue for the change  
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why  
- [ ] Write necessary unit-test to verify your logic correction (not added; transport layer behavior validated by build and manual verification)  
- [x] Make sure GitHub actions can pass

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/stream/AbstractTripleClientStream.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/stream/AbstractTripleClientStream.java
@@ -369,6 +369,7 @@ public abstract class AbstractTripleClientStream extends AbstractStream implemen
         }
 
         void onHeaderReceived(Http2Headers headers) {
+
             if (transportError != null) {
                 transportError.appendDescription("headers:" + headers);
                 return;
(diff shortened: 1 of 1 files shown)


SURYAS1306:
Hi maintainers,

This PR fixes an issue where the Triple client continued deframing even when response headers were invalid (for example, non-gRPC HTTP responses returned by gateways like APISIX).

The client now fails fast when header validation fails and immediately cancels the stream, ensuring invalid responses never reach the gRPC deframer.

Thanks for reviewing.

zrlw:
this PR has nothing to do with https://github.com/apache/dubbo/issues/15968

Issue #16046, opened 2026-01-24
[Bug] Compatibility issues when migrating from Dubbo 2.5.x to 3.3.x


jk-tonycui:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.5, JDK 1.8, Linux

### Steps to reproduce this issue

When migrating Dubbo2.5.x project up to 3.3.x, there are two issues found. 
1. If there are two registry defined, but their properties are all same except the Id of registry.  This is ok with Dubbo 2.5.x, but the project cannot start with Dubbo 3.3.x.

2. If there is dubbo:annoation in the xml config, the project cannot start with Dubbo 3.3.x

### What you expected to happen

It should be smooth during migration up to 3.3.x.

### Do you have a (mini) reproduction demo?

- [ ] Yes, I have a minimal reproduction demo to help resolve this issue more effectively!

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Sreejaan:
Taking this issue. I'll handle the initial compatibility fix first. There are a few more compatibility gaps here that I can address in follow-up PRs.

jk-tonycui:
@Sreejaan hi  Sreejaan, thanks for paying attention on this issue, but I would like to summit PR on my own. It is pleasure to review it together.

zrlw:
The namespace of  ```dubbo:annoation``` had changed since 2.7.x, see details at
https://github.com/apache/dubbo/blob/2.7.x/dubbo-config/dubbo-config-spring/src/main/resources/META-INF/dubbo.xsd

As for me, the ```compat/dubbo.xsd``` at model ```dubbo-config-spring``` defined for migrating from 2.7.x to 3, not for previous versions.

zrlw:
```If there are two registry defined, but their properties are all same except the Id of registry```

Why configure it this way?

jk-tonycui:
> `If there are two registry defined, but their properties are all same except the Id of registry`
> Why configure it this way?

[...]

PR #16059, opened 2026-01-29
Feature 2.5.x compatible fix

jk-tonycui:
fix two compatible issues when migration from dubbo 2.5.x up to 3.3.x.
1. If there are two registry defined, but their properties are all same except the Id of registry. This is ok with Dubbo 2.5.x, but the project cannot start with Dubbo 3.3.x.
2. If there is dubbo:annoation in the xml config, the project cannot start with Dubbo 3.3.x

related issue #16046

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractConfig.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/AbstractConfig.java
@@ -1096,6 +1096,10 @@ public abstract class AbstractConfig implements Serializable {
 
     @Override
     public boolean equals(Object obj) {
+        return equalsWithId(obj, false);
+    }
+
+    public boolean equalsWithId(Object obj, boolean ignoreId) {
         if (obj == null || obj.getClass() != this.getClass()) {
             return false;
         }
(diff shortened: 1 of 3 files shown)


zrlw:
unit tests should be added for the PR.

zrlw:
This pr might cause compatible problems for migrating 2.7.x to 3.3 as the namespace of ``` dubbo:annoation``` had changed since 2.7.x, see details at
https://github.com/apache/dubbo/blob/2.7.x/dubbo-config/dubbo-config-spring/src/main/resources/META-INF/dubbo.xsd
so you'd better modify the xmls by chaning the xml namespace to solve your problem:
```
<?xml version="1.0" encoding="UTF-8"?>
<beans xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xmlns:dubbo="http://dubbo.apache.org/schema/dubbo"
       xmlns="http://www.springframework.org/schema/beans"
       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-4.3.xsd
       http://dubbo.apache.org/schema/dubbo http://dubbo.apache.org/schema/dubbo/dubbo.xsd">
    <dubbo:annotation package="xxx"/>
</beans>
```

zrlw:
This pr violated the consistency rule between method ```equals``` and ```hashCode``` as "getId" is still ignored at later method.

zrlw:
Method ```getId``` was ignored since https://github.com/apache/dubbo/pull/8289 for the sake of performance, so you'd better change your codes by removing redundant registry definitions.

Issue #16054, opened 2026-01-27
[Bug] NettyConnectionHandler.channelInactive() does not remove channel from CHANNEL_MAP causing memory leak


liuhao1117:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

- Dubbo version: dubbo-dubbo.3.2.19
- Java version: 1.8
- OS: Linux
- Protocol: Triple (tri://)

### Steps to reproduce this issue

1. Configure a Dubbo consumer to call a Triple protocol service
2. The Triple service endpoint is unreachable or TLS handshake fails
3. Connection continuously fails and reconnects every second
4. Observe CHANNEL_MAP size growing unbounded

### What you expected to happen

When `channelInactive` is called, the channel should be removed from `CHANNEL_MAP` to prevent memory leak.

`NettyConnectionHandler.channelInactive()` only calls `reconnect()` but does NOT call `NettyChannel.removeChannel()`, causing channels to accumulate in `CHANNEL_MAP`.

**NettyConnectionHandler.java (BUG):**
@Override
public void channelInactive(ChannelHandlerContext ctx) throws Exception {
    super.channelInactive(ctx);
    final Attribute<Boolean> goawayAttr = ctx.channel().attr(GO_AWAY_KEY);
    if (!Boolean.TRUE.equals(goawayAttr.get())) {
        reconnect(ctx.channel());
    }
    // ❌ Missing: NettyChannel.removeChannel(ctx.channel())
}

### Do you have a (mini) reproduction demo?

- [x] Yes, I have a minimal reproduction demo to help resolve this issue more effectively!

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

LI123456mo:
@liuhao1117 Has this been resolved ?
Is it supposed to be "NettyChannel.removeChannel(ch);" ?

SURYAS1306:
Hi @liuhao1117 ,

I’ve investigated this issue and confirmed that the channel cleanup logic has already been added in the current codebase.

[...]

PR #16068, opened 2026-02-03
Add regression test for NettyChannel cleanup on repeated disconnects

SURYAS1306:

This PR adds a regression test to validate the NettyChannel lifecycle cleanup behavior.
The test ensures that channels can be safely removed and re-created across repeated
add/remove cycles, preventing channel map leaks observed in issue #16054.

## What is changed?

- Added a unit test in `NettyChannelTest` to cover repeated channel add/remove scenarios.

## Checklist

- [x] The issue has been investigated and is already fixed in the codebase
- [x] Added a regression test to prevent future regressions
- [x] All tests pass locally

### DIFF
--- a/dubbo-remoting/dubbo-remoting-netty4/src/test/java/org/apache/dubbo/remoting/transport/netty4/NettyChannelTest.java
+++ b/dubbo-remoting/dubbo-remoting-netty4/src/test/java/org/apache/dubbo/remoting/transport/netty4/NettyChannelTest.java
@@ -138,4 +138,24 @@ class NettyChannelTest {
         Assertions.assertEquals(nettyChannel, nettyChannel);
         Assertions.assertNotEquals(nettyChannel, nettyChannel2);
     }
+
+    @Test
+    void testChannelMapCleanupOnRepeatedAddRemove() {
+        Channel channel = Mockito.mock(Channel.class);
+        Mockito.when(channel.isActive()).thenReturn(true);
+
+        URL url = URL.valueOf("test://127.0.0.1/test");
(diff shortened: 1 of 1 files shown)


dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/stream/AbstractTripleClientStream.java
```diff
@@ -390,6 +391,13 @@ void onHeaderReceived(Http2Headers headers) {
             headerReceived = true;
             transportError = validateHeaderStatus(headers);
 
+            if (transportError != null) {
```
zrlw: why add such codes?

SURYAS1306:
Thanks for pointing this out.

That change is not related to the NettyChannel regression test and should not be part of this PR. It was brought in unintentionally from earlier local branch history.

I’ll clean up the PR to keep it focused only on the NettyChannel regression test.

SURYAS1306:
Thanks for the review @zrlw ,

I’ve cleaned up the PR history to remove the unrelated commits.  
This PR now contains only a single commit that adds the NettyChannel regression test
covering repeated add/remove scenarios (no functional changes).

All checks pass locally and in CI.

uuuyuqi:
Hi! Any ETA on when this PR might be merged?

uuuyuqi:
This looks like it's just the UTs. Did the actual code files get left out of the commit by mistake?

Issue #15597, opened 2025-07-30
[Feature] Implement full zero-copy output for Dubbo Triple protocol


heliang666s:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar feature requirement.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Descriptions

The main reason for implementing zero-copy output in the Dubbo Triple protocol is to significantly improve performance: it reduces multiple memory copies, lowers CPU and memory usage, and lessens GC pressure, resulting in lower latency and higher throughput for service communication—making it ideal for high-concurrency and large-scale distributed systems.

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

YouthVoyager:
Hello, I am interested in your work, may I ask if I can join you to help with your work

LI123456mo:
Hi @heliang666s Will like to work on this

LI123456mo:
> ### Pre-check
> * [x]  I am sure that all the content I provide is in English.
> 
> ### Search before asking
> * [x]  I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar feature requirement.
> 
> ### Apache Dubbo Component
> Java SDK (apache/dubbo)
> 
> ### Descriptions
> The main reason for implementing zero-copy output in the Dubbo Triple protocol is to significantly improve performance: it reduces multiple memory copies, lowers CPU and memory usage, and lessens GC pressure, resulting in lower latency and higher throughput for service communication—making it ideal for high-concurrency and large-scale distributed systems.
> 
> ### Related issues
> _No response_
> 
> ### Are you willing to submit a pull request to fix on your own?
> * [x]  Yes I am willing to submit a pull request on my own!
> 
> ### Code of Conduct
> * [x]  I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Hi @heliang666s, I'm looking into this feature. Are we planning to implement this by utilizing Netty's CompositeByteBuf to avoid the extra copy during the header/payload assembly?

Anon2Tokyo:
[...]

PR #16071, opened 2026-02-04
[Feature] Implement Zero-Copy infrastructure in DataQueueCommand #15597

LI123456mo:
**Problem:** Currently, the DataQueueCommand in the Triple protocol uses byte[] for data transfer. This triggers a System.arraycopy inside the writeBytes() call when stitching the frame header with the payload. For large messages or high-throughput scenarios, this constant memory copying creates significant CPU overhead.

Solution: This PR refactors the outbound data path to support Zero-Copy transport:

DataQueueCommand: Updated to use Netty ByteBuf. It now uses CompositeByteBuf to stitch the 5-byte Triple header and the message payload without performing a memory copy.

Transport Bridge: Updated AbstractTripleClientStream to wrap outgoing payloads in Unpooled.wrappedBuffer. This ensures that even though the upper layers still use byte[], the transport layer handles them as buffers, enabling "gathering writes."

Why I limited the scope: I have intentionally avoided modifying the PackableMethod and Compressor interfaces in this PR. Updating them to ByteBuf or ByteBuffer would require breaking public APIs and introducing Netty dependencies into core modules. I believe this PR provides the necessary infrastructure for Zero-Copy first; we can discuss evolving the Serialization/Compression layers in a separate follow-up to keep this change safe and backward-compatible.

Tests: Verified that dubbo-rpc-triple compiles and core transport logic remains intact.

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/command/DataQueueCommand.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/command/DataQueueCommand.java
@@ -19,20 +19,21 @@ package org.apache.dubbo.rpc.protocol.tri.command;
 import org.apache.dubbo.rpc.protocol.tri.stream.TripleStreamChannelFuture;
 
 import io.netty.buffer.ByteBuf;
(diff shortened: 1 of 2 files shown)


EarthChen:
replace https://github.com/apache/dubbo/pull/16072

PR #16116, opened 2026-03-03
fix(common): close input stream in ConfigUtils#loadProperties fallback

### LINKED ISSUE #16098: [Bug] Fix resource leaks and improve code quality in ConfigUtils
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.x, branch 3.3 (source code review)

1. Open `dubbo-common/src/main/java/org/apache/dubbo/common/utils/ConfigUtils.java`
2. In `replaceProperty()` method (line 142): `StringBuffer` is used for a local variable, `StringBuilder` would be more appropriate
3. In `loadMigrationRule()` method: `InputStream` opened via `url.openStream()` is never closed after `readString()` returns — this is a confirmed resource leak
4. In `loadProperties()` multi-file loop: uses nested try-finally with empty catch blocks instead of try-with-resources

1. Local variables should use `StringBuilder` instead of `StringBuffer` when thread-safety is not needed
2. All `InputStream` resources should be properly closed using try-with-resources to prevent file handle leaks
3. Empty catch blocks should be eliminated

### Anything else

The most critical issue is in `loadMigrationRule()` where an InputStream opened via `url.openStream()` is never closed. In long-running applications this can lead to file descriptor exhaustion ("Too many open files" error).

I will submit a PR to fix all these issues.

### Code of Conduct

LouisLau-art:
## What
Fixes an input stream resource leak in `ConfigUtils.loadProperties(...)` when `allowMultiFile == false` and fallback loading path is used.

Previously the code called:
- `properties.load(ClassUtils.getClassLoader().getResourceAsStream(fileName));`

This could leave the stream unclosed. The fallback now uses try-with-resources and null-checks the stream before loading.

## Why
This addresses issue #16098 and keeps resource handling consistent with other load paths in `ConfigUtils`.

## Verification
- `./mvnw -pl dubbo-common -Dtest=ConfigUtilsTest test -DskipITs`
- Result: BUILD SUCCESS (`Tests run: 25, Failures: 0, Errors: 0, Skipped: 1`)

Fixes #16098

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/ConfigUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/ConfigUtils.java
@@ -257,8 +257,10 @@ public class ConfigUtils {
             }
 
             // fall back to use method getResourceAsStream
-            try {
-                properties.load(ClassUtils.getClassLoader().getResourceAsStream(fileName));
+            try (InputStream input = ClassUtils.getClassLoader().getResourceAsStream(fileName)) {
+                if (input != null) {
+                    properties.load(input);
+                }
             } catch (Throwable e) {
                 logger.warn(
                         COMMON_IO_EXCEPTION,


zrlw (approved):
LGTM

PR #16144, opened 2026-03-16
Fix issue 16083: Consumer is not registered as a Spring bean when the Dubbo generic invocation mode is used and the consumer is created with ```@Bean``` annotation

### LINKED ISSUE #16083: [Bug] When Dubbo is used in conjunction with the Spring container, if the Dubbo generic invocation mode is used and the bean is created using the @bean annotation, it can result in the consumer not being registered as a Spring bean.
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

dubbo-dubbo-3.2.9  open jdk 8

https://github.com/apache/dubbo/tree/dubbo-3.2.9

When Dubbo is used in conjunction with the Spring container, if the Dubbo generic invocation mode is used and the bean is created using the @bean annotation, it can result in the consumer not being registered as a Spring bean.

For example, when registering a consumer using Dubbo's generic invocation mode, if the @bean annotation is used to create this bean, it can result in the consumer not being registered as a Spring bean.

The reference.get() method returns a proxy object, and the tClass of this proxy object is 'com.xxx.XxxService'.

Image

The reference.get() method returns a proxy object, and the tClass of this proxy object is 'com.xxx.XxxService'. Additionally, this bean is injected using @Autowired and @qualifier as follows:
@Autowired
@qualifier("XxxService")
private GenericService xxxService;

The code for generating a proxy object by Dubbo is as follows: If com.xxx.XxxService exists in the current code package, but reference.setGeneric(true); is still set, it will cause the actual object of the proxy to be com.xxx.XxxService instead of org.apache.dubbo.rpc.service.GenericService.

Image

spring start error text e.g ：

Unsatisfied dependency expressed through field 'membershipUserTagTestService'; nested exception is org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'membershipUserTagTestServiceImpl': Unsatisfied dependency expressed through field 'aeMarketingStrategyShowFacadeGeneric'; nested exception is org.springframework.beans.factory.BeanNotOfRequiredTypeException: Bean named 'aeMarketingStrategyShowFacadeGeneric' is expected to be of type 'org.apache.dubbo.rpc.service.GenericService' but was actually of type 'com.sun.proxy.$Proxy524'

### Anything else

_No response_

### Code of Conduct

ban-xiu:

Close https://github.com/apache/dubbo/issues/16083

## Checklist


ban-xiu:
In this mode, ensure that invoker.getInterface() (i.e., GenericService) is placed at the forefront.

zrlw:
this pr might cause problem if the dubbo generice service references are created without annotation ```@Bean```.
Dubbo registers reference beans into the Spring container automatically no matter using annotation ```@Bean``` or not.

ban-xiu:
> this pr might cause problem if the dubbo generice service references are created without annotation `@Bean`. Dubbo registers reference beans into the Spring container automatically no matter using annotation `@Bean` or not.

It seems that the only scenario where the interface order of AbstractProxyFactory needs to take effect for Spring is when users entirely bypass ReferenceBean and directly call referenceConfig.get() in a @Bean method—which is precisely the case this PR aims to fix.

ban-xiu:
It seems that scenarios without @Bean won't cause issues either, because those cases are protected by the lazy proxy layer.

ban-xiu:
@zrlw I looked into this concern. This change only reorders two code blocks — the final set of interfaces the proxy implements stays exactly the same, just the iteration   
  order in LinkedHashSet changes.                                                                                                                                       
                  
  My understanding is that for references created without @Bean (via @DubboReference or XML), Spring resolves bean type through ReferenceBean.getObjectType(), which
  returns interfaceClass directly and doesn't depend on proxy interface order. So I think this change should not affect those scenarios. But please correct me if I'm   
  missing something here.

  As for why the reorder helps the @Bean case: Javassist Proxy.getProxy() uses ics[0].getClassLoader() to create the proxy class. Before this change, ics[0] was the    
  user's business interface (e.g. com.xxx.XxxService), whose classloader might not be compatible with GenericService. After this change, ics[0] becomes GenericService, 
  and its classloader (Dubbo's classloader) generally has better visibility.

  Would love to hear your thoughts — happy to adjust the approach if there are cases I haven't considered.

zrlw (approved):
LGTM

zrlw:
@AlbumenJ PTAL

PR #16198, opened 2026-04-09
fix: preserve generic parameter types during deserialization to fix List<Byte> → List<Integer> [3.3]

### LINKED ISSUE #16197: [Bug] List<Byte>, List<Short>, Map<String, Byte> deserialized as List<Integer>, Map<String, Integer> on provider side
### Environment

- Dubbo version: 3.2.x / 3.3.x (all versions affected)
- Protocol: dubbo (hessian2) and triple (hessian2)
- JDK: 17

### Steps to reproduce

1. Define an RPC interface with `List<Byte>` or `Map<String, Byte>` parameters:

```java
public interface GreetingService {
    String processByteCollection(List<Byte> byteList, Map<String, Byte> byteMap);
}
```

2. Consumer sends `List<Byte>` with values like `(byte) 1, (byte) 2, (byte) 127`
3. Provider receives the parameters and checks element types

### Expected behavior

Provider receives `List<Byte>` with elements of type `java.lang.Byte`.

### Actual behavior

Provider receives `List<Integer>` — all `Byte` elements are deserialized as `Integer`. This causes `ClassCastException` when the provider code tries to use the elements as `Byte`.

Same issue affects `List<Short>`, `List<Float>`, `Map<K, Byte>`, `Map<K, Short>`, etc. — any collection/map with narrow number type arguments.

### Root cause

Dubbo passes only the erased `Class` (e.g., `List.class`) to the serialization framework during request deserialization, discarding the generic `Type` (e.g., `List<Byte>`). The serialization framework (hessian2) has no way to know the element type should be `Byte`, so it defaults to `Integer` for small numbers.

Specifically:
- `ReflectionMethodDescriptor` stores `method.getParameterTypes()` (raw `Class[]`) but not `method.getGenericParameterTypes()` (which includes `ParameterizedType`).
- `DecodeableRpcInvocation.drawArgs()` only calls `in.readObject(Class)`, never `in.readObject(Class, Type)`.
- `ReflectionPackableMethod.WrapRequestUnpack` similarly only passes `Class<?>[]` to `MultipleSerialization.deserialize()`.
- `Hessian2ObjectInput.readObject(Class, Type)` ignores the `Type` parameter entirely.

### Tested & verified

- Reproduced with Dubbo `3.3.6` (official release) on both dubbo and triple protocol
- Fix verified with both dubbo and triple protocol via E2E testing
- Verified with Arthas that the new code paths are executed at runtime

uuuyuqi:
Closes #16197

## Summary

When RPC methods have parameters like `List<Byte>`, `Map<String, Byte>`, or `List<Short>`, the provider-side deserialization incorrectly produces `List<Integer>` / `Map<String, Integer>` because the generic type information is discarded during deserialization.

This PR fixes the issue across three layers:

- **Model layer**: Add `getGenericParameterTypes()` to `MethodDescriptor` (default method for backward compat) and implement it in `ReflectionMethodDescriptor` to expose `Type[]` for method parameters.
- **Protocol layer**: Update `DecodeableRpcInvocation.drawArgs()` (Dubbo protocol) and `ReflectionPackableMethod.WrapRequestUnpack` (Triple protocol) to pass generic `Type` to the serialization framework when available.
- **Serialization layer**: Fix `Hessian2ObjectInput.readObject(Class, Type)` to post-convert collection/map elements to the correct narrow number type (`Byte`, `Short`, `Float`).

No wire protocol changes needed. The fix leverages existing local reflection information on the provider side.

## Changes

| Layer | File | Change |
|---|---|---|
| Model | `MethodDescriptor.java` | Add `default getGenericParameterTypes()` |
| Model | `ReflectionMethodDescriptor.java` | Store and return generic parameter types |
| Dubbo Protocol | `DecodeableRpcInvocation.java` | Pass generic `Type` in `drawArgs()` |
| Triple Protocol | `ReflectionPackableMethod.java` | Pass generic `Type` in `WrapRequestUnpack` |
| Serialization API | `MultipleSerialization.java` | Add `deserialize(url, type, clz, genericType, is)` overload |
| Serialization API | `DefaultMultipleSerialization.java` | Implement the new overload |
| Serialization Impl | `Hessian2ObjectInput.java` | Post-convert collection/map elements to narrow types |

## Test plan

- [x] Unit tests for `ReflectionMethodDescriptor.getGenericParameterTypes()`
- [x] Unit tests for `Hessian2ObjectInput.readObject(Class, Type)` with `List<Byte>`, `List<Short>`, `List<Float>`, `Map<String, Byte>`
- [x] E2E tested with Dubbo protocol + hessian2: all Byte elements correctly deserialized
- [x] E2E tested with Triple protocol + hessian2: all Byte elements correctly deserialized
- [x] Verified with Arthas that new code paths are executed at runtime
- [x] Reproduced the bug on official Dubbo 3.3.6, confirmed the fix resolves it

### DIFF
--- a/dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2ObjectInput.java
+++ b/dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2ObjectInput.java
@@ -24,15 +24,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.LinkedList;
-import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
-import java.util.TreeSet;
 
 import com.alibaba.com.caucho.hessian.io.Hessian2Input;
 
@@ -136,88 +128,15 @@ public class Hessian2ObjectInput implements ObjectInput, Cleanable {
             mH2i.setSerializerFactory(hessian2FactoryManager.getSerializerFactory(
                     Thread.currentThread().getContextClassLoader()));
         }
-        T result = readObject(cls);
-        if (type instanceof ParameterizedType && result != null) {
-            result = (T) convertCollectionElementsIfNeeded(result, (ParameterizedType) type);
-        }
-        return result;
-    }
-
-    private Object convertCollectionElementsIfNeeded(Object obj, ParameterizedType type) {
-        Type[] typeArgs = type.getActualTypeArguments();
-
-        if (obj instanceof Collection && typeArgs.length >= 1 && typeArgs[0] instanceof Class) {
-            Class<?> elementType = (Class<?>) typeArgs[0];
-            if (isNarrowNumberType(elementType)) {
-                Collection<?> src = (Collection<?>) obj;
-                Collection<Object> converted = createCompatibleCollection(src, src.size());
-                for (Object e : src) {
-                    converted.add(convertNumber(e, elementType));
-                }
-                return converted;
+        if (type instanceof ParameterizedType) {
+            Type[] typeArgs = ((ParameterizedType) type).getActualTypeArguments();
+            Class<?>[] expectedTypes = new Class<?>[typeArgs.length];
+            for (int i = 0; i < typeArgs.length; i++) {
+                expectedTypes[i] = typeArgs[i] instanceof Class ? (Class<?>) typeArgs[i] : Object.class;
             }
+            return (T) mH2i.readObject(cls, expectedTypes);
         }
-
-        if (obj instanceof Map && typeArgs.length >= 2) {
-            Class<?> keyType = typeArgs[0] instanceof Class ? (Class<?>) typeArgs[0] : null;
-            Class<?> valType = typeArgs[1] instanceof Class ? (Class<?>) typeArgs[1] : null;
-            boolean convertKey = keyType != null && isNarrowNumberType(keyType);
-            boolean convertVal = valType != null && isNarrowNumberType(valType);
-            if (convertKey || convertVal) {
-                Map<Object, Object> src = (Map<Object, Object>) obj;
-                Map<Object, Object> converted = new java.util.LinkedHashMap<>(src.size());
-                for (Map.Entry<Object, Object> entry : src.entrySet()) {
-                    Object k = convertKey ? convertNumber(entry.getKey(), keyType) : entry.getKey();
-                    Object v = convertVal ? convertNumber(entry.getValue(), valType) : entry.getValue();
-                    converted.put(k, v);
-                }
-                return converted;
-            }
-        }
-        return obj;
-    }
-
-    private static boolean isNarrowNumberType(Class<?> type) {
-        return type == Byte.class
-                || type == byte.class
-                || type == Short.class
-                || type == short.class
-                || type == Float.class
-                || type == float.class;
-    }
-
-    private static Object convertNumber(Object value, Class<?> targetType) {
-        if (!(value instanceof Number)) {
-            return value;
-        }
-        Number num = (Number) value;
-        if (targetType == Byte.class || targetType == byte.class) {
-            return num.byteValue();
-        }
-        if (targetType == Short.class || targetType == short.class) {
-            return num.shortValue();
-        }
-        if (targetType == Float.class || targetType == float.class) {
-            return num.floatValue();
-        }
-        return value;
-    }
-
-    @SuppressWarnings({"unchecked", "rawtypes"})
-    private static Collection<Object> createCompatibleCollection(Collection<?> source, int size) {
-        if (source instanceof LinkedList) {
-            return new LinkedList<>();
-        }
-        if (source instanceof LinkedHashSet) {
-            return new LinkedHashSet<>(size);
-        }
-        if (source instanceof TreeSet) {
-            return new TreeSet(((TreeSet) source).comparator());
-        }
-        if (source instanceof Set) {
-            return new HashSet<>(size);
-        }
-        return new ArrayList<>(size);
+        return readObject(cls);
     }
 
     public InputStream readInputStream() throws IOException {
--- a/dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2ObjectInput.java
+++ b/dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2ObjectInput.java
@@ -131,27 +131,30 @@ public class Hessian2ObjectInput implements ObjectInput, Cleanable {
         if (type instanceof ParameterizedType) {
             Type[] typeArgs = ((ParameterizedType) type).getActualTypeArguments();
             Class<?>[] expectedTypes = new Class<?>[typeArgs.length];
-            boolean hasNarrowType = false;
+            boolean hasExpectedType = false;
             for (int i = 0; i < typeArgs.length; i++) {
-                if (typeArgs[i] instanceof Class && isNarrowNumberType((Class<?>) typeArgs[i])) {
+                if (typeArgs[i] instanceof Class && isPrimitive((Class<?>) typeArgs[i])) {
                     expectedTypes[i] = (Class<?>) typeArgs[i];
-                    hasNarrowType = true;
+                    hasExpectedType = true;
                 }
             }
-            if (hasNarrowType) {
+            if (hasExpectedType) {
                 return (T) mH2i.readObject(cls, expectedTypes);
             }
         }
         return readObject(cls);
     }
 
-    private static boolean isNarrowNumberType(Class<?> type) {
-        return type == Byte.class
-                || type == byte.class
+    private static boolean isPrimitive(Class<?> type) {
+        return type.isPrimitive()
+                || type == Boolean.class
+                || type == Character.class
+                || type == Byte.class
                 || type == Short.class
-                || type == short.class
+                || type == Integer.class
+                || type == Long.class
                 || type == Float.class
-                || type == float.class;
+                || type == Double.class;
     }
 
     public InputStream readInputStream() throws IOException {
--- a/dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2ObjectInput.java
+++ b/dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2ObjectInput.java
@@ -22,8 +22,17 @@ import org.apache.dubbo.rpc.model.FrameworkModel;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
+import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
+import java.util.TreeSet;
 
 import com.alibaba.com.caucho.hessian.io.Hessian2Input;
 
@@ -119,6 +128,7 @@ public class Hessian2ObjectInput implements ObjectInput, Cleanable {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public <T> T readObject(Class<T> cls, Type type) throws IOException, ClassNotFoundException {
         if (!Objects.equals(
                 mH2i.getSerializerFactory().getClassLoader(),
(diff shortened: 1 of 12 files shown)


zrlw (dismissed):
LGTM

oxsean (changes requested):
I don't think this is a problem. Could you provide a complete demo to reproduce it? 

When making an RPC call, the type of each value should be included during serialization, rather than being resolved by the provider. For generic calls, the type is handle by PojoUtils.
https://github.com/apache/dubbo/blob/3.3/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/GenericFilter.java

uuuyuqi:
@oxsean Thanks for the review! Here is the complete reproduction demo:

**Demo repo**: https://github.com/uuuyuqi/dubbo3-list-byte-bug-demo

Clone, start Nacos, run `mvn spring-boot:run` for provider & consumer, then `curl http://localhost:8082/testByte`. With Dubbo 3.3.6, all `Byte` elements become `Integer` on the provider side.

### Root cause explained

The issue is specific to **hessian2 serialization** with narrow number types (`Byte`, `Short`, `Float`) inside collections/maps. It is **not** about generic calls or `PojoUtils` — this happens with normal strongly-typed RPC calls.

Here is why:

1. **Hessian2 does not preserve the concrete type for small integers during serialization.** When hessian2 serializes a `Byte` value like `(byte) 1`, it writes it as a compact integer (single-byte encoding `0x91`). There is no type tag distinguishing `Byte` from `Integer` — it's just a number in the wire format. This is by design in the hessian2 protocol for compactness.

2. **During deserialization, hessian2 reads this compact integer and produces `Integer` by default**, because it has no information about the target element type. The call chain is:

   ```
   DecodeableRpcInvocation.drawArgs()
     → in.readObject(List.class)          // only raw Class, no generic info
       → hessian2 deserializes the list
         → reads each element as Integer   // no way to know it should be Byte
   ```

3. **The generic type information IS available** via `Method.getGenericParameterTypes()` on the provider side (local reflection, no wire transmission needed), but Dubbo never passes it to the serialization layer. The `ObjectInput.readObject(Class, Type)` API already exists in Dubbo's serialization interface but was never utilized in the deserialization path.

This fix simply connects the existing dots:
- `ReflectionMethodDescriptor` already calls `method.getParameterTypes()` → now also calls `method.getGenericParameterTypes()`
- `DecodeableRpcInvocation.drawArgs()` already calls `in.readObject(Class)` → now calls `in.readObject(Class, Type)` when generic info is available
- `Hessian2ObjectInput.readObject(Class, Type)` already exists but ignored the `Type` param → now uses it to post-convert narrow number types

No changes to the wire protocol, no impact on non-generic parameters (the `Type` equals the raw `Class` for non-parameterized types, so the original code path is used).

zrlw:
@uuuyuqi  you'd better add testcase that  the interface has complex object which has narrow number bytes fields. e.g.,
```
public interface DemoService {
    void processPerson(Person person);
}

public static class Person {
    String name;
    byte age;
    short height;
    float salary;
}
```

dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2ObjectInput.java
```diff
@@ -119,14 +128,96 @@ public <T> T readObject(Class<T> cls) throws IOException, ClassNotFoundException
...
+    @SuppressWarnings("unchecked")
     public <T> T readObject(Class<T> cls, Type type) throws IOException, ClassNotFoundException {
         if (!Objects.equals(
                 mH2i.getSerializerFactory().getClassLoader(),
                 Thread.currentThread().getContextClassLoader())) {
             mH2i.setSerializerFactory(hessian2FactoryManager.getSerializerFactory(
                     Thread.currentThread().getContextClassLoader()));
         }
-        return readObject(cls);
+        T result = readObject(cls);
```
oxsean: Consider using readObject(Class cl, Class<?>... expectedTypes) to avoid a second conversion.

uuuyuqi:
@zrlw Thanks for the suggestion! I've added a test case with `NarrowNumberPojo` that has `byte age`, `short height`, `float salary` fields, plus `List<Byte> scores` and `Map<String, Byte> attributes` fields.

The test (`testReadObjectWithGenericType_pojoWithNarrowNumberFields`) verifies that:
- Primitive narrow number fields (`byte`, `short`, `float`) are deserialized correctly
- `List<Byte>` and `Map<String, Byte>` fields inside the POJO are also deserialized correctly
- The fix does not affect POJO deserialization in any way

This is expected because when the method parameter is a POJO (e.g., `processPerson(Person person)`), `genericParameterTypes[i]` equals `pts[i]` (both are `Person.class`), so the code falls through to the original `readObject(Class)` path — POJO fields are handled by hessian2's `JavaDeserializer` which already reads field type info from the class definition.

All 16 tests pass locally.

oxsean (dismissed):

uuuyuqi:
Now it‘s using `Hessian2Input.readObject(Class cl, Class<?>... expectedTypes)`, eliminating the need for a secondary conversion. Updated.

zrlw:
jobs/testjob_1-result-java21.txt:[11/35] [dubbo-samples-triple-rest-springmvc:1/1] TEST FAILURE: Run tests failed, version: -Ddubbo.version=3.3.7-SNAPSHOT -Dspring.version=6.1.5 -Djava.version=21, please check logs: /home/runner/work/dubbo/dubbo/2-advanced/dubbo-samples-triple-rest/dubbo-samples-triple-rest-springmvc/target/logs
jobs/testjob_2-result-java21.txt:[11/35] [dubbo-samples-triple-rest-jaxrs:1/1] TEST FAILURE: Run tests failed, version: -Ddubbo.version=3.3.7-SNAPSHOT -Dspring.version=6.1.5 -Djava.version=21, please check logs: /home/runner/work/dubbo/dubbo/2-advanced/dubbo-samples-triple-rest/dubbo-samples-triple-rest-jaxrs/target/logs

zrlw:
2026-04-11T13:28:00.3369064Z org.apache.dubbo.remoting.http12.exception.DecodeException: status=500, Internal Server Error
2026-04-11T13:28:00.3369388Z 	at org.apache.dubbo.rpc.protocol.tri.h12.grpc.GrpcCompositeCodec.decode(GrpcCompositeCodec.java:105)
2026-04-11T13:28:00.3369710Z 	at org.apache.dubbo.rpc.protocol.tri.h12.grpc.GrpcCompositeCodec.decode(GrpcCompositeCodec.java:111)
2026-04-11T13:28:00.3370019Z 	at org.apache.dubbo.remoting.http12.message.HttpMessageDecoder.decode(HttpMessageDecoder.java:56)
2026-04-11T13:28:00.3370387Z 	at org.apache.dubbo.remoting.http12.message.DefaultListeningDecoder.decode(DefaultListeningDecoder.java:41)
2026-04-11T13:28:00.3371064Z 	at org.apache.dubbo.rpc.protocol.tri.h12.http2.GenericHttp2ServerTransportListener$DefaultFragmentListener.onFragmentMessage(GenericHttp2ServerTransportListener.java:131)
2026-04-11T13:28:00.3371666Z 	at org.apache.dubbo.remoting.http12.message.LengthFieldStreamingDecoder.invokeListener(LengthFieldStreamingDecoder.java:257)
2026-04-11T13:28:00.3372305Z 	at org.apache.dubbo.rpc.protocol.tri.h12.grpc.GrpcHttp2ServerTransportListener$DetermineMethodDescriptorListener.onFragmentMessage(GrpcHttp2ServerTransportListener.java:212)
2026-04-11T13:28:00.3372740Z 	at org.apache.dubbo.remoting.http12.message.LengthFieldStreamingDecoder.invokeListener(LengthFieldStreamingDecoder.java:257)
2026-04-11T13:28:00.3373164Z 	at org.apache.dubbo.remoting.http12.message.LengthFieldStreamingDecoder.processBody(LengthFieldStreamingDecoder.java:249)
2026-04-11T13:28:00.3373574Z 	at org.apache.dubbo.remoting.http12.message.LengthFieldStreamingDecoder.deliver(LengthFieldStreamingDecoder.java:172)
2026-04-11T13:28:00.3373971Z 	at org.apache.dubbo.remoting.http12.message.LengthFieldStreamingDecoder.decode(LengthFieldStreamingDecoder.java:79)
2026-04-11T13:28:00.3374517Z 	at org.apache.dubbo.rpc.protocol.tri.h12.grpc.GrpcHttp2ServerTransportListener$LazyFindMethodListener.onMessage(GrpcHttp2ServerTransportListener.java:180)
2026-04-11T13:28:00.3374949Z 	at org.apache.dubbo.rpc.protocol.tri.h12.AbstractServerTransportListener.doOnData(AbstractServerTransportListener.java:183)
2026-04-11T13:28:00.3375394Z 	at org.apache.dubbo.rpc.protocol.tri.h12.AbstractServerTransportListener.lambda$onData$1(AbstractServerTransportListener.java:168)
2026-04-11T13:28:00.3375715Z 	at org.apache.dubbo.common.threadpool.serial.SerializingExecutor.run(SerializingExecutor.java:111)
2026-04-11T13:28:00.3375996Z 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
2026-04-11T13:28:00.3376269Z 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
2026-04-11T13:28:00.3376537Z 	at org.apache.dubbo.common.threadlocal.InternalRunnable.run(InternalRunnable.java:39)
2026-04-11T13:28:00.3376670Z 	at java.base/java.lang.Thread.run(Thread.java:1583)
2026-04-11T13:28:00.3377099Z Caused by: com.alibaba.com.caucho.hessian.io.HessianProtocolException: expected map/object at org.apache.dubbo.rest.demo.pojo.User
2026-04-11T13:28:00.3377392Z 	at com.alibaba.com.caucho.hessian.io.AbstractDeserializer.error(AbstractDeserializer.java:141)
2026-04-11T13:28:00.3377737Z 	at com.alibaba.com.caucho.hessian.io.AbstractMapDeserializer.readObject(AbstractMapDeserializer.java:68)
2026-04-11T13:28:00.3378091Z 	at com.alibaba.com.caucho.hessian.io.CollectionDeserializer.readLengthList(CollectionDeserializer.java:121)
2026-04-11T13:28:00.3378351Z 	at com.alibaba.com.caucho.hessian.io.Hessian2Input.readObject(Hessian2Input.java:2347)
2026-04-11T13:28:00.3378765Z 	at org.apache.dubbo.common.serialize.hessian2.Hessian2ObjectInput.readObject(Hessian2ObjectInput.java:137)
2026-04-11T13:28:00.3379306Z 	at org.apache.dubbo.common.serialize.DefaultSerializationExceptionWrapper$ProxyObjectInput.readObject(DefaultSerializationExceptionWrapper.java:170)
2026-04-11T13:28:00.3379761Z 	at org.apache.dubbo.common.serialize.DefaultMultipleSerialization.deserialize(DefaultMultipleSerialization.java:58)
2026-04-11T13:28:00.3380236Z 	at org.apache.dubbo.rpc.protocol.tri.ReflectionPackableMethod$WrapRequestUnpack.deserializeFromWrapper(ReflectionPackableMethod.java:521)
2026-04-11T13:28:00.3380632Z 	at org.apache.dubbo.rpc.protocol.tri.ReflectionPackableMethod$WrapRequestUnpack.unpack(ReflectionPackableMethod.java:505)
2026-04-11T13:28:00.3380852Z 	at org.apache.dubbo.rpc.model.WrapperUnPack.unpack(WrapperUnPack.java:37)
2026-04-11T13:28:00.3381098Z 	at org.apache.dubbo.rpc.model.PackableMethod.parseRequest(PackableMethod.java:73)
2026-04-11T13:28:00.3381414Z 	at org.apache.dubbo.rpc.protocol.tri.h12.grpc.GrpcCompositeCodec.decode(GrpcCompositeCodec.java:101)

RainYuY:
I'm sorry, but I don't think this problem needs to be fixed. Because the real reason why we cannot serialize it is that the Hessian types do not include byte. So the byte and Integer use the same tag value. If you fix this one, it will have the same problem like `List<List<Integer>>`, so I think the best way is to fix it in the Hessian protocol (hard), or change the serialization protocol.

zrlw:
> I'm sorry, but I don't think this problem needs to be fixed. Because the real reason why we cannot serialize it is that the Hessian types do not include byte. So the byte and Integer use the same tag value. If you fix this one, it will have the same problem like `List<List<Integer>>`, so I think the best way is to fix it in the Hessian protocol (hard), or change the serialization protocol.

how about ```List<Map<String, Byte>>``` or ```Map<String, List<Byte>>```?

uuuyuqi:
@zrlw Fixed. The previous implementation passed all generic type arguments as `expectedTypes` to `readObject(Class, Class<?>...)`, but the `expectedTypes` mechanism in hessian-lite only works for primitive types. For complex POJOs (like `User`), it bypasses tag-driven deserialization and causes `HessianProtocolException`.

Now only `isPrimitive()` types are passed as `expectedTypes`, consistent with hessian-lite's `FieldDeserializer2Factory`. POJO types fall back to default tag-driven deserialization and are not affected.

uuuyuqi:
@RainYuY You're right that Hessian protocol doesn't have a native byte type — byte and int share the same tag. But hessian-lite already handles this through `readObject(Class, Class<?>... expectedTypes)`: when `CollectionDeserializer` receives expectedTypes, it calls `Number.byteValue()` / `Number.shortValue()` etc. to narrow the type.

The issue was that Dubbo's invocation chain didn't pass generic type info down to hessian-lite, so this existing conversion capability was never triggered. This fix simply propagates the generic parameter types that are already available via reflection, letting hessian-lite's own type narrowing logic work as designed. No new conversion logic is introduced.

For nested generics like `List<List<Byte>>`, yes, that's a limitation of hessian-lite's `expectedTypes` which only supports one level of type arguments. But that's a separate issue and not in scope here — this fix addresses the common single-level cases (`List<Byte>`, `Map<String, Byte>`, etc.) which are the most frequently encountered.

zrlw (approved):

PR #16199, opened 2026-04-09
fix: preserve generic parameter types during deserialization to fix List<Byte> → List<Integer> [3.2]

### LINKED ISSUE #16197: [Bug] List<Byte>, List<Short>, Map<String, Byte> deserialized as List<Integer>, Map<String, Integer> on provider side

uuuyuqi:
Closes #16197

## Summary

Backport of #16198 to the 3.2 branch.

When RPC methods have parameters like `List<Byte>`, `Map<String, Byte>`, or `List<Short>`, the provider-side deserialization incorrectly produces `List<Integer>` / `Map<String, Integer>` because the generic type information is discarded during deserialization.

This PR fixes the issue across three layers:

- **Model layer**: Add `getGenericParameterTypes()` to `MethodDescriptor` (default method for backward compat) and implement it in `ReflectionMethodDescriptor` to expose `Type[]` for method parameters.
- **Protocol layer**: Update `DecodeableRpcInvocation.drawArgs()` (Dubbo protocol) and `ReflectionPackableMethod.WrapRequestUnpack` (Triple protocol) to pass generic `Type` to the serialization framework when available.
- **Serialization layer**: Fix `Hessian2ObjectInput.readObject(Class, Type)` to post-convert collection/map elements to the correct narrow number type (`Byte`, `Short`, `Float`).

No wire protocol changes needed. The fix leverages existing local reflection information on the provider side.

## Changes

| Layer | File | Change |
|---|---|---|
| Model | `MethodDescriptor.java` | Add `default getGenericParameterTypes()` |
| Model | `ReflectionMethodDescriptor.java` | Store and return generic parameter types |
| Dubbo Protocol | `DecodeableRpcInvocation.java` | Pass generic `Type` in `drawArgs()` |
| Triple Protocol | `ReflectionPackableMethod.java` | Pass generic `Type` in `WrapRequestUnpack` |
| Serialization API | `MultipleSerialization.java` | Add `deserialize(url, type, clz, genericType, is)` overload |
| Serialization API | `DefaultMultipleSerialization.java` | Implement the new overload |
| Serialization Impl | `Hessian2ObjectInput.java` | Post-convert collection/map elements to narrow types |

## Test plan

- [x] Unit tests for `ReflectionMethodDescriptor.getGenericParameterTypes()`
- [x] Unit tests for `Hessian2ObjectInput.readObject(Class, Type)` with `List<Byte>`, `List<Short>`, `List<Float>`, `Map<String, Byte>`
- [x] E2E tested with Dubbo protocol + hessian2: all Byte elements correctly deserialized
- [x] E2E tested with Triple protocol + hessian2: all Byte elements correctly deserialized
- [x] Verified with Arthas that new code paths are executed at runtime
- [x] Reproduced the bug on official Dubbo 3.3.6, confirmed the fix resolves it

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/rpc/model/MethodDescriptor.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/rpc/model/MethodDescriptor.java
@@ -32,6 +32,18 @@ public interface MethodDescriptor {
 
     Class<?>[] getParameterClasses();
 
+    /**
+     * Retrieves the generic parameter types of the method.
+     * <p>
+     * For parameterized parameters like {@code List<Byte>}, this returns the
+     * {@link java.lang.reflect.ParameterizedType} instead of just the raw {@code Class}.
+     *
+     * @return the generic parameter types
+     */
+    default Type[] getGenericParameterTypes() {
+        return getParameterClasses();
+    }
+
     Class<?> getReturnClass();
 
     Type[] getReturnTypes();
--- a/dubbo-common/src/main/java/org/apache/dubbo/rpc/model/ReflectionMethodDescriptor.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/rpc/model/ReflectionMethodDescriptor.java
@@ -41,6 +41,7 @@ public class ReflectionMethodDescriptor implements MethodDescriptor {
     public final String methodName;
     private final String[] compatibleParamSignatures;
     private final Class<?>[] parameterClasses;
+    private final Type[] genericParameterTypes;
     private final Class<?> returnClass;
     private final Type[] returnTypes;
     private final String paramDesc;
@@ -52,6 +53,7 @@ public class ReflectionMethodDescriptor implements MethodDescriptor {
         this.method = method;
         this.methodName = method.getName();
         this.parameterClasses = method.getParameterTypes();
+        this.genericParameterTypes = method.getGenericParameterTypes();
         this.returnClass = method.getReturnType();
         Type[] returnTypesResult;
         try {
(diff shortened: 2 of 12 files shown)


zrlw:
3.2 might not be maintained anymore.

PR #16200, opened 2026-04-09
Create QuicSslEngine for client with parameter peer host to meet HostnameChecker matching rules

zrlw:
1. Since Netty 4.2.11.Final, hostname checker was enabled by default, the checker will get expected hostname from the session created by client QuicSslEngine, but current dubbo created client QuicSslEngine without parameter ```peerhost``` and the CertificateException will be thrown at ```sun.security.util.HostnameChecker#match``` due to null ```expectedName``` ,
```
    public void match(String expectedName, X509Certificate cert,
                      boolean chainsToPublicCA) throws CertificateException {
        if (expectedName == null) {
            throw new CertificateException("Hostname or IP address is " +
                    "undefined.");
        }
        if (isIpAddress(expectedName)) {
           matchIP(expectedName, cert);
        } else {
           matchDNS(expectedName, cert, chainsToPublicCA);
        }
    }
```
so we should create QuicSslEngine for client with parameter peer host to support hostname checker.

2. The ```server.pem``` of ```TripleHttp3ProtocolTest``` was changed by adding SAN (set ip address to 127.0.0.1) configuration to meet HostnameChecker matching rules.

## Checklist


zrlw:
@RainYuY

PR #16202, opened 2026-04-09
Close channel on TLS exception to prevent half-open connections

### LINKED ISSUE #16201: [Bug] SslServerTlsHandler.exceptionCaught() swallows exceptions without closing channel, causing permanent half-open TLS connections
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo Java 3.3 (also affects 3.2.x). Netty 4.1.x.

**Scenario**: Provider has a broken Netty dependency (e.g., incompatible `netty-buffer` version causing `NoClassDefFoundError: Could not initialize class io.netty.buffer.PooledUnsafeDirectByteBuf`).

1. Consumer connects to Provider — TCP three-way handshake succeeds (handled by OS kernel, unaffected by the Netty bug).
2. `NettyClient.doConnect()` only waits for TCP handshake completion, so it considers the connection successful. `DubboInvoker` is created and added to `validInvokers`.
3. TLS handshake begins asynchronously — Consumer sends `ClientHello`.
4. Provider's Netty read loop tries to allocate a `ByteBuf` to read the incoming data → `NoClassDefFoundError` is thrown.
5. Netty's `NioByteUnsafe.handleReadException()` fires `pipeline.fireExceptionCaught(cause)` but does **not** close the channel (Netty only auto-closes for `IOException` or `OutOfMemoryError`).
6. The exception reaches `SslServerTlsHandler.exceptionCaught()`, which **only logs the error** — it neither closes the channel nor propagates the exception.
7. The channel remains TCP-active but is completely non-functional at the application layer.
8. Consumer's `DubboInvoker.isAvailable()` returns `true` (it only checks `channel.isActive()`), so the invoker is never removed from `validInvokers`.
9. All RPC requests routed to this Provider time out after 10 seconds.

When `SslServerTlsHandler.exceptionCaught()` is invoked, the channel should be closed (via `ctx.close()`), just like the `userEventTriggered()` method in the **same class** already does on TLS handshake failure. This would allow:

- The Consumer to detect `channelInactive` → `isConnected()=false` → `isAvailable()=false`
- Dubbo's `addInvalidateInvoker` mechanism to remove the broken invoker from `validInvokers`
- The self-healing loop to work as designed

**Current behavior of `SslServerTlsHandler.exceptionCaught()`** (line 60-68):
```java
@Override
public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
    logger.error(INTERNAL_ERROR, "unknown error in remoting module", "",
            "TLS negotiation failed when trying to accept new connection.", cause);
    // BUG: no ctx.close() and no ctx.fireExceptionCaught(cause)
    // The exception is silently swallowed, channel stays open but broken
}
```

**Compare with `userEventTriggered()` in the same class** (line 81-89), which correctly closes the channel:
```java
} else {
    logger.error(INTERNAL_ERROR, "", "",
            "TLS negotiation failed when trying to accept new connection.",
            handshakeEvent.cause());
    ctx.close();  // ← correctly closes the channel
}
```

Similarly, `SslClientTlsHandler.userEventTriggered()` on the Consumer side fires `ctx.fireExceptionCaught()` on TLS failure but does not close the channel, which can also lead to half-open connections.

### Anything else

**Root cause analysis:**

The exception propagation chain breaks at `SslServerTlsHandler.exceptionCaught()`:

```
Netty read loop: allocate ByteBuf → NoClassDefFoundError
    ↓
NioByteUnsafe.handleReadException() → pipeline.fireExceptionCaught(cause)
    (Netty does NOT auto-close: NoClassDefFoundError is not IOException/OutOfMemoryError)
    ↓
SslServerTlsHandler.exceptionCaught() → logs error, BUT:
    ✗ Does NOT call ctx.close()
    ✗ Does NOT call ctx.fireExceptionCaught(cause)
    → Exception is silently swallowed
    → Channel remains TCP-active but application-dead
    ↓
NettyServerHandler.exceptionCaught() → NEVER reached (exception stopped above)
    ↓
Consumer side: channel still active → isAvailable()=true → invoker never removed
    → Continuous timeout on every RPC call routed to this Provider
```

This is not limited to `NoClassDefFoundError` — any non-`IOException`/non-`OutOfMemoryError` exception during the Netty read loop would trigger the same behavior, leaving the channel in a zombie state.

### Code of Conduct

uuuyuqi:

Fixes #16201

`SslServerTlsHandler.exceptionCaught()` only logged errors without closing the channel or propagating the exception. This caused channels to remain TCP-active but application-dead (half-open) when non-IOException/non-OutOfMemoryError exceptions (e.g., `NoClassDefFoundError`) occurred during Netty's read loop.

**Impact**: Consumer-side `DubboInvoker.isAvailable()` always returned `true` for these broken connections (it only checks TCP-level `channel.isActive()`), so the `addInvalidateInvoker` mechanism never triggered. The broken Provider was never removed from `validInvokers`, causing continuous RPC timeouts.

**Changes**:
- **`SslServerTlsHandler.exceptionCaught()`**: Added `ctx.close()` after logging, consistent with the `userEventTriggered()` method in the same class which already closes the channel on TLS handshake failure.
- **`SslClientTlsHandler.userEventTriggered()`**: Changed from `ctx.fireExceptionCaught()` to `ctx.close()` on handshake failure, consistent with server-side behavior and ensuring the channel is properly closed so the Consumer can detect the failure.

## Checklist
- [x] Make sure there is a [GitHub_issue](https://github.com/apache/dubbo/issues) field for the change. (#16201)

### DIFF
--- a/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/ssl/SslClientTlsHandler.java
+++ b/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/ssl/SslClientTlsHandler.java
@@ -70,7 +70,7 @@ public class SslClientTlsHandler extends ChannelInboundHandlerAdapter {
                         "",
                         "TLS negotiation failed when trying to accept new connection.",
                         handshakeEvent.cause());
-                ctx.fireExceptionCaught(handshakeEvent.cause());
+                ctx.close();
             }
         }
     }
--- a/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/ssl/SslClientTlsHandler.java
+++ b/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/ssl/SslClientTlsHandler.java
@@ -68,8 +68,9 @@ public class SslClientTlsHandler extends ChannelInboundHandlerAdapter {
                         INTERNAL_ERROR,
                         "unknown error in remoting module",
                         "",
-                        "TLS negotiation failed when trying to accept new connection.",
+                        "TLS negotiation failed when connecting to remote server.",
                         handshakeEvent.cause());
+                ctx.fireExceptionCaught(handshakeEvent.cause());
                 ctx.close();
             }
         }
--- a/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/ssl/SslServerTlsHandler.java
+++ b/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/ssl/SslServerTlsHandler.java
@@ -65,6 +65,7 @@ public class SslServerTlsHandler extends ByteToMessageDecoder {
                 "",
                 "TLS negotiation failed when trying to accept new connection.",
                 cause);
+        ctx.close();
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


zrlw (dismissed):
LGTM

chickenlj (approved):
LGTM.

zrlw (approved):

PR #16211, opened 2026-04-13
[Bug] Fix TOCTOU race condition in ZookeeperClientManager.getInstance()

LiyunZhang10:
Fixes a Time-Of-Check to Time-Of-Use (TOCTOU) race condition in `ZookeeperClientManager.getInstance()`.

In the current implementation, `getInstance(ApplicationModel)` uses a non-atomic check-then-act pattern (`containsKey` followed by `put`) on a `ConcurrentHashMap`. 

When multiple threads call `getInstance()` concurrently for the first time with the same `ApplicationModel`, each thread can bypass the `containsKey` check, create a separate `ZookeeperClientManager` instance, and register duplicate `addDestroyListener` callbacks. The later threads will overwrite the earlier instances in `managerMap`, causing the overwritten instances to be lost but their destroy listeners to remain registered, which leads to redundant Zookeeper client cleanup operations and resource waste.

## What is changed?
- Replaced the `containsKey` + `put` pattern with `managerMap.computeIfAbsent()` to ensure that exactly one manager instance and exactly one destroy listener are created per `ApplicationModel`, regardless of concurrency.
- Added `testGetInstanceConcurrentRaceCondition` to verify that concurrent calls safely return exactly 1 instance and create only 1 entry in the map.

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/test/java/org/apache/dubbo/registry/client/AbstractServiceDiscoveryFactoryTest.java
+++ b/dubbo-registry/dubbo-registry-api/src/test/java/org/apache/dubbo/registry/client/AbstractServiceDiscoveryFactoryTest.java
@@ -22,6 +22,7 @@ import org.apache.dubbo.rpc.model.ApplicationModel;
 
 import java.util.List;
 
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Assertions;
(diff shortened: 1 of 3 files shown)


zrlw:
run mvn spotless:apply to format your pr codes

zrlw (approved):
LGTM

PR #16236, opened 2026-04-27
Improve generic type safety of CollectionUtils.sort

### LINKED ISSUE #16235: [Bug] The parameters of method ```CollectionUtils#sort```  should be comparable type
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo - 3.3, JDK - 21, OS - MacOS(26.3.1 (a) (25D771280a)) 

Improve generic type safety of CollectionUtils.sort

### Motivation

`CollectionUtils.sort` currently accepts any `List<T>` and internally calls `Collections.sort` using a raw cast / suppressed warnings.

This allows code like `CollectionUtils.sort(List<NonComparableType>)` to compile, but it may fail at runtime because natural ordering requires elements to implement `Comparable`.

### Proposed change

Change the method signature from:

```java
public static <T> List<T> sort(List<T> list)
```
to 

```java
public static <T extends Comparable<? super T>> List<T> sort(List<T> list)
```

and remove the raw cast

```java
Collections.sort(list);
```

### Anything else

This aligns the method with java.util.Collections.sort, improves compile-time type safety, and preserves the original return type List<T>.

### Code of Conduct

codingkiddo:
### What is changed

This PR updates `CollectionUtils.sort` to use a bounded generic type:

```java
<T extends Comparable<? super T>>
```
To 

```java
public static <T extends Comparable<? super T>> List<T> sort(List<T> list)
```

The raw cast is also removed:

```java
Collections.sort((List) list);
```

to 

```java
Collections.sort(list);
```

## Why

CollectionUtils.sort currently accepts any List<T> and relies on a raw cast with suppressed warnings.

This means callers can pass a list whose elements do not implement Comparable, and the code still compiles. However, sorting such a list can fail at runtime.

By using:

```java
<T extends Comparable<? super T>>
```

the method now follows the same generic constraint style as java.util.Collections.sort.

This ensures that only naturally comparable elements can be passed to the method.

## Benefit

- Improves compile-time type safety
- Removes raw type usage
- Removes unchecked/raw warning suppression
- Preserves the original return type as List<T>
- Aligns the method with Java collection sorting conventions
- Supports inherited comparison cases, for example when a subtype inherits Comparable<SuperType>

## Compatibility

The erased method signature remains sort(List), so this change should be binary-compatible.

However, this may be source-incompatible for callers that currently pass non-comparable element types to CollectionUtils.sort. Those usages would already be unsafe because they may fail at runtime.

## Tests

Added/updated tests to verify:

Sorting a normal comparable list
Sorting a subtype whose parent implements Comparable<Parent>

Fixes #16235

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/CollectionUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/CollectionUtils.java
@@ -68,10 +68,9 @@ public class CollectionUtils {
 
     private CollectionUtils() {}
 
-    @SuppressWarnings({"unchecked", "rawtypes"})
-    public static <T> List<T> sort(List<T> list) {
+    public static <T extends Comparable<? super T>> List<T> sort(List<T> list) {
         if (isNotEmpty(list)) {
-            Collections.sort((List) list);
+            Collections.sort(list);
         }
         return list;
     }
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/utils/CollectionUtilsTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/utils/CollectionUtilsTest.java
@@ -244,4 +244,39 @@ class CollectionUtilsTest {
         expectedSet.add("C");
         assertEquals(expectedSet, set);
     }
+
+    @Test
+    void sortShouldSupportComparableSuperType() {
+        List<Student> students = new ArrayList<>(Arrays.asList(new Student("b"), new Student("a")));
+
+        List<Student> sorted = CollectionUtils.sort(students);
+
+        assertEquals("a", sorted.get(0).getName());
+        assertEquals("b", sorted.get(1).getName());
+    }
+
+    private static class Person implements Comparable<Person> {
+
+        private final String name;
+
+        private Person(String name) {
+            this.name = name;
+        }
+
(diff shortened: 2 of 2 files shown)


zrlw (approved):

zrlw:
LGTM

nilesh-2005:
I reviewed the changes, and the use of generics looks good. One small suggestion: consider adding null checks or clarifying behavior for non-comparable inputs.

codingkiddo:
Hi @nilesh-2005 

Thanks for reviewing.

For null input, the existing behavior is preserved because the method uses `isNotEmpty(list)`, so null or empty lists are not sorted and are returned as-is.

For non-comparable inputs, this PR intentionally moves the failure from runtime to compile time by changing the signature to:

```java
<T extends Comparable<? super T>>
```

So callers passing `List<NonComparableType>` will no longer compile, instead of compiling and potentially failing at runtime through `Collections.sort`.

I can add a short JavaDoc clarification if maintainers prefer, for example:

```java
/**
 * Sorts the given list according to the natural ordering of its elements.
 * Returns the input list as-is when it is null or empty.
 */
```

kimmking (approved):

kimmking:
nice polish job.

PR #16248, opened 2026-04-28
feat: isolate dual-registration failure handling per mode

### LINKED ISSUE #16247: [Bug]   Dual registration: one registry's failure aborts the sibling arm under register-mode=all
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

 Dubbo Java 3.3.7-SNAPSHOT

Pre-requisite: a provider configured with `register-mode=all` so Dubbo fans out the same provider URL to two registries — an interface-level one
  (`registry://`, e.g. ZooKeeper) and an instance-level one (`service-discovery-registry://`).

  1. Configure the provider so both registries are in use, for example:

     ```properties
     dubbo.application.register-mode=all
     dubbo.registries.zk.address=zookeeper://127.0.0.1:2181
     dubbo.registries.sd.address=zookeeper://127.0.0.1:2181?registry-type=service
     ```

  2. Make the **first** registry fail at registration time (easiest repro:
     start the provider with ZooKeeper unreachable, or let the interface-level
     registry throw `NoNode` during the write). The failing registry can be
     either arm — whichever is processed first.

  3. Export a service:

     ```java
     ServiceConfig<DemoService> sc = new ServiceConfig<>();
     sc.setInterface(DemoService.class);
     sc.setRef(new DemoServiceImpl());
     sc.export();
     ```

  4. Observe that `export()` throws the underlying `RuntimeException` from the
     failing arm and the **second** registry is never registered to, even
     though the two registrations are semantically independent and the user
     explicitly opted into dual mode with `register-mode=all`.

  Root cause is in `RegistryProtocol#export`:

  ```java
  // dubbo-registry-api/.../RegistryProtocol.java
  boolean register = providerUrl.getParameter(REGISTER_KEY, true)
          && registryUrl.getParameter(REGISTER_KEY, true);
  if (register) {
      register(registry, registeredProviderUrl);  // <-- not guarded per-arm
  }
  ```

  The single call fans out to both arms and any `RuntimeException` from one
  arm aborts the whole export.

  A minimal unit-test reproducer (no external registry needed) using a mocked
  `Registry`:

  ```java
  URL registryUrl = new ServiceConfigURL("zookeeper", "127.0.0.1", 2181,
          Map.of("check", "false"));
  URL providerUrl = new ServiceConfigURL("dubbo", "127.0.0.1", 20880,
          "com.example.DemoService", new HashMap<>());
  Registry registry = Mockito.mock(Registry.class);
  Mockito.doThrow(new IllegalStateException("NoNode for /dubbo/..."))
         .when(registry).register(providerUrl);

  // Through the current export() -> register() path, this RuntimeException
  // propagates and any sibling registration never runs; there is also no
  // per-arm outcome emitted to FrameworkStatusReportService.
  ```

 Under `register-mode=all`, each arm of the dual registration should be
  handled independently, and outcomes should be observable per arm:

  1. **Failure isolation.** A failure on one arm (interface-level OR
     instance-level) must not abort the other arm. Propagation should only
     happen when the registry URL explicitly asks for `check=true` — which
     is the convention Dubbo already uses elsewhere for `CHECK_KEY`.
  2. **Per-arm outcome reporting.** `FrameworkStatusReportService` should
     emit an outcome that identifies which arm was attempted, against which
     registry address, for which service, and whether it succeeded or
     failed (and why). Today only a coarse `{application, status}` payload
     is emitted, so operators cannot tell which side of the dual
     registration failed.
  3. **Accurate log tags.** The success log currently hardcodes
     `[INSTANCE_REGISTER]` in `ExporterChangeableWrapper#register`
     regardless of whether the registry is actually an instance-level one,
     which makes grep-based troubleshooting misleading for interface-level
     arms.

  What currently happens instead:

  - **(1)** The `RuntimeException` from the first arm aborts `export()` and
    the second arm is never tried. Users who opted into dual mode lose the
    healthy arm because the other arm was transiently unhealthy.
  - **(2)** Observers subscribed to the `registration` report key only see
    `{application, status=interface|instance}`. They cannot attribute a
    partial failure to a specific arm/registry/service, and they never see
    an error message.
  - **(3)** Relevant log line:

    ```
    [INSTANCE_REGISTER] Registered dubbo service ... to registry zookeeper://...
    ```

    …is printed even when the registry is `zookeeper://` (interface-level).

  Concretely, the expected payload shape for the outcome report is:

  ```json
  {
    "application": "APP",
    "mode": "INTERFACE_REGISTER",          // or INSTANCE_REGISTER
    "registry": "127.0.0.1:2181",
    "service": "GroupA/DemoService:1.0.0",
    "status": "SUCCESS",                   // or FAILED
    "error": "NoNode for /dubbo/..."       // present only on FAILED
  }
  ```

  emitted under the same `REGISTRATION_STATUS` key so existing
  `FrameworkStatusReporter` extensions keep working (new fields are
  additive).

### Anything else

- Occurs deterministically: every time one arm of a dual registration
    throws a `RuntimeException` synchronously. Not a flake.
  - The existing logger code `CONFIG_REGISTER_INSTANCE_ERROR` (5-11) is a
    natural fit for the failure path — no new code is needed.
  - Backward compatibility:
    - No public API removed.
    - `reportRegistrationStatus(...)` / `createRegistrationReport(...)`
      untouched; reporters that only read `application` / `status` keep
      working.
    - Users who relied on "one failure aborts everything" can keep that by
      setting `check=true` on the registry URL, which is the documented
      knob for "registration must succeed or fail fast".
  - Consumer-side `registry.register(...)` (for consumer URLs) is **not**
    part of this report; the scope here is provider-side dual registration
    only.

### Code of Conduct

ban-xiu:

Close https://github.com/apache/dubbo/issues/16247.

 Problem
  -----                                                                                                        
  In dual registration mode (register-mode=all) Dubbo emits a single                                                                                               provider URL through two registries:                                                                                                                           
    - interface-level  -> registry:// (ZooKeeper, Nacos, ...)                                                                                                        - instance-level   -> service-discovery-registry://                                                                                                          
                                                                                                                                                                 
  RegistryProtocol#export calls register() once; the registry wrapper                                                                                            
  fans out to both arms. Any RuntimeException from the first arm aborts
  the second, so a transient failure on the metadata center silently
  takes down the interface-level registration that was otherwise healthy
  (or vice versa). The pre-existing FrameworkStatusReportService only
  reports a coarse "instance/interface" status and cannot tell observers
  which arm failed, which swallowed the error, or why.

  Fix
  -----
  1. FrameworkStatusReportService
     - add reportRegistrationOutcome(mode, registryAddress, serviceKey,
       success, errorMessage) + createRegistrationOutcomeReport() so each
       registration attempt emits an independently tagged payload
       {application, mode, registry, service, status, error?}.

  2. RegistryProtocol
     - new registerWithModeTag(registry, registryUrl, providerUrl) wraps
       the existing register() call with per-arm failure isolation:
         * success -> report SUCCESS tagged with the resolved mode
         * RuntimeException:
             - report FAILED with the cause message
             - honor registryUrl check= parameter: rethrow iff check=true,
               otherwise log a warn under CONFIG_REGISTER_INSTANCE_ERROR
               (5-11) and return false so the sibling arm keeps going
     - resolveRegisterModeTag() centralizes the INTERFACE_REGISTER vs
       INSTANCE_REGISTER label (previously hardcoded in one log line).
     - route export() and ExporterChangeableWrapper.register() through
       the new wrapper; replace the hardcoded "[INSTANCE_REGISTER]" log
       prefix with the resolved tag.
     - reporting failures never leak into the registration flow
       (wrapped in try/catch(Throwable)).

  Tests
  -----
  - dubbo-common: FrameworkStatusReportServiceTest gains
    testReportRegistrationOutcomeSuccess / Failure covering the new
    payload shape.
  - dubbo-registry-api: new RegistryProtocolDualRegisterTest with 4
    reflection-driven cases over registerWithModeTag — success paths for
    both protocols, check=true rethrow path, check=false swallow path.
    A local CapturingFrameworkStatusReporter + SPI registration under
    src/test/resources provides the capture sink (MockFrameworkStatus-
    Reporter is scoped to dubbo-common tests and not reachable here).

Behavior change
  -----
  | Scenario | Before | After |
  |---|---|---|
  | Both arms succeed | 1 coarse `registration` report | 2 outcome reports, one per mode, `status=SUCCESS` |
  | One arm fails, `check=false` on that registry URL | exception bubbles out, sibling arm may be skipped | failure is tagged + reported `FAILED`, returns false,
   sibling arm continues |
  | One arm fails, `check=true` | exception bubbles out (no outcome report) | outcome still reported `FAILED` **then** the original exception is rethrown |      
  | Reporter itself throws | would leak into registration flow | swallowed — reporting never affects registration |


## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/status/reporter/FrameworkStatusReportService.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/status/reporter/FrameworkStatusReportService.java
@@ -80,6 +80,32 @@ public class FrameworkStatusReportService implements ScopeModelAware {
         return JsonUtils.toJson(registration);
     }
 
+    /**
+     * Report a registration outcome tagged with the register mode (INTERFACE / INSTANCE), the
+     * registry address it targets, and a success flag. Used to surface partial-failure cases when
+     * Dubbo is running in dual registration mode, so observers can tell which side succeeded.
+     */
+    public void reportRegistrationOutcome(
+            String mode, String registryAddress, String serviceKey, boolean success, String errorMessage) {
+        doReport(
+                REGISTRATION_STATUS,
+                createRegistrationOutcomeReport(mode, registryAddress, serviceKey, success, errorMessage));
+    }
+
+    public String createRegistrationOutcomeReport(
+            String mode, String registryAddress, String serviceKey, boolean success, String errorMessage) {
+        HashMap<String, String> registration = new HashMap<>();
+        registration.put("application", applicationModel.getApplicationName());
+        registration.put("mode", mode);
+        registration.put("registry", registryAddress);
+        registration.put("service", serviceKey);
+        registration.put("status", success ? "SUCCESS" : "FAILED");
+        if (errorMessage != null) {
+            registration.put("error", errorMessage);
+        }
+        return JsonUtils.toJson(registration);
+    }
+
     public String createConsumptionReport(String interfaceName, String version, String group, String status) {
         HashMap<String, String> migrationStatus = new HashMap<>();
         migrationStatus.put("type", "consumption");
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/status/reporter/FrameworkStatusReportService.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/status/reporter/FrameworkStatusReportService.java
@@ -80,6 +80,10 @@ public class FrameworkStatusReportService implements ScopeModelAware {
         return JsonUtils.toJson(registration);
     }
 
+    public static final String OUTCOME_SUCCESS = "SUCCESS";
+    public static final String OUTCOME_FAILED = "FAILED";
+    public static final String OUTCOME_PENDING_RETRY = "PENDING_RETRY";
+
     /**
      * Report a registration outcome tagged with the register mode (INTERFACE / INSTANCE), the
      * registry address it targets, and a success flag. Used to surface partial-failure cases when
@@ -87,19 +91,36 @@ public class FrameworkStatusReportService implements ScopeModelAware {
      */
     public void reportRegistrationOutcome(
             String mode, String registryAddress, String serviceKey, boolean success, String errorMessage) {
+        reportRegistrationOutcome(
+                mode, registryAddress, serviceKey, success ? OUTCOME_SUCCESS : OUTCOME_FAILED, errorMessage);
+    }
+
+    /**
+     * String-status overload so callers can report outcomes that don't fit the success/failure
+     * binary — notably {@link #OUTCOME_PENDING_RETRY} for the FailbackRegistry silent-retry path
+     * where {@code register()} returned normally but the URL was queued for retry.
+     */
+    public void reportRegistrationOutcome(
+            String mode, String registryAddress, String serviceKey, String status, String errorMessage) {
         doReport(
                 REGISTRATION_STATUS,
-                createRegistrationOutcomeReport(mode, registryAddress, serviceKey, success, errorMessage));
+                createRegistrationOutcomeReport(mode, registryAddress, serviceKey, status, errorMessage));
     }
 
     public String createRegistrationOutcomeReport(
             String mode, String registryAddress, String serviceKey, boolean success, String errorMessage) {
+        return createRegistrationOutcomeReport(
+                mode, registryAddress, serviceKey, success ? OUTCOME_SUCCESS : OUTCOME_FAILED, errorMessage);
+    }
+
+    public String createRegistrationOutcomeReport(
+            String mode, String registryAddress, String serviceKey, String status, String errorMessage) {
         HashMap<String, String> registration = new HashMap<>();
         registration.put("application", applicationModel.getApplicationName());
         registration.put("mode", mode);
         registration.put("registry", registryAddress);
         registration.put("service", serviceKey);
-        registration.put("status", success ? "SUCCESS" : "FAILED");
+        registration.put("status", status);
         if (errorMessage != null) {
             registration.put("error", errorMessage);
         }
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/status/reporter/FrameworkStatusReportServiceTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/status/reporter/FrameworkStatusReportServiceTest.java
@@ -112,4 +112,54 @@ class FrameworkStatusReportServiceTest {
 
         frameworkModel.destroy();
     }
+
+    @Test
+    void testReportRegistrationOutcomeSuccess() {
+        FrameworkModel frameworkModel = new FrameworkModel();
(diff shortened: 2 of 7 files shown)


SavitarC:
Thanks for working on this. I noticed a couple of edge cases around how registration failures are reported and recorded.

The first case is mainly related to standard `FailbackRegistry` implementations. In `FailbackRegistry.register()`, a failed `doRegister()` does not always propagate to the caller. When `check=false`, the failure is converted into a retry task and `register()` returns normally:

```java
try {
    doRegister(url);
} catch (Exception e) {
    boolean check = getUrl().getParameter(Constants.CHECK_KEY, true)
            && url.getParameter(Constants.CHECK_KEY, true)
            && (url.getPort() != 0);
    if (check || skipFailback) {
        throw new IllegalStateException(...);
    } else {
        logger.error(... "waiting for retry" ...);
    }

    addFailedRegistered(url);
}
```

`addFailedRegistered(url)` then schedules a retry:

```java
FailedRegisteredTask newTask = new FailedRegisteredTask(url, this);
oldOne = failedRegistered.putIfAbsent(url, newTask);
if (oldOne == null) {
    retryTimer.newTimeout(newTask, retryPeriod, TimeUnit.MILLISECONDS);
}
```

So for failback registries such as `ZookeeperRegistry`, `NacosRegistry`, or `ServiceDiscoveryRegistry`, `register()` may return normally even though the immediate registration attempt failed and is only waiting for retry. In that case, reporting `SUCCESS` based only on the absence of an exception would be misleading.

The second case is about the caller-side recorded state. `registerWithModeTag()` may return `false` when registration fails with `check=false`, but `export()` appears to ignore that result and record the stated URL/exporter as registered based on the configured `register` flag:

```java
boolean register = providerUrl.getParameter(REGISTER_KEY, true)
        && registryUrl.getParameter(REGISTER_KEY, true);
if (register) {
    registerWithModeTag(registry, registryUrl, registeredProviderUrl);
}

registerStatedUrl(registryUrl, registeredProviderUrl, register);
exporter.setRegistered(register);
```

Similarly, `ExporterChangeableWrapper.register()` should probably avoid keeping its atomic registered flag set when registration returns `false` or throws. Otherwise, later manual register/re-register logic may incorrectly believe registration has already succeeded.

Could we propagate the actual registration outcome into these recorded states and add regression coverage for these paths? In particular, it would be helpful to cover a realistic `FailbackRegistry` path where `doRegister()` fails under `check=false` but `Registry.register()` returns normally, since the current failure coverage appears to focus on the direct-throwing `Registry.register()` path.

zrlw (approved):
LGTM

zrlw (changes requested):
see comments

SavitarC (approved):
LGTM

zrlw (approved):
LGTM

zrlw (approved):

ban-xiu:
@SavitarC PTAL

PR #16255, opened 2026-05-04
[FIX] Close channel on TLS negotiation failure to prevent zombie conn…

LI123456mo:
This pull request addresses a bug in SslServerTlsHandler where exceptions during the TLS handshake were swallowed, leaving TCP channels active but dead. By adding ctx.close() to exceptionCaught, the PR ensures that failed connections are closed, enabling the consumer to detect the failure and initiate self-healing.

### DIFF
--- a/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/ssl/SslServerTlsHandler.java
+++ b/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/ssl/SslServerTlsHandler.java
@@ -65,8 +65,8 @@ public class SslServerTlsHandler extends ByteToMessageDecoder {
                 "",
                 "TLS negotiation failed when trying to accept new connection.",
                 cause);
-        ctx.close();
         ctx.fireExceptionCaught(cause);
+        ctx.close();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


zrlw:
it might be better to add a testcase for it.

dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/ssl/SslServerTlsHandler.java
```diff
@@ -65,6 +65,8 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
                 "",
                 "TLS negotiation failed when trying to accept new connection.",
                 cause);
+        ctx.close();
```
zrlw: should close context after calling fireExceptionCaught

LI123456mo:
> it might be better to add a testcase for

Thanks for suggestion , will add

dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/ssl/SslServerTlsHandler.java
LI123456mo: > should close context after calling fireExceptionCaught

Thank you for the feedback, @zrlw. That makes sense—firing the exception first ensures the error is fully propagated through the pipeline before the channel is terminated. I will update the PR to swap the order: ctx.fireExceptionCaught(cause) followed by ctx.close().

LI123456mo:
> it might be better to add a testcase for it.



> it might be better to add a testcase for it.
@zrlw , done

zrlw (approved):

RainYuY (approved):
LGTM

PR #16298, opened 2026-05-29
fix: throw clear exception when request parameter is not Serializable

### LINKED ISSUE #16293: Misleading NullPointerException when request parameter is not Serializable
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Descriptions

When calling a Dubbo service with a parameter that does not​ implement java.io.Serializable, the request fails with a confusing NullPointerException.

java.lang.NullPointerException: Cannot invoke "Object.hashCode()" because "key" is null
    at java.util.concurrent.ConcurrentHashMap.get(...)
    at org.apache.dubbo.rpc.support.GroupServiceKeyCache.getServiceKey(...)
    at org.apache.dubbo.rpc.protocol.dubbo.DubboProtocol.getInvoker(...)

Root Cause:​

The actual problem is that the entity class is not Serializable.

Adding implements Serializablefixes the issue immediately.

Expected Behavior:​

Dubbo should throw a clear exception (e.g. NotSerializableException) instead of an internal NPE.

Suggestion:​

Fail fast with a meaningful error message indicating which class is not serializable.

### Related issues

_No response_

### Code of Conduct

joaohmalves:
Previously, passing a non-Serializable parameter to a Dubbo service resulted in a confusing error message. Now a clear IOException with the class name and 'must implement java.io.Serializable' is thrown at encode time in both request and response paths.

Fixes #16293

This PR fixes a misleading error message when calling a Dubbo service 
with a parameter or return value that does not implement 
`java.io.Serializable`.

Previously, the error was buried inside internal serialization layers 
and did not clearly indicate the root cause, making it hard to diagnose.
Now a clear `IOException` containing the class name and 
`must implement java.io.Serializable` is thrown at encode time in both 
request and response paths.

### What is changed?

- `DubboCodec.encodeRequestData()`: wraps `writeObject()` in a 
try/catch that intercepts the internal `IllegalArgumentException` and 
rethrows a clear `IOException` with the offending class name.
- `DubboCodec.encodeResponseData()`: same treatment for the response path.
- `DubboProtocolTest`: re-enabled `testNonSerializedParameter` and 
`testReturnNonSerialized` which were previously `@Disabled` because 
the expected behavior was not implemented.

### Verifying this change

All tests in `dubbo-rpc-dubbo` module pass locally, including the 
two previously disabled tests that now validate the correct behavior.

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboCodec.java
+++ b/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboCodec.java
@@ -294,7 +294,19 @@ public class DubboCodec extends ExchangeCodec {
         Object[] args = inv.getArguments();
         if (args != null) {
             for (int i = 0; i < args.length; i++) {
-                out.writeObject(callbackServiceCodec.encodeInvocationArgument(channel, inv, i));
+                Object arg = callbackServiceCodec.encodeInvocationArgument(channel, inv, i);
+                try {
+                    out.writeObject(arg);
+                } catch (IOException e) {
+                    Throwable cause = e.getCause() != null ? e.getCause().getCause() : null;
+                    if (cause instanceof IllegalArgumentException
+                            && cause.getMessage() != null
+                            && cause.getMessage().contains("has not implement Serializable")) {
+                        throw new IOException(arg.getClass().getName()
+                                + " must implement java.io.Serializable", e);
+                    }
+                    throw e;
+                }
             }
         }
         out.writeAttachments(inv.getObjectAttachments());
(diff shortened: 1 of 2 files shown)


RainYuY:
Have you checked the situation of the issue?I couldn't reproduce. Moreover, it seems that such changes risk swallowing non-Serializer exceptions?

joaohmalves:
After checking a bit further, I was unable to reproduce the exact NPE from the stack trace (GroupServiceKeyCache.getServiceKey). 
What I found instead is that Dubbo already detects the non-Serializable class in DefaultSerializeClassChecker and logs an ERROR, but does not propagate a clear exception. 

Also as you pointed out this fix does risks swallowing unrelated exceptions so I would like to work on it a bit more.

As this is my first contribution to this project, would it make sense for me to go back to the issue and ask the reporter for more details on how to reproduce the exact stack trace before reworking this PR?

zrlw:
As for me, https://github.com/apache/dubbo/pull/16299 might be better.

PR #16299, opened 2026-05-31
Fix misleading NPE when request parameter is not Serializable

### LINKED ISSUE #16293: Misleading NullPointerException when request parameter is not Serializable

chensishang:
When a Dubbo service is called with non-serializable parameters, the invocation deserialization fails silently and `path` becomes null. This null path is then passed to GroupServiceKeyCache which uses a ConcurrentHashMap that does not accept null keys, resulting in a confusing NullPointerException.

Added a null check for `path` in DubboProtocol.getInvoker() to throw a clear RemotingException with a meaningful message pointing users to the actual cause (non-serializable parameters).

Fixes #16293

## Checklist

### DIFF
--- a/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
+++ b/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
@@ -290,6 +290,14 @@ public class DubboProtocol extends AbstractProtocol {
         boolean isStubServiceInvoke;
         int port = channel.getLocalAddress().getPort();
         String path = (String) inv.getObjectAttachmentWithoutConvert(PATH_KEY);
+        if (path == null) {
+            throw new RemotingException(
+                    channel,
+                    "Failed to resolve service path from invocation. "
+                            + "This may be caused by non-serializable request parameters. "
+                            + "Please ensure all parameter types implement java.io.Serializable, "
+                            + "channel: " + channel.getRemoteAddress() + " --> " + channel.getLocalAddress());
+        }
 
         // if it's stub service on client side(after enable stubevent, usually is set up onconnect or ondisconnect
         // method)
--- a/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
+++ b/dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
@@ -293,9 +293,10 @@ public class DubboProtocol extends AbstractProtocol {
         if (path == null) {
             throw new RemotingException(
                     channel,
-                    "Failed to resolve service path from invocation. "
-                            + "This may be caused by non-serializable request parameters. "
-                            + "Please ensure all parameter types implement java.io.Serializable, "
+                    "Service path is missing from the invocation, which indicates the invocation metadata is "
+                            + "missing or corrupted. Possible causes include a request decode failure "
+                            + "(e.g. parameter types that failed to deserialize), an incompatible protocol "
+                            + "version, or a custom codec/invocation implementation that does not set the path, "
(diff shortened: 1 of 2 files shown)


chensishang:
Hi reviewers, I noticed the Codecov check failed with "0% coverage". After checking, this seems to be due to an incomplete report upload ("HEAD比BASE少7次上传") rather than a problem with the code change itself.

I have already added a unit test (testGetInvokerThrowsOnNullPath) which covers the new null-check logic, and it passes locally. The Codecov result appears to be a false negative caused by the CI infrastructure.

Could you please take another look? Thanks!

zrlw (dismissed):
LGTM

chensishang:
Thanks for the review!

chensishang:
Hi @zrlw, thanks for the review!

  The PR contains two commits: the actual fix (`6a870129`) and a merge commit for syncing with the latest 3.3 branch. The only real change is in `DubboProtocol.java` and `DubboProtocolTest.java`.

  The maintainer can squash on merge if needed. Sorry for the noise!

chensishang:
@RainYuY

Hi, thanks for your time.

This PR has already received one approval and all checks are passing. It fixes the misleading NPE issue described in #16293 and includes a corresponding unit test.

Could you please help take a look when you have time?
Thank you!

dubbo-rpc/dubbo-rpc-dubbo/src/main/java/org/apache/dubbo/rpc/protocol/dubbo/DubboProtocol.java
```diff
@@ -290,6 +290,14 @@ Invoker<?> getInvoker(Channel channel, Invocation inv) throws RemotingException
         boolean isStubServiceInvoke;
         int port = channel.getLocalAddress().getPort();
         String path = (String) inv.getObjectAttachmentWithoutConvert(PATH_KEY);
+        if (path == null) {
+            throw new RemotingException(
```
RainYuY, line 294: Is non-serializable parameter really the only scenario that can make path null here? In the normal Dubbo request format, the service path is encoded/decoded before arguments, and DubboInvoker also sets PATH_KEY before sending the request. So a null path seems more generally like missing/malformed invocation metadata, custom codec/invocation behavior, or decode/protocol incompatibility. Could we make this error message less specific than “Please ensure all parameter types implement Serializable” to avoid another misleading diagnosis?

chensishang: Thanks for the careful review — you're right, and I appreciate you pointing this out. A null path indicates missing or corrupted invocation metadata in general, and my original message attributing   ▎ it solely to non-serializable parameters could indeed cause another misleading diagnosis
 
   I've updated the message to describe the problem generically and only mention deserialization failure as one of several possible causes (along with protocol incompatibility and custom 
   codec/invocation implementations). Please let me know if the wording could be further improved — happy to adjust.

chensishang:
@zrlw

Sorry for the extra round — the review above pointed out that my error message was too specific, so I've updated the wording, which dismissed your earlier approval. Could you take another    
  ▎ look when you have time? Thanks!

RainYuY (approved):

zrlw (approved):

PR #16306, opened 2026-06-02
Improve type safety of CollectionUtils.toTreeSet

### LINKED ISSUE #16297: [Feature] Improve type safety of CollectionUtils.toTreeSet
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Motivation

`CollectionUtils.toTreeSet` currently accepts any `Set<T>`:

```java
public static <T> Set<T> toTreeSet(Set<T> set)
```

When the input set is not empty and is not already a `TreeSet`, the method creates a new `TreeSet`:

```java
set = new TreeSet<>(set);
```

This relies on natural ordering. If the elements are not mutually comparable, the code compiles successfully but may fail at runtime with `ClassCastException`.

This is similar to the recently improved `CollectionUtils.sort` method, where the generic signature was changed to require comparable elements.

### Current behavior

The current method allows code like this to compile:

```java
class Item {
}

Set<Item> items = new HashSet<>();
items.add(new Item());

CollectionUtils.toTreeSet(items);
```

However, converting this set to a `TreeSet` may fail at runtime because `Item` does not implement `Comparable`.

### Proposed improvement

One possible improvement is to make the natural-ordering requirement explicit in the method signature:

```java
public static <T extends Comparable<? super T>> Set<T> toTreeSet(Set<T> set)
```

This would catch invalid usages at compile time instead of allowing them to fail at runtime.

### Compatibility note

This may be source-incompatible for callers that currently pass non-comparable element types.

There is also one subtle case: an existing `TreeSet` can use a custom comparator even when its element type does not implement `Comparable`.

For example:

```java
class Item {
    private final String name;

    Item(String name) {
        this.name = name;
    }

    String name() {
        return name;
    }
}

TreeSet<Item> items = new TreeSet<>(Comparator.comparing(Item::name));
items.add(new Item("b"));
items.add(new Item("a"));

CollectionUtils.toTreeSet(items);
```

In this case, the existing `TreeSet` is valid because it already has a comparator, even though `Item` does not implement `Comparable`.

Because of this compatibility concern, I would like maintainer feedback before opening a PR.

#### Option 1: Add a comparable bound to the existing method

```java
public static <T extends Comparable<? super T>> Set<T> toTreeSet(Set<T> set)
```

This makes the natural-ordering requirement explicit and catches invalid usages at compile time.

#### Option 2: Keep the existing method and add a comparator-based overload

```java
public static <T> Set<T> toTreeSet(Set<T> set, Comparator<? super T> comparator)
```

This would allow callers to safely convert sets of non-comparable elements by providing an ordering.

### Expected benefit

- Improves type safety
- Makes the natural-ordering requirement explicit
- Avoids hidden runtime `ClassCastException` for non-comparable elements
- Provides a possible path for comparator-based conversion
- Keeps the discussion open for maintainers to choose the preferred compatibility approach

I can work on this once maintainers confirm the preferred direction.

### Related issues

_No response_

### Code of Conduct

skt-shinyruo:

GitHub_issue: Fixes #16297

This pull request improves the type safety of `CollectionUtils.toTreeSet`.

The existing single-argument overload accepted any `Set<T>`, but its implementation delegates to `TreeSet`, which requires naturally comparable elements unless a comparator is provided. The updated signature makes that requirement explicit at compile time:

```java
public static <T extends Comparable<? super T>> Set<T> toTreeSet(Set<T> set)
```

This pull request also adds a comparator-based overload for custom ordering and non-`Comparable` element types:

```java
public static <T> Set<T> toTreeSet(Set<T> set, Comparator<? super T> comparator)
```

For non-null empty inputs, the comparator overload returns an empty `TreeSet` that preserves the provided comparator.

Verification:

- `mvn -pl dubbo-common -Dtest=CollectionUtilsTest test`
- `git diff --check`

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/CollectionUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/CollectionUtils.java
@@ -419,7 +419,7 @@ public class CollectionUtils {
         return values.get(0);
     }
 
-    public static <T> Set<T> toTreeSet(Set<T> set) {
+    public static <T extends Comparable<? super T>> Set<T> toTreeSet(Set<T> set) {
         if (isEmpty(set)) {
             return set;
         }
@@ -427,6 +427,15 @@ public class CollectionUtils {
             set = new TreeSet<>(set);
         }
         return set;
+    }
+
+    public static <T> Set<T> toTreeSet(Set<T> set, Comparator<? super T> comparator) {
+        if (set == null) {
+            return set;
+        }
+        Set<T> treeSet = new TreeSet<>(comparator);
+        treeSet.addAll(set);
+        return treeSet;
     }
 
     public static <T> Set<T> newHashSet(int expectedSize) {
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/utils/CollectionUtilsTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/utils/CollectionUtilsTest.java
@@ -278,6 +278,13 @@ class CollectionUtilsTest {
         Assertions.assertSame(comparator, ((TreeSet<PriorityItem>) sorted).comparator());
     }
 
+    @Test
+    void toTreeSetShouldReturnNullForNullInputWithComparator() {
+        Comparator<PriorityItem> comparator = Comparator.comparingInt(PriorityItem::getPriority);
+
+        assertNull(CollectionUtils.toTreeSet(null, comparator));
+    }
+
     private static class Person implements Comparable<Person> {
 
         private final String name;
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/utils/CollectionUtilsTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/utils/CollectionUtilsTest.java
@@ -21,12 +21,14 @@ import org.apache.dubbo.config.ProtocolConfig;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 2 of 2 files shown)


zrlw (dismissed):
LGTM

dubbo-common/src/main/java/org/apache/dubbo/common/utils/CollectionUtils.java
```diff
@@ -429,6 +429,15 @@ public static <T> Set<T> toTreeSet(Set<T> set) {
         return set;
     }
 
+    public static <T> Set<T> toTreeSet(Set<T> set, Comparator<? super T> comparator) {
+        if (set == null) {
```
RainYuY, line 433: Why did it become null instead of checking for empty?

RainYuY (changes requested):

dubbo-common/src/main/java/org/apache/dubbo/common/utils/CollectionUtils.java
skt-shinyruo: The null check is intentional here. For this comparator overload, null and empty inputs need different handling: null keeps the existing null-return behavior, while an empty non-null set should still return an empty TreeSet initialized with the provided comparator. Returning the original set from isEmpty(set) would lose that comparator. The addAll call is a no-op for an empty set.

I also added a null-input test for this overload in 630215ff94.

RainYuY (approved):

skt-shinyruo:
Hi @zrlw, RainYuY has approved the latest commit and all checks are passing. Could you please take another look when available?

zrlw (approved):

PR #16307, opened 2026-06-03
Fix Hessian2 writeReplace Serializable check bypass

### LINKED ISSUE #16287: Classes with writeReplace() bypass Hessian2 Serializable security check
### Environment

- Dubbo version: 3.2.15 (also affects 3.3.x — the relevant code path is in hessian-lite)
- hessian-lite version: 3.2.13
- JDK: 17

### Description

`Hessian2SerializerFactory` hooks into `getDefaultSerializer()` to enforce the `checkSerializable` security check. However, in hessian-lite's `SerializerFactory.getSerializer(Class)`, classes that define a `writeReplace()` method are handled **before** `getDefaultSerializer()` is reached — they get a `JavaSerializer` directly and skip the `getDefaultSerializer` override entirely.

This means:
1. A class **without** `Serializable` and **without** `writeReplace()` → correctly rejected by `checkSerializable`
2. A class **without** `Serializable` but **with** `writeReplace()` → **bypasses the check**, serialization succeeds

The relevant priority chain in `SerializerFactory.getSerializer(Class)`:

```
1. _staticSerializerMap (built-in types)
2. Custom serializer factories
3. EnumSet check
4. writeReplace check → new JavaSerializer(cl, loader); goto cache   ← short-circuits here
5. Map / Collection / Iterator / Enum / ...
6. getDefaultSerializer(cl)   ← checkSerializable is here, never reached for step 4
```

### How to reproduce

Define two DTOs:

```java
// No Serializable, no writeReplace
public class NormalDto {
    private String name;
    // getter/setter
}

// No Serializable, HAS writeReplace
public class WriteReplaceDto {
    private String name;
    // getter/setter

    private Object writeReplace() {
        return this;
    }
}
```

Define a Dubbo service that accepts both as parameters:

```java
public interface TestService {
    String sendNormal(NormalDto dto);
    String sendWriteReplace(WriteReplaceDto dto);
}
```

Call both methods from a consumer:

- `sendNormal(new NormalDto("test"))` → **Fails on consumer side** with `"Serialized class ... has not implement Serializable interface"`
- `sendWriteReplace(new WriteReplaceDto("test"))` → **Consumer serializes successfully** and sends the message (provider-side deserialization catches it via a different path — `loadSerializedClass`)

### Observed behavior

The security check is **inconsistent** between serialization and deserialization:

| Side | Path | writeReplace class caught? |
|------|------|---------------------------|
| Serialization (sender) | `getSerializer()` → writeReplace detected → `JavaSerializer` directly → **bypasses** `getDefaultSerializer` / `checkSerializable` | **No** |
| Deserialization (receiver) | `loadSerializedClass()` → `DefaultSerializeClassChecker.loadClass()` | **Yes** |

### Expected behavior

The `checkSerializable` enforcement should be consistent regardless of whether a class has `writeReplace()`. A class that doesn't implement `Serializable` should be rejected on both the serialization and deserialization sides.

### Suggested fix

Override `getSerializer(Class)` in `Hessian2SerializerFactory` to ensure the security check runs before delegating to `super.getSerializer()`, so that writeReplace classes are also covered. Alternatively, the check could be moved into `SerializerFactory.getSerializer(Class)` in hessian-lite itself, before the writeReplace branch.

yuzhiyang1:

Fixes #16287.

`Hessian2SerializerFactory` currently enforces Dubbo's serialization class policy from `getDefaultSerializer()`. However, hessian-lite resolves classes with `writeReplace()` before it reaches `getDefaultSerializer()`, so non-Serializable classes with `writeReplace()` can bypass the sender-side Serializable check.

This change enforces the same class policy in `getSerializer(Class)` for replacement-capable classes before delegating to hessian-lite, while keeping the existing default serializer/deserializer checks shared through one helper.

## Brief changelog

- Enforce Dubbo's serialization class policy for Hessian2 classes that define `writeReplace()`.
- Keep existing `getDefaultSerializer()` and `getDefaultDeserializer()` checks by sharing the same `checkClass()` helper.
- Add regression coverage for:
  - non-Serializable class whose `writeReplace()` returns itself;
  - Serializable class whose `writeReplace()` returns a non-Serializable replacement with its own `writeReplace()`;
  - valid Serializable replacement path.

## Verifying this change

- `git diff --check`
- `mvn -am -pl dubbo-serialization/dubbo-serialization-hessian2 -Dtest=Hessian2SerializationTest -Dsurefire.failIfNoSpecifiedTests=false -Dcheckstyle.skip=true -Drat.skip=true test`

## Checklist

- [x] Make sure there is a GitHub issue field for the change. (#16287)
- [x] Make sure GitHub Actions can pass.

### DIFF
--- a/dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2SerializerFactory.java
+++ b/dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2SerializerFactory.java
@@ -51,9 +51,9 @@ public class Hessian2SerializerFactory extends SerializerFactory {
     @Override
     public Serializer getSerializer(Class cl) throws HessianProtocolException {
         // SerializerFactory handles writeReplace before getDefaultSerializer(), so enforce Dubbo's
-        // class policy here for replacement-capable classes as well.
+        // Serializable requirement here without expanding strict allow-list checks to JDK replacements.
         if (JavaSerializer.getWriteReplace(cl) != null) {
-            checkClass(cl);
+            checkSerializable(cl);
         }
         return super.getSerializer(cl);
     }
--- a/dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2SerializerFactory.java
+++ b/dubbo-serialization/dubbo-serialization-hessian2/src/main/java/org/apache/dubbo/common/serialize/hessian2/Hessian2SerializerFactory.java
@@ -22,6 +22,7 @@ import java.io.InputStream;
 import java.io.Serializable;
 
 import com.alibaba.com.caucho.hessian.io.Deserializer;
+import com.alibaba.com.caucho.hessian.io.HessianProtocolException;
 import com.alibaba.com.caucho.hessian.io.InputStreamDeserializer;
 import com.alibaba.com.caucho.hessian.io.JavaDeserializer;
 import com.alibaba.com.caucho.hessian.io.JavaSerializer;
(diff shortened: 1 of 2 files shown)


yuzhiyang1:
The CI failure was caused by the previous follow-up enforcing the full strict allow-list check for classes with writeReplace(). That also rejected JDK immutable collection implementations such as java.util.ImmutableCollections in Java 21 samples. I narrowed the writeReplace path to enforce only the Serializable requirement, which is the bypass covered by this issue, and added a regression test for JDK immutable lists under STRICT mode. The latest Build and Test run is now green.

LI123456mo (approved):
LGTM

PR #16309, opened 2026-06-03
Fix IPv4 address port parsing in NetUtils

### LINKED ISSUE #16308: [Bug] NetUtils.matchIpExpression(String, String) fails to parse normal IPv4 address with port
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.7 , OpenJDK 17 , Ubuntu

1. Add the following test case to `dubbo-common/src/test/java/org/apache/dubbo/common/utils/NetUtilsTest.java`:

```java
@Test
void testMatchIpExpressionWithIpv4AddressAndPort() throws UnknownHostException {
    assertTrue(NetUtils.matchIpExpression("*.*.*.*:90", "192.168.1.63:90"));
    assertTrue(NetUtils.matchIpExpression("192.168.1.*:90", "192.168.1.63:90"));
    assertFalse(NetUtils.matchIpExpression("192.168.1.*:80", "192.168.1.63:90"));
}
```

2. Run the test:
```bash
./mvnw -pl dubbo-common -Dtest=NetUtilsTest#testMatchIpExpressionWithIpv4AddressAndPort test
```

3. The test fails because `192.168.1.63:90` is treated as the host instead of being parsed into host `192.168.1.63` and port `90`.
```bash
./mvnw -pl dubbo-common -Dtest=NetUtilsTest#testMatchIpExpressionWithIpv4AddressAndPort test
```

`NetUtils.matchIpExpression(String pattern, String address)` should correctly parse a normal IPv4 address with port, such as `192.168.1.63:90`, into host `192.168.1.63` and port `90`.

For example:

- `NetUtils.matchIpExpression("*.*.*.*:90", "192.168.1.63:90")` should return `true`.
- `NetUtils.matchIpExpression("192.168.1.*:90", "192.168.1.63:90")` should return `true`.
- `NetUtils.matchIpExpression("192.168.1.*:80", "192.168.1.63:90")` should return `false`.

### Anything else

This appears to be a small parsing bug in `NetUtils.matchIpExpression(String, String)`. The fix should avoid affecting IPv6 addresses, because IPv6 addresses also contain colons.

### Code of Conduct

yin-bo-Final:
Fixes #16308

This PR fixes `NetUtils.matchIpExpression(String, String)` so it can correctly parse normal IPv4 `host:port` addresses, such as `192.168.1.63:90`.

Previously, the method only split the address when it ended with `:`, so a normal IPv4 address with port was treated as the full host string and could fail with `UnknownHostException`.

The fix parses the address only when it contains exactly one colon, avoiding impact on IPv6 addresses.

## Checklist

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/mesh/rule/virtualservice/match/AddressMatch.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/mesh/rule/virtualservice/match/AddressMatch.java
@@ -18,6 +18,7 @@ package org.apache.dubbo.rpc.cluster.router.mesh.rule.virtualservice.match;
 
 import org.apache.dubbo.common.logger.ErrorTypeAwareLogger;
 import org.apache.dubbo.common.logger.LoggerFactory;
+import org.apache.dubbo.common.utils.StringUtils;
 
 import java.net.UnknownHostException;
 
@@ -60,7 +61,7 @@ public class AddressMatch {
     public boolean isMatch(String input) {
         if (getCird() != null && input != null) {
             try {
-                return input.equals(getCird()) || matchIpExpression(getCird(), input);
+                return input.equals(getCird()) || matchCird(input);
             } catch (UnknownHostException e) {
                 logger.error(
                         CLUSTER_FAILED_EXEC_CONDITION_ROUTER,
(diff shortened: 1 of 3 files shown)


yin-bo-Final:
The failed Samples Test seems caused by Docker Hub timeout while pulling `eclipse-temurin:21-jdk` and `mysql:5.7.40`.

zrlw (dismissed):
LGTM

yin-bo-Final:
Thanks for the review!

RainYuY (dismissed):
LGTM,But I think a better way is to directly deprecate this method. In fact, we have a three-parameter method that supports normal calls, just migrate the calls from mesh over

yin-bo-Final:
Thanks for the suggestion😊, I updated the PR to deprecate the two-argument method and migrated the mesh AddressMatch usage to the three-argument matchIpExpression API.

I also added AddressMatch tests to cover IPv4 host:port matching, port mismatch, and IPv6 matching.

And I noticed the field/method name is currently `cird`, which looks like a typo of `cidr`🤔, but I kept it unchanged in this PR to avoid introducing an unrelated compatibility change.

If needed, I can handle the naming cleanup in a separate PR while keeping backward-compatible getCird/setCird methods.

zrlw (approved):

RainYuY (approved):

PR #16350, opened 2026-06-19
Fix StringUtils.substringAfter to return empty string when the separator is absent

### LINKED ISSUE #16349: [Bug] StringUtils.substringAfter returns the original string instead of an empty string when the separator is absent
### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

`3.3` branch (current HEAD). The method was introduced in #14924.

`StringUtils.substringAfter("abc", ':')` returns `"abc"`.

The method's Javadoc states *"If nothing is found, the empty string is returned."*, but when the
separator is absent it returns the original string instead:

```
substringAfter("a:b", ':')  -> "b"     // ok
substringAfter("abc", ':')  -> "abc"   // expected ""
```

`StringUtils.substringAfter("abc", ':')` should return `""` (the empty string), as documented.

### Root cause

In `substringAfter`, the not-found branch returns `str`:

```java
int index = str.indexOf(separator);
return index == INDEX_NOT_FOUND ? str : str.substring(index + 1);
```

This was copy-pasted from the adjacent `substringBefore`, whose contract is to return the original
string. The sibling method `substringAfterLast` correctly returns `EMPTY_STRING` in the same
situation, so the two "after" methods are inconsistent. The behavior also diverges from Apache
commons-lang3's `substringAfter`, which returns the empty string when the separator is absent.

### Impact

`substringAfter` is a public utility method. Although it currently has no internal callers, any
future caller relying on the documented contract would silently receive the original string instead
of an empty string.

ns3154:

`StringUtils.substringAfter(String, int)` is documented to return the empty string when the
separator is not found ("If nothing is found, the empty string is returned."), but it currently
returns the **original string** instead:

```
substringAfter("a:b", ':')  -> "b"     // ok
substringAfter("abc", ':')  -> "abc"   // expected "", actually returns "abc"
```

The not-found branch was copy-pasted from the neighbouring `substringBefore` (whose contract *is*
to return the original string), which left `substringAfter` inconsistent with both its own Javadoc
and its sibling `substringAfterLast` (which correctly returns the empty string). It also diverges
from commons-lang3's `substringAfter`.

This PR makes `substringAfter` return `EMPTY_STRING` when the separator is absent, and adds a unit
test (`testSubstringAfter`) covering the documented behaviour. The new test fails before the fix
(`expected: <> but was: <abc>`) and passes after it; the full `StringUtilsTest` suite stays green.

Fixes #16349

## Checklist

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/StringUtils.java
@@ -1308,7 +1308,7 @@ public final class StringUtils {
             return str;
         }
         int index = str.indexOf(separator);
-        return index == INDEX_NOT_FOUND ? str : str.substring(index + 1);
+        return index == INDEX_NOT_FOUND ? EMPTY_STRING : str.substring(index + 1);
     }
 
     /**
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/utils/StringUtilsTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/utils/StringUtilsTest.java
@@ -503,4 +503,18 @@ class StringUtilsTest {
         assertTrue(startsWithIgnoreCase("dubbo.Application.name", "dubbo.application."));
         assertTrue(startsWithIgnoreCase("Dubbo.application.name", "dubbo.application."));
     }
+
+    @Test
+    void testSubstringAfter() {
+        // null or empty input is returned as-is
+        assertThat(StringUtils.substringAfter(null, ':'), is(nullValue()));
+        assertEquals("", StringUtils.substringAfter("", ':'));
+        // returns the substring after the first occurrence of the separator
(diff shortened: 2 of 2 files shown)


zrlw:
both substringBefore and substringAfter are not used, maybe we should delete them instead of fixing them.

zrlw:
@oxsean

oxsean (dismissed):

zrlw (dismissed):
LGTM

zrlw (approved):
LGTM

oxsean (approved):

ns3154:
Thanks for the suggestion, @zrlw! That's a fair point — both methods have no internal callers today. I chose to fix rather than delete because `StringUtils` is a public class and these are public methods: removing them could silently break any downstream project that happens to use them. Fixing the documented contract felt like the safer, non-breaking option. Happy to defer to the team's preference if deletion is favoured, though!

PR #16358, opened 2026-06-27
Detect ip:port with contains in NetUtils.matchIpExpression

vasiliy-mikhailov:
The deprecated two-arg `matchIpExpression(pattern, address)` uses `address.endsWith(":")` to detect the `ip:port` format. A real address such as `192.168.1.63:90` ends with the port, not a colon, so the host/port split is skipped — the full `host:port` string is then used as the host and matching fails. Use `contains(":")` to detect the separator.

Added a test covering `ip:port` addresses against patterns with and without a port.

## Verifying this change

The added `NetUtilsTest#testMatchIpExpressionDeprecatedTwoArgsWithIpPort` fails before this change and passes after:

Before the fix:

```
$ mvn test -Dtest=NetUtilsTest -pl dubbo-common
[INFO] Running org.apache.dubbo.common.utils.NetUtilsTest
[ERROR] Tests run: 49, Failures: 0, Errors: 1, Skipped: 2 <<< ERROR!
java.net.UnknownHostException: 192.168.1.63:90: invalid IPv6 address literal
	at org.apache.dubbo.common.utils.NetUtils.matchIpRange(NetUtils.java:784)
	at org.apache.dubbo.common.utils.NetUtils.matchIpExpression(NetUtils.java:753)
	at org.apache.dubbo.common.utils.NetUtilsTest.testMatchIpExpressionDeprecatedTwoArgsWithIpPort(NetUtilsTest.java:357)
[INFO] BUILD FAILURE
```

After the fix:

```
$ mvn test -Dtest=NetUtilsTest -pl dubbo-common
[INFO] Running org.apache.dubbo.common.utils.NetUtilsTest
[INFO] Tests run: 49, Failures: 0, Errors: 0, Skipped: 2
[INFO] BUILD SUCCESS
```

## AI assistance disclosure

This contribution was produced with the help of an AI pipeline. The pipeline processed a large amount of source code to surface suspected bugs, reproduced a subset of them with failing unit tests and generated candidate fixes, and prepared pull requests from the ones that held up. Each PR was then reviewed and verified by a human before being opened: the fix and test were checked by hand and the test was confirmed to fail before the change and pass after.

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/utils/NetUtils.java
@@ -738,7 +738,7 @@ public final class NetUtils {
         String host = address;
         int port = 0;
         // only works for ipv4 address with 'ip:port' format
-        if (address.endsWith(":")) {
+        if (address.contains(":")) {
             String[] hostPort = address.split(":");
             host = hostPort[0];
             port = StringUtils.parseInteger(hostPort[1]);
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/utils/NetUtilsTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/utils/NetUtilsTest.java
@@ -347,6 +347,22 @@ class NetUtilsTest {
     }
 
     @Test
+    void testMatchIpExpressionDeprecatedTwoArgsWithIpPort() throws UnknownHostException {
+        // The deprecated 2-arg matchIpExpression should correctly parse host and port from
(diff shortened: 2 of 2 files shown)


vasiliy-mikhailov:
The single failing check (Unit Test, Java 8) is `AbstractMetadataReportTest.testPublishAll` (expected 3 but was 2), an async retry-timer count race in `dubbo-metadata`. This change only touches `dubbo-common`'s `NetUtils.matchIpExpression`, and the Java 11/17/21/25 jobs all pass, so the failure looks unrelated and flaky. Could a maintainer re-run the Java 8 job when convenient? Happy to rebase if that helps.

Issue #16330, opened 2026-06-15
[Bug] findReflectionMethodDescriptor generic block should return early like echo block


LI123456mo:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Any - this is a code inconsistency, not version specific. 
Observed in current main branch.

### Steps to reproduce this issue

This is a code inconsistency found in StubServiceDescriptor.java 
in the findReflectionMethodDescriptor method.

Location:
dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/model/StubServiceDescriptor.java

The isGeneric block assigns to a variable and returns at the end of the method:
    methodDescriptor = ServiceDescriptorInternalCache.genericService()
            .getMethods(methodName)
            .get(0);

While the isEcho block returns immediately:
    return ServiceDescriptorInternalCache.echoService()
            .getMethods(methodName)
            .get(0);

Both blocks do the same logical thing so they should be consistent.
The generic block should also return immediately like the echo block.

Suggested fix:
    return ServiceDescriptorInternalCache.genericService()
            .getMethods(methodName)
            .get(0);

### What you expected to happen

Both the isGeneric and isEcho blocks should behave consistently 
since they do the same logical thing. The isGeneric block should 
return immediately like the isEcho block does, instead of assigning 
to a variable and waiting to return at the end of the method.

### Anything else

This is a minor code inconsistency, not a functional bug. 
The code produces correct results either way, but the inconsistency 
makes the code harder to read and understand.

The fix is a one line change in findReflectionMethodDescriptor:

Current:
    methodDescriptor = ServiceDescriptorInternalCache.genericService()
            .getMethods(methodName)
            .get(0);

Should be:
    return ServiceDescriptorInternalCache.genericService()
            .getMethods(methodName)
            .get(0);

I am happy to submit a PR for this if maintainers agree.

### Do you have a (mini) reproduction demo?

- [x] Yes, I have a minimal reproduction demo to help resolve this issue more effectively!

### Are you willing to submit a pull request to fix on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

[...]

PR #16325, opened 2026-06-09
Fix Triple method resolution for overloaded wrapper methods

skt-shinyruo:
## Summary
- Prefer exact Triple wrapper arg-type matching when resolving overloaded methods.
- Only fall back to the generated unary/server-stream pair when the pair is verified.
- Fail fast on ambiguous overloads instead of guessing.

Fixes #14939

## Verification
- `mvn -pl dubbo-rpc/dubbo-rpc-triple -Dtest=DescriptorUtilsTest,ReflectionPackableMethodTest test`
- Full `mvn test -Djdk.net.hosts.file=/tmp/dubbo-test-hosts` reached `dubbo-metrics-registry` and failed there in `RegistryMetricsTest.testAvgResponseTime` with an assertion mismatch unrelated to this patch.

PR #16359, opened 2026-06-27
Simplify generic return in DescriptorUtils

Zhengcy05:

This PR makes `DescriptorUtils#findReflectionMethodDescriptor` return directly from the generic branch, matching the existing echo branch.

fixs #16330

The change is a small consistency cleanup in `dubbo-rpc-triple`. It does not intend to change behavior. The generic and echo paths both resolve a single cached descriptor, so the generic branch can return immediately instead of assigning to a local variable and falling through to the final return.

## Checklist
- [x] Make sure there is a [[GitHub_issue](https://github.com/apache/dubbo/issues)](https://github.com/apache/dubbo/issues) field for the change.
- [x] Write necessary unit-test to verify your logic correction. No new unit test was added because this is a behavior-preserving refactor.
- [x] Make sure gitHub actions can pass. Local verification was attempted with `mvn -pl dubbo-rpc/dubbo-rpc-triple -am -DskipTests compile`, but the build was blocked by an existing `dubbo-maven-plugin` plugin descriptor issue (`No plugin descriptor found at META-INF/maven/plugin.xml`), which is unrelated to this change.

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/DescriptorUtils.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/DescriptorUtils.java
@@ -86,10 +86,9 @@ public final class DescriptorUtils {
 
     public static MethodDescriptor findReflectionMethodDescriptor(
             ServiceDescriptor serviceDescriptor, String methodName) {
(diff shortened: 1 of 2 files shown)


LI123456mo (commented):
Good Idea, but ,  the new PR https://github.com/apache/dubbo/pull/16325, has already covered this

Zhengcy05:
> Good Idea, but , the new PR #16325, has already covered this

Thank you for your reply, I will close this PR.

PR #16370, opened 2026-07-04
Fix VirtualThreadPool benchmark methodology issue (#16174)

### LINKED ISSUE #16174: [Feature] Request to revisit VirtualThreadPool optimization: Potential benchmark methodology issue
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Descriptions

**Summary**
While reviewing the VirtualThreadPool refactoring (#16042, merged via #16055), I noticed a potential issue in the benchmark methodology that may have affected the performance conclusions. I'd like to bring this to the attention of the original authors and reviewers for re-evaluation.

**Observation**
In the benchmark code referenced in the discussion, there is a comment indicating a correction:

countDownLatch2.await();  // Fixed: should await countDownLatch2 (not the first one)
I ran some tests and found that:

When correctly awaiting countDownLatch2, the non-pooled approach (Executors.newThreadPerTaskExecutor) appears to show better performance
The pooled approach only appeared faster when the benchmark was using the incorrect latch (countDownLatch1)
This suggests the performance comparison may have been affected by synchronization timing issues rather than actual execution efficiency.

**Questions**
@funky-eyes  @heliang666s @RainYuY @zrlw  — Could you help verify:

Was the benchmark data in #16042 collected before or after the latch fix?
Would it be possible to re-run the benchmark with the corrected synchronization to confirm the results?

**Suggestion**
Given that the pooled implementation adds configuration complexity (THREADS_VIRTUAL_CORE), it would be valuable to ensure the performance benefits are validated with the correct measurement approach. I'm happy to assist with re-testing if needed.

Thank you for your time and for the work on this feature. Looking forward to your thoughts.

### Related issues

_No response_

### Code of Conduct

anushkagupta200615-jpg:

Fixes #16174

This PR addresses the benchmark methodology issue and the root cause of the `VirtualThreadPool` configuration bug introduced in #16055.

**Problem:**
1. **Broken `keepAliveTime`:** The pooled virtual thread implementation used `keepAliveTime = 0L MILLISECONDS`. Because of this, non-core threads died immediately after completing a task. This defeated the entire purpose of the pooled mode (as outlined in #16042), which is to keep threads warm so that libraries like FastJSON and Aerospike can reuse their heavy `ThreadLocal` byte buffers across requests.
2. **Flawed Benchmark:** The benchmark in the original PR used a flawed two-latch synchronization pattern. It awaited the `startGate` (which was already zero) instead of the `completionLatch`. As a result, the timing window closed *before* the tasks were actually completed, masking the `keepAlive` overhead and falsely inflating the performance metrics of the pooled executor.

**Solution:**
- **Fixed `keepAliveTime`:** Updated `VirtualThreadPool` to use `60L SECONDS` for `keepAliveTime` in the pooled branch. This ensures warm virtual threads survive long enough to reuse their `ThreadLocal` caches between subsequent requests.
- **Added Regression Guards:** Enhanced `VirtualThreadPoolTest` to verify that `keepAliveTime > 0` and added a test `getExecutor4_threadLocalReuseInPooledMode` proving that `ThreadLocal` state is actually preserved and reused across sequential tasks on the pooled executor.
- **Added Correct Benchmark Test:** Introduced `VirtualThreadPoolBenchmarkTest.java` that uses the correct two-latch pattern (awaiting the completion latch) to strictly verify that all submitted tasks run to completion within the benchmark window for both pooled and unpooled modes.

## Checklist

### DIFF
--- a/dubbo-plugin/dubbo-plugin-loom/src/main/java/org/apache/dubbo/common/threadpool/support/loom/VirtualThreadPool.java
+++ b/dubbo-plugin/dubbo-plugin-loom/src/main/java/org/apache/dubbo/common/threadpool/support/loom/VirtualThreadPool.java
@@ -23,28 +23,69 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 import static org.apache.dubbo.common.constants.CommonConstants.DEFAULT_THREAD_NAME;
 import static org.apache.dubbo.common.constants.CommonConstants.THREADS_VIRTUAL_CORE;
 import static org.apache.dubbo.common.constants.CommonConstants.THREAD_NAME_KEY;
 
 /**
- * Creates a thread pool that use virtual thread
+ * Creates a thread pool that uses virtual threads (Project Loom).
+ *
+ * <p>Two operating modes are supported:
+ *
+ * <ul>
+ *   <li><b>Unpooled mode (default)</b>: When {@code threads.virtual.core} is not set (or {@code
+ *       <= 0}), a new virtual thread is created for every submitted task via {@link
+ *       Executors#newThreadPerTaskExecutor(java.util.concurrent.ThreadFactory)}. This is the
+ *       simplest and most scalable mode.
+ *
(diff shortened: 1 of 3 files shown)


zrlw:
@funky-eyes PTAL

zrlw (approved):
LGTM

PR #16374, opened 2026-07-09
Fix PU plaintext detection when AuthPolicy.NONE is configured

### LINKED ISSUE #16373: [Bug] Port unification skips plaintext detection when AuthPolicy.NONE is configured
### Pre-check

- [x] I am sure that all the content I provide in English is directly translated by me and not produced by machine translation.

### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/dubbo/issues) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Latest `3.3` branch, HEAD `316df8e57677cc8c1ed6d92bd40ca0daba27393f` (`3.3.7-SNAPSHOT`).

1. Enable port unification on the provider side.
2. Configure a `CertProvider` that returns a `ProviderCert` with `AuthPolicy.NONE`, which means TLS is allowed but not required.
3. Send a plaintext Dubbo request to the port-unification port. The first bytes are non-TLS, for example the Dubbo magic bytes `0xda 0xbb`.
4. `NettyPortUnificationServerHandler` enters the TLS-detection branch because provider TLS config exists and enough bytes are available for SSL detection.
5. Since the first packet is not TLS and `AuthPolicy` is `NONE`, the current handler neither closes the channel nor continues with plaintext protocol detection.

When `AuthPolicy.NONE` is configured, a non-TLS first packet should be treated as a permitted plaintext connection and should continue through port-unification protocol detection.

### What actually happened

The plaintext `WireProtocol` detector is not invoked. The connection remains open and waits for more data, so a plaintext Dubbo consumer can hang or time out when it connects to a provider that uses port unification with optional TLS.

### Anything else

The root cause is in `NettyPortUnificationServerHandler#decode`.

When provider TLS config exists and the inbound bytes are enough for SSL detection, the handler checks:

- TLS first packet: enable SSL.
- Non-TLS first packet with `AuthPolicy` not equal to `NONE`: log and close.
- Non-TLS first packet with `AuthPolicy.NONE`: no action is taken and the code does not fall through to `detectProtocol(...)`.

I have a regression test that reproduces the issue on current `3.3`: before the fix, the test fails because the plaintext detector is called `0` times instead of `1` time.

I will submit a pull request with the fix and the regression test.

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

uuuyuqi:

GitHub_issue: #16373

Fixes #16373

This pull request fixes port-unification protocol detection when provider TLS is configured with `AuthPolicy.NONE`.

`AuthPolicy.NONE` means TLS is optional. If a client sends a non-TLS first packet to the port-unification port, the server should continue with plaintext protocol detection. The current `3.3` branch enters the TLS-detection branch, sees that the packet is not TLS, and then does nothing when the auth policy is `NONE`. As a result, plaintext Dubbo detection is skipped and the connection waits until the caller times out.

The fix keeps the existing strict-TLS behavior unchanged:

- TLS first packet: enable SSL and return.
- Non-TLS first packet with required TLS: log, discard data, close the channel, and return.
- Non-TLS first packet with `AuthPolicy.NONE`: fall through to plaintext `detectProtocol(...)`.

This PR also adds a regression test that provides an optional-TLS test `CertProvider`, sends a Dubbo plaintext first packet, and verifies that the plaintext protocol detector is invoked.

Verification:

- Before the production fix, the new regression test failed because the detector count was `0` instead of `1`.
- `mvn -q -pl dubbo-remoting/dubbo-remoting-netty4 -am -Dtest=NettyPortUnificationServerHandlerTest,PortUnificationServerTest,SslServerTlsHandlerTest -Dsurefire.failIfNoSpecifiedTests=false -DskipITs -DskipIntegrationTests=true -Djacoco.skip=true test`
- `mvn -q -pl dubbo-remoting/dubbo-remoting-netty4 -am -DskipTests -DskipIntegrationTests=true -Djacoco.skip=true -Dcheckstyle.skip=false -Dcheckstyle_unix.skip=false -Drat.skip=false verify`
- `mvn -q -pl dubbo-remoting/dubbo-remoting-netty4 -am spotless:check`

## Checklist

- [x] Make sure there is a [GitHub_issue](https://github.com/apache/dubbo/issues) field for the change (usually before the description, using `GitHub_issue: #123456`).
- [x] Write necessary unit-test to verify your logic correction, more mock a little better when cross module dependency exists. If the new feature or significant change is added, please remember to add integration-test in [dubbo-samples](https://github.com/apache/dubbo-samples) project.
- [x] Make sure gitHub actions can pass, only after CI passed will we review it.

### DIFF
--- a/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyPortUnificationServerHandler.java
+++ b/dubbo-remoting/dubbo-remoting-netty4/src/main/java/org/apache/dubbo/remoting/transport/netty4/NettyPortUnificationServerHandler.java
@@ -127,27 +127,29 @@ public class NettyPortUnificationServerHandler extends ByteToMessageDecoder {
         if (providerConnectionConfig != null && canDetectSsl(in)) {
             if (isSsl(in)) {
                 enableSsl(ctx, providerConnectionConfig);
-            } else {
-                // check server should load TLS or not
-                if (providerConnectionConfig.getAuthPolicy() != AuthPolicy.NONE) {
-                    byte[] preface = new byte[in.readableBytes()];
-                    in.readBytes(preface);
-                    LOGGER.error(
-                            CONFIG_SSL_CONNECT_INSECURE,
-                            "client request server without TLS",
-                            "",
-                            String.format(
-                                    "Downstream=%s request without TLS preface, but server require it. " + "preface=%s",
-                                    ctx.channel().remoteAddress(), Bytes.bytes2hex(preface)));
-
-                    // Untrusted connection; discard everything and close the connection.
-                    in.clear();
-                    ctx.close();
-                }
+                return;
             }
-        } else {
-            detectProtocol(ctx, url, channel, in);
+            // check server should load TLS or not
+            if (providerConnectionConfig.getAuthPolicy() != AuthPolicy.NONE) {
+                byte[] preface = new byte[in.readableBytes()];
+                in.readBytes(preface);
+                LOGGER.error(
+                        CONFIG_SSL_CONNECT_INSECURE,
+                        "client request server without TLS",
+                        "",
+                        String.format(
+                                "Downstream=%s request without TLS preface, but server require it. " + "preface=%s",
+                                ctx.channel().remoteAddress(), Bytes.bytes2hex(preface)));
+
+                // Untrusted connection; discard everything and close the connection.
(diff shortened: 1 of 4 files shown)


CrazyHZM (approved):
LGTM

zrlw (approved):
LGTM

PR #16388, opened 2026-07-20
Skills/add dubbo overview skill

Sourabh-Majhee:
## Checklist

### DIFF
--- a/skills/dubbo-overview/SKILL.md
+++ b/skills/dubbo-overview/SKILL.md
@@ -36,8 +36,8 @@ Apache Dubbo is a Java RPC and microservice framework. Its architecture
 divides into two layers:
 
 **Data Plane** — Provider and Consumer JVM processes that communicate
-directly via RPC protocols (Triple/HTTP2, Dubbo/TCP, REST). Every remote
-call travels through the data plane.
+directly via RPC protocols (Triple/HTTP2 with REST support, Dubbo/TCP).
+Every remote call travels through the data plane.
 
 **Control Plane** — Three optional centers that govern how data plane
 processes discover each other, share configuration, and exchange metadata:
@@ -56,8 +56,7 @@ Consumer with a direct connection and no registry at all.
 
 Handles registration and discovery of service instances.
 
-Supported registries: Nacos, Zookeeper, Consul, Kubernetes (native),
-Multicast (dev only), Etcd, Redis.
+Supported registries: Nacos, Zookeeper, Multicast (dev only).
 
 **Interface-level discovery (Dubbo 2 default)**
 One registry entry per interface per instance.
@@ -88,7 +87,7 @@ Stores the detailed interface metadata (methods, parameters, return types)
 that application-level service discovery no longer puts in the registry.
 Consumers retrieve this to know exactly what methods a provider exposes.
 
-Supported: Nacos, Zookeeper, Redis. Optional if using interface-level
+Supported: Nacos, Zookeeper. Optional if using interface-level
 discovery. **Required for any Dubbo 2 to Dubbo 3 migration**, because
 without it, consumers cannot resolve the interface-to-application mapping.
 
@@ -155,7 +154,7 @@ dubbo:
     # register-mode: instance
   protocol:
     name: tri          # Triple (HTTP/2, gRPC-compatible) — recommended for Dubbo 3
-    port: 50051
+    port: -1           # -1 = auto-assign random available port (recommended)
   registry:
     address: nacos://127.0.0.1:8848
   # Required if migrating from Dubbo 2, or if mixing Dubbo 2 and Dubbo 3 consumers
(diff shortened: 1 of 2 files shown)


skills/dubbo-overview/SKILL.md
```diff
@@ -0,0 +1,311 @@
...
+annotations. For consumers, `@DubboReference` injection works via
+Spring Boot auto-configuration (`DubboAutoConfiguration`) even without
+`@EnableDubbo`, but adding it is still recommended for consistency.
+
+**4. Defaulting to `dubbo` protocol instead of `tri` in Dubbo 3**
+Dubbo 3 recommends the `tri` (Triple) protocol. It is HTTP/2 based,
+gRPC-compatible, and works through standard API gateways. New services
+should use `tri` unless they need backward TCP compatibility.
+
+**5. Omitting MetadataCenter during Dubbo 2→3 migration**
```
RainYuY: During mixed Dubbo 2/3 deployments, providers must retain interface-level registrations using register-mode=all; Dubbo 2 consumers cannot interpret application-level registrations or service-name mappings. Following the current guidance could cause No provider available. The same misconception appears at L90-L92, L160, and README L33. See the [official migration guide](https://dubbo.apache.org/en/overview/mannual/java-sdk/reference-manual/upgrades-and-compatibility/migration-service-discovery/).

skills/dubbo-overview/SKILL.md
```diff
@@ -0,0 +1,311 @@
...
+does not enable dynamic configuration.
+
+**7. Setting `register-mode: interface` in Dubbo 3 for new services**
+The Dubbo 3 default is `all` (dual registration at both interface and
+application level). For new greenfield services, prefer `instance`
+(application-level only) for better scalability. Only use `interface`
+mode when you need compatibility with existing Dubbo 2 consumers that
+have not been upgraded.
+
+**8. QoS port conflict when running multiple Dubbo apps locally**
```
RainYuY, line 322: [QosProtocolWrapper](https://github.com/apache/dubbo/blob/0ce4bb6f997b6df483b48daa5b20454b1ae8e36c/dubbo-plugin/dubbo-qos/src/main/java/org/apache/dubbo/qos/protocol/QosProtocolWrapper.java#L99-L151) defaults qos.check to false, catches the bind exception, and only throws when QoS checking is enabled on the provider. The text should state that the QoS server fails to start and logs a warning, while the application normally continues.

RainYuY (changes requested):

skills/dubbo-overview/SKILL.md
```diff
@@ -0,0 +1,311 @@
...
+
+**5. Omitting MetadataCenter during Dubbo 2→3 migration**
+When Dubbo 3 providers register at the application level, Dubbo 2
+consumers cannot discover them without the Metadata Center bridging the
+interface-to-application mapping. Without it, consumers throw
+`No provider available`.
+
+**6. Confusing Registry Center with Config Center**
+Registry = where services are. Config = how services behave.
+A service can use Nacos as both, but they are configured separately under
```
RainYuY: The statement that registry-only configuration does not enable dynamic configuration is the opposite of the current 3.3 behavior. When no Config Center is explicit, DefaultApplicationDeployer.useRegistryAsConfigCenterIfNecessary() synthesizes one from a supported registry whenever use-as-config-center is unset or true; Nacos and Zookeeper therefore enable this reuse for the sample above. Please document the default reuse and the use-as-config-center: false opt-out instead, otherwise the skill will teach assistants to misdiagnose working governance configuration.

RainYuY (changes requested):
Re-reviewed at f7e49cbb. The latest update only merges 3.3; git diff 0ce4bb6f..f7e49cbb -- skills is empty, so both existing blocking comments remain unaddressed: (1) mixed Dubbo 2/3 migration must retain interface-level registration, because a Metadata Center does not make Dubbo 2 consumers understand application-level registrations; and (2) a QoS bind conflict only logs a warning and lets the application continue by default because qos.check defaults to false. I found one additional configuration-center correctness issue inline. Requesting changes until all three are corrected.

skills/dubbo-overview/SKILL.md
Sourabh-Majhee: Good catch, thanks. Updated to document the default registry-as-config-center reuse via useRegistryAsConfigCenterIfNecessary() and added the use-as-config-center: false opt-out. Should accurately reflect current 3.3 behavior now.

skills/dubbo-overview/SKILL.md
Sourabh-Majhee: You're right, fixed. Now states that qos.check defaults to false, so the bind conflict just logs a warning and the app continues, and only throws if qos-check=true is set.

skills/dubbo-overview/SKILL.md
Sourabh-Majhee: You're right, thanks. Updated section 5 (renamed to "Breaking Dubbo 2 consumers"), the Metadata Center description, and the migration YAML comment to state providers must keep register-mode: all so Dubbo 2 consumers can still discover services via interface-level entries — a Metadata Center alone doesn't fix this. Also confirmed skills/README.md L33 already reflects this correctly.

Sourabh-Majhee:
All three points addressed in the latest commits — migration guidance (section 5, Metadata Center description, migration YAML, README L33), QoS behavior, and Config Center default reuse. Replied inline on each. Let me know if anything else needs another pass.

PR #16393, opened 2026-07-23
Fix instance push loss during initial subscribe in ServiceDiscoveryRegistry

### LINKED ISSUE #16392: [Bug] Instance push loss during initial subscribe in ServiceDiscoveryRegistry
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

- Dubbo: **3.2.19** in production, reproduced against **apache/3.3** HEAD
- Registry: **Nacos 2.x** (gRPC client, application-level discovery)
- Discovery mode: `APPLICATION_FIRST`

The bug is deterministic and can be reproduced without a Nacos server, using only a failing unit test that captures the race.

1. Clone apache/dubbo at the 3.3 branch:

       git clone https://github.com/apache/dubbo.git
       cd dubbo
       git checkout 3.3

2. Add my fork as a remote and fetch the branch:

       git remote add hnpkso https://github.com/hnpkso/dubbo.git
       git fetch hnpkso fix/subscribe-push-loss

3. Check out just the test commit (source stays as unmodified 3.3):

       git checkout f9755db2

4. Run the tests:

       mvn -pl dubbo-registry/dubbo-registry-api test \
           -Dtest=ServiceDiscoveryRegistryTest

5. Observe 3 failures out of 6:

    - testSubscribeURLsRegistersPushCallbackBeforePull
      InOrder verify: addServiceInstancesChangedListener must precede
      getInstances. Fails today because it runs afterwards.

    - testSubscribeURLsSkipsPullWhenPushAlreadyPopulated
      A mocked push populates allInstances the moment the callback is
      registered; the code proceeds to pull anyway and would overwrite
      the fresher push snapshot.

    - testSubscribeURLs
      A pre-existing test's `verify(..., times(2))` on the register call
      captures a related idempotency issue: register is invoked on every
      subscribeURLs() call, not once per listener lifetime.

6. Check out the fix commit `8e0f8f2d81` — all 6 tests pass.

The failing tests directly model the production race:
ServiceDiscoveryRegistry.subscribeURLs() calls getInstances() and runs
the resulting onEvent() (which synchronously fetches metadata over the
network) BEFORE registering the push callback via
addServiceInstancesChangedListener(). Any push arriving in this window
is absorbed by the discovery SDK's local cache and never reaches Dubbo.

### Expected

When Nacos removes a provider instance during the consumer's initial subscribe, the removal push should be delivered to Dubbo's ServiceInstancesChangedListener and update `allInstances`. The consumer's
invoker table should reflect the current state of the registry.

### What actually happened

`ServiceDiscoveryRegistry.subscribeURLs()` performs its initial `getInstances()` pull before registering the push callback via `serviceDiscovery.addServiceInstancesChangedListener(...)`.

The pull path also runs `ServiceInstancesChangedListener.onEvent()`, which fetches metadata over the network — in production this took ~500 ms. Any push arriving between the pull and the register call is
absorbed by the discovery SDK's local cache and never delivered to Dubbo. From that point on, `allInstances` holds a permanently stale snapshot until either:

- the same appName has a later, independent push event (which then reconciles), or
- the consumer restarts.

Neither is guaranteed to happen; we saw a stale entry survive for the entire process lifetime.

### Impact — production incident

Consumer routes RPCs to a provider that no longer exports the service, producing recurring errors:

```
RpcException: Failed to invoke the method queryList in the service ...
  provider: DefaultServiceInstance{host='10.87.175.50', port=20880, ...}
  cause: RemotingException: Fail to decode request due to:
         RpcInvocation [methodName=queryList, parameterTypes=null]
```

The address `10.87.175.50` had already been removed from Nacos before the consumer's subscribe completed, but the removal push landed inside the loss window.

### Timeline (production log)

The Nacos client log and Dubbo log run in different timezones; both refer to the same wall clock. Offsets below are relative to when `subscribeURLs` entered.

| Offset | Source | Event |
|---|---|---|
| T+0ms | Dubbo | `Trying to subscribe from apps [3 apps]` — `subscribeURLs` enters |
| T+27ms | Nacos client | `init new ips(4)` — snapshot returned to `selectInstances`, contains the ghost |
| T+45ms | Dubbo | `Received instance notification, instances: 4` — pull result written into `allInstances` |
| **T+119ms** | **Nacos client** | **`removed ips(1)` — server pushes removal, absorbed by SDK cache** |
| T+844ms | Dubbo | `Start NettyClient /...:20880` — synchronous metadata fetch begins |
| T+1345ms | Dubbo | metadata fetch completes (~500 ms elapsed) |
| T+1.4s+ | Dubbo | `addServiceInstancesChangedListener` finally called — too late |

The removal push arrives 119 ms after `subscribeURLs` starts and ~1.2 s before the push callback is registered.

### Root cause (code reference on apache/dubbo `3.3`)

`dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/ServiceDiscoveryRegistry.java`:

```java
protected void subscribeURLs(URL url, NotifyListener listener, Set<String> serviceNames) {
    ...
    if (serviceInstancesChangedListener == null) {
        serviceInstancesChangedListener = serviceDiscovery.createListener(serviceNames);
        for (String serviceName : serviceNames) {
            List<ServiceInstance> serviceInstances =
                    serviceDiscovery.getInstances(serviceName);            // (1) pull
            if (CollectionUtils.isNotEmpty(serviceInstances)) {
                serviceInstancesChangedListener.onEvent(...);              // (2) blocks on metadata fetch
            }
        }
        serviceListeners.put(serviceNamesKey, serviceInstancesChangedListener);
    }
    ...
    serviceDiscovery.addServiceInstancesChangedListener(...);              // (3) register — too late
}
```

Between (1) and (3), the discovery SDK is holding pushes without a subscriber attached to Dubbo. Those pushes never reach `ServiceInstancesChangedListener.doOnEvent()`, so `allInstances` is never refreshed.

### Affected versions

Same code path exists unchanged on 3.2, 3.3, and 3.4 — the last touch to this method predates the fix window. The `MappingListener` sibling race in the same method was fixed by #14851, but the instance-list side (this issue) was left untouched.

### Proposed fix

PR #16393 is open. Summary:

- register `addServiceInstancesChangedListener` before the initial pull;
- serialize the pull loop with `synchronized(listener)` so it shares the intrinsic lock that `doOnEvent` already holds;
- skip the pull for any `serviceName` a push has already populated (push wins over stale pull).

The PR ships two commits (test + fix) so reviewers can `git checkout` each and observe the tests transition from failing to passing, without any Nacos setup.

### Code of Conduct

hnpkso:
# PR: Fix instance push loss during initial subscribe in ServiceDiscoveryRegistry

**Base branch**: `apache/dubbo:3.3` (also applies cleanly to `3.2` and `3.4`)
**Commits**: 2 (test-only `f9755db2`, then fix `ec2ba983`). Check out either to see the failing-then-passing tests.

---

## What this fixes

Fixes #16392.

`ServiceDiscoveryRegistry.subscribeURLs()` currently:

1. calls `serviceDiscovery.getInstances(serviceName)` — the initial pull,
2. runs `listener.onEvent(...)` which synchronously fetches metadata (can take seconds),
3. **only then** calls `serviceDiscovery.addServiceInstancesChangedListener(...)` to arm the push callback.

Any push arriving between step 1 and step 3 is absorbed by the discovery SDK's local cache and never reaches Dubbo. `allInstances` is then permanently stale until either (a) a later, independent push happens for the same appName, or (b) the process restarts.

We hit this in production (Nacos 2.x, Dubbo 3.2.19): a provider Nacos had removed 92 ms after our subscribe started remained a ghost invoker for the entire process lifetime, producing recurring `Fail to decode request` errors.

Timeline and full analysis are in the linked issue.

## The change

`dubbo-registry-api/…/ServiceDiscoveryRegistry.java` — `subscribeURLs()`:

1. **Register the push callback before the initial pull.** `addServiceInstancesChangedListener(...)` moves inside the `if (listener == null)` guard, ahead of the pull loop. Any push arriving during/after the pull now flows through the listener's already-armed `doOnEvent` path.

2. **Serialize the pull with the push callback.** Wrap the pull loop in `synchronized(serviceInstancesChangedListener)`. `doOnEvent` is a `synchronized` method on the same object; both now share the same intrinsic lock, so a push arriving during initialization queues cleanly instead of interleaving writes to `allInstances`.

3. **Push wins over stale pull.** Inside the synchronized region, check `listener.getAllInstances().containsKey(serviceName)` before pulling. If a push has already populated this service, treat push as authoritative and skip the pull. This prevents a slow pull (reading a stale SDK cache) from overwriting a fresher push snapshot.

Diff is under 30 lines.

## Before/after verification

Two commits are separated intentionally so the failing-then-passing tests can be observed directly:

```bash
git remote add hnpkso https://github.com/hnpkso/dubbo.git
git fetch hnpkso fix/subscribe-push-loss

git checkout  f9755db2       # test commit only

mvn -pl dubbo-registry/dubbo-registry-api test \
    -Dtest=ServiceDiscoveryRegistryTest
# Tests run: 6, Failures: 3
#   ✗ testSubscribeURLs (times(2) assertion on register)
#   ✗ testSubscribeURLsRegistersPushCallbackBeforePull  ← MAIN
#   ✗ testSubscribeURLsSkipsPullWhenPushAlreadyPopulated

git checkout ec2ba983       # fix on top

mvn -pl dubbo-registry/dubbo-registry-api test \
    -Dtest=ServiceDiscoveryRegistryTest
# Tests run: 6, Failures: 0
```

The red-to-green transition **is** the reproduction — no Nacos or external system needed.

Full module test suite: `Tests run: 111, Failures: 0`.

## Notes on behavior changes

1. **`addServiceInstancesChangedListener` cardinality**: now called once per listener lifetime, previously once per `subscribeURLs()` call. All in-tree `ServiceDiscovery` implementations already treat duplicate calls as no-ops via `instanceListeners.add()` early-return, so the change is observationally equivalent. The `testSubscribeURLs` assertion updates from `times(2)` to `times(1)` accordingly.

2. **`RegistryEvent.toSsEvent` metric cardinality** rides along with (1): it now fires once per listener creation instead of once per subscribe. If per-subscribe metric cardinality is preferred, `MetricsEventBus.post(...)` can be split back out of the register lambda without affecting correctness — happy to adjust.

3. **Push executor threads may briefly park** on the listener intrinsic lock during initial subscribe, bounded by the metadata fetch duration (~500 ms in our production). Previously such pushes were silently dropped, so this is a strict improvement.

## Scope

This PR only fixes the push-loss race in `subscribeURLs()`. `NacosServiceDiscovery` and other `ServiceDiscovery` implementations are untouched; no new SPI contract. Any unrelated concerns in the same code path are out of scope for this change.

## Related

- #14851 fixed a sibling race on the `MappingListener` side of the same subscribe path (initial-apps was `null` when the listener was constructed). The instance-list side, addressed here, has the same shape of bug (populate-before-arm) but a different failure mode (silent instance loss vs. spurious mapping event).

## Commits

- `f9755db2` — `Add failing tests for instance push loss during initial subscribe`. Tests only, fails on unmodified 3.3.
- `ec2ba983` — `Fix instance push loss during initial subscribe in ServiceDiscoveryRegistry`. Source fix, turns tests green.

Feel free to squash on merge; the two commits exist purely to make the red→green transition observable during review.

### DIFF
--- a/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/ServiceDiscoveryRegistry.java
+++ b/dubbo-registry/dubbo-registry-api/src/main/java/org/apache/dubbo/registry/client/ServiceDiscoveryRegistry.java
@@ -352,31 +352,44 @@ public class ServiceDiscoveryRegistry extends FailbackRegistry {
             ServiceInstancesChangedListener serviceInstancesChangedListener = serviceListeners.get(serviceNamesKey);
             if (serviceInstancesChangedListener == null) {
                 serviceInstancesChangedListener = serviceDiscovery.createListener(serviceNames);
+                serviceListeners.put(serviceNamesKey, serviceInstancesChangedListener);
+
+                // Register the push callback BEFORE the initial pull, so any push arriving
+                // during or right after the pull is delivered to our listener instead of
+                // being swallowed by the discovery SDK's local cache. See DUBBO issue on
+                // "instance push loss during initial subscribe".
+                ServiceInstancesChangedListener listenerForRegister = serviceInstancesChangedListener;
+                String serviceDiscoveryName =
+                        url.getParameter(RegistryConstants.REGISTRY_CLUSTER_KEY, url.getProtocol());
+                MetricsEventBus.post(
+                        RegistryEvent.toSsEvent(
+                                url.getApplicationModel(), serviceKey, Collections.singletonList(serviceDiscoveryName)),
+                        () -> {
+                            serviceDiscovery.addServiceInstancesChangedListener(listenerForRegister);
+                            return null;
+                        });
+
                 for (String serviceName : serviceNames) {
-                    List<ServiceInstance> serviceInstances = serviceDiscovery.getInstances(serviceName);
-                    if (CollectionUtils.isNotEmpty(serviceInstances)) {
-                        serviceInstancesChangedListener.onEvent(
-                                new ServiceInstancesChangedEvent(serviceName, serviceInstances));
+                    // Hold the same intrinsic lock as ServiceInstancesChangedListener.doOnEvent()
+                    // to serialize with any concurrent push callback. If a push has already
+                    // populated this serviceName, treat push as authoritative and skip the pull —
+                    // otherwise a stale pull result could overwrite fresher push data.
+                    synchronized (serviceInstancesChangedListener) {
+                        if (serviceInstancesChangedListener.getAllInstances().containsKey(serviceName)) {
+                            continue;
+                        }
+                        List<ServiceInstance> serviceInstances = serviceDiscovery.getInstances(serviceName);
+                        if (CollectionUtils.isNotEmpty(serviceInstances)) {
+                            serviceInstancesChangedListener.onEvent(
+                                    new ServiceInstancesChangedEvent(serviceName, serviceInstances));
+                        }
                     }
                 }
-                serviceListeners.put(serviceNamesKey, serviceInstancesChangedListener);
             }
 
             if (!serviceInstancesChangedListener.isDestroyed()) {
                 listener.addServiceListener(serviceInstancesChangedListener);
                 serviceInstancesChangedListener.addListenerAndNotify(url, listener);
-                ServiceInstancesChangedListener finalServiceInstancesChangedListener = serviceInstancesChangedListener;
-
-                String serviceDiscoveryName =
-                        url.getParameter(RegistryConstants.REGISTRY_CLUSTER_KEY, url.getProtocol());
-
-                MetricsEventBus.post(
-                        RegistryEvent.toSsEvent(
-                                url.getApplicationModel(), serviceKey, Collections.singletonList(serviceDiscoveryName)),
-                        () -> {
-                            serviceDiscovery.addServiceInstancesChangedListener(finalServiceInstancesChangedListener);
-                            return null;
-                        });
             } else {
                 logger.info(String.format("Listener of %s has been destroyed by another thread.", serviceNamesKey));
                 serviceListeners.remove(serviceNamesKey);
--- a/dubbo-registry/dubbo-registry-api/src/test/java/org/apache/dubbo/registry/client/ServiceDiscoveryRegistryTest.java
+++ b/dubbo-registry/dubbo-registry-api/src/test/java/org/apache/dubbo/registry/client/ServiceDiscoveryRegistryTest.java
@@ -41,6 +41,7 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.ArgumentCaptor;
+import org.mockito.InOrder;
 
 import static org.apache.dubbo.common.constants.CommonConstants.CHECK_KEY;
 import static org.apache.dubbo.common.constants.CommonConstants.DUBBO;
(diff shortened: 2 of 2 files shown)


zrlw (approved):

hnpkso:
Thanks @zrlw  for the review!

@EarthChen  would appreciate a second look when you have a moment — production race fix for the initial-subscribe path against Nacos 2.x, deterministic tests included, one approval already in.

PR #16395, opened 2026-07-25
Fix condition route leaking tail invokers

### LINKED ISSUE #16130: [Bug] When configuring condition route to route traffic to particular instance, there is still traffic in others
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.6, JDK 1.8

When configuring condition route to route traffic to particular instance, there is still traffic in others

The route config should be executed exactly. When configuration is set, traffic should follow the configuration to route traffic to the specified instance.

### Anything else

After debugging, the root cause is the code below in `AbstractStateRouter`
```
routeResult = doRoute(invokers, url, invocation, needToPrintMessage, nodeHolder, messageHolder);
        if (routeResult != invokers) {
            routeResult = invokers.and(routeResult);
        }
```

`invokers.and` should only return the invokes that contains in both lists, but the implementation of and function is adding the tailList together. Therefore ,the final routeResults are redundant.  

### Code of Conduct

014-code:

Fixes #16130.

Condition routing may receive a BitList that contains additional tail invokers. After the router filters the cloned result, AbstractStateRouter intersects the original invokers with the routed result. BitList.and() previously preserved tailList entries even when they were not present in the target result, which could reintroduce filtered invokers and send traffic to non-target instances.

This change makes BitList.and() apply true intersection semantics to both origin entries and tail entries.

## Brief changelog

- Update BitList.and() to keep only elements present in both lists.
- Add BitList regression coverage for tail-list intersection.
- Add condition-route regression coverage to ensure filtered tail invokers do not leak into route results.

## Verifying this change

- .\mvnw.cmd -pl dubbo-cluster "-Dtest=BitListTest,ConditionStateRouterTest" test

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
@@ -114,16 +114,26 @@ public class BitList<E> extends AbstractList<E> implements Cloneable {
     }
 
     /**
-     * And operation between two bitList. Return a new cloned list.
-     * TailList in source bitList will be totally saved even if it is not appeared in the target bitList.
+     * And operation between two bitList.
      *
      * @param target target bitList
-     * @return this bitList only contains those elements contain in both two list and source bitList's tailList
+     * @return this bitList only contains elements contained in both lists
      */
     public synchronized BitList<E> and(BitList<E> target) {
-        rootSet.and(target.rootSet);
-        if (target.getTailList() != null) {
-            target.getTailList().forEach(this::addToTailList);
+        if (originList == target.originList) {
+            rootSet.and(target.rootSet);
+        } else {
+            BitSet resultSet = new BitSet();
+            for (int bitIndex = rootSet.nextSetBit(0); bitIndex >= 0; bitIndex = rootSet.nextSetBit(bitIndex + 1)) {
+                if (target.contains(originList.get(bitIndex))) {
+                    resultSet.set(bitIndex);
+                }
+            }
+            rootSet.clear();
+            rootSet.or(resultSet);
+        }
+        if (CollectionUtils.isNotEmpty(tailList)) {
+            tailList.removeIf(e -> !target.contains(e));
         }
         return this;
     }
--- a/dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionStateRouterTest.java
+++ b/dubbo-cluster/src/test/java/org/apache/dubbo/rpc/cluster/router/condition/ConditionStateRouterTest.java
@@ -197,6 +197,33 @@ class ConditionStateRouterTest {
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
```diff
@@ -114,16 +114,26 @@ public synchronized E getByIndex(int index) {
...
+            for (int bitIndex = rootSet.nextSetBit(0); bitIndex >= 0; bitIndex = rootSet.nextSetBit(bitIndex + 1)) {
+                if (target.contains(originList.get(bitIndex))) {
+                    resultSet.set(bitIndex);
+                }
+            }
+            rootSet.clear();
+            rootSet.or(resultSet);
+        }
+        if (CollectionUtils.isNotEmpty(tailList)) {
+            tailList.removeIf(e -> !target.contains(e));
```
zrlw, line 136: Conflict with the method comment?
```
TailList in source bitList will be totally saved even if it is not appeared in the target bitList.
```

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
```diff
@@ -114,16 +114,26 @@ public synchronized E getByIndex(int index) {
...
      *
      * @param target target bitList
-     * @return this bitList only contains those elements contain in both two list and source bitList's tailList
+     * @return this bitList only contains elements contained in both lists
      */
     public synchronized BitList<E> and(BitList<E> target) {
-        rootSet.and(target.rootSet);
-        if (target.getTailList() != null) {
-            target.getTailList().forEach(this::addToTailList);
+        if (originList == target.originList) {
```
zrlw, line 123: using ```equals``` if you want to compare two different list objects which contain same items, 
and if you don't care the item order in them , you should sort them before calling ```equals```.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
014-code: Thanks! The JavaDoc has been updated to match the new behavior.

dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/state/BitList.java
014-code: `==` is intentional here.

This branch is only used as a fast path when both `BitList`s share the same backing `originList` instance, where the bit indexes have the same meaning and `BitSet.and()` is safe.

If the two `BitList`s are backed by different `originList`s, even if they contain the same elements, we still fall back to the element-based intersection logic below.

zrlw (approved):
LGTM

EarthChen (approved):
LGTM

PR #16398, opened 2026-07-28
Fix #16397: return 503 (UNAVAILABLE) instead of 404 (UNIMPLEMENTED) w…

### LINKED ISSUE #16397: [Bug] 3.x Graceful Shutdown.  "UNIMPLEMENTED : Invoker for gRPC not found"(consumer)/ "Invoker for gRPC not found"(provider)
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.6, Triple protocol (`dubbo.protocol.name=tri`), instance mode , JDK 8. 

Graceful shutdown tears down the Triple `PathResolver` mapping *before* in-flight requests have drained.
A request that has already been accepted by the server, but has not yet resolved its invoker, will fail with
`UNIMPLEMENTED` once `afterUnExport` runs.

The race can be reproduced deterministically with two breakpoints on the **provider** side — no timing luck
required. Use a **thread-level** (suspend-thread, not suspend-all) breakpoint so the shutdown thread and the
request thread can be released independently.

Provider service (anything works, the body is irrelevant):

```java
@DubboService
public class DemoServiceImpl implements DemoService {
    @Override
        return "hello " + name;
    }
}

Consumer: a single call is enough.

@DubboReference(timeout = 500000)
private DemoService demoService;

demoService.sayHello("dubbo");

Reproduction procedure (provider JVM, debugger attached):

1. Set a thread-suspending breakpoint at
org.apache.dubbo.rpc.protocol.tri.h12.grpc.GrpcRequestHandlerMapping#getRequestHandler, on the line

Invoker<?> invoker = pathResolver.resolve(path.getPath(), group, version);   // GrpcRequestHandlerMapping.java:62

2. Issue the consumer call. The request thread is now suspended before the path is resolved — i.e. the
request has already been accepted by a fully healthy provider.
3. Send TERM to the provider process to start graceful shutdown, with a second thread-suspending breakpoint at

pathResolver.unregister(invoker);   // TripleProtocol.java:111, in TripleProtocol$1#afterUnExport

4. Step over that line so the unregister actually executes, then leave the shutdown thread suspended.
TriplePathResolver no longer contains the path mapping for the service.
5. Release the request thread from step 1 and step over line 62. resolve(...) now returns null, so control
falls into

if (invoker == null) {
    throw notFound();               // GrpcRequestHandlerMapping.java:64
}

Result

- Provider log:
Invoker for gRPC not found
- Consumer:
org.apache.dubbo.rpc.RpcException: ... UNIMPLEMENTED : Invoker for gRPC not found

Without the debugger, the same thing happens naturally on every shutdown that overlaps with concurrent
traffic; the breakpoints only make the window deterministic.

The request accepted in step 1 was accepted while the service was still exported, and the configured graceful
shutdown wait (dubbo.service.shutdown.wait) had not elapsed. It should have been served normally.

Concretely:

1. PathResolver.unregister(invoker) should not take effect until in-flight requests have drained — the path
mapping must outlive the request-draining phase rather than being removed at the start of it.
2. If the server has genuinely stopped accepting requests, the failure should be retriable: UNAVAILABLE,
or an HTTP/2 GOAWAY that lets the consumer re-route to another provider instance.

UNIMPLEMENTED is semantically wrong here — it means "this server does not implement the service", so
consumer-side cluster fault tolerance treats it as a permanent definition error and does not fail over. Every
rolling restart of a Triple provider therefore surfaces as business-visible errors on the consumer, which
defeats the purpose of graceful shutdown.

### Anything else

Code paths involved (3.3.6):

- org.apache.dubbo.rpc.protocol.tri.TripleProtocol$1#afterUnExport — pathResolver.unregister(invoker)
(TripleProtocol.java:111), registered by TripleProtocol#export
- org.apache.dubbo.rpc.protocol.tri.h12.grpc.GrpcRequestHandlerMapping#getRequestHandler
(GrpcRequestHandlerMapping.java:62 resolve, :64 throw notFound())
- org.apache.dubbo.rpc.PathResolver#unregister / #resolve,
org.apache.dubbo.rpc.protocol.tri.TriplePathResolver

The same window applies to any RequestHandlerMapping that resolves through PathResolver during shutdown,
not only the gRPC one.

### Code of Conduct

LI123456mo:
Fixes part of #16397.

## Problem
When `pathResolver.resolve()` fails to find an invoker because it was
unregistered during graceful shutdown, `GrpcRequestHandlerMapping` throws
a 404, which maps to gRPC `UNIMPLEMENTED`. Consumer-side cluster fault
tolerance treats `UNIMPLEMENTED` as a permanent, non-retriable error, so
it never fails over to another provider instance — turning a routine
rolling restart into a business-visible error.

## Fix
When resolution fails specifically because the application is stopping
(`ApplicationDeployer#isStopping()`), return 503 instead of 404, which
maps to gRPC `UNAVAILABLE` — a retriable status consumers already know
how to fail over on.

## Testing
Added `GrpcRequestHandlerMappingTest#returnsServiceUnavailableWhenStoppingAndInvokerNotFound`,
verifying a 503 is thrown when the app is stopping and resolution fails.

### DIFF
--- a/dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/HttpStatus.java
+++ b/dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/HttpStatus.java
@@ -32,7 +32,8 @@ public enum HttpStatus {
     REQUEST_TIMEOUT(408),
     CONFLICT(409),
     UNSUPPORTED_MEDIA_TYPE(415),
-    INTERNAL_SERVER_ERROR(500);
+    INTERNAL_SERVER_ERROR(500),
+    SERVICE_UNAVAILABLE(503);
 
     private final int code;
 
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcRequestHandlerMapping.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcRequestHandlerMapping.java
@@ -61,6 +61,9 @@ public final class GrpcRequestHandlerMapping implements RequestHandlerMapping {
         String version = request.header(TripleHeaderEnum.SERVICE_VERSION.getKey());
         Invoker<?> invoker = pathResolver.resolve(path.getPath(), group, version);
         if (invoker == null) {
+            if(url.getOrDefaultApplicationModel().getDeployer().isStopping()){
+                throw unavailable();
+            }
             throw notFound();
         }
 
@@ -77,6 +80,11 @@ public final class GrpcRequestHandlerMapping implements RequestHandlerMapping {
 
     private static HttpStatusException notFound() {
         return new HttpStatusException(HttpStatus.NOT_FOUND.getCode(), "Invoker for gRPC not found");
+    }
+
+    private static HttpStatusException unavailable() {
+        return new HttpStatusException(
+                HttpStatus.SERVICE_UNAVAILABLE.getCode(), "Invoker for gRPC not found, service is shutting down");
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


LI123456mo:
@zrlw , @RainYuY

PR #16391, opened 2026-07-21
[ISSUE #16389] Close late input after streaming decoder shutdown

gaoxiaolei-s59:
## What is the purpose of the change

Fixes #16389. Triple HTTP/2 wraps DATA frames in an auto-closing input stream. When a DATA frame arrives after the streaming decoder starts closing or has closed, the decoder previously returned without closing that stream, leaving the underlying ByteBuf unreleased.

## Brief changelog

- Close input streams rejected while the decoder is closing or closed.
- Propagate input close failures as DecodeException.
- Add regression coverage for both closing and closed decoder states.

## Verifying this change

- LengthFieldStreamingDecoderTest: 2 tests passed.
- All dubbo-remoting-http12 tests: 23 tests passed, 2 skipped.
- Spotless and reactor compilation checks passed.

The broader dependency-reactor test run reached 1,165 dubbo-common tests and was stopped by the pre-existing environment-sensitive NetUtilsTest.testMatchIpRangeMatchWhenIpWrongException failure; the target module had not started in that run.

PR #16400, opened 2026-07-29
 [ISSUE #16399] Close late input after streaming decoder shutdown

### LINKED ISSUE #16399: DefaultStreamingDecoder ByteBuf leak when decode() is called after close
## Bug Description

`DefaultStreamingDecoder.decode()` does not close the input stream when the decoder has already been closed (`closed == true`). In the Triple HTTP/2 path, DATA frames are wrapped in `ByteBufInputStream(content, true)` (i.e., `releaseOnClose = true`), meaning the underlying pooled `ByteBuf` is released only when `InputStream.close()` is called. If a late DATA frame arrives asynchronously (via `executor.execute()` in `AbstractServerTransportListener.onData()`) after the decoder has started closing, the `ByteBuf` is never released, causing a leak.

## Symmetry with #16389 / #16391

This is the exact same bug pattern as #16389, which was fixed by #16391 for `LengthFieldStreamingDecoder`. The fix in #16391 added `inputStream.close()` in the `decode()` method when the decoder is closing or closed. `DefaultStreamingDecoder` has the same pattern but was not included in that fix:

- **LengthFieldStreamingDecoder** (gRPC path): fixed in #16391 ✅
- **DefaultStreamingDecoder** (non-gRPC path): still leaks ❌

## Steps to Reproduce

1. Create a `DefaultStreamingDecoder` instance
2. Call `onStreamClosed()` to mark the decoder as closed
3. Call `decode(inputStream)` with any input stream
4. The input stream is never closed, and any wrapped `ByteBuf` is leaked

## Impact

ByteBuf memory leak in the non-gRPC Triple HTTP/2 streaming path when late DATA frames arrive after the decoder has been closed. Under sustained traffic with connection churn, leaked `ByteBuf`s accumulate and can eventually trigger Netty's `ResourceLeakDetector` warnings or cause `OutOfMemoryError`.

## Suggested Fix

Mirror the #16391 fix: close the input stream in `decode()` when `closed == true`, propagating close failures as `DecodeException`.

dong0713:

Fixes #16399. Mirror the #16391 fix for **both** streaming decoder implementations: when `decode()` is called after the decoder has been shut down, the input stream is now closed to release the underlying pooled `ByteBuf` wrapped with `releaseOnClose = true`.

#16391 introduced an early-return guard in both decoders but did not close the input on that guard, so the `ByteBuf` leaked. This PR closes that gap in both implementations.

### 1. `DefaultStreamingDecoder` (dubbo-remoting-http12)
- Close input streams rejected after the decoder has been closed (`closed == true` early-return branch).
- Propagate input close failures as `DecodeException` (consistent with #16391).

### 2. `FinalFragmentStreamingDecoder` (dubbo-remoting-websocket)
- Close input streams rejected after the decoder has been closed or is closing (`closing || closed` early-return branch) — the symmetric counterpart of the `DefaultStreamingDecoder` fix, in the websocket module.

## Verifying this change

- `DefaultStreamingDecoderTest` (http12): closed-path + exception propagation.
- `FinalFragmentStreamingDecoderTest` (websocket, new test dir): closed-path + closing-path + exception propagation.
- Spotless and module compilation checks passed.

### DIFF
--- a/dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/message/DefaultStreamingDecoder.java
+++ b/dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/message/DefaultStreamingDecoder.java
@@ -38,7 +38,13 @@ public class DefaultStreamingDecoder implements StreamingDecoder {
     @Override
     public void decode(InputStream inputStream) throws DecodeException {
         if (closed) {
-            // ignored
+            // Close late input rejected after the decoder has been closed, so the
+            // underlying ByteBuf (wrapped with releaseOnClose) is released.
+            try {
+                inputStream.close();
+            } catch (IOException e) {
+                throw new DecodeException(e);
+            }
             return;
         }
         accumulate.addInputStream(inputStream);
--- a/dubbo-remoting/dubbo-remoting-http12/src/test/java/org/apache/dubbo/remoting/http12/message/DefaultStreamingDecoderTest.java
+++ b/dubbo-remoting/dubbo-remoting-http12/src/test/java/org/apache/dubbo/remoting/http12/message/DefaultStreamingDecoderTest.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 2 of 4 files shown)


LI123456mo (approved):
LGTM @zrlw

zrlw:
What about ```FinalFragmentStreamingDecoder```?

LI123456mo:
> What about `FinalFragmentStreamingDecoder`?

Ooops , didn't see that , @dong0713 , here is the path "org.apache.dubbo.remoting.websocket".

dong0713:
`FinalFragmentStreamingDecoder` shares the same late-input path (`closing || closed` returning without closing the input). Fixed symmetrically and added regression tests covering both the `closing` and `closed` paths. Pushed.

PR #16404, opened 2026-07-31
fix: reject oversized telnet payload before header copy

RainYuY:
When the server-side `ExchangeCodec` receives non-Dubbo input, it currently allocates and copies the entire readable buffer before the Telnet payload limit is checked. An oversized input can therefore cause an unnecessary large byte-array allocation and copy.

This change validates the readable payload size in `ExchangeCodec.decode()` before allocating the header buffer for non-magic server-side input.

### What has changed?

- Added an early `checkPayload(channel, readable)` for non-magic server-side input.
- The validation is performed before `new byte[]` and `buffer.readBytes(...)`.
- Inputs with the Dubbo magic header continue through the existing Dubbo header/body decoding path.
- Client-side decoding and Telnet command parsing behavior remain unchanged.
- No cross-packet scan state or decoder structure was introduced.

### Tests

Added a regression test to verify that:

- An oversized non-magic input throws `ExceedPayloadLimitException`.
- The input buffer is not consumed when validation fails.

Executed:

```bash
mvn -pl dubbo-remoting/dubbo-remoting-api \
  -DskipITs \
  -Dcheckstyle.skip \
  -Drat.skip=true \
  -Dtest=ExchangeCodecTest \
  test

### DIFF
--- a/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
+++ b/dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java
@@ -87,6 +87,13 @@ public class ExchangeCodec extends TelnetCodec {
     @Override
     public Object decode(Channel channel, ChannelBuffer buffer) throws IOException {
         int readable = buffer.readableBytes();
+        if (readable > 0
+                && isServerSide(channel)
+                && (readable < 2
+                        || buffer.getByte(buffer.readerIndex()) != MAGIC_HIGH
+                        || buffer.getByte(buffer.readerIndex() + 1) != MAGIC_LOW)) {
+            checkPayload(channel, readable);
+        }
         byte[] header = new byte[Math.min(readable, HEADER_LENGTH)];
         buffer.readBytes(header);
         return decode(channel, buffer, readable, header);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: dubbo-remoting/dubbo-remoting-api/src/main/java/org/apache/dubbo/remoting/exchange/codec/ExchangeCodec.java [hunk 1]
```diff
@@ -87,6 +87,13 @@ public void encode(Channel channel, ChannelBuffer buffer, Object msg) throws IOE
     @Override
     public Object decode(Channel channel, ChannelBuffer buffer) throws IOException {
         int readable = buffer.readableBytes();
+        if (readable > 0
+                && !isClientSide(channel)
```
zrlw: why not use ```isServerSide``` directly?
```
        if (isServerSide(channel)
                && (readable == 1
                        || buffer.getByte(buffer.readerIndex()) != MAGIC_HIGH
                        || buffer.getByte(buffer.readerIndex() + 1) != MAGIC_LOW)) {
            checkPayload(channel, readable);
        }
```

RainYuY: I missed that. My mistake.

zrlw:
```
        if (isServerSide(channel)
                && readable > 1
                && (buffer.getByte(buffer.readerIndex()) != MAGIC_HIGH
                        || buffer.getByte(buffer.readerIndex() + 1) != MAGIC_LOW)) {
            checkPayload(channel, readable);
        }
```
It may be sufficient to only handle cases greater than 1

RainYuY:
> ```
>         if (isServerSide(channel)
>                 && readable > 1
>                 && (buffer.getByte(buffer.readerIndex()) != MAGIC_HIGH
>                         || buffer.getByte(buffer.readerIndex() + 1) != MAGIC_LOW)) {
>             checkPayload(channel, readable);
>         }
> ```
> 
> It may be sufficient to only handle cases greater than 1

 I don't think this change provides any meaningful behavioral benefit. When readable == 1, checkPayload is effectively a no-op because the smallest positive payload limit is 1, and the check only rejects size > payload. The current condition is correct and keeps the zero-readable case explicit, so I'd prefer to leave it as is.

PR #16407, opened 2026-07-31
[ISSUE #16381] Retain registry configs with different explicit ids

### LINKED ISSUE #16381: [Bug] removing the so-called registry config duplication leads to exception
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3.6，openjdk 21

if the dubbo consumer xml config is as follow

```
     <dubbo:registry` id="zk1" protocol="zookeeper"
                        address="${ZK1_ADDRESS:10.47.181.23:2181,10.47.181.24:2181,10.47.181.25:2181}"/>

        <dubbo:registry id="zk2" protocol="zookeeper"
                        address="${ZK2_ADDRESS:10.47.181.20:2181,10.47.181.21:2181,10.47.181.22:2181}"/>

        <dubbo:reference id="app1" interface="com.test.app1" registry="zk1"/>

        <dubbo:reference id="app2" interface="com.test.app2" registry="zk2"/>
```

in which ZK1_ADDRESS and ZK2_ADDRESS are env variables.

what is exactly needed is that ZK1_ADDRESS and ZK2_ADDRESS may be the same or different.

when ZK1_ADDRESS and ZK2_ADDRESS are the same, current implementation will only register zk1 registy config successfully, the zk2 registry config is identified as duplication，then it throws exception `java.lang.IllegalStateException: Registry not found: zk2` when loading app2 reference config

zk1 and zk2 are all registered successfully.

### Anything else

_No response_

### Code of Conduct

cyforkk:
Fixes #16381

When two `RegistryConfig` instances have different explicit ids (e.g. `zk1`, `zk2`) but the same address, the previous dedup logic incorrectly treated the second one as a duplicate of the first and dropped it. This caused `IllegalStateException: Registry not found: zk2` when later resolving the reference's registry id.

The root cause is that `AbstractConfig.equals()` intentionally ignores the `id` field, so the dedup check in `addConfig()` and `findConfigByValue()` could not distinguish configs that the user explicitly identified as different.

## Brief changelog

- In `AbstractConfigManager.addConfig()` fast-path check: skip the early return when the existing config has a different explicit id than the new config.
- In `AbstractConfigManager.findConfigByValue()`: exclude matches where both configs have non-null, different explicit ids.
- Add helper `hasDifferentExplicitId()` to encapsulate the comparison.
- Add regression tests in `ConfigManagerTest`:
  - `testRegistryConfigWithDifferentIdsSameAddress`: two registries with different ids and same address are both retained.
  - `testRegistryConfigWithSameIdSameAddress`: same-id registries are still deduplicated (override).

## Verifying this change

- `ConfigManagerTest`: 21 tests passed (including 2 new regression tests).

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/context/AbstractConfigManager.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/context/AbstractConfigManager.java
@@ -170,7 +170,7 @@ public abstract class AbstractConfigManager extends LifecycleAdapter {
         // fast check duplicated equivalent config before write lock
         if (!(config instanceof ReferenceConfigBase || config instanceof ServiceConfigBase)) {
             for (AbstractConfig value : configsMap.values()) {
-                if (value.equals(config)) {
+                if (value.equals(config) && !hasDifferentExplicitId(value, config)) {
                     return (T) value;
                 }
             }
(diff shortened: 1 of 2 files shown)


LI123456mo (commented):
Hi @cyforkk , this PR https://github.com/apache/dubbo/pull/16382 already fixes the issue

cyforkk:
ok，bro

PR #16411, opened 2026-08-01
fix(triple): release REST form request decoder

014-code:

Closes #16403.

Release the cached `HttpPostRequestDecoder` when a Triple REST request reaches a terminal state. This removes the synthetic request and parsed form data from Netty's static `DefaultHttpDataFactory`, preventing retained entries from accumulating after form requests.

The cleanup is invoked after successful invocation completion and on error, cancellation, and HTTP/2 transport close paths. It is intentionally not performed immediately after argument decoding so multipart data remains available to application code during request handling.

### GitHub_issue

Closes #16403

### Tests

Added `completed form request releases post data` to `RestProtocolTest`.

The test submits an `application/x-www-form-urlencoded` request, verifies the normal response, then asserts that `DefaultHttpDataFactory.requestFileDeleteMap` has no retained request entries.

Passed:

- `RestProtocolTest`: 104 tests, 0 failures
- JaCoCo `verify`: passed
- `DefaultHttpRequest.close()`: 100% line and branch coverage
- `AbstractServerCallListener.closeRequest()`: 100% line coverage

## Checklist

- [x] Make sure gitHub actions can pass.

### DIFF
--- a/dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/AbstractServerHttpChannelObserver.java
+++ b/dubbo-remoting/dubbo-remoting-http12/src/main/java/org/apache/dubbo/remoting/http12/AbstractServerHttpChannelObserver.java
@@ -26,6 +26,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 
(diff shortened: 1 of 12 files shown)


LI123456mo (commented):
Hi @zrlw , have tried reviewing this PR  which fixes the issue . But , there is also another PR https://github.com/apache/dubbo/pull/16412 which also targets the same issue , can you help review that one so that we compare the two

014-code:
> Hi @zrlw , have tried reviewing this PR which fixes the issue . But , there is also another PR #16412 which also targets the same issue , can you help review that one so that we compare the two

Thanks for pointing out #16412. I compared both implementations and updated
this PR to combine their strengths.

This PR keeps the generic, idempotent HttpRequest.close() abstraction and the
single-commit history based on the current 3.3 branch. It now adopts the safer
response-termination lifecycle from #16412: remote cancellation no longer
destroys multipart data immediately, and the decoder is released only after
application or response processing reaches a terminal state.

The updated tests cover normal form completion, HTTP/1 exceptional completion,
HTTP/2 remote cancellation with multipart data remaining readable, cleanup
after asynchronous success and failure, idempotent observer termination, and
custom request implementations.

This avoids premature multipart cleanup while ensuring that Netty's
DefaultHttpDataFactory entries are released on every terminal path.

PR #16416, opened 2026-08-07
Fix Triple gRPC decoder handoff

### LINKED ISSUE #16414: [Triple][3.3.6] First client-streaming call after consumer restart may fail with malformed gRPC frame and IllegalReferenceCountException
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Apache Dubbo：3.3.6
JDK：17

1. Start a Triple provider in one JVM with a client-streaming RPC:
   `rpc Upload(stream UploadFrame) returns (UploadResult);`

2. Start a new consumer JVM and immediately invoke `Upload`:
   send the metadata frame first, then send one or more file-content frames.

3. The first call after the consumer restart may fail. Subsequent calls usually succeed without restarting either side.

4. The issue is reproducible with both:
   - a single Dubbo port with `ext-protocol: tri`
   - native `tri` protocol

5. The payload can be very small (for example, 16 KB); it is not dependent on a large gRPC frame.

## Actual result

The consumer receives:

```text
org.apache.dubbo.rpc.StatusRpcException: UNKNOWN : gRPC frame header malformed:
reserved bits not zero
    at org.apache.dubbo.rpc.TriRpcStatus.asException(TriRpcStatus.java:260)
    at org.apache.dubbo.rpc.protocol.tri.call.ObserverToClientCallListenerAdapter.onClose(
        ObserverToClientCallListenerAdapter.java:52)
    at org.apache.dubbo.rpc.protocol.tri.call.TripleClientCall.onComplete(
        TripleClientCall.java:126)
    at org.apache.dubbo.rpc.protocol.tri.stream.AbstractTripleClientStream$
        ClientTransportListener.finishProcess(AbstractTripleClientStream.java:224)


The provider 

<!-- Failed to upload "dubbo-triple-first-call-repro-standalone.zip" -->

receives:

[dubbo-triple-first-call-repro-standalone.zip](https://github.com/user-attachments/files/30777940/dubbo-triple-first-call-repro-standalone.zip)

io.netty.util.IllegalReferenceCountException: refCnt: 0
    at io.netty.buffer.AbstractByteBuf.ensureAccessible(AbstractByteBuf.java:1454)
    at io.netty.buffer.ByteBufInputStream.read(ByteBufInputStream.java:183)
    at org.apache.dubbo.remoting.http12.CompositeInputStream.read(
        CompositeInputStream.java:82)
    at org.apache.dubbo.remoting.http12.message.LengthFieldStreamingDecoder
        .processHeader(LengthFieldStreamingDecoder.java:147)
    at org.apache.dubbo.remoting.http12.message.LengthFieldStreamingDecoder
        .deliver(LengthFieldStreamingDecoder.java:117)
    at org.apache.dubbo.remoting.http12.message.LengthFieldStreamingDecoder
        .decode(LengthFieldStreamingDecoder.java:69)
    at org.apache.dubbo.rpc.protocol.tri.h12.http2.
        GenericHttp2ServerTransportListener$StreamingHttpMessageListener.onMessage(
        GenericHttp2ServerTransportListener.java:220)

The first client-streaming Triple call after a consumer restart should be handled
the same as subsequent calls.

The provider should retain the received buffers until
`LengthFieldStreamingDecoder` finishes decoding them, so no malformed gRPC frame
is reported and no `IllegalReferenceCountException: refCnt: 0` occurs.

If a transport failure does occur, it should be propagated through the normal
stream error callback so that application resources can be released safely.

### Anything else

- Apache Dubbo: 3.3.6
- JDK: 17
- Transport: Triple / HTTP2
- Provider and consumer run in separate JVM processes.
- The failure also occurs with a 16 KB payload, so it does not appear to be
  caused by exceeding the default gRPC frame-size limit.
- We reproduced it with both native `tri` and single-port `dubbo` plus
  `ext-protocol: tri`; therefore it does not appear to be specific to the
  single-port configuration.
- The minimal two-process demo is nondeterministic, but the production symptom
  consistently occurs on the first call after the consumer restarts. We can
  provide the demo if needed.

### Code of Conduct

### PR DESCRIPTION
**Pull request by Anon2Tokyo**

Fixes #16414.

This change fixes the gRPC no-stub method discovery path for Triple streaming requests. The lazy method discovery listener now reuses the stream's existing `StreamingDecoder` instead of creating a temporary `GrpcStreamingDecoder`, so buffered bytes from a subsequent message are preserved when the listener switches to the resolved business handler.

The close callback on the temporary method-discovery listener is also made a no-op because it only exists to resolve the method descriptor and should not close the actual request stream.

A regression test covers a DATA-frame layout where the first decode contains one complete gRPC message followed by the prefix of the next message, then verifies the second message is still decoded after the listener switch.

## Checklist

Test:

```powershell
mvn.cmd -pl dubbo-rpc/dubbo-rpc-triple -Dtest=GrpcStreamingDecoderTest test
```

### DIFF
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcHttp2ServerTransportListener.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcHttp2ServerTransportListener.java
@@ -172,7 +172,6 @@ public class GrpcHttp2ServerTransportListener extends GenericHttp2ServerTranspor
         private LazyFindMethodListener() {
             streamingDecoder = getStreamingDecoder();
             streamingDecoder.setFragmentListener(new DetermineMethodDescriptorListener());
-            streamingDecoder.request(Integer.MAX_VALUE);
         }
 
         @Override
--- a/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcHttp2ServerTransportListener.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcHttp2ServerTransportListener.java
@@ -170,7 +170,7 @@ public class GrpcHttp2ServerTransportListener extends GenericHttp2ServerTranspor
         private final StreamingDecoder streamingDecoder;
 
         private LazyFindMethodListener() {
-            streamingDecoder = new GrpcStreamingDecoder();
+            streamingDecoder = getStreamingDecoder();
             streamingDecoder.setFragmentListener(new DetermineMethodDescriptorListener());
             streamingDecoder.request(Integer.MAX_VALUE);
         }
@@ -195,7 +195,7 @@ public class GrpcHttp2ServerTransportListener extends GenericHttp2ServerTranspor
 
         @Override
         public void onClose() {
-            getStreamingDecoder().close();
+            // no op
         }
 
         @Override
--- a/dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcStreamingDecoderTest.java
+++ b/dubbo-rpc/dubbo-rpc-triple/src/test/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcStreamingDecoderTest.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.dubbo.rpc.protocol.tri.h12.grpc;
+
+import org.apache.dubbo.remoting.http12.message.StreamingDecoder;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class GrpcStreamingDecoderTest {
+
+    @Test
+    void continueDecodingWithBufferedPartialFrameAfterListenerSwitch() {
+        GrpcStreamingDecoder decoder = new GrpcStreamingDecoder();
+        List<String> messages = new ArrayList<>();
+
+        StreamingDecoder.FragmentListener normalListener = new StreamingDecoder.FragmentListener() {
+            @Override
+            public void bytesRead(int numBytes) {}
+
+            @Override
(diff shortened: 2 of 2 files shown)


Anon2Tokyo:
@zrlw Hello, if you have some time, could you please review this PR for me? Thank you for your time.

zrlw (dismissed):
LGTM

Anon2Tokyo:
@EarthChen Hello, if you have some time, could you please review this PR for me? Thank you for your time.

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcHttp2ServerTransportListener.java
```diff
@@ -170,7 +170,7 @@ private class LazyFindMethodListener implements HttpMessageListener {
         private final StreamingDecoder streamingDecoder;
 
         private LazyFindMethodListener() {
-            streamingDecoder = new GrpcStreamingDecoder();
+            streamingDecoder = getStreamingDecoder();
             streamingDecoder.setFragmentListener(new DetermineMethodDescriptorListener());
             streamingDecoder.request(Integer.MAX_VALUE);
```
RainYuY: Could this change cause the existing backpressure mechanism to stop working?

RainYuY (changes requested):

dubbo-rpc/dubbo-rpc-triple/src/main/java/org/apache/dubbo/rpc/protocol/tri/h12/grpc/GrpcHttp2ServerTransportListener.java
Anon2Tokyo: @RainYuY I took a look and it's indeed possible. The `request(max)` part in the diff above was from the previous temporary decoder logic, meaning giving the decoder an extremely large message quota. This indeed had no impact on the temporary decoder, but the change I made here actually set up the main decoder, and the subsequent automatic backpressure mechanism works in an accumulative way, so it will affect the backpressure mechanism. Currently, my understanding is to remove the `request max`.Based on your review, I hope I've made myself clear. I'll submit a commit to fix this backpressure issue when I get off work tonight

RainYuY: > @RainYuY I took a look and it's indeed possible. The `request(max)` part in the diff above was from the previous temporary decoder logic, meaning giving the decoder an extremely large message quota. This indeed had no impact on the temporary decoder, but the change I made here actually set up the main decoder, and the subsequent automatic backpressure mechanism works in an accumulative way, so it will affect the backpressure mechanism. Currently, my understanding is to remove the `request max`.Based on your review, I hope I've made myself clear. I'll submit a commit to fix this backpressure issue when I get off work tonight

I think so, but I don’t have time right now to check whether removing this will have any impact. You can go ahead and do it. I’ll check it later, and if there are no problems, I’ll approve it.

LI123456mo (commented):
@RainYuY , @Anon2Tokyo 

Traced this through fully to make sure backpressure/cleanup isn't lost — it isn't, it just moved.

Before this change, streamingDecoder was created fresh inside LazyFindMethodListener() per-call, so closing it there made sense — it was privately owned.

After this change, streamingDecoder is a single instance created once in GenericHttp2ServerTransportListener's constructor and shared with the response side via prepareResponseObserver() → responseObserver.setStreamingDecoder(streamingDecoder). Since it's now shared, this method shouldn't unilaterally close it.

Here's the full chain confirming it's still torn down correctly, every time:

Netty channel physically closes (channel.closeFuture())
  → NettyHttp2ProtocolSelectorHandler: http2TransportListener.close()
    → GenericHttp2ServerTransportListener.close()
      → responseObserver.close()   [Http2ServerChannelObserver]
        → streamingDecoder.onStreamClosed()
          → accumulate.close()     [CompositeInputStream]
            → each buffered InputStream.close()
              → Netty ByteBuf.release()

So removing getStreamingDecoder().close() from onClose() here isn't dropping cleanup — it was actually a redundant/unsafe extra close on a resource this method no longer exclusively owns (risking a double-close on a shared decoder). The real teardown path is via onStreamClosed(), triggered reliably off Netty's own channel-close event.

Anon2Tokyo:
@RainYuY Hello, I have already deleted it. My initial review shows no impact. Could you please check it? Thank you for your time

RainYuY (approved):

RainYuY:
@zrlw @EarthChen @oxsean

zrlw (approved):

PR #16434, opened 2026-08-24
chore: upgrade actions/setup-java to v6

brunoborges:

Upgrade every active GitHub workflow reference from setup-java v1/v4 to v6.

GitHub_issue: N/A (routine CI dependency maintenance).

## Checklist
- [x] Write necessary unit-test to verify your logic correction. Not applicable; this only updates the CI action version. `git diff --check` passes and no setup-java references older than v6 remain.
- [ ] Make sure gitHub actions can pass. [Why the workflow is failing and how to fix it?](../CONTRIBUTING.md)

### DIFF
--- a/.github/workflows/build-and-test-pr.yml
+++ b/.github/workflows/build-and-test-pr.yml
@@ -52,7 +52,7 @@ jobs:
       - name: "Checkout"
         uses: actions/checkout@v4
       - name: "Setup maven"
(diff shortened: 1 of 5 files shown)


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #16435, opened 2026-08-25
fix: JavaBeanSerializeUtil handles null key in Map serialization

### LINKED ISSUE #16432: [Bug] JavaBeanSerializeUtil.serialize() throws IllegalArgumentException on a Map containing a null key
### Pre-check

### Search before asking

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

dubbo 3.2.20 (tag dubbo-3.2.20)

JavaBeanSerializeUtil.serialize() fails on a java.util.HashMap that contains a null key. HashMap permits exactly one null key, so this is a legal argument, and the serializer's own code has an explicit branch for it — but the value that branch produces is then rejected by a null check further down the same call chain.

import org.apache.dubbo.common.beanutil.JavaBeanSerializeUtil;
import java.util.HashMap;
import java.util.Map;

public class NullKeyRepro {
        Map<Object, Object> ok = new HashMap<>();
        ok.put("k", "v");
        System.out.println("without null key: " + (JavaBeanSerializeUtil.serialize(ok) != null));

        Map<Object, Object> withNullKey = new HashMap<>();
        withNullKey.put(null, "value1");          // HashMap permits one null key
        JavaBeanSerializeUtil.serialize(withNullKey);   // throws
    }
}

The map is serialised, with the null key represented as a null descriptor — which is what the code appears to intend.

### Actual
```
without null key: true
Exception in thread "main" java.lang.IllegalArgumentException: Property name is null
```

### Root cause
`dubbo-common/src/main/java/org/apache/dubbo/common/beanutil/JavaBeanSerializeUtil.java`, lines 162–166:

```java
map.forEach((key, value) -> {
    Object keyDescriptor   = key   == null ? null : createDescriptorIfAbsent(key,   accessor, cache);
    Object valueDescriptor = value == null ? null : createDescriptorIfAbsent(value, accessor, cache);
    descriptor.setProperty(keyDescriptor, valueDescriptor);
});
```

Line 163 deliberately produces `null` for a null key. Line 165 passes it to
`JavaBeanDescriptor.setProperty`, whose first statement (`JavaBeanDescriptor.java`,
lines 118–120) is:

```java
public Object setProperty(Object propertyName, Object propertyValue) {
    notNull(propertyName, "Property name is null");
    return properties.put(propertyName, propertyValue);
}
```

So the null-producing branch and the null-rejecting guard sit in the same call
chain. Either the ternary on line 163 is dead for keys, or the guard is too strict
for this call site.

### Notes
Not introduced by 3.2.20 — the code path is long-standing. The nearest existing
issue I found, #12248, is a different trigger (a null *parameter* causing an NPE
in generic-call bean mode).

### Code of Conduct

waterWang:

Fixes #16432: `JavaBeanSerializeUtil.serialize()` throws `IllegalArgumentException("Property name is null")` when serializing a `Map` that contains a `null` key.

`HashMap` permits exactly one `null` key, so this is a legal input. The Map branch in `serializeInternal` explicitly handles null keys (`key == null ? null : createDescriptorIfAbsent(...)`), but the descriptor produced is then rejected by the `notNull` check in `JavaBeanDescriptor.setProperty()`.

## Root cause

In `JavaBeanSerializeUtil.serializeInternal()` Map branch:

```java
map.forEach((key, value) -> {
    Object keyDescriptor = key == null ? null : createDescriptorIfAbsent(key, accessor, cache);
    Object valueDescriptor = value == null ? null : createDescriptorIfAbsent(value, accessor, cache);
    descriptor.setProperty(keyDescriptor, valueDescriptor);
});
```

When the map has a null key, `keyDescriptor` is `null`, and `JavaBeanDescriptor.setProperty(null, value)` throws because of the `notNull(propertyName, "Property name is null")` guard.

## Fix

Relax the `notNull` guard in `JavaBeanDescriptor.setProperty()` for map-type descriptors only — a `null` key is a legal Map key and `LinkedHashMap` (the backing storage) supports it. Bean/array/collection/class/enum descriptors keep the original null check.

## Verification

- Added `testSerialize_MapWithNullKey` (round-trip test: map with a null key serializes, deserializes, and preserves both key-value pairs)
- Existing `JavaBeanSerializeUtilTest` suite unaffected

## Checklist

- [x] I have read and agree to the Contributor Guidelines
- [x] I have added tests that prove my fix is effective
- [x] The PR targets the `3.3` branch

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/common/beanutil/JavaBeanDescriptor.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/common/beanutil/JavaBeanDescriptor.java
@@ -116,7 +116,9 @@ public final class JavaBeanDescriptor implements Serializable, Iterable<Map.Entr
     }
 
     public Object setProperty(Object propertyName, Object propertyValue) {
-        notNull(propertyName, "Property name is null");
+        if (!isMapType()) {
+            notNull(propertyName, "Property name is null");
+        }
         return properties.put(propertyName, propertyValue);
     }
 
--- a/dubbo-common/src/test/java/org/apache/dubbo/common/beanutil/JavaBeanSerializeUtilTest.java
+++ b/dubbo-common/src/test/java/org/apache/dubbo/common/beanutil/JavaBeanSerializeUtilTest.java
@@ -225,7 +225,26 @@ class JavaBeanSerializeUtilTest {
     }
 
     @Test
-    void testConstructorArg() {
+    public void testSerialize_MapWithNullKey() {
+        Map<String, String> map = new HashMap<>();
+        map.put("key1", "value1");
+        map.put(null, "nullKeyValue");
+
+        JavaBeanDescriptor descriptor = JavaBeanSerializeUtil.serialize(map);
+        Assertions.assertTrue(descriptor.isMapType());
+        Assertions.assertEquals(HashMap.class.getName(), descriptor.getClassName());
+        Assertions.assertEquals(2, descriptor.propertySize());
+
+        // round-trip
+        Object result = JavaBeanSerializeUtil.deserialize(descriptor);
+        Assertions.assertTrue(result instanceof Map);
+        Map<String, String> resultMap = (Map<String, String>) result;
+        Assertions.assertEquals("value1", resultMap.get("key1"));
+        Assertions.assertEquals("nullKeyValue", resultMap.get(null));
+    }
+
+    @Test
+    public void testConstructorArg() {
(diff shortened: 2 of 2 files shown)


LI123456mo (commented):
Hi @waterWang , Nice fix . I've gone through the both issue and PR ;  confirmed the deserialize round-trip path doesn't hit this (it iterates properties directly via iterator(), bypassing getProperty/containsProperty). But since JavaBeanDescriptor is a public class, wouldn't calling getProperty(null) or containsProperty(null) directly on a map-type descriptor still throw, given they don't have the same isMapType() exemption? Worth applying the same exemption there for API consistency, even if nothing internal currently exercises that path.

PR #16446, opened 2026-09-04
Validate condition router YAML before parsing

minnnjuuu:
`ConditionRuleParser.parse()` assumes that SnakeYAML always returns a YAML mapping. However, an empty YAML document returns `null`, while a sequence or scalar document returns a non-map value. This can result in a `NullPointerException` or `ClassCastException` instead of a clear validation error.

Some configuration-center adapters filter a Java `null` or an exactly empty string, but that does not cover every input that produces a null SnakeYAML result. Non-empty inputs such as a comment-only document, `---`, or the explicit YAML value `null` can pass string-level checks and still load as `null`.

This change validates the top-level YAML value before parsing it and rejects non-mapping documents with a clear `IllegalArgumentException` instead of relying on incidental `NullPointerException` or `ClassCastException` failures.

**What does this change do?**

- Parse the YAML document into an `Object` before casting it.
- Require the top-level YAML value to be a mapping.
- Report a clear `IllegalArgumentException` for empty, comment-only, sequence, and scalar documents.
- Add regression coverage for null and non-mapping YAML documents.

## Checklist

### DIFF
--- a/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/config/model/ConditionRuleParser.java
+++ b/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/router/condition/config/model/ConditionRuleParser.java
@@ -49,10 +49,15 @@ public class ConditionRuleParser {
 
     private static final ErrorTypeAwareLogger logger = LoggerFactory.getErrorTypeAwareLogger(ConditionRuleParser.class);
 
+    @SuppressWarnings("unchecked")
     public static AbstractRouterRule parse(String rawRule) {
         AbstractRouterRule rule;
         Yaml yaml = new Yaml(new SafeConstructor(new LoaderOptions()));
-        Map<String, Object> map = yaml.load(rawRule);
+        Object parsedRule = yaml.load(rawRule);
+        if (!(parsedRule instanceof Map)) {
+            throw new IllegalArgumentException("Condition router rule must be a YAML mapping.");
+        }
+        Map<String, Object> map = (Map<String, Object>) parsedRule;
         String confVersion = (String) map.get(CONFIG_VERSION_KEY);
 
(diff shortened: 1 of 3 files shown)


LI123456mo (commented):
Verified the fix — yaml.load() can return null/non-Map for bad input, old code cast blindly (NPE/CCE), new instanceof Map check + clear exception is correct.

Note: testRejectNonMappingRule actually proves the fix (fails on old code). testCommentOnlyRuleUpdatePreservesPreviousRule passes either way — ListenableStateRouter.process() already has a try/catch around the parse & assign, so routerRule was never actually at risk of getting nulled out, worth checking that @minnnjuuu . PR description overstates that part a bit — real value here is clearer error messages, not data-loss prevention. Not a blocker.

LGTM , @zrlw

minnnjuuu:
Thanks for pointing this out. You're right: the right-hand side of the assignment is evaluated before `routerRule` is updated, so the existing exception handling already preserves the previous rule.

I've updated the PR description to focus on explicit YAML validation and clearer exception reporting, and removed the redundant state-preservation test.

LI123456mo:
@minnnjuuu , your test classes are well  you can keep both for test purposes, the only issue was the description which you have already modified it clearly

minnnjuuu:
Thanks for clarifying. The second test has already been removed in the latest commit, while `testRejectNonMappingRule` remains as the focused regression test. I'll keep the current version unless you would prefer both tests to remain.

LI123456mo:
> Thanks for clarifying. The second test has already been removed in the latest commit, while `testRejectNonMappingRule` remains as the focused regression test. I'll keep the current version unless you would prefer both tests to remain.

yees keep all the tests

minnnjuuu:
Got it. I’ll restore testCommentOnlyRuleUpdatePreservesPreviousRule so that both tests remain. Thank you

PR #16345, opened 2026-06-16
Fix #16344: graceful HTTP/2 GOAWAY migration to eliminate request failure window

devjue:
## What is the purpose of the change

Fixes #16344.

Eliminate the single-request failure window that occurs every time a Dubbo Triple Consumer receives an HTTP/2 `GOAWAY(errorCode=0, lastStreamId=MAX_INT)` frame from gateways / reverse proxies / providers that gracefully rotate connections (e.g. `max_requests_per_connection`, `idle_timeout`, hot-restart drain).

Previously, `AbstractNettyConnectionClient#onGoaway` nulled the channel reference immediately and `NettyConnectionHandler#onGoAway` then scheduled a reconnect, leaving a brief window in which `AbstractClusterInvoker#checkInvokers` throws `RpcException` before FailoverCluster's retry loop — so `retries=N` did NOT mitigate the issue. This PR implements graceful migration: keep the old channel serving until a new channel is ready, then atomically swap.

## Brief changelog

- **`NettyConnectionHandler#onGoAway`**: keep old channel alive; schedule `attemptGracefulMigration` on `connectivityExecutor` after `GRACEFUL_RECONNECT_DELAY_MS = 200ms`; one retry on failure, then fall back to `AbstractNettyConnectionClient#onGoaway`.
- **`NettyConnectionClient#initBootstrap`**: `closeFuture` listener now calls `compareAndClearNettyChannel(ch)` instead of `clearNettyChannel()` — prevents the old channel's close listener from wiping a freshly swapped-in new channel.
- **`AbstractNettyConnectionClient`**: add `compareAndClearNettyChannel`, plus package-private `getConnectivityExecutor()` / `getReconnectDuration()` accessors.
- **`NettyChannel#getChannelIfPresent`**: lookup-only API used by `channelInactive` to avoid allocating a transient `NettyChannel` purely for logging.
- **`TripleGoAwayHandler`**: log `errorCode` and `lastStreamId` so the GOAWAY trigger is observable.

## Verifying this change

Validated against a local Triple demo with sustained 1 QPS traffic:

- 26 GOAWAY frames received → 100% graceful migration success → **zero** `No provider available` errors and **zero** request failures.
- Channel rotation chain (first 18 migrations):
  `:45956 → :45110 → :53222 → :58366 → :56372 → :55436 → :55188 → :52926 → :50682 → :48682 → :47626 → :44014 → :43116 → :45044 → :48602 → :51880 → :50992 → :51672`
[...]

Issue #16344, opened 2026-06-16
[Bug] Single-request failure window when receiving HTTP/2 GOAWAY due to immediate channel nullification


devjue:
### Pre-check

- [x] I am sure that all the content I provide is in English.

### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/dubbo/issues?q=is%3Aissue) and found no similar issues.

### Apache Dubbo Component

Java SDK (apache/dubbo)

### Dubbo Version

Dubbo 3.3 (also reproducible on 3.2.x), JDK 8/17, Linux

### Steps to reproduce this issue

1. A Dubbo Triple Consumer talks to any HTTP/2 server (gateway / reverse proxy / direct Provider) that periodically emits `GOAWAY(errorCode=0, lastStreamId=MAX_INT)`. Common triggers: `max_requests_per_connection`, `idle_timeout`, `max_connection_duration`, hot-restart drain, overload manager.
2. Send sustained traffic (e.g. 1 QPS) over the connection.
3. Every GOAWAY produces exactly one failed request, then traffic recovers within ~1s.

Captured call chain on the Consumer side:

```
SslHandler.decode
  -> DefaultHttp2FrameReader.readGoAwayFrame
  -> Http2FrameCodec.onHttp2Frame
  -> Http2MultiplexHandler
  -> TripleGoAwayHandler.channelRead(DefaultHttp2GoAwayFrame)
  -> NettyConnectionHandler.onGoAway
  -> AbstractNettyConnectionClient.onGoaway(channel)   // sets channel ref to null
```

### What you expected to happen

A graceful GOAWAY (`errorCode=0`, `lastStreamId=MAX_INT`) means the server still accepts new streams on the old channel, so it should be transparent to callers — no request failure window.

Actual behaviour: `AbstractNettyConnectionClient#onGoaway` immediately nulls the channel reference, then `NettyConnectionHandler#onGoAway` schedules a reconnect. In the gap, the application logs:

```
WARN  SingleRouterChain  - filter result is empty
WARN  AbstractDirectory  - validInvokers: 0, invokersToReconnect: 1
ERROR ...                - No provider available for service ...
```

`AbstractClusterInvoker#checkInvokers` throws `RpcException` **before** entering FailoverCluster's retry loop, so configuring `retries=N` does **not** mitigate the issue.

Root causes:

1. The channel is nulled immediately instead of being kept alive during a graceful migration; with `lastStreamId=MAX_INT` the old channel could still serve.
[...]

PR #16454, opened 2026-09-06
Add max-connection-age support for Triple server (connection rotation for load balancing)

digyear:

Add server-side `max-connection-age` support for the Triple protocol, aligned with the standard gRPC capability ([gRFC A9: Server-Side Connection Management](https://github.com/grpc/proposal/blob/master/A9-server-side-conn-mgt.md)).

HTTP/2 multiplexes all requests over a single long-lived connection, so a Triple consumer stays pinned to the same provider instance forever — even after new instances are scaled up or traffic should be rebalanced. gRPC solves this with `maxConnectionAge` / `maxConnectionAgeGrace`: when a connection exceeds the configured age, the server sends an advisory GOAWAY (`NO_ERROR`, `last-stream-id = MAX_INT`) so in-flight requests keep running while clients migrate to a new connection (and get redistributed by the load balancer), then closes the connection after the grace period.

Triple currently has no equivalent: there is no way for a Triple server to proactively rotate long-lived connections. This PR adds it, disabled by default.

## Brief changelog

- **`TripleConfig`**: add `maxConnectionAge` (ms, default `-1` = disabled) and `maxConnectionAgeGrace` (ms, default `10000`) with validation, following the existing field conventions (`getXxxOrDefault()`); matching `TripleBuilder` methods in dubbo-config-api.
- **`TripleServerConnectionHandler`**: on `channelActive`, if `maxConnectionAge > 0`, schedule a one-shot task on the channel's EventLoop with **±10% jitter** (same as grpc-java) to avoid mass simultaneous reconnections when many connections are established at the same time. On expiry:
  1. send the advisory GOAWAY via the existing `GracefulShutdown.sendGoAwayFrame` (reuses the production code path — in-flight streams are unaffected);
  2. after the grace period, initiate close through `ctx.channel().close()` so the request traverses this handler's `close()` override and runs the existing graceful-shutdown sequence (final GOAWAY + PING) instead of an abrupt disconnect;
  3. pending tasks are cancelled on `channelInactive` (no leaks when the client disconnects first).
- **`TripleHttp2Protocol`**: pass the config into the handler on both server pipeline paths (direct h2 and h1-upgrade); raise the codec's `gracefulShutdownTimeoutMillis` to at least the configured grace so Netty's built-in backstop never fires before it.

Usage:

```yaml
dubbo:
  protocol:
    name: tri
    triple:
      max-connection-age: 3600000      # rotate connections after ~1h
      max-connection-age-grace: 10000  # let in-flight requests finish
```

## Verifying this change

New tests (all passing):

- `TripleServerConnectionHandlerTest` — EmbeddedChannel with virtual clock: advisory GOAWAY (`NO_ERROR` + `extraStreamIds=MAX_INT`) fires within the jitter window and never before it; disabled by default (24h, no frames); tasks cancelled on `channelInactive`; graceful-shutdown sequence (final GOAWAY + PING) initiated after the grace period.
- `TripleServerConnectionHandlerConcurrencyTest` — real NIO server/client, 100 connections sharing 2 EventLoop threads: concurrent rotation of all connections; client disconnects racing age expiry (random delays within ±20% of the window); server-side `close()` racing pending age tasks; 5 rounds of connection churn under `ResourceLeakDetector.PARANOID` with zero leak reports.
- `TripleConfigTest` / `TripleBuilderTest` — defaults, validation and builder wiring.

Regression: full test suites of `dubbo-rpc-triple` (492), `dubbo-common` (1169) and `dubbo-config-api` (695) pass locally.

Note for reviewers: with the current consumer-side GOAWAY handling (#16344), each rotation costs one request failure window on the consumer, so this feature is best paired with the graceful consumer migration in #16345. This PR is independent and safe to merge on its own (default: disabled).

## Does this pull request potentially affect one of the following parts

- [ ] Dependencies (does it add or upgrade a dependency)
- [x] The public API (new optional `TripleConfig` fields; default values keep behavior unchanged)
- [ ] The persistence format of the configurations
- [x] The default values of configurations (new fields only; existing defaults untouched)
- [ ] The serialization protocol
- [ ] The compatibility with previous versions

## Documentation

- Does this pull request introduce a new feature? **yes**
- If yes, how is the feature documented? **JavaDoc on the new `TripleConfig` fields** (website docs can follow after merge)

### DIFF
--- a/dubbo-common/src/main/java/org/apache/dubbo/config/nested/TripleConfig.java
+++ b/dubbo-common/src/main/java/org/apache/dubbo/config/nested/TripleConfig.java
@@ -43,6 +43,8 @@ public class TripleConfig implements Serializable {
     public static final int DEFAULT_MAX_HEADER_LIST_SIZE = 32_768;
     public static final int DEFAULT_MAX_MESSAGE_SIZE = 50 * 1024 * 1024;
     public static final float DEFAULT_WINDOW_UPDATE_RATIO = 0.5f;
+    public static final long DEFAULT_MAX_CONNECTION_AGE = -1L;
+    public static final long DEFAULT_MAX_CONNECTION_AGE_GRACE = 10_000L;
 
     public static final String H2_SETTINGS_MAX_MESSAGE_SIZE_KEY = "dubbo.protocol.triple.max-message-size";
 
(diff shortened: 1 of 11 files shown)


LI123456mo (commented):
Have just finished today going through your diff , confirmed all tests passed and works as per expected beahaviour,  hope  mantainers take time to see this

LGTM @zrlw @RainYuY

